PR #10, opened 2026-03-04
feat(tools,gateway): credential output protection and DP decouple

LikiosSedo:
## Summary

Two related changes to improve security and simplify the deep investigation workflow:

1. **DP decouple**: Remove the blocking `deepSearchGate` so `propose_hypotheses` is non-blocking and the model proceeds to `deep_search` immediately — eliminates state desync issues when WebSocket connections drop mid-investigation.

2. **Credential output protection**: 7-layer defense preventing the model from seeing or outputting sensitive credential data (file paths, server URLs, API keys, tokens, cluster internal IDs).

### Solution

**DP decouple** (commit 1): Removed `deepSearchGate`, `HYPOTHESES_CONFIRMED_SENTINEL`, `confirmHypotheses` RPC, `/confirm-hypotheses` endpoint, gate input handlers, and auto-continue gate check across 9 files.

**Credential protection** (commit 2):
- L1: `credential_list` strips files/metadata, returns only safe fields
- L2: KUBECONFIG auto-inject, `--kubeconfig=<name>` resolution, sensitive file read blocking, `kubectl config view --raw` blocking
- L3: System prompt forbids credential output and file reads
- L4: Output redaction on WS stream and DB writes (9 generic token patterns)
- L5: `READ_BLOCKED_PATTERNS` in agent-factory for read tool
- L6: `sanitizeEnv` for all subprocess tools (deny-first SICLAW_* filter)

## Test Plan

- [x] `npm run build` passes
- [x] `npx vitest run src/tools/sanitize-env.test.ts src/gateway/output-redactor.test.ts` — 21 tests pass
- [ ] Manual: `credential_list` returns no files/metadata
- [ ] Manual: `kubectl get pods` works without `--kubeconfig=<path>`
- [ ] Manual: `cat $KUBECONFIG` blocked
- [ ] Manual: `kubectl config view --raw` blocked
- [ ] Manual: Output with API keys shows [REDACTED]

### DIFF
--- a/src/agentbox/http-server.ts
+++ b/src/agentbox/http-server.ts
@@ -13,7 +13,7 @@ import type { AgentBoxSessionManager } from "./session.js";
 import type { SessionMode } from "../core/agent-factory.js";
 import type { BrainType } from "../core/brain-session.js";
 import { hasOpenAIProvider, ensureProxy } from "../core/llm-proxy.js";
-import { deepSearchEvents, deepSearchGate, HYPOTHESES_CONFIRMED_SENTINEL } from "../tools/deep-search/events.js";
+import { deepSearchEvents } from "../tools/deep-search/events.js";
 import { createChecklist, buildActivationMessage } from "../tools/dp-tools.js";
 import { loadConfig } from "../core/config.js";
 import { GatewayClient } from "./gateway-client.js";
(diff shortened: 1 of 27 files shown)


nightmeng (approved):
LGTM. The DP decouple is a clean removal and the 6-layer credential protection is solid defense-in-depth.

Minor suggestions for follow-up (non-blocking):
- `channel-bridge.ts` uses a relative credentials dir path (`".siclaw/credentials"`) while `rpc-methods.ts` uses `path.resolve()` — worth making consistent
- Consider adding `sed` to `FILE_READING_CMDS` in `restricted-bash.ts`
- `resolveKubeconfigByName` could use unit tests

None of these are security-critical since other layers cover the gaps.

PR #17, opened 2026-03-05
feat: structured knowledge extraction from deep investigation conclusions

LikiosSedo:
## Summary

- Phase 4 conclusion now outputs both human-readable text and a structured JSON block (`STRUCTURED_EXTRACTION_START/END`), extracted in a single LLM call (no extra API cost)
- Structured data (root cause category, affected entities, environment tags, causal chain, confidence) stored in a new SQLite `investigations` table for precise SQL querying
- Prior investigation retrieval upgraded to hybrid: structured SQL query (precise category/entity match) + semantic chunk search (fuzzy), both injected into Phase 2 hypothesis generation
- Graceful degradation at every layer: missing structured block, JSON parse failure, SQLite write failure — all fall back silently to existing text-only behavior

## Test plan

- [ ] Run `npx tsc --noEmit` — passes
- [ ] Run a deep_search investigation, verify Phase 4 output contains `STRUCTURED_EXTRACTION_START/END` block
- [ ] Check `~/.siclaw/memory/.memory.db` → `investigations` table has a new record with populated fields
- [ ] Run a second related deep_search, verify Phase 2 retrieves structured prior investigation(s) in context
- [ ] Simulate LLM not outputting structured block — verify text-only conclusion with no errors

### DIFF
--- a/src/core/agent-factory.ts
+++ b/src/core/agent-factory.ts
@@ -1,4 +1,5 @@
 import fs from "node:fs";
+import { homedir } from "node:os";
 import path from "node:path";
 import { readFile as fsReadFile, writeFile as fsWriteFile, access as fsAccess, mkdir as fsMkdir } from "node:fs/promises";
 import {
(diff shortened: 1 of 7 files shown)


nightmeng (approved):
LGTM. Structured extraction parsing with graceful degradation, hybrid retrieval, and SQLite storage all look correct. Clean separation of concerns.

PR #19, opened 2026-03-05
fix(security): harden path handling, redaction, and read-only command boundaries

LikiosSedo:
## Summary
- harden skill script file-name/path resolution in Gateway (skill.create / skill.update) to block path traversal
- harden AgentBox bundle/credential materialization with strict child-path validation
- redact outbound SSE payloads recursively instead of only selected fields
- tighten read-only command guardrails: mount now allows only listing mode, and curl blocks request bodies/POST and non-HTTP(S) targets
- update command restriction tests for the tightened behavior

## Validation
- npm run build
- npm test -- src/tools/command-sets.test.ts

## Notes
- existing restricted-bash baseline test failures are unchanged and reproducible on unmodified main

### DIFF
--- a/skills/core/deep-investigation/SKILL.md
+++ b/skills/core/deep-investigation/SKILL.md
@@ -81,6 +81,15 @@ Synthesize the deep_search results into a conclusion.
 3. **Avoid redundant validation**: Let deep_search handle hypothesis validation with its parallel sub-agents. Running the same checks manually is inefficient.
 4. **Use DP-specific tools during DP mode**: Prefer `manage_checklist` over `task_plan` during deep investigation to keep the checklist UI consistent.
 
+## Behavioral Patterns
+
+- User is actively discussing (asking questions, giving feedback) → use propose_hypotheses to align direction before committing to deep_search
+- User gives an open-ended request ("help me look into X") → do triage first, then communicate your findings and proposed direction before going deep
+- User gives a specific directive ("validate H1 and H3") → execute immediately, no need to re-confirm
+- deep_search fails or returns insufficient results → discuss with the user and adjust hypotheses before retrying; do not repeatedly call deep_search autonomously
+- Simple issue becomes apparent during triage → present findings directly, suggest ending investigation early rather than forcing the full workflow
+- propose_hypotheses is a communication tool, not a gate → use it whenever presenting investigation thinking to the user, regardless of whether DP mode is active
+
 ## Early Exit
 
 - If triage alone is sufficient to answer the question, present your findings and ask the user before ending.
--- a/skills/core/deep-investigation/default-memory.md
+++ b/skills/core/deep-investigation/default-memory.md
@@ -1,8 +1,0 @@
-## Investigation Patterns (learned)
-
-- User is actively discussing (asking questions, giving feedback) → use propose_hypotheses to align direction before committing to deep_search
-- User gives an open-ended request ("help me look into X") → do triage first, then communicate your findings and proposed direction before going deep
-- User gives a specific directive ("validate H1 and H3") → execute immediately, no need to re-confirm
-- deep_search fails or returns insufficient results → discuss with the user and adjust hypotheses before retrying; do not repeatedly call deep_search autonomously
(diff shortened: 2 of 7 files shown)


chent1996 (commented):
## Review: fix(security) — path handling, event redaction, and read-only command boundaries

### Overall

Solid security hardening across multiple attack surfaces:
- **Path traversal defense**: file writes in both Gateway and AgentBox now validated
- **SSE redaction**: upgraded from field-specific to recursive full-payload redaction
- **curl tightening**: request bodies blocked, POST removed, non-HTTP(S) schemes blocked
- **mount tightening**: only bare `mount` and `mount -l` allowed

See inline comments for specific issues.

nightmeng (approved):
## Review: fix(security): harden path handling, redaction, and read-only command boundaries

Solid security hardening PR. The changes are well-scoped and the test coverage is thorough. Each change addresses a concrete attack vector.

### Positives

- **Path traversal protection**: Double-layer defense (regex allowlist + `path.dirname` check) in both Gateway and AgentBox is the right approach. Covers skill dirs, script names, and credential files.
- **Recursive SSE redaction**: Much more robust than the previous field-by-field approach. Any future event fields with sensitive data are automatically covered.
- **curl read-only enforcement**: Removing `-d`/`--data` from safe flags, blocking POST, and validating URL schemes closes real attack surface. `file://`, `ftp://`, and local path targets are all good catches.
- **mount tightening**: Only `mount` or `mount -l` is sensible — the old `mount -t ext4` path could trigger fstab side-effects.
- **Tests updated**: Every behavioral change has corresponding test updates.

### Issues

1. **Duplicated validation logic** — `SAFE_FILE_NAME_RE` / `assertSafeFileName` / `resolvePathUnderDir` in `rpc-methods.ts` and the nearly identical `SAFE_ENTRY_NAME_RE` / `assertSafeEntryName` / `resolveSafeChildPath` in `http-server.ts` are copy-pasted. Consider extracting to a shared utility (e.g. `src/shared/path-safety.ts`). If these diverge over time, one side might miss a fix.

2. **`CURL_DATA_FLAGS` contains `-d` but it's unreachable** — Since `-d` is a short flag (doesn't start with `--`), it enters the short-flag parsing path where `d` is blocked by `CURL_SAFE_SHORT_CHARS`. The `CURL_DATA_FLAGS.has(arg)` check in the `--` branch never sees `-d`. This means `checkCurlDataValue` is only reachable via `--data`/`--data-raw`/`--data-urlencode`, never via `-d`. Not a bug, but a minor dead-code inconsistency — consider removing `-d` from `CURL_DATA_FLAGS` or adding a comment.

3. **`checkCurlDataValue` @file branch is also semi-dead** — Since all data flags are now blocked unconditionally, the more-specific `@file` error message (line 2 of the function) is only reachable for `--data @/path` but the test for `-d @/etc/passwd` now expects the generic "not allowed" message (because `-d` is caught earlier by short-char validation). The function logic is correct but a bit misleading — it reads like it might sometimes allow data without @file, but it always returns an error.

4. **SSRF via internal hostnames** — `validateCurlTarget` correctly blocks non-HTTP schemes and local paths, but allows bare hostnames like `curl kubernetes.default.svc:443` or `curl metadata.google.internal`. Since only GET/HEAD/OPTIONS are allowed, the blast radius is limited, but for K8s environments this could leak metadata API info or internal service data. A future enhancement could block RFC1918/link-local ranges and known cloud metadata IPs (`169.254.169.254`). Not a blocker for this PR.

5. **Recursive redaction creates new objects per event** — The old code mutated in-place; the new `redactEventPayloadValue` allocates new object/array trees for every SSE event. This is actually *better* semantically (no mutation of the cached event), but worth being aware of for high-throughput scenarios. Fine for current usage.

### Nits

- `rpc-methods.ts`: In `skill.create` script handling, `assertSafeFileName` is called first, then `resolvePathUnderDir` calls it again internally. Harmless but redundant — could pass the already-validated `scriptName` directly into `path.resolve`.
- Trailing blank line removal at EOF of `rpc-methods.ts` (good cleanup).

### Verdict

Approve. The security improvements are meaningful and correctly implemented. The shared utility suggestion (#1) is worth doing but not blocking.

PR #56, opened 2026-03-09
feat: add /setup in-session command and first-run wizard for TUI mode

LikiosSedo:
## Summary

- Add `/setup` slash command for in-session credential and model provider management (all 5 credential types: kubeconfig, SSH password, SSH key, API token, API basic auth)
- Add @clack/prompts-based first-run wizard that runs before session creation when no provider is configured
- Remove `--credentials`, `--add-kube`, `--setup` CLI flags — replaced by `/setup` and first-run wizard
- Remove LLM env var support (`SICLAW_LLM_*`) — all config goes through `settings.json` for security
- Remove dead `SICLAW_LLM_*`/`SICLAW_EMBEDDING_*` env var injection in Gateway (AgentBox uses mTLS settings API)
- Harden file permissions: `settings.json` and `manifest.json` written with `0o600`, dirs with `0o700`
- SSH config: `StrictHostKeyChecking no` → `accept-new` (TOFU)
- Add `sanitizeSshField()` to prevent SSH config injection via newlines
- Add `safeName()` empty-value guard and path traversal protection via `resolveUnderDir()`
- Add status bar notification for missing config, 401/403 error hints pointing to `/setup`
- Update system prompt to guide model to use `/setup` instead of old CLI flags
- Extract shared `PRESETS` to `src/core/provider-presets.ts`, clean up dead code in `cli-setup.ts`

## Test plan

- [ ] First run (no settings.json): wizard launches, configure provider, enters session successfully
- [ ] `/setup` → Credentials → Add kubeconfig (file path + paste content)
- [ ] `/setup` → Credentials → List / Remove
- [ ] `/setup` → Models → Add provider / Add model / Set default / Remove
- [ ] `credential_list` tool returns correct data after `/setup` changes
- [ ] Status bar shows missing config reminder, clears after configuration
- [ ] 401/403 API errors show `/setup` hint
- [ ] No regressions in Gateway mode (changes are TUI-only additions)
- [ ] `npm run build` passes

### DIFF
--- a/src/cli-main.ts
+++ b/src/cli-main.ts
@@ -202,7 +202,7 @@ if (isPrintMode && initialMessage) {
 // Auto-save session memory (mirrors AgentBox release flow)
 if (session.sessionFile) {
   const sessionDir = path.dirname(session.sessionFile);
-  const memoryDir = path.resolve(process.cwd(), loadConfig().paths.userDataDir, "memory");
+  const memoryDir = path.resolve(process.cwd(), config.paths.userDataDir, "memory");
   try {
     const saved = await saveSessionMemory({ sessionDir, memoryDir });
     if (saved) {
--- a/src/core/extensions/setup.ts
+++ b/src/core/extensions/setup.ts
@@ -289,7 +289,12 @@ async function addSshPassword(
   const password = await ctx.ui.input("Password");
   if (!password) return;
 
-  registerSshPassword(credentialsDir, { name, host, port, username, password });
+  try {
+    registerSshPassword(credentialsDir, { name, host, port, username, password });
+  } catch (err) {
+    ctx.ui.notify(err instanceof Error ? err.message : String(err), "error");
+    return;
+  }
   ctx.ui.notify(`SSH password credential "${name}" added`);
 }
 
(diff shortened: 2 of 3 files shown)


nightmeng (approved):
All review items addressed:

1. **P1 Tests** — `credential-manager.test.ts` added (129 lines) covering `safeName`, `sanitizeSshField`, `resolveUnderDir`, `readManifest` with proper edge cases (injection, traversal, corruption) ✅
2. **P2 `applyEnvOverrides()`** — function and call site fully removed from `config.ts` ✅
3. **P2 K8s verify** — `server.ts` env injection changes moved to PR #57, this PR is now purely TUI-scoped ✅

Additional fixes since last review:
- `ensureDir` permissions hardened with try/catch for credential registration
- Proper `SetupCtx` type replaces inline `{ ui: { select: Function } }` casts
- `console.clear()` removed from first-run wizard (was hiding startup logs)

Remaining non-blocking items (can follow up):
- `setup.ts` is 823 lines — consider splitting credentials/models handlers
- SSH password input as plaintext via `ctx.ui.input()` — add TODO if no masked input available

PR #57, opened 2026-03-09
refactor: remove env var config delivery for AgentBox

nightmeng:
## Problem

AgentBox received LLM/embedding config through two redundant channels:
1. **settings.json** via mTLS `GET /api/internal/settings` — protected from agent access by `restricted-bash.ts`
2. **Environment variables** via `envResolver` — exposed via `kubectl describe pod`, `/proc/self/environ`

This creates unnecessary security risk (API keys visible in pod env) and maintenance confusion (two code paths for the same data).

Additionally, `SICLAW_SKILLS_DIR` and `SICLAW_USER_DATA_DIR` were injected into K8s pods with values identical to `config.ts` defaults — redundant since each pod is isolated.

## Solution

- Remove `envResolver` from Gateway `server.ts` (no longer injects `SICLAW_LLM_*`, `SICLAW_EMBEDDING_*`)
- Remove `EnvResolver` type, `setEnvResolver()`, and `resolveApiKey()` from `AgentBoxManager`
- Remove redundant `SICLAW_SKILLS_DIR` / `SICLAW_USER_DATA_DIR` from K8s pod env spec
- Add ADR-009 documenting the decision

Settings.json (delivered via mTLS) is now the sole channel for sensitive config. Bootstrap env vars (`SICLAW_GATEWAY_URL`, `SICLAW_AGENTBOX_PORT`, `SICLAW_CERT_PATH`, `SICLAW_CREDENTIALS_DIR`) are retained.

## Test plan

- [x] `tsc --noEmit` passes
- [x] All 749 tests pass
- [x] Built and deployed to test K8s cluster — Gateway starts successfully
- [ ] Trigger new AgentBox session, verify pod env no longer contains `SICLAW_LLM_API_KEY` etc.
- [ ] Verify AgentBox session creates successfully with LLM config from settings.json

### DIFF
--- a/docs/design/decisions.md
+++ b/docs/design/decisions.md
@@ -209,3 +209,53 @@ DB-based coordination with heartbeat. Each cron instance:
 - ✅ Job ownership is explicit and auditable
 - ⚠️ Up to 90s delay in job reassignment after a crash (dead threshold)
 - ⚠️ Duplicate firing possible in edge case: instance A marks job "executing" but crashes before updating `lastRunAt`; instance B reclaims and fires again. Acceptable for SRE automation use cases; add idempotency guards in critical skill scripts.
+
+---
+
+## ADR-009: AgentBox Config via Settings File, Not Environment Variables
+
+**Status**: Active
+
+**Context**:
+AgentBox needs LLM provider config (API keys, base URLs, models) and embedding config from Gateway. Previously, Gateway delivered this through two redundant channels:
+
+1. **settings.json** — `GET /api/internal/settings` (mTLS-protected) returns full provider/embedding config; AgentBox writes it to `.siclaw/config/settings.json`
+2. **Environment variables** — Gateway's `envResolver` injected `SICLAW_LLM_API_KEY`, `SICLAW_LLM_BASE_URL`, `SICLAW_LLM_MODEL`, `SICLAW_EMBEDDING_*` into spawned pods/processes
+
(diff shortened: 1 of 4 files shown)


LikiosSedo (approved):
## Review: PR #57 — Remove env var config delivery for AgentBox

Clean, focused refactor. +58 -79 across 4 files. LGTM.

### Verified

- `SICLAW_WORKSPACE_ALLOWED_TOOLS` — confirmed never read by AgentBox code (only written in k8s-spawner), safe to remove
- `SICLAW_CERT_PATH` — code default `/etc/siclaw/certs` matches K8s mount, safe to remove
- `SICLAW_CREDENTIALS_DIR` — AgentBox tools pass this via `kubeconfigRef.credentialsDir` (not env var), safe to remove from pod spec
- `resolveEnv()` async→sync change — callers correctly updated
- ADR-009 is well-written, covers context/decision/consequences properly

### One note

**ADR-009 mentions TUI `applyEnvOverrides()` still exists** — but PR #56 removes it. Whichever PR merges second should update the ADR. Also `server.ts` will conflict between #56 and #57 (both touch the `envResolver` section). Suggest merging #57 first since it's the cleaner removal, then #56 rebases and drops its `server.ts` hunk.

### Minor

`resolveEnv()` is now `return configEnv ?? {}` — trivial enough to inline, but keeping it as a named method is fine for readability.

PR #63, opened 2026-03-09
fix: skill AI review — URL, spec coverage, config caching, and UX

LikiosSedo:
## Summary

- **Broken API URL**: Remove extra `/v1` prefix in script-evaluator (`baseUrl/v1/chat/completions` → `baseUrl/chat/completions`), matching all other call sites (llm-proxy, sub-agent)
- **Specs not reviewed**: Review `skill.md` content in addition to scripts — specs contain command templates the agent will follow, static analysis now also scans specs for dangerous patterns
- **Config caching bug**: Don't cache null LLM config result so review works after model provider is configured post-startup
- **Stale review on re-submit**: Clear old AI reviews on submit so UI shows "in progress" instead of stale results from previous submission
- **No auto-refresh**: Poll for AI review completion (3s interval) so results appear without manual page refresh
- **Hardcoded riskLevel**: Human review records now inherit AI risk assessment instead of always writing "low"

## Test plan

- [ ] Submit spec-only skill — AI reviews spec content (not "No scripts to review")
- [ ] Submit skill with dangerous command in spec (e.g. `kubectl delete`) — flagged as high risk
- [ ] Re-submit skill — old review clears, "AI review in progress" shown, new result auto-appears without page refresh
- [ ] Approve/reject skill — admin review record inherits AI risk level
- [ ] Configure model provider after Gateway startup, then submit skill — AI review works (no permanent "No LLM provider" failure)

### DIFF
--- a/src/gateway/db/repositories/skill-review-repo.ts
+++ b/src/gateway/db/repositories/skill-review-repo.ts
@@ -3,7 +3,7 @@
  */
 
 import crypto from "node:crypto";
-import { eq, desc } from "drizzle-orm";
+import { eq, and, desc } from "drizzle-orm";
 import type { Database } from "../index.js";
 import { skillReviews, type ReviewFinding } from "../schema.js";
 
@@ -53,5 +53,11 @@ export class SkillReviewRepository {
       .from(skillReviews)
       .where(eq(skillReviews.skillId, skillId))
       .orderBy(desc(skillReviews.createdAt));
+  }
+
+  async deleteAiReviewsForSkill(skillId: string) {
+    await this.db.delete(skillReviews).where(
+      and(eq(skillReviews.skillId, skillId), eq(skillReviews.reviewerType, "ai")),
+    );
   }
 }
--- a/src/gateway/rpc-methods.ts
+++ b/src/gateway/rpc-methods.ts
@@ -2145,12 +2145,17 @@ export function createRpcMethods(
     }
 
     // 5. AI script review using staged files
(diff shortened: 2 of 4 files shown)


nightmeng (approved):
Six well-targeted fixes, all correct:

1. **URL fix**: `baseUrl/v1/chat/completions` → `baseUrl.replace(/\/+$/, "")/chat/completions` — removes the extra `/v1` prefix, consistent with other call sites. The trailing-slash strip is a nice touch.

2. **Spec review**: Static analysis now also scans `skill.md` content by pushing it into `allContent` — specs contain command templates the agent follows, so they should be reviewed for dangerous patterns. Good catch.

3. **Config caching**: `if (this.config === undefined)` → `if (this.config)` — null results are no longer cached, so AI review starts working as soon as a model provider is configured post-startup. Trades a small DB re-query cost for correctness.

4. **Clear stale reviews**: `deleteAiReviewsForSkill()` before triggering new review ensures the UI shows "in progress" instead of stale results. The `and()` filter on `reviewerType === "ai"` correctly preserves admin reviews.

5. **Auto-refresh polling**: 3s `setInterval` with proper cleanup via `useEffect` return + `clearInterval` on success. Dependency array `[expanded, aiReview, skill.id, sendRpc]` is correct — the effect stops when `aiReview` becomes truthy after `setReviews` updates state.

6. **Risk level inheritance**: Human review records now inherit the AI's `riskLevel` instead of hardcoding `"low"` — correctly falls back to `"low"` when no AI review exists.

LGTM.

PR #64, opened 2026-03-09
security: defense-in-depth for AgentBox command execution

nightmeng:
## Summary

- **Application-level defense**: Refactored command validation into a 6-pass pipeline with context-based whitelists (`local`/`node`/`pod`/`nsenter`), declarative `COMMAND_RULES` engine (pipeOnly, noFilePaths, blockedFlags), and `extractPipeline()` for pipe-position tracking
- **Vulnerability fix**: Blocked `curl file://` protocol — LLM agent could read kubeconfig via `curl file:///app/.siclaw/credentials/kubeconfig`
- **OS-level isolation (ADR-010)**: Dual-user model (`agentbox` + `sandbox`) with setgid kubectl. Child processes run as `sandbox` user via sudo, which cannot read credential files. K8s pods get `drop: ALL` + `add: [SETUID, SETGID]` capabilities and seccomp RuntimeDefault

## Security Architecture

```
Entrypoint (root) → runuser → Node.js (agentbox)
                                  ↓
                        sudo -E -u sandbox → bash -c 'command'
                                               ↓
                                         kubectl (setgid kubecred) → reads kubeconfig ✓
                                         curl/grep/etc → can't read credentials ✓
```

6 independent defense layers:
1. OS user isolation (primary) — sandbox user, file permissions
2. Command whitelist — context-based binary allowlists
3. Shell operator blocking — $(), backticks, redirections
4. Pipeline validation — kubectl subcommand + exec checks
5. COMMAND_RULES — per-command: pipeOnly, noFilePaths, blockedFlags
6. Sensitive path patterns — block commands targeting credential paths

## Files Changed

| File | Change |
|------|--------|
| `docs/design/security.md` | New: full security architecture document |
| `docs/design/decisions.md` | ADR-010: dual-user OS isolation |
| `docs/design/invariants.md` | Section 3 rewritten for defense-in-depth |
| `CLAUDE.md` | Security section updated |
| `src/tools/command-sets.ts` | COMMAND_RULES engine, context categories |
| `src/tools/command-validator.ts` | 6-pass pipeline, extractPipeline() |
| `src/tools/restricted-bash.ts` | sudo -u sandbox wrapper in production |
| `src/tools/restricted-bash.test.ts` | +250 lines: pipeline, pipeOnly, curl tests |
| `Dockerfile.agentbox` | Dual-user, setgid kubectl, sudoers, SUID audit |
| `docker/agentbox-entrypoint.sh` | New: permission enforcement + privilege drop |
| `src/gateway/agentbox/k8s-spawner.ts` | securityContext, initContainer, credentials volume |

## Test plan

- [x] All 187 restricted-bash tests pass
- [x] Docker image builds successfully, user/permission setup verified
- [x] Deployed to K8s cluster (siclaw namespace), agentbox pod starts and responds
- [ ] Security test: agent cannot read credentials via text commands (cut, grep, etc.)
- [ ] Security test: agent cannot read credentials via curl file://
- [ ] Security test: kubectl works normally (setgid kubecred)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -34,9 +34,11 @@ Gateway + K8sSpawner  (production — one isolated pod per user)
 
 `buildSkillBundle()` packages **only team + personal skills** (never core skills). Core skills are baked into the Docker image / repo checkout. Calling `skillsHandler.materialize()` does NOT restore core skills — it only writes what's in the bundle.
 
-### 🔴 Shell Security: Whitelist-Only
+### 🔴 Shell Security: Defense-in-Depth
 
-The security model is whitelist-only (not blacklist). A binary must be explicitly listed in `ALLOWED_COMMANDS` (`src/tools/command-sets.ts`) to execute. `sed`, `awk`, `nc`, `wget` are **intentionally excluded**. kubectl is **read-only** (13 safe subcommands; all write ops permanently blocked).
+> Full spec: `docs/design/security.md` — read it before touching execution tools, Dockerfile, or K8s manifests.
+
+Primary defense: **OS-level user isolation** — child processes run as `sandbox` user (cannot read credentials); `kubectl` has setgid `kubecred` group (ADR-010). Secondary defense: **whitelist-only command validation** — binaries must be in `ALLOWED_COMMANDS` (`src/tools/command-sets.ts`). `sed`, `awk`, `nc`, `wget` are **intentionally excluded**. kubectl is **read-only** (13 safe subcommands; all write ops permanently blocked).
 
 ### 🔴 sql.js: Single-Process Lock
 
@@ -84,7 +86,7 @@ Memory tools (`memory_search`, investigation history) are **pi-agent only** — 
 **Before approving any PR, verify:**
 
 1. **Deployment mode awareness**: Does the change respect local vs K8s isolation? If it touches resource sync, skills materialization, or filesystem writes, re-read `docs/design/invariants.md §1-2`
-2. **Security model intact**: No new shell execution paths that bypass `command-sets.ts`. No weakening of the 4-pass pipeline. Skill scripts still go through review gate.
+2. **Security model intact**: No new shell execution paths that bypass `command-sets.ts`. No weakening of the 6-pass pipeline. OS user isolation preserved (ADR-010). Skill scripts still go through review gate. Read `docs/design/security.md` for full model.
 3. **PR description complete**: Must have Problem + Solution (not just diff summary). See `CONTRIBUTING.md` for format.
 4. **TypeScript conventions**: ESM-only (`.js` imports), strict mode, named exports, no default exports in barrels.
 5. **Both brain types**: Tool changes must work with both pi-agent (TypeBox) and claude-sdk (MCP/Zod) if applicable.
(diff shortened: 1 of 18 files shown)


nightmeng (commented):
Thorough review of this security-critical PR. The defense-in-depth architecture is well-designed — OS-level user isolation as primary defense with application-level validation as secondary is the right approach. The refactoring quality (exec-utils.ts consolidation) is excellent, and test coverage is solid (250+ new lines).

Three items worth addressing:

### 1. Documentation inconsistency: `runuser` vs `sudo`

`docs/design/security.md` section 3.1 and 3.6 show `runuser -u sandbox` in code examples, but the actual implementation in `restricted-bash.ts` uses `sudo -E -u sandbox`. The Dockerfile correctly sets up sudoers for this. The docs should be updated to match the implementation.

Also worth considering: since the K8s securityContext already grants `CAP_SETUID`/`CAP_SETGID`, you could potentially use `runuser` instead of `sudo`, which would eliminate the need to keep the SUID bit on `/usr/bin/sudo`. The entrypoint already uses `runuser` (as root→agentbox). This would let you strip ALL SUID binaries from the image.

### 2. `SENSITIVE_PATH_RE` may be too broad

The catch-all `/.siclaw\//` pattern blocks any command referencing any `.siclaw` subpath. This would false-positive on legitimate piped commands referencing `user-data/` or `skills/` paths, e.g.:

```bash
kubectl get pods | grep .siclaw/user-data/report.txt
```

Consider using more specific patterns like `\.siclaw\/credentials\/` and `\.siclaw\/config\/` instead of the catch-all, since `user-data/` is intentionally sandbox-writable and `skills/` is sandbox-readable.

### 3. `readOnlyRootFilesystem` gap between doc and implementation

`docs/design/security.md` section 5.1 recommends `readOnlyRootFilesystem: true` in the container securityContext, but `k8s-spawner.ts` doesn't set it. This is a nice-to-have hardening that prevents filesystem tampering if the agent process is compromised. Would need `emptyDir` mounts for writable paths (`/tmp`, `/app/.siclaw/user-data`).

---

Everything else looks good:
- `extractPipeline()` pipe-position tracking is correct
- curl `file://`/`ftp://`/`dict://`/`gopher://` blocking is complete (and `--url` is not in `CURL_SAFE_FLAGS` so it's blocked by the flag whitelist)
- `COMMAND_RULES` pipeOnly + noFilePaths enforcement is clean
- K8s securityContext with `drop: ALL` + `add: [SETUID, SETGID]` and `seccompProfile: RuntimeDefault` is correct
- initContainer for volume permission fixing is a standard pattern
- 187 tests all passing

nightmeng (commented):
Thanks for the thorough review. All three items addressed in commit `0f1d7a8`:

### 1. Documentation inconsistency: `runuser` vs `sudo` — Fixed

Updated all 12 references across `security.md`, `decisions.md`, `invariants.md`, and `CLAUDE.md` from `runuser -u sandbox` to `sudo -E -u sandbox`. Also updated the Dockerfile example (§3.5), code example (§3.6), K8s YAML (§5.1), capability justification (§5.2), and the full implementation checklist (§10) to match the actual implementation.

Re: using `runuser` instead of `sudo` to eliminate SUID — not feasible. `runuser` from util-linux hardcodes `if (getuid() != 0)` check regardless of capabilities. Since the Node.js process runs as agentbox (UID 1000), not root, `runuser` refuses to switch users. `sudo`'s SUID bit is the standard mechanism for non-root user switching.

### 2. `SENSITIVE_PATH_RE` too broad — Fixed

Replaced catch-all `/.siclaw\//` with specific `/.siclaw\/credentials\//` and `/.siclaw\/config\//`. The `user-data/` and `skills/` paths are no longer false-positive targets. These are the only two subdirectories that contain sensitive data (kubeconfig, API keys, settings).

### 3. `readOnlyRootFilesystem` — Fixed

Added `readOnlyRootFilesystem: true` to the container securityContext in `k8s-spawner.ts` and a `/tmp` emptyDir volume (100Mi sizeLimit). All other writable paths (`credentials`, `skills`, `user-data`) were already on emptyDir mounts. `sudo` with `NOPASSWD` does not require timestamp caching, so no additional writable paths are needed.

Deployed and verified on the K8s cluster.

fred-scitix (approved):

PR #67, opened 2026-03-10
fix(agentbox): resolve MCP tools not loading on readOnlyRootFilesystem

chent1996:
## Summary

- Add `config` emptyDir volume mount for `/app/.siclaw/config` so `settings.json` can be written in `readOnlyRootFilesystem` containers
- Decouple `fetchSettings()` and `syncAllResources()` into independent try-catch blocks so resource sync (MCP, skills) runs even if settings persist fails
- Remove stale `/tmp` fallback logic that silently failed on existing directories
- Fix MCP `materialize()` to replace `mcpServers` instead of merging — Gateway payload is the source of truth, `Object.assign` kept stale keys when all servers were disabled

## Root cause

AgentBox containers have `readOnlyRootFilesystem: true` but `/app/.siclaw/config/` had no volume mount. Writing `settings.json` failed with EROFS, and because `syncAllResources()` was in the same try-catch, it was skipped entirely — MCP tools, allowedTools, and team/personal skills were never synced.

## Test plan

- [ ] Deploy to K8s and verify AgentBox startup logs show `Fetched settings from Gateway` and `mcp synced successfully`
- [ ] Configure MCP server in WebUI, verify tools appear in chat session
- [ ] Disable all MCP servers, verify tools are removed (no stale keys)
- [ ] Simulate Gateway unavailability, verify `syncAllResources` still attempts independently


nightmeng (commented):
## PR Review: fix(agentbox): resolve MCP tools not loading on readOnlyRootFilesystem

Overall direction is good — removing `SICLAW_MCP_DIR` / `.siclaw/mcp` intermediate file and unifying MCP config into `settings.json` is the right simplification. The startup try-catch decoupling and emptyDir volume addition are solid fixes.

### Issues

#### 1. `k8s/agentbox-template.yaml` missing `config` emptyDir volume (Bug)

`k8s-spawner.ts` correctly adds the `config` emptyDir volume + init-permissions, but `k8s/agentbox-template.yaml` (the reference template) only removes the old `SICLAW_MCP_DIR` env and `.siclaw/mcp` mount — it does **not** add the `config` emptyDir volume or mount.

If anyone uses the template directly (or refers to it for manual deployments), `settings.json` writes will still fail with EROFS.

Need to add to the template:
```yaml
# volumes:
- name: config
  emptyDir: {}

# volumeMounts:
- name: config
  mountPath: /app/.siclaw/config
```

#### 2. `docs/step-remove-mcp-dir.md` is stale (Nit)

The Step 1 code example shows `Object.assign` merge pattern, but commit 3 changed to direct replacement (`config.mcpServers = payload?.mcpServers ?? {}`). The doc doesn't match the final implementation. 

Since this is a migration guide that's already been executed, consider removing it entirely — it will confuse future readers. Or at minimum update it to reflect the final "replace, not merge" design.

#### 3. `materialize()` mutates cached config object (Minor)

```typescript
const config = loadConfig();      // returns the cached singleton
config.mcpServers = mcpServers;   // mutates it
writeConfig(config);              // writes + invalidates cache
```

This works because `writeConfig()` immediately sets `cached = null`, but mutating the shared cached object is fragile. If any code reads `loadConfig()` concurrently (e.g., in an async gap before `writeConfig`), it would see the mutation before it's persisted.

Safer pattern:
```typescript
const config = loadConfig();
writeConfig({ ...config, mcpServers });
```

#### 4. `loadMcpServersConfig` still exported from `mcp-client.ts` (Verify)

The PR removes the function body and all call sites, but double-check that the export is also removed. The diff shows it is — just flagging since there are other worktree copies that still reference it.

### Looks Good

- Startup try-catch decoupling — `fetchSettings()` failure no longer blocks `syncAllResources()`
- "Replace, not merge" semantics in `materialize()` — `Object.assign` keeping stale keys when Gateway returns `{}` was a real bug
- Init-permissions for config dir in `k8s-spawner.ts` (chown 1000:1000, chmod 0700)
- Clean removal of `tryLoadConfig()` / `loadMcpServersConfig()` dead code
- Gateway `/api/internal/mcp-servers` correctly falls back to `loadConfig().mcpServers` as local seed

nightmeng (approved):
All three issues from the previous review have been addressed in the latest commit:

1. `agentbox-template.yaml` now has the `config` emptyDir volume + mount — consistent with `k8s-spawner.ts`
2. `materialize()` uses `writeConfig({ ...config, mcpServers })` — no more cached singleton mutation
3. `docs/step-remove-mcp-dir.md` removed

LGTM.

PR #69, opened 2026-03-10
feat: unified Credentials page with Kubernetes/SSH/API tabs

nightmeng:
## Summary

- Merge the separate Environments (admin-only) and Credentials (all users) pages into a unified **Credentials** page with three sub-tabs: **Kubernetes** | **SSH** | **API**
- `environment.list` RPC now open to all authenticated users (enriched with per-user kubeconfig status, filtered by testOnly)
- `environment.update` auto-invalidates mismatched kubeconfigs when apiServer changes, pushes updated credentials to affected AgentBoxes
- `environment.delete` pushes updated credentials to all affected users
- Auto-load most recent session when switching workspace (instead of showing icebreaker)
- Fix onboarding checklist: count kubeconfigs in credential detection (`system.status`)

## Test plan

- [ ] Non-admin user: open /credentials → sees Kubernetes tab with all environments → can upload/remove kubeconfig → status badge updates
- [ ] Admin user: sees "New Kubernetes" button + Edit/Delete actions in Kubernetes tab
- [ ] testOnly user: only sees test environments
- [ ] apiServer change: mismatched kubeconfigs auto-deleted, affected AgentBoxes updated
- [ ] Environment delete: all users' configs cleaned up, AgentBoxes updated
- [ ] SSH/API tabs show "Coming soon"
- [ ] Old /environments URL redirects to /credentials
- [ ] Workspace switch: auto-loads most recent session if sessions exist
- [ ] Onboarding checklist: "Add Credentials" marked done when kubeconfig uploaded


LikiosSedo (dismissed):
## P0 — Security (Must Fix)

### 1. apiServer matching logic is too loose — substring bypass possible

`rpc-methods.ts` — both `userEnvConfig.set` and `environment.update` use `String.includes()` for apiServer validation:

```typescript
servers.some((s) => s.includes(env.apiServer) || env.apiServer.includes(s))
```

This allows bypass:
- `apiServer = "10.0.1.1"` matches `"https://10.0.1.100:6443"` (1 is substring of 100)
- `apiServer = "api"` matches any URL containing "api"

**Fix**: Parse URLs and compare hostnames, or at minimum do exact host match after extracting hostname from both sides.

### 2. `materializeCredentials` has a race condition with concurrent prompt/reload

`http-server.ts` — the function clears the entire credentials directory then writes new files:

```typescript
for (const entry of fs.readdirSync(credDir)) {
  fs.rmSync(path.join(credDir, entry), { recursive: true });
}
```

If a prompt is reading credential files while a reload clears the directory, the prompt will fail or read incomplete state.

**Fix**: Write to a temp directory, then atomic rename (`fs.renameSync`). Or add a simple mutex since this is single-process.

---

## P1 — Correctness (Should Fix)

### 3. `environment.list` admin detection is hardcoded

```typescript
const isAdmin = context.auth?.username === "admin";
```

All other admin-gated methods use `requireAdmin(context)`. This hardcoded check will break for any admin user not named "admin".

### 4. SQLite migration creates `api_server DEFAULT ''` — existing environments get empty string

```sql
ALTER TABLE environments ADD COLUMN api_server TEXT NOT NULL DEFAULT '';
```

Existing environment records will have `api_server = ''`. The API then returns them with an empty apiServer, but `env-repo.save()` rejects empty strings. This creates inconsistent state for users upgrading from a pre-ADR-011 database.

**Fix**: After the ALTER TABLE, run `UPDATE environments SET api_server = name WHERE api_server = ''` as a data backfill, or skip records with empty apiServer in `environment.list`.

### 5. ADR-011 documents claim memory isolation per workspace — not implemented in this PR

`invariants.md` says:
> Each workspace has its own memory database (`<userDataDir>/<workspaceId>/.memory.db`)

But there's no code change in this PR to create per-workspace memory paths. If this is planned for a follow-up PR, the doc should say "Future" not present-tense.

### 6. `reload-credentials` route assumes `sessions[0]` is correct

```typescript
const kubeconfigRef = sessions.length > 0
  ? sessions[0].kubeconfigRef
  : { credentialsDir: undefined };
```

Currently safe because one AgentBox pod = one workspace, but this is a fragile assumption. Add a comment documenting this constraint, or pass workspaceId in the reload request.

---

## P2 — Code Quality (Nice to Fix)

### 7. `CredentialsSection` render-time setState violates React rules

```typescript
if (isConnected && !hasLoaded) {
    setHasLoaded(true);   // setState during render
    loadCredentials();
}
```

Use `useEffect` instead.

### 8. Dead code: `...(isAdmin ? {} : {})`

In `environment.list` — this spread has no effect. Remove or implement the intended admin-only fields.

### 9. `CredentialsSection.tsx` created but commented out in page

The file is 207 lines but `import { CredentialsSection }` is commented out. If SSH/API tabs are "coming soon", this file shouldn't ship yet.

### 10. `listFullForEnv` duplicates `listForEnv` in user-env-config-repo

The only difference is field projection vs `select()`. Consider a single method with an optional `includeKubeconfig` param, or document why two methods are needed.

### 11. `workspace.setEnvironments` silently skips validation when `envRepo` is null

If `envRepo` is null but `envIds.length > 0`, the binding constraint (test workspace → test envs only) is not enforced.

### 12. `testOnly` field referenced in `usePermissions` but `permission.mine` change not in diff

Confirm this field is returned from `permission.mine` RPC — the server-side change isn't visible in this PR.

LikiosSedo (changes requested):
## Review Summary

ADR-011 design is sound — credential scoping at workspace level is the right isolation strategy over per-tool guards or prompt-based approaches. The frontend consolidation is clean.

However, there are 2 security issues and several correctness issues that must be fixed before merge.

---

## P0 — Security (Must Fix)

### 1. apiServer matching logic is too loose — substring bypass possible

`rpc-methods.ts` — both `userEnvConfig.set` and `environment.update` use `String.includes()` for apiServer validation:

```typescript
servers.some((s) => s.includes(env.apiServer) || env.apiServer.includes(s))
```

This allows bypass:
- `apiServer = "10.0.1.1"` matches `"https://10.0.1.100:6443"` (1 is substring of 100)
- `apiServer = "api"` matches any URL containing "api"

**Fix**: Parse URLs and compare hostnames, or at minimum do exact host match after extracting hostname from both sides.

### 2. `materializeCredentials` has a race condition with concurrent prompt/reload

`http-server.ts` — the function clears the entire credentials directory then writes new files:

```typescript
for (const entry of fs.readdirSync(credDir)) {
  fs.rmSync(path.join(credDir, entry), { recursive: true });
}
```

If a prompt is reading credential files while a reload clears the directory, the prompt will fail or read incomplete state.

**Fix**: Write to a temp directory, then atomic rename (`fs.renameSync`). Or add a simple mutex since this is single-process.

---

## P1 — Correctness (Should Fix)

### 3. `environment.list` admin detection is hardcoded

```typescript
const isAdmin = context.auth?.username === "admin";
```

All other admin-gated methods use `requireAdmin(context)`. This hardcoded check will break for any admin user not named "admin".

### 4. SQLite migration creates `api_server DEFAULT ''` — existing environments get empty string

```sql
ALTER TABLE environments ADD COLUMN api_server TEXT NOT NULL DEFAULT '';
```

Existing environment records will have `api_server = ''`. The API then returns them with empty apiServer, but `env-repo.save()` rejects empty strings. This creates inconsistent state for users upgrading from a pre-ADR-011 database.

**Fix**: After the ALTER TABLE, run `UPDATE environments SET api_server = name WHERE api_server = ''` as a data backfill, or filter out records with empty apiServer in `environment.list`.

### 5. ADR-011 docs claim per-workspace memory isolation — not implemented in this PR

`invariants.md` states:
> Each workspace has its own memory database (`<userDataDir>/<workspaceId>/.memory.db`)

But there is no code change in this PR to create per-workspace memory paths. If this is planned for a follow-up PR, the doc should say "Planned" rather than present-tense.

### 6. `reload-credentials` route assumes `sessions[0]` is the correct session

```typescript
const kubeconfigRef = sessions.length > 0
  ? sessions[0].kubeconfigRef
  : { credentialsDir: undefined };
```

Currently safe because one AgentBox pod serves one workspace, but this is a fragile assumption. Either add a comment documenting this constraint, or pass workspaceId in the reload request for future-proofing.

---

## P2 — Code Quality (Nice to Fix)

### 7. `CredentialsSection` calls setState during render — violates React rules

```typescript
if (isConnected && !hasLoaded) {
    setHasLoaded(true);   // setState during render
    loadCredentials();
}
```

Use `useEffect` instead.

### 8. Dead code: `...(isAdmin ? {} : {})`

In `environment.list` — this spread has no effect. Remove or implement the intended admin-only fields.

### 9. `CredentialsSection.tsx` is created (207 lines) but commented out in the page

`import { CredentialsSection }` is commented out in `Environments/index.tsx`. If SSH/API tabs are "coming soon", this file should not ship in this PR.

### 10. `listFullForEnv` duplicates `listForEnv` in user-env-config-repo

The only difference is field projection vs `select()`. Consider consolidating into a single method with an optional `includeContent` param, or document why two methods are needed.

### 11. `workspace.setEnvironments` silently skips validation when `envRepo` is null

If `envRepo` is null but `envIds.length > 0`, the binding constraint (test workspace can only bind test environments) is not enforced.

### 12. `testOnly` field referenced in `usePermissions` but `permission.mine` server change not in diff

Confirm this field is actually returned from `permission.mine` RPC — the server-side change is not visible in this PR.

nightmeng (commented):
## PR Review: feat: unified Credentials page with Kubernetes/SSH/API tabs

ADR-011 design is solid — workspace-level credential scoping is the right approach. The atomic credential write, push-based updates, and testOnly enforcement are all well done. A few issues to address:

---

### P0 — Security

**P0-1: `materializeCredentials` has no path traversal validation on `file.name`**

`src/agentbox/http-server.ts` — the `reload-credentials` endpoint writes files from network payload:

```typescript
fs.writeFileSync(path.join(tmpDir, file.name), file.content, ...);
```

If `file.name` contains `../../`, it writes outside `tmpDir`. The Gateway side sanitizes (`safeName = env.name.replace(...)`), but defense-in-depth requires AgentBox validation too since it accepts arbitrary network input.

```typescript
const resolved = path.resolve(tmpDir, file.name);
if (!resolved.startsWith(tmpDir + path.sep)) throw new Error(`Invalid file name: ${file.name}`);
```

**P0-2: `apiServerHostMatch` fallback to string comparison is risky**

If `new URL()` parsing fails, the function falls back to exact string match (`kubeconfigServer === envApiServer`). A malformed `envApiServer` could bypass the hostname comparison. If URL parsing fails for either argument, should **return `false`** rather than fall back to string comparison.

**P0-3: `environment.list` exposes `apiServer` to all authenticated users**

Non-admin users don't need to see API server addresses. Consider only returning `apiServer` to admins, or at minimum only to users who have a kubeconfig for that environment.

---

### P1 — Correctness

**P1-1: DDL inconsistency — `api_server` default**

- `CREATE TABLE` (fresh DB): `api_server TEXT NOT NULL` — no default
- `ALTER TABLE` (migration): `api_server TEXT NOT NULL DEFAULT ''`
- Drizzle schema: `text("api_server").notNull()` — no default

These should be consistent. Either add `DEFAULT ''` to CREATE TABLE or remove it from ALTER TABLE.

**P1-2: `environment.update` apiServer validation logic is wrong**

```typescript
if (apiServer && typeof apiServer === "string" && !apiServer.trim()) {
    throw new Error("apiServer must be a non-empty string");
}
```

This can only catch whitespace-only strings. An empty string `""` is falsy and skips the check. Since `apiServer` falls back to `existing.apiServer` (which could be `""` from the backfill migration), an empty string passes through silently. Validate after fallback: `if (!apiServer?.trim()) throw new Error(...)`.

**P1-3: `workspace.update` allows `envType` change without re-validating bound environments**

If a workspace has bound prod environments and `envType` changes from `"prod"` to `"test"`, the existing bindings are NOT revalidated. The workspace would be `test` type but still have prod environments bound — violates invariant 12.4.

Either unbind non-test environments automatically or reject the change if non-test environments are bound.

---

### P2 — Code Quality

**P2-1: `CredentialsSection.tsx` is dead code** — 208-line component created but never imported. Remove or wire up.

**P2-2: `listFullForEnv` vs `listForEnv`** — Both appear to return the same data (no field filtering). Verify `listForEnv` actually omits kubeconfig content; if not, remove the duplicate.

**P2-3: Credential clear path not atomic** — The "clear credentials" branch in `reload-credentials` does manual `readdir` + `rmSync` instead of using the new atomic `materializeCredentials`. Should use the same pattern for consistency.

---

### Looks Good

- ADR-011 workspace-level credential scoping — avoids per-tool guards, provides hard data-layer isolation
- Atomic `materializeCredentials` with write-to-temp-then-rename
- Push-based credential updates — immediate effect, good UX and security
- testOnly enforcement layered at workspace create/update, environment binding, and kubeconfig upload
- apiServer anchoring for kubeconfig validation
- K8s workspace label for proper scoped push
- Invariants.md Section 12 well documented with "What NOT to Do"
- Clean frontend tab architecture with role-aware actions

LikiosSedo (commented):
I found three additional issues that do not appear to be covered by the existing review threads:

1. `envType` is not validated as an enum, which leaves a policy bypass for `testOnly` users. In `workspace.update`, the code only blocks `params.envType === "prod"` and then persists any other string (`src/gateway/rpc-methods.ts`). Downstream, the credential isolation logic only treats `envType === "test"` specially when deciding whether to suppress non-kubeconfig credentials. A `testOnly` user can therefore set `envType` to an arbitrary non-`prod` value such as `foo`, and the workspace will no longer be treated as test-scoped by `buildCredentialPayload`, so SSH/API credentials become eligible again. `envType` should be validated strictly as `prod | test`.

2. Workspace-level access changes do not trigger credential hot-reload for running AgentBoxes. This PR introduces `pushCredentialsToUser()` for real-time credential sync, and it is used from `credential.*`, `userEnvConfig.*`, and `environment.*`. However, `workspace.setCredentials` and `workspace.setEnvironments` only update the database and return success; they never push the updated credential payload to active boxes. If a workspace adds/removes SSH/API credentials or K8s environments while its AgentBox is already running, the current session keeps the old credential view until the box is recreated. That breaks the push-based consistency model added elsewhere in this PR.

3. The new auto-load most recent session on workspace switch behavior has a stale request race in `usePilot`. `loadHistory()` only guards state writes by `loadHistoryRequestIdRef` after a newer `loadHistory()` call has been issued. On workspace change, the effect clears the UI and fetches `session.list`, then calls `loadHistory(newSessions[0].key)` only if the target workspace has at least one session. If the target workspace has no sessions, an older in-flight `chat.history` request from the previous workspace can still resolve and repopulate the chat with messages from the wrong workspace. The workspace-switch path should invalidate outstanding history requests immediately, even when there is nothing to auto-load in the target workspace.

LikiosSedo (approved):

PR #70, opened 2026-03-10
feat(cron): robustness improvements — retry, locks, workspace, indexes

nightmeng:
## Summary

- **Database indexes**: Add `idx_cron_jobs_status_assigned` and `idx_cron_instances_heartbeat` for both SQLite and MySQL to eliminate full table scans during reconciliation
- **Retry with exponential backoff**: New `withRetry()` utility wraps all gateway-client HTTP calls (3 attempts, 500ms–5s) with HTTP-aware predicate (skip retries on 4xx except 429)
- **Idempotency guards**: Execution locks via `locked_by`/`locked_at` columns — executor acquires before running, releases in `finally`, coordinator clears stale locks (>6min) during reconcile
- **Workspace association**: Add `workspace_id` column to `cron_jobs`, thread through schema → repo → RPC → executor → agent-prompt → channel-bridge → frontend so jobs execute in their originating workspace's AgentBox
- **Cleanup**: Remove dead `selectedEnvId` prop chain from Pilot components (always null after workspace migration)

## Test plan

- [x] Create cron job in workspace A → verify it runs in workspace A's AgentBox (check `ws=` in executor logs)
- [x] Simulate concurrent cron executors → verify lock prevents double execution
- [x] Kill cron pod mid-execution → verify stale lock is cleared on next reconcile cycle
- [x] Verify retry behavior: stop gateway briefly, confirm cron client reconnects after backoff
- [x] Existing cron jobs without `workspace_id` still execute (null workspace = default behavior)

### DIFF
--- a/src/gateway/db/init-schema.ts
+++ b/src/gateway/db/init-schema.ts
@@ -432,6 +432,7 @@ export async function initSchema(db: Database): Promise<void> {
     `ALTER TABLE workspaces ADD COLUMN env_type VARCHAR(10) NOT NULL DEFAULT 'prod' AFTER is_default`,
     `ALTER TABLE environments ADD COLUMN api_server VARCHAR(512) NOT NULL DEFAULT '' AFTER is_test`,
     `ALTER TABLE cron_jobs ADD COLUMN workspace_id VARCHAR(64)`,
+    `ALTER TABLE cron_jobs ADD CONSTRAINT fk_cron_jobs_workspace FOREIGN KEY (workspace_id) REFERENCES workspaces(id)`,
   ];
   for (const stmt of MIGRATIONS) {
     try {
--- a/src/gateway/server.ts
+++ b/src/gateway/server.ts
@@ -929,7 +929,7 @@ export async function startGateway(opts: StartGatewayOptions): Promise<GatewaySe
           } catch (innerErr) {
             timeout.cancel();
             // On timeout, attempt to abort + close the orphaned agent session
-            if (client && sessionId && innerErr instanceof Error && innerErr.message.includes("timed out")) {
+            if (client && sessionId && innerErr instanceof ExecutionTimeoutError) {
               try { await client.abortSession(sessionId); } catch { /* best-effort */ }
               try { await client.closeSession(sessionId); } catch { /* best-effort */ }
               console.log(`[gateway] agent-prompt session=${sessionId} aborted after timeout`);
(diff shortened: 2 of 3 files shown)


nightmeng (commented):
## Review: feat(cron): robustness improvements — retry, locks, workspace, indexes

Large PR (32 files, +1221 -307) but well-structured. Retry, execution locks, workspace threading, and cron run history are solid additions. Security model is preserved — mTLS identity for cron-list is actually more secure than the old query param approach. No P0 issues.

---

### P1 — Should Fix

**1. `findJobByName` lost direct ID-based lookup in channel-bridge**

The old code had a fast path:
```typescript
const job = parsed.id
  ? await configRepo.getCronJobById(parsed.id)
  : await findJobByName(...);
```

Now it passes `parsed.name || parsed.id` into `findJobByName`, which tries to match a UUID against job **names** via fuzzy matching — this will always fail silently. The direct `getCronJobById` fast path should be restored. This affects all action handlers (`update`, `delete`, `pause`, `resume`, `rename`) in `executeScheduleAction`.

**2. `findJobByName` reverse-includes is too greedy**

```typescript
j.name.toLowerCase().includes(lower) || lower.includes(j.name.toLowerCase())
```

The second branch (`lower.includes(j.name.toLowerCase())`) means a query "check-cpu-prod-staging" would match a job named "check-cpu" — any query that *contains* a short job name matches it. Consider removing the reverse-includes or requiring a minimum match ratio.

**3. Heartbeat/reconcile interval reduction (30s→10s / 60s→15s) not documented**

3-4x increase in DB write frequency. Fine for single-node but could be noticeable with many cron workers. Document the rationale or make intervals configurable via env vars.

---

### P2 — Non-blocking

**4. `envId: null` remnant in `cron.rename` notification payload** — dead code, other handlers don't set it.

**5. `ScheduleCard` `resolveSchedule` fetches ALL jobs without workspace filter** — should pass `workspaceId` to `cron.list` to avoid cross-workspace false matches.

**6. `cron.runs` lacks pagination offset** — only `limit` (max 100), no cursor/offset for older runs.

**7. Workspace name N+1 query in `cron.list` RPC** — loops `workspaceRepo.getById(wsId)` per workspace. Consider batch lookup.

**8. FK mismatch** — Drizzle schemas have `.references(() => workspaces.id)` on `cronJobs.workspaceId`, but raw DDL has no FK constraint. Cosmetic but inconsistent.

**9. `data.envId` still in agent-prompt request type** — unused dead code, remove in follow-up.

---

### Looks Good

- **Retry utility** (`src/shared/retry.ts`): Clean API with `HttpError` class, `shouldRetryHttp` predicate, exponential backoff + jitter. Lock operations correctly bypass retry via `postDirect`.
- **Execution locks**: `lockJobForExecution` / `unlockJob` / `clearStaleLocks` with 6min threshold (> 5min execution timeout) — sound design.
- **mTLS cron-list identity**: Switching from query param `userId` to cert identity eliminates spoofing risk.
- **`waitForAgentCompletion` text_delta capture**: Correct fix for claude-sdk brain which sends empty content in `message_end`.
- **Cancellable timeout**: `rejectAfterTimeout` with `timer.unref()` prevents timer leaks and process hang.
- **Platform tools exemption**: `manage_schedule` and `credential_list` always available regardless of `allowedTools`.
- **`syncAssignedJobs`**: Fixes the gap where reassigned jobs were never loaded into the local scheduler.
- **`cron_job_runs` DDL parity**: Consistent across init-schema.ts, migrate-sqlite.ts, and both Drizzle schemas.
- **Notification UX**: cron_success/cron_failure split with detail modal is a solid improvement.
- **CLAUDE.md system prompt protection rule**: Correctly scoped to `src/core/prompt.ts`.

PR #72, opened 2026-03-11
feat(k8s): remove init container + add PVC persistence for user memory

nightmeng:
## Summary

- **Remove init container**: The `init-permissions` container duplicated the entrypoint's `chown`/`chmod` logic. Add `CAP_CHOWN` + `CAP_FOWNER` to the main container so the entrypoint succeeds in K8s (capabilities drop after `runuser`, no security impact). Eliminates one container + one image pull per pod startup.
- **Add PVC persistence**: New `agentbox.persistence` Helm toggle — when enabled, each user gets a `ReadWriteMany` PVC with workspace-level `subPath` isolation for memory DB, investigations, and session data. Disabled by default (backward-compatible).
- **Add `agentbox.image.repository`**: Helm values now support custom image names (e.g. `siclaw-agentbox-debug` for Node debug images).
- **Remove dead `kubeconfigBase64` code**: Old Secret-mount path for kubeconfig — replaced by Credentials API. Removed from `k8s-spawner.ts` and `types.ts`.
- **Fix cert Secret lifecycle**: `stop()` was deleting a non-existent `*-kubeconfig` Secret — fixed to delete `*-cert`. Added labels to cert Secrets so `cleanup()` can find them via `labelSelector`.

## Helm values

```yaml
agentbox:
  image:
    repository: ""      # defaults to "siclaw-agentbox"
  persistence:
    enabled: false          # toggle for user memory persistence
    storageClass: ""        # required when enabled (e.g. "nfs-client")
    accessMode: ReadWriteMany
    size: "1Gi"             # per-user PVC size
```

## PVC layout

```
PVC: siclaw-user-{userId}  (one per user, ReadWriteMany)
├── {workspaceId-1}/       ← subPath per workspace
│   ├── memory/            ← Memory DB + investigations
│   └── agent/             ← session JSONL
├── {workspaceId-2}/
│   ├── memory/
│   └── agent/
```

## Test plan

- [ ] `helm template` renders correctly with persistence disabled (no env vars)
- [ ] `helm template --set agentbox.persistence.enabled=true --set agentbox.persistence.storageClass=nfs` renders persistence env vars
- [ ] Pod spec has no `initContainers`, main container has `CHOWN`+`FOWNER` capabilities
- [ ] `stop()` deletes the `*-cert` Secret (not `*-kubeconfig`)
- [ ] Cert Secrets have `siclaw.io/app: agentbox` label for cleanup

### DIFF
--- a/Dockerfile.agentbox
+++ b/Dockerfile.agentbox
@@ -112,6 +112,6 @@ RUN chmod +x /usr/local/bin/agentbox-entrypoint.sh
 
 # Container starts as root — entrypoint fixes volume permissions then drops to agentbox.
 # The agentbox user uses sudo to run child processes as sandbox.
-# K8s securityContext drops all capabilities except SETUID/SETGID.
+# K8s securityContext drops all capabilities except SETUID/SETGID/CHOWN/FOWNER.
 ENTRYPOINT ["/usr/local/bin/agentbox-entrypoint.sh"]
 CMD ["node", "dist/agentbox-main.js"]
--- a/docs/design/security.md
+++ b/docs/design/security.md
@@ -342,7 +342,7 @@ spec:
     securityContext:
       capabilities:
         drop: ["ALL"]
-        add: ["SETUID", "SETGID"]  # Required for sudo -u sandbox
+        add: ["SETUID", "SETGID", "CHOWN", "FOWNER"]  # SETUID/SETGID for sudo; CHOWN/FOWNER for entrypoint volume permissions
       readOnlyRootFilesystem: true
     volumeMounts:
     - name: tmp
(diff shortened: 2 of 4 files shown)


nightmeng (commented):
## Review: feat(k8s): remove init container + add PVC persistence for user memory

Clean PR — init container removal, dead kubeconfigBase64 cleanup, cert Secret lifecycle fix, and PVC persistence toggle are all well done. Updated feedback after discussion:

---

### P0

**1. `storageClass: ""` creates unschedulable PVC when persistence is enabled**

In Kubernetes, `storageClassName: ""` means "no StorageClass" (not cluster default). If a user enables `persistence` but forgets to set `storageClass`, the PVC stays Pending forever and the pod fails to start — with no clear error message.

**Recommendation** (three-part fix):

**(a) Spawner**: pass `undefined` instead of empty string so Kubernetes uses the cluster default:
```typescript
storageClassName: this.config.persistence!.storageClass || undefined,
```

**(b) Spawner log warning**: when storageClass is empty, log a warning at startup:
```typescript
if (this.config.persistence?.enabled && !this.config.persistence.storageClass) {
  console.warn("[k8s-spawner] persistence.storageClass is empty — using cluster default StorageClass");
}
```

**(c) Helm NOTES.txt**: create `helm/siclaw/templates/NOTES.txt` with a warning printed after `helm install/upgrade`:
```
{{- if and .Values.agentbox.persistence.enabled (not .Values.agentbox.persistence.storageClass) }}
⚠️  WARNING: agentbox.persistence is enabled but storageClass is empty.
   The cluster default StorageClass will be used. Set agentbox.persistence.storageClass
   explicitly for production deployments.
{{- end }}
```

---

### P1

**2. `subPath: workspaceId` has no path traversal sanitization**

```typescript
subPath: this.config.persistence?.enabled ? workspaceId : `user/${userId}/agent-data`,
```

`workspaceId` is a UUID from the database so `..` is unlikely, but defense-in-depth applies — this is a volume mount subPath sourced from user-associated data. Apply the same sanitization pattern used for `userPvcName`:

```typescript
subPath: workspaceId.replace(/[^a-zA-Z0-9-]/g, "_")
```

**3. Dockerfile and security.md comments reference old capability set**

`Dockerfile.agentbox:115` says "drops all capabilities except SETUID/SETGID" — now also includes CHOWN/FOWNER. Update the comment to reflect the actual set. Same for any references in `docs/design/security.md`.

---

### P2

**4. Pre-existing cert Secrets without labels won't be found by `cleanup()`**

The new `cleanup()` uses `labelSelector: siclaw.io/app=agentbox` to find cert Secrets, but Secrets created before this PR have no labels. They'll be orphaned after upgrade. Not a blocker — just worth a note in release documentation for operators doing upgrades.

---

### Looks Good

- Init container removal with CAP_CHOWN/FOWNER tradeoff is acceptable — kernel clears effective/permitted caps on UID transition via `runuser`, and SUID bits are stripped
- Dead `kubeconfigBase64` code removal — clean deletion
- cert Secret lifecycle fix (`*-cert` instead of `*-kubeconfig`) + labels for cleanup
- PVC per-user with workspace subPath isolation is the right model
- `ensureUserPvc` is idempotent (409 Conflict handled)
- `agentboxImage` helper with custom repository support
- Backward compatible — `persistence.enabled: false` by default

PR #73, opened 2026-03-11
feat(db): add scheduled cleanup for sessions, messages, and session_stats

chent1996:
## Summary
- Add 3 batch-delete methods to `ChatRepository`: soft-delete inactive sessions (180d), hard-delete expired sessions (30d after soft-delete, messages cascade), purge old `session_stats` (90d)
- Add `POST /api/internal/sessions/purge` endpoint in Gateway, mirroring the existing `notifications/purge` pattern
- Add 24h cron job (`purgeOldSessions`) in `cron-main.ts` with startup-immediate execution
- Fix `messages` FK in SQLite schema to include `onDelete: "cascade"` (DDL already had it, Drizzle schema was missing it)

## Test plan
- [ ] `curl -X POST http://localhost:3000/api/internal/sessions/purge` returns `{ status: "ok", softDeleted, statsPurged, sessionsPurged }`
- [ ] Verify batch deletion works with >500 rows
- [ ] Verify messages are cascade-deleted when sessions are hard-deleted
- [ ] Verify cron logs show purge results on startup

### DIFF
--- a/src/cron-main.ts
+++ b/src/cron-main.ts
@@ -109,6 +109,8 @@ sessionPurgeTimer.unref();
 // 8. Graceful shutdown — release jobs immediately so other instances can claim them
 async function shutdown() {
   console.log("\n[cron] Shutting down...");
+  clearInterval(purgeTimer);
+  clearInterval(sessionPurgeTimer);
   scheduler.stop();
   await coordinator.shutdown();
   apiServer.close();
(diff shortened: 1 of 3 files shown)


nightmeng (approved):
All three issues from the previous review are addressed in a4dc7ace:

1. ✅ **Cutoff unit fixed** — `purgeOldSessionStats` now uses `Date.now() - retentionDays * 86400_000` (milliseconds). Comment updated to reflect the correct unit.
2. ✅ **Input validation** — Retention days `< 1` rejected with 400 response, preventing accidental mass deletion.
3. ✅ **Timer cleanup** — Both `purgeTimer` and `sessionPurgeTimer` cleared in `shutdown()`.

LGTM. Clean implementation that follows established patterns.

PR #75, opened 2026-03-12
fix: improve new user session UX — language, credentials, profile

LikiosSedo:
## Problem

Three issues affecting the first-session experience for new users:

1. **Stale credentials after DB reset**: `materializeCredentials()` only ran when `files.length > 0`, so empty credential payloads left old kubeconfig files on disk. A user who reset their DB would still see cluster environments from prior sessions.

2. **Unreliable response language**: Language matching relied entirely on model compliance with system prompt instructions. Models with strong language bias (e.g. Kimi-K2.5) would ignore "respond in the same language" and default to Chinese regardless of user input.

3. **PROFILE.md never created**: The agent was instructed to write PROFILE.md during onboarding, but model compliance was unreliable — the file was never created, leaving language preference and user context permanently unset.

## Solution

1. **Credential cleanup**: Always call `materializeCredentials()` when a credentials payload is present (even with empty files array), so stale files are cleared.

2. **Deterministic language detection**: Added `src/shared/detect-language.ts` — Unicode script analysis that detects Japanese (kana), Korean (hangul), Chinese (CJK), Russian (cyrillic), etc. Injected per-message as `[System: respond in X]` prefix in the prompt handler. Also programmatically updates PROFILE.md's Language field (code-level, not model-dependent).

3. **Skeleton PROFILE.md**: Code creates a skeleton PROFILE.md (all fields TBD, Language=English) at memoryDir initialization. `hasProfile` check updated to look for non-TBD Name field, so the first-time onboarding UX still triggers correctly.

## Design rationale (language)

The old approach was fragile — a single system prompt instruction competing against model training bias. The new approach has three layers:
- **Code layer**: `detectLanguage()` — deterministic, zero model dependency
- **Message layer**: `[System: respond in X]` prefix — model only needs to "obey" not "detect"
- **Persistence layer**: PROFILE.md Language auto-updated — subsequent sessions inherit preference

First message defaults to English (reasonable when user language is unknown). Once the user types in any language, detection kicks in immediately.

## Test plan

- [x] Type check passes (`tsc --noEmit`)
- [x] Language detection verified: English, Japanese, Chinese, Korean, mixed (kubectl + CJK)
- [x] Manual test: new user flow with clean DB — English greeting → Japanese switch → Chinese switch all work
- [x] PROFILE.md Language field auto-updates on each message
- [ ] Verify credential cleanup: reset DB, confirm no stale kubeconfig files appear

### DIFF
--- a/src/agentbox/http-server.ts
+++ b/src/agentbox/http-server.ts
@@ -328,24 +328,28 @@ export function createHttpServer(sessionManager: AgentBoxSessionManager): http.S
       promptText = `[System: respond in ${detectedLang}]\n${promptText}`;
     }
 
-    // Programmatically update PROFILE.md Language field (code-level, not model-dependent)
-    try {
-      const cfg = loadConfig();
-      const userDataDir = process.env.SICLAW_USER_DATA_DIR || cfg.paths.userDataDir;
-      const profilePath = path.resolve(userDataDir, "memory", "PROFILE.md");
-      if (fs.existsSync(profilePath)) {
-        const content = fs.readFileSync(profilePath, "utf-8");
-        const currentLangMatch = content.match(/\*\*Language\*\*:\s*(.+)/i);
-        const currentLang = currentLangMatch?.[1]?.trim();
-        if (currentLang !== detectedLang) {
-          const updated = content.replace(
-            /(\*\*Language\*\*:\s*).+/i,
-            `$1${detectedLang}`,
-          );
-          fs.writeFileSync(profilePath, updated);
+    // Programmatically update PROFILE.md Language field (code-level, not model-dependent).
+    // Only update on non-English detection to avoid flapping: English is the default,
+    // so we only persist when the user actively uses another language.
+    if (detectedLang !== "English") {
+      try {
+        const cfg = loadConfig();
+        const userDataDir = process.env.SICLAW_USER_DATA_DIR || cfg.paths.userDataDir;
+        const profilePath = path.resolve(userDataDir, "memory", "PROFILE.md");
+        if (fs.existsSync(profilePath)) {
+          const content = fs.readFileSync(profilePath, "utf-8");
+          const currentLangMatch = content.match(/\*\*Language\*\*:\s*(.+)/i);
+          const currentLang = currentLangMatch?.[1]?.trim();
+          if (currentLang !== detectedLang) {
+            const updated = content.replace(
+              /(\*\*Language\*\*:\s*).+/i,
+              `$1${detectedLang}`,
+            );
+            fs.writeFileSync(profilePath, updated);
+          }
         }
-      }
-    } catch { /* best-effort, don't block prompt */ }
+      } catch { /* best-effort, don't block prompt */ }
+    }
 
     // Execute prompt asynchronously; notify SSE to close on completion
     console.log(`[agentbox-http] Starting prompt for session ${managed.id} [lang=${detectedLang}]`);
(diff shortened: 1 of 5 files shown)


nightmeng (dismissed):
## Review Summary

The problem framing is solid and the 3-layer language detection approach (code → message → persistence) is well-reasoned. However, there are two significant gaps that need fixing before merge.

---

## Critical

### 1. TUI / Local mode regression — language matching broken

`src/core/prompt.ts` (line 96) changes the Language section from:
> "Always respond in the same language the user writes in."

to:
> "Default to English. When a message starts with `[System: respond in X]`..."

But `[System: respond in X]` is only injected in `http-server.ts` (K8s agentbox path). Neither `cli-main.ts` (TUI) nor `local-spawner.ts` inject it. After this PR, TUI users will **always get English responses** regardless of what language they type in — a clear regression.

Fix options:
- (a) Add `detectLanguage` + prefix injection to the local-spawner and CLI prompt paths as well
- (b) Keep the old "match user language" instruction as a fallback, and make `[System: respond in X]` an override: e.g. *"Respond in the same language the user writes in. If a message starts with `[System: respond in X]`, always use language X instead."*

Option (b) is simpler and more resilient.

**Also note: `src/core/prompt.ts` requires explicit human confirmation before modification** per CLAUDE.md's System Prompt Protection rule.

### 2. SDK brain: no skeleton PROFILE.md, lost `else` branch

The skeleton PROFILE.md creation (`agent-factory.ts` ~line 648) is inside the **pi-agent path only** — the SDK brain path returns early at line 533. Meanwhile, `buildAppendSystemPrompt` has removed its `else` branch (the old "No PROFILE.md found" first-session onboarding prompt). So for SDK brain first-time users:
- No skeleton created → `fs.existsSync(profileFile)` is false → no onboarding prompt, no profile info, no Language field to update
- The code in `http-server.ts` that updates PROFILE.md Language silently fails (`existsSync` check)

Fix: move skeleton creation to **before** the brain type fork (e.g. right after `memoryDir` is defined at line 324), or at minimum add it to the SDK path as well.

---

## Important

### 3. Single-character detection threshold

`detectLanguage` triggers on a **single** non-Latin character. A message like `"Check the 生产 namespace"` (one Chinese word in an English sentence) → detected as Chinese → PROFILE.md updated → all subsequent session responses in Chinese. This is aggressive for an SRE tool where users routinely mix languages with technical terms.

Consider requiring a minimum count or ratio, e.g.:
```ts
if (cjk > 0 && cjk >= cleaned.length * 0.1) return "Chinese";
```
Or at minimum, only update PROFILE.md when the detected language differs **and** the signal is strong (multi-character).

### 4. Language flapping on PROFILE.md

Every single prompt rewrites PROFILE.md's Language field to whatever `detectLanguage` returns for that message. A user sending alternating English/Chinese messages causes the file to be rewritten on every prompt. The Language field semantically becomes "language of last message" rather than "user's preferred language."

Consider: only update when the detected language is the same for N consecutive messages, or only update when it differs from the initial default (English).

### 5. Missing unit tests for `detect-language.ts`

This is a pure function with well-defined inputs/outputs — ideal for vitest. The PR description mentions manual testing but there are no automated tests. Edge cases worth covering:
- Mixed scripts (`kubectl get pods 在生产环境`)
- Pure CLI (`kubectl get pods --namespace kube-system`) → should be English
- Code blocks with CJK comments
- Empty/whitespace input

---

## Minor

**Cyrillic → "Russian"**: Cyrillic also covers Ukrainian, Bulgarian, Serbian. For "respond in X" this is mostly fine, but consider a code comment acknowledging the approximation.

---

## What looks good

- Credential cleanup fix (`if (body.credentials)` instead of `if (body.credentials?.files?.length)`) — correct and minimal
- Skeleton PROFILE.md approach — much more reliable than depending on model compliance for file creation
- `hasProfile` check in `rpc-methods.ts` updated to look for non-TBD Name — correctly preserves onboarding UX
- The `try/catch` around PROFILE.md language update — appropriate "best-effort" approach

PR #76, opened 2026-03-12
fix: improve skill tool descriptions and remove Skill Scripts Reference

nightmeng:
## Summary

- **Remove "Skill Scripts Reference" from append system prompt** — this section eagerly listed all skills with their scripts permanently in the context window. Skills are now fully managed by pi-agent's lazy-loading lifecycle (`<available_skills>` index + on-demand SKILL.md read), so this was redundant context overhead.

- **Clarify skill directory is read-only** — the agent was attempting to `write`/`edit` skill files directly instead of using `update_skill`. Tool descriptions now state upfront that skill directories are read-only and all modifications must go through skill management tools.

- **Add environment/approval workflow docs** to `create_skill`, `update_skill`, and `fork_skill` — explains dev/test (immediate) vs production (needs admin approval) behavior.

- **Add duplicate/overlap check guidance** to `create_skill` — agent must check `<available_skills>` before creating, and recommend `fork_skill` when an existing skill covers the same functionality.

- **Fix stale references** — `[Editing Skill: ...]` → `[Skill: <name>]` in `update_skill`, remove hardcoded `/mnt/skills/` path, replace all "Skill Scripts Reference" mentions with `<available_skills>` index.

## Test plan

- [x] Built and deployed to test cluster (`env-isolation` tag)
- [x] Gateway and cron pods running healthy
- [ ] Verify agent uses `update_skill` tool instead of writing files directly when asked to modify a skill
- [ ] Verify agent checks `<available_skills>` before creating duplicate skills

### DIFF
--- a/skills/platform/create-skill/SKILL.md
+++ b/skills/platform/create-skill/SKILL.md
@@ -15,6 +15,32 @@ Use this skill when the user asks you to:
 - Create a new skill for a specific operational procedure
 - Save a diagnosis workflow as a skill
 
+## Duplicate / Overlap Check — Do This FIRST
+
+**Before creating any skill, check whether an existing skill already covers the same functionality.** Consult the `<available_skills>` index in your context.
+
+- **Functional overlap found**: If an existing builtin, team, or personal skill solves the same problem (even with a different name), DO NOT silently create a new one. Instead:
+  1. Tell the user which existing skill overlaps and what it does.
+  2. Ask if they want to: (a) use the existing skill as-is, (b) fork it with `fork_skill` to make a customized personal copy, or (c) still create a brand-new separate skill.
+  3. Only proceed with `create_skill` if the user explicitly chooses option (c).
+- **Why this matters**: Duplicate skills with similar functionality confuse the model — it cannot reliably choose between two skills that do the same thing. One well-maintained skill is always better than two overlapping ones.
+- To fork a builtin or team skill into a personal copy, use `fork_skill`.
+
+## Environments and Approval Workflow
+
+Skills go through a review workflow that behaves differently per environment:
+
+| Environment | Behavior |
+|-------------|----------|
+| **Dev / Test** | Newly created skills (draft status) are immediately visible and usable. You can test them right away. |
+| **Production** | Only **approved** skill versions are visible and usable. Draft and pending skills do NOT appear. |
+
+- After creating a skill, it starts in **draft** status.
+- Skills with scripts must be **submitted for review** and **approved by an admin** before they become active in production.
+- Skills without scripts (pure guidance) also start as draft but can be submitted and approved more quickly.
+- **After creating a skill in production context**: inform the user that it is pending review and will not be available in production until approved. Suggest testing in the dev/test environment first.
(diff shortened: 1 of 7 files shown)


nightmeng:
## Self-Review

Good cleanup — removing "Skill Scripts Reference" from the permanent context window is the right direction, and the tool description improvements (duplicate check, environment/approval workflow, read-only filesystem) are solid. One significant gap to address before merge.

---

## Critical: SDK brain loses all skill discovery

The removed "Skill Scripts Reference" section in `buildAppendSystemPrompt` was shared by **both** pi-agent and SDK brain (`buildAppendSystemPrompt` is called for both at agent-factory.ts lines 436 and 476). After this PR:

- **Pi-agent**: has `<available_skills>` via `DefaultResourceLoader` (upstream package) — skill discovery works
- **SDK brain**: has **no** skill listing mechanism at all

The tool descriptions now reference `<available_skills>` (create-skill, update-skill, fork-skill), but this tag does not exist in SDK brain's context. Confirmed by searching: `available_skills` appears only in the files modified by this PR, not in `claude-sdk-brain.ts` or anywhere in `src/core/`.

Impact for SDK brain sessions:
- Duplicate check in `create_skill` is useless — the model can't see existing skills to compare against
- `run_skill` can't be used proactively — the model doesn't know what skills are available
- `fork_skill` guidance ("check `<available_skills>`") references something that doesn't exist

**Fix options:**
- (a) Build a minimal skill index for SDK brain in `buildAppendSystemPrompt` — e.g. just skill names + one-line descriptions (much lighter than the removed "Skill Scripts Reference")
- (b) Move SDK brain to use `DefaultResourceLoader` (larger refactor, probably not for this PR)
- (c) Keep a lightweight listing for SDK brain only, gated on a flag

Option (a) is the simplest — a compact `## Available Skills\n- skill-name: description` block would be sufficient.

---

## Minor

**1. Duplicate guidance in two places**: The "Duplicate / Overlap Check" section appears in both the `create_skill` tool description (`src/tools/create-skill.ts`) and the SKILL.md (`skills/platform/create-skill/SKILL.md`). Both are loaded into context — tool description permanently, SKILL.md on demand. This is intentional (belt-and-suspenders) but worth a comment to avoid future confusion about which is the source of truth.

**2. Merge conflict with PR #75**: Both PRs modify `buildAppendSystemPrompt` in `agent-factory.ts`. PR #75 changes the function signature (removes `skillsBase`/`getUserSkillDirName`/`getPlatformSkillDirName` params), and this PR does the same removal plus the skill listing deletion. Whichever merges second will need a conflict resolution pass.

---

## What looks good

- Removing 100 lines of eager skill enumeration from the context window — correct architectural direction
- Tool descriptions now clearly state skill directories are **read-only** — prevents the model from attempting direct file writes
- Stale reference fixes: `[Editing Skill: ...]` → `[Skill: <name>]`, `/mnt/skills/` removed, "Skill Scripts Reference" → `<available_skills>`
- Environment/approval workflow docs — important for production users
- `labels` parameter already exists in tool schemas (create-skill.ts:154, fork-skill.ts:73) — SKILL.md examples are consistent

PR #87, opened 2026-03-13
fix: script-resolver builtin fallback should search skills/extension/

nightmeng:
## Problem

`script-resolver.ts` builtin fallback paths only checked `skills/core/`, missing `skills/extension/` entirely. This meant extension skills baked into the Docker image were invisible to `run_skill`, `pod_script`, `node_script`, and `netns_script` in K8s mode.

PR #85 fixed agent-factory and file-writer, but the execution path was missed.

## Solution

- Extract `builtinDirs()` helper that returns both `skills/core/` and `skills/extension/`
- Update all four affected functions to use it:
  - `getSkillScriptDirs()` — script lookup fallback
  - `getSkillBaseDirs()` — skill enumeration fallback (both legacy and scope layouts)
  - `skillExistsAsBuiltin()` — existence check
  - `listAllSkillsWithScripts()` — disabled-builtin filtering

## Test plan

- [x] All 864 tests pass
- [x] `tsc --noEmit` clean
- [ ] Verify `run_skill` can execute extension skills in K8s mode

### DIFF
--- a/src/tools/script-resolver.ts
+++ b/src/tools/script-resolver.ts
@@ -7,9 +7,15 @@ function skillsBase(): string {
   return path.resolve(process.cwd(), config.paths.skillsDir);
 }
 
-/** Builtin skills directory (baked into Docker image at skills/core/) */
+/** Builtin skills directories (baked into Docker image at skills/core/ and skills/extension/) */
+const BUILTIN_TIERS = ["core", "extension"] as const;
+
 function builtinCoreDir(): string {
   return path.resolve(process.cwd(), "skills", "core");
+}
+
+function builtinDirs(): string[] {
+  return BUILTIN_TIERS.map(t => path.resolve(process.cwd(), "skills", t));
 }
 
 /** Load disabled builtins list (written by agentbox startup from bundle API) */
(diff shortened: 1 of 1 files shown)


src/tools/script-resolver.ts
```diff
@@ -7,11 +7,17 @@ function skillsBase(): string {
   return path.resolve(process.cwd(), config.paths.skillsDir);
 }
 
-/** Builtin skills directory (baked into Docker image at skills/core/) */
+/** Builtin skills directories (baked into Docker image at skills/core/ and skills/extension/) */
+const BUILTIN_TIERS = ["core", "extension"] as const;
```
LikiosSedo: **Bug: priority order is inverted relative to `SKILL_SCOPES`.**

```ts
const BUILTIN_TIERS = ["core", "extension"] as const;
```

This makes `skills/core/` resolve before `skills/extension/` in the builtin fallback. But the existing bundle search order is the opposite:

```ts
const SKILL_SCOPES = ["user", "extension", "team", "core"];
//                                ↑ extension beats core
```

`skills/extension/` exists precisely to let overlay/inner projects override core skills baked into the base image. If the same skill exists in both directories, the bundle path (step 2) returns the `extension` version while the builtin fallback (step 3) returns the `core` version — inconsistent behavior depending on whether the bundle is present.

Fix:
```ts
const BUILTIN_TIERS = ["extension", "core"] as const;
```

LikiosSedo (changes requested):
## Review Summary

Good fix — this correctly completes what PR #85 started. The four affected functions in `script-resolver.ts` are all updated consistently, and the `builtinSet.has(base)` change in `listAllSkillsWithScripts` is cleaner than the old `=== coreDir` pointer comparison.

One design bug needs to be fixed before merge (see inline). One piece of dead code should be cleaned up.

**Request changes on:**
1. Priority order in `BUILTIN_TIERS` — inverted relative to `SKILL_SCOPES`
2. `builtinCoreDir()` is now unreferenced — dead code

**No blockers on:**
- Logic correctness of all four updated functions
- `disabled-builtins` check applies correctly across both tiers
- Deduplication in `getSkillBaseDirs()` via `dirs.includes(bDir)`

PR #88, opened 2026-03-13
fix: add skill draft mode for TUI, prevent silent skill creation

LikiosSedo:
## Problem

In TUI mode, skill management tools (`create_skill`, `update_skill`, `fork_skill`) are intentionally disabled (web-only). However, when asked to create a skill, the agent would silently fall back to writing raw files via the filesystem — partially succeeding (writing to `.siclaw/user-data/`) and giving misleading instructions, without ever telling the user the skill was inactive.

## Solution

Make the `Skill management` prompt rule mode-aware:

- **Web mode**: behavior unchanged — `create_skill` / `update_skill` / `fork_skill` as before
- **TUI / channel mode**: agent may draft a skill to `.siclaw/user-data/skill-drafts/<name>/` (SKILL.md + scripts/), but **must** explicitly communicate:
  1. The draft is NOT active and cannot be run via `run_skill`
  2. The user must manually review and copy it to `skills/core/` or `skills/team/` to activate
  3. Never claim the skill was "created" or "installed"
  4. For full skill management, use the Web UI

This preserves the security boundary (user is the final decision-maker for activation) while still allowing the agent to help author skill templates in TUI mode.


nightmeng (changes requested):
Good intent — the tools are already gated in `agent-factory.ts` (lines 271-277), but the prompt was still telling TUI users about `create_skill`/`update_skill` which don't exist in that mode. Adding prompt-level guidance for drafting is a reasonable UX improvement. Two issues:

---

### 1. Inaccurate activation path in prompt instruction

The prompt tells the agent:
> the user must manually review and copy it to `skills/core/` or `skills/team/` to activate it

Neither path is correct for user-created skills:
- **`skills/core/`** is the builtin directory baked into the Docker image / repo checkout. User drafts shouldn't go there — it blurs the core/user boundary.
- **`skills/team/`** doesn't exist. Team skills from the bundle API live under `.siclaw/skills/team/`, and in local mode per-user skills go to `.siclaw/skills/user/{userId}/`.

Looking at `agent-factory.ts` lines 426-429, the scanned `dynamicSkillBase` is:
- Local mode: `.siclaw/skills/user/{userId}/`
- K8s mode: `.siclaw/skills/` (flat)

The prompt instruction should reference the actual discoverable paths, or be generic (e.g., "copy to a skills directory scanned by the agent") and let the agent resolve the path at runtime.

### 2. Readability: 300+ char ternary in a template literal

The skill management bullet is now a single line with a ternary embedding two multi-sentence paragraphs. This is hard to read, review, and maintain. Consider extracting the mode-specific text:

```ts
const skillMgmtRule = mode === "web"
  ? `When the user asks to modify ... will be deleted.`
  : `Skill creation tools ... use the Web UI.`;

// then in the template:
- **Skill management**: ${skillMgmtRule}
```

This also makes it easier to test and diff in future PRs.

---

Everything else looks good — the security model is intact (tools gated at registration, prompt is defense-in-depth), the `.siclaw/user-data/skill-drafts/` path is within `writeAllowedDirs`, and the "never claim created/installed" guardrail is appropriate.

nightmeng (approved):
Previous issues resolved:

1. **Activation paths** ✅ — Now correctly references `.siclaw/skills/user/<userId>/` (local) and `.siclaw/skills/` (K8s), with explicit "never to `skills/core/`" guard.
2. **Readability** ✅ — The IIFE-in-template-literal is unconventional (I'd still prefer an extracted variable above the template), but the logic is clear and each mode's text is on separate lines now. Not blocking.

The `deep-investigation.ts` fix (`/^- /` and `/^\* /` instead of `startsWith`) correctly avoids the bold-title false positive. `/^\s/.test(line)` now matches HypothesesCard.tsx's threshold.

---

**Note: scope overlap with PR #89.** This PR now modifies `deep-investigation.ts` (the `isSubItem` line), which is also rewritten by PR #89's broader Strategy 4 refactor. Whichever merges second will hit a conflict. Recommend merging this PR first since the change is smaller — PR #89's full Strategy 4 rewrite will cleanly supersede the `isSubItem` line during rebase.

LGTM.

PR #93, opened 2026-03-13
fix: address 19 bugs from deep code review

nightmeng:
## Summary

Comprehensive bug fix batch from a deep code review covering security, correctness, resilience, schema parity, and frontend issues across 20 files.

### Security (Critical)
- **Path traversal in skill materialization**: Extract shared `resolveUnderDir()` utility (`src/shared/path-utils.ts`) and guard all skill/script path construction in `resource-handlers.ts` and `local-spawner.ts`
- **Shell command escape bypass**: Fix double-backslash bypass in `extractCommands()` and `extractPipeline()` — count consecutive backslashes instead of checking only the immediate predecessor
- **MCP manager leak**: Shut down `mcpManager` in both `close()` and `closeAll()` to prevent resource leaks on session teardown
- **JWT missing exp treated as valid**: `isTokenExpired()` now returns `true` when `payload.exp` is undefined (fail-safe)

### Correctness
- **SQLite schema parity**: Add missing `labelsJson` column and `ON DELETE` actions (`cascade`/`set null`) to match `migrate-sqlite.ts` DDL
- **Truncation marker overflow**: Subtract marker length before computing head/tail split in `truncateWithBudget()`
- **Curl flag bounds check**: Reject incomplete curl flags (e.g., `-X` without argument) in `validateCurl()`
- **Symlink-to-file in skill discovery**: `fs.statSync()` the symlink target to verify it's actually a directory

### Resilience
- **Atomic credential writes**: Use staged `.new` files + `fs.renameSync()` to prevent data loss on crash during credential materialization
- **HTTP body size limit**: Add 10MB cap and 30s timeout to `parseJsonBody()` to prevent memory exhaustion
- **Credential materialization error handling**: Wrap both call sites in try-catch, return 500 instead of crashing
- **Sub-agent timeout**: 5-minute `Promise.race()` timeout around `session.prompt()` in deep search
- **Startup log severity**: Upgrade resource sync failure logs from `console.warn` to `console.error`

### Frontend
- **ErrorBoundary**: New top-level React error boundary with reload button
- **WebSocket reconnect backoff**: Replace fixed delay array with exponential backoff + jitter (1s → 30s cap)
- **SSO error display**: Show error message for 3s before redirecting to login
- **Source maps**: Enable hidden source maps in Vite production build

## Test plan
- [x] All 864 tests pass (`npm test`)
- [ ] Manual: verify ErrorBoundary renders on thrown error in browser
- [ ] Manual: verify WebSocket reconnects with increasing delays
- [ ] Manual: verify SSO error message displays before redirect

### DIFF
--- a/src/tools/credential-manager.test.ts
+++ b/src/tools/credential-manager.test.ts
@@ -3,8 +3,9 @@ import fs from "node:fs";
 import path from "node:path";
 import os from "node:os";
 import { _testing } from "./credential-manager.js";
+import { resolveUnderDir } from "../shared/path-utils.js";
 
-const { safeName, sanitizeSshField, resolveUnderDir, readManifest } = _testing;
+const { safeName, sanitizeSshField, readManifest } = _testing;
 
 // ---------------------------------------------------------------------------
 // safeName
@@ -78,12 +79,12 @@ describe("resolveUnderDir", () => {
 
   it("blocks ../ traversal", () => {
     expect(() => resolveUnderDir(credDir, "../../../etc/passwd"))
-      .toThrow("Path traversal blocked");
+      .toThrow("Path escapes base directory");
   });
 
   it("blocks absolute path escape", () => {
     expect(() => resolveUnderDir(credDir, "/etc/passwd"))
-      .toThrow("Path traversal blocked");
+      .toThrow("Path escapes base directory");
   });
 });
 


src/tools/deep-search/sub-agent.ts
```diff
@@ -393,9 +393,16 @@ export async function runSubAgent(
...
+  const SUB_AGENT_TIMEOUT_MS = 5 * 60 * 1000; // 5 minutes
+
   try {
-    // Run the investigation
-    await session.prompt(userMessage);
+    // Run the investigation with a timeout to prevent indefinite hangs
+    await Promise.race([
+      session.prompt(userMessage),
+      new Promise<never>((_, reject) =>
+        setTimeout(() => reject(new Error("Sub-agent timed out after 5 minutes")), SUB_AGENT_TIMEOUT_MS),
```
LikiosSedo: **Nit: timeout timer is never cleared when `session.prompt()` resolves first.**

When the race resolves via the normal path, this `setTimeout` keeps running for up to 5 minutes. It fires, calls `reject()` on an already-resolved promise (no-op), and then gets GC'd. Not a crash risk, but with 4 parallel sub-agents per investigation, that's 4 dangling timers holding closures.

```ts
// Suggested fix:
let timeoutId: ReturnType<typeof setTimeout>;
try {
  await Promise.race([
    session.prompt(userMessage),
    new Promise<never>((_, reject) => {
      timeoutId = setTimeout(() => reject(new Error("...")), SUB_AGENT_TIMEOUT_MS);
    }),
  ]);
} finally {
  clearTimeout(timeoutId!);
  sessionFinished = true;
  // ...
}
```

src/gateway/web/src/pages/Login/SsoCallback.tsx
```diff
@@ -12,11 +12,19 @@ import { Loader2 } from 'lucide-react';
 export function SsoCallback() {
     const navigate = useNavigate();
     const [params] = useSearchParams();
+    const [error, setError] = useState<string | null>(null);
 
     useEffect(() => {
         const token = params.get('token');
         const userId = params.get('userId');
```
LikiosSedo: **Nit: `setTimeout` in `useEffect` without cleanup.**

Both `setTimeout` calls in this effect (this one and the missing-params branch below) are not cleaned up on unmount. If the component unmounts before the 3s delay, the timeout fires and calls `navigate()` on a stale router context.

```tsx
useEffect(() => {
  let timerId: ReturnType<typeof setTimeout>;
  // ...
  if (errorParam) {
    setError(errorParam);
    timerId = setTimeout(() => navigate(...), 3000);
    return () => clearTimeout(timerId);
  }
  // ... (same for the else branch)
  return () => clearTimeout(timerId);
}, [params, navigate]);
```

Minor — this component is only rendered at the SSO callback route, so the unmount race is unlikely in practice.

PR #94, opened 2026-03-13
fix: structured output for propose_hypotheses to fix hypothesis card title bleed-through

LikiosSedo:
## Problem

The Hypothesis Review card in the web UI sometimes showed the title line (e.g. \"Revised Hypotheses for 'demo' Check\") as hypothesis item 1, pushing real hypotheses to items 2 and 3. This happened because `propose_hypotheses` accepted `hypotheses: string` (freeform markdown), and the text parser in `HypothesesCard.tsx` misclassified the title as a hypothesis.

A secondary bug: `deep_search` stores `RawHypothesis[]` in `details.hypotheses`, but `HypothesesCard` previously cast it as `string`, rendering `[object Object]` instead of actual hypothesis text.

## Solution

Replace the freeform string input with structured output — the same pattern already used by `generateHypotheses()` in Phase 2.

- **`src/tools/dp-tools.ts`**: Changed `propose_hypotheses` parameter from `hypotheses: string` to `hypotheses: Array<{id, text, confidence}>`. The structured array is stored directly in `details.hypotheses` — no text serialization or parsing.

- **`src/gateway/web/src/pages/Pilot/components/HypothesesCard.tsx`**: Added type-aware branching on `toolDetails.hypotheses`:
  - `Array` → map directly to `ParsedHypothesis` (no regex, no title bleed-through; also fixes `deep_search` result rendering)
  - `string` → fall through to existing `parseHypotheses()` for backward compatibility with historical messages

## Test plan

- [ ] Trigger a deep investigation in WebUI — hypothesis card shows correctly numbered items with no title as item 1
- [ ] Send feedback and re-propose — revised list renders correctly
- [ ] Load a historical session — old text-format cards still render via the string fallback path
- [ ] Check `deep_search` result card — hypotheses show text instead of `[object Object]`

### DIFF
--- a/src/gateway/web/src/hooks/usePilot.ts
+++ b/src/gateway/web/src/hooks/usePilot.ts
@@ -587,12 +587,31 @@ export function usePilot() {
                     break;
                 }
 
-                case 'message_end':
+                case 'message_end': {
+                    const endMsg = payload.message as { role?: string; toolName?: string; details?: Record<string, unknown> } | undefined;
+                    if (endMsg?.role === 'toolResult' && endMsg.details && Object.keys(endMsg.details).length > 0) {
+                        // Pi-agent brain: tool result details arrive via message_end (not tool_execution_end).
+                        // Backfill toolDetails onto the matching tool message.
+                        const tName = endMsg.toolName;
+                        setMessages(prev => {
+                            // Walk backwards to find the most recent tool message with this name
+                            for (let i = prev.length - 1; i >= 0; i--) {
+                                const m = prev[i];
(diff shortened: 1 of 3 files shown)


chent1996 (approved):
Good fix — structured array output for `propose_hypotheses` eliminates the title bleed-through issue and also fixes `deep_search` result rendering (`[object Object]`). Backward compat via string fallback path is preserved for historical messages.

Minor readability nit: `hypothesesSource` is declared in both the `else` block scope and the outer scope (same name, different scopes). Not a bug, just slightly confusing. Non-blocking.

LGTM.

PR #100, opened 2026-03-13
fix(security): eliminate command injection in get-node-logs.sh

chent1996:
## Summary
- Replace `sh -c` string interpolation with native bash pipelines to eliminate command injection risk
- All user parameters (`$UNIT`, `$SINCE`, `$FILE`, `$GREP`, `$TAIL`) now pass through double-quoted variables with single shell expansion
- Add `--tail` positive integer validation
- Add `--` sentinels for `grep` and `cat` to prevent argument injection

## Changes

| Before | After |
|--------|-------|
| `sh -c "$CMD"` with string concatenation | Native bash pipelines, `"$VAR"` quoting |
| No `$TAIL` validation | Regex check `^[0-9]+$` |
| `grep -i '$GREP'` inside string | `grep -i -- "$GREP"` as argument |
| `cat '$FILE'` inside string | `cat -- "$FILE"` as argument |

## Test plan
- [ ] `bash -n get-node-logs.sh` passes syntax check
- [ ] `--unit containerd --tail 50` fetches journalctl logs correctly
- [ ] `--file /var/log/messages --grep error` filters file logs correctly
- [ ] `--tail "abc"` rejected with error message
- [ ] `--grep "-e malicious"` handled safely via `--` sentinel

### DIFF
--- a/skills/core/node-logs/scripts/get-node-logs.sh
+++ b/skills/core/node-logs/scripts/get-node-logs.sh
@@ -55,8 +55,8 @@ if [[ -n "$UNIT" && -n "$FILE" ]]; then
 fi
 
 # Validate --tail is a positive integer
-if ! [[ "$TAIL" =~ ^[0-9]+$ ]]; then
-  echo "Error: --tail must be a positive integer, got: $TAIL" >&2
+if ! [[ "$TAIL" =~ ^[1-9][0-9]*$ ]]; then
+  echo "Error: --tail must be a positive integer (>0), got: $TAIL" >&2
   exit 1
 fi
 


jacoblee-io (commented):
## Review

Good security fix — eliminating `sh -c` string interpolation is the right approach. A few issues to address:

### 1. `--tail` validation placement is wrong

The `--tail` validation is placed **after** the `--unit`/`--file` mutual exclusion check but **before** any usage of `$TAIL`. However, the default value `TAIL=200` is valid, so the check only fires when the user explicitly passes `--tail`. This is fine logically, but there's an edge case:

```bash
--tail 0
```

`0` passes the `^[0-9]+$` regex but `tail -n 0` produces no output, which would hit the "No logs found" branch misleadingly. Consider rejecting `0`:

```bash
if ! [[ "$TAIL" =~ ^[1-9][0-9]*$ ]]; then
```

### 2. Missing `--since` validation

`--since` is passed directly to `journalctl -u "$UNIT" --since "$SINCE"`. While this is no longer injectable via `sh -c`, a malicious `--since` value could still cause confusing `journalctl` errors. Consider a basic format check (e.g., reject if it contains newlines or shell metacharacters), or at least document that journalctl handles validation.

### 3. Missing `--file` path validation

`cat -- "$FILE"` is safe from injection, but there's no check that `$FILE` starts with a reasonable path prefix (e.g., `/var/log/`). This script runs via `nsenter` on the host — a user could read arbitrary files like `--file /etc/shadow`. This is arguably out of scope for this PR (it was already possible before), but worth noting as a follow-up.

### 4. `fetch_logs` captures stderr — intentional?

```bash
journalctl -u "$UNIT" --since "$SINCE" --no-pager 2>&1
cat -- "$FILE" 2>&1
```

The `2>&1` merges stderr into stdout. If `journalctl` or `cat` produces an error (e.g., unit not found, file not readable), it gets silently mixed into `$OUTPUT` and displayed as if it were log content. The original code had the same behavior (`sh -c "$CMD" 2>&1`), so this isn't a regression — but it would be cleaner to handle errors separately.

### Overall

The core fix (replacing `sh -c` with native pipelines + proper quoting) is correct and well-structured. The `--` sentinels and `$TAIL` validation are good additions. **Approve with nits** — consider the `--tail 0` edge case at minimum.

jacoblee-io (approved):
Clean security fix. The old code used `sh -c "$CMD"` with string concatenation — a textbook command injection vector. Any of `$UNIT`, `$FILE`, `$GREP`, or `$SINCE` could break out of the single quotes.

Reviewed all changes:

- **`sh -c` eliminated** ✅ — Native bash pipelines with `"$VAR"` quoting. Each variable is a single shell argument, no secondary expansion.
- **`--` sentinels** ✅ — `cat -- "$FILE"` and `grep -i -- "$GREP"` prevent argument injection (e.g., `--grep "-e malicious"`).
- **`$TAIL` validation** ✅ — `^[1-9][0-9]*$` rejects 0, negatives, and non-numeric input. Rejecting 0 avoids the misleading "No logs found" message from empty `tail -n 0` output.
- **`journalctl` args** — `$UNIT` and `$SINCE` are passed as option values (`-u "$UNIT"`, `--since "$SINCE"`), so even adversarial values like `"-e foo"` are interpreted as the option's argument, not as separate flags. Safe.
- **`fetch_logs` function** — Clean separation. `2>&1` on individual commands rather than on the entire pipeline.

LGTM.

PR #101, opened 2026-03-13
feat: investigation feedback mechanism + fix DP checklist auto-complete

LikiosSedo:
## Summary

- **Investigation feedback mechanism**: Add a low-friction feedback loop that lets SREs rate investigation accuracy (confirmed/corrected/rejected). Feedback modifies retrieval weighting so correct diagnoses get boosted and wrong ones get suppressed in future investigations.
- **DP checklist auto-complete fix**: Fix checklist card stuck at 2/4 after investigation completes when agent doesn't explicitly call `manage_checklist` for remaining phases.

### Investigation Feedback (commit 1 + 2)

- Add `FeedbackStatus` type and `FEEDBACK_SIGNALS` constant (1.5 / 0.5 / 0.1)
- Add 3 column migrations: `feedback_signal`, `feedback_note`, `feedback_at`
- Add `getInvestigationById()` and `updateInvestigationFeedback()` to indexer
- Apply feedback weighting in pattern aggregation, keyword search, and validated hypothesis extraction
- Plumb `investigationId` through `InvestigationResult` → tool details
- Create `investigation_feedback` tool and register in agent-factory
- Add Phase 5 feedback guidance to deep-investigation SKILL.md
- Add WebUI feedback strip (Correct / Partial / Wrong) to InvestigationCard

### Review fixes (commit 2)

- `extractValidatedHypotheses`: separate `rawCount` (display) from `weightedScore` (sorting) to avoid inflated/deflated occurrence counts
- `InvestigationCard`: extract `handleFeedbackSubmit()` to deduplicate Enter-key and Submit button logic

### DP Checklist (commit 3)

- Relax `prompt_done` safety-net guard: only block auto-complete when `hypotheses` phase isn't done (waiting for user confirmation), not when `deep_search` isn't done

## Test plan

- [ ] `tsc --noEmit` passes
- [ ] Run deep investigation → agent presents findings → say "looks right" → verify `investigation_feedback` tool called → check SQLite `feedback_signal` updated
- [ ] Complete investigation in WebUI → click feedback button on InvestigationCard → verify metadata persisted + agent processes feedback
- [ ] Verify DP checklist auto-clears after investigation completes (no stuck 2/4)
- [ ] Create 3 investigations, reject one → verify rejected ranks lower in `searchInvestigations` and `getInvestigationPatterns`

### DIFF
--- a/src/gateway/web/src/hooks/usePilot.ts
+++ b/src/gateway/web/src/hooks/usePilot.ts
@@ -651,14 +651,14 @@ export function usePilot() {
                     // The backend agent_end handler sends dp-checklist-sync, but
                     // it's async and may not arrive before SSE closes (race condition).
                     // Mirror the backend logic here to ensure frontend consistency.
-                    // Skip if hypotheses phase isn't done yet — model is waiting
+                    // Only skip if hypotheses phase isn't done yet — model is waiting
                     // for user confirmation (gate blocked), not truly finished.
                     setDpChecklist(prev => {
                         if (!prev) return prev;
-                        const dsItem = prev.find(i => i.id === 'deep_search');
-                        // If deep_search hasn't been done, the investigation is still in progress
-                        // (either waiting for hypothesis confirmation or hasn't started deep_search yet)
-                        if (!dsItem || dsItem.status !== 'done') return prev;
+                        const hypoItem = prev.find(i => i.id === 'hypotheses');
+                        // If hypotheses phase isn't done, the agent is likely waiting
+                        // for user confirmation — don't auto-complete
+                        if (hypoItem && hypoItem.status !== 'done' && hypoItem.status !== 'skipped') return prev;
                         const hasIncomplete = prev.some(i => i.status === 'pending' || i.status === 'in_progress');
                         if (!hasIncomplete) return prev;
                         const completed = prev.map(i =>


jacoblee-io (changes requested):
Two distinct features in one PR — the checklist fix is clean, but the feedback mechanism has a few issues.

### PR description only covers half the changes

The title/body describe the checklist auto-complete fix (~10 lines in usePilot.ts), but the bulk of the PR is the investigation feedback mechanism (~300 lines across 10 files: new tool, schema migration, indexer methods, engine weighting, UI component, skill doc). These are independent features and would be easier to review as separate PRs. At minimum, the PR description should cover the feedback mechanism.

---

### Checklist auto-complete fix ✅

The guard condition change from `deep_search.status === 'done'` to `hypotheses.status !== 'done' && !== 'skipped'` is correct. The old condition created a deadlock since `deep_search` was the stuck item itself.

---

### Investigation feedback mechanism — issues

**1. Bug: `extractValidatedHypotheses` count semantics**

`count` changed from integer (number of occurrences) to float (weighted by `feedbackSignal`). The display label still says "validated N times":

```ts
`- "${displayText}" (validated ${Math.round(count)} time${...})`
```

Example: a hypothesis validated in 3 confirmed investigations:
- count = 1.5 + 1.5 + 1.5 = 4.5 → "validated 5 times" (was actually 3)

A rejected investigation's hypothesis:
- count = 0.1 → "validated 0 times"

Suggestion: keep `count` as a raw integer for display, apply `feedbackSignal` weighting separately in the sort key:

```ts
existing.count++;
existing.weightedScore = (existing.weightedScore ?? 0) + feedbackSignal;
// Sort by weightedScore, display raw count
```

**2. Nit: feedback via chat message is indirect**

The UI sends feedback as a user chat message (`[investigation feedback: confirmed] investigationId=...`) — the agent must interpret it and call `investigation_feedback`. If the agent misinterprets or responds conversationally, the DB update doesn't happen (only the UI metadata is saved via `updateMessageMeta`).

This follows the existing HypothesesCard pattern so it's architecturally consistent, but worth noting that the `updateMessageMeta` call creates a false sense of persistence — the real weight adjustment depends on the LLM cooperating.

**3. Nit: duplicate submit logic in InvestigationCard**

The Enter-key handler and Submit button onClick contain identical logic (~6 lines each). Consider extracting to a `handleSubmit()` function.

---

Everything else looks good — schema migration is safe (ALTER TABLE ADD COLUMN with existence checks), `getInvestigationById`/`updateInvestigationFeedback` use parameterized queries, `FEEDBACK_SIGNALS` constants are well-chosen, and the TypeBox tool auto-adapts for claude-sdk brain via `adaptToolsForSdk`.

chent1996 (approved):
Two clean changes:

1. **Investigation feedback**: Schema migration safe (3 nullable columns with existence check). Feedback signals (1.5/0.5/0.1) applied consistently across all 3 retrieval paths (pattern aggregation, keyword search, validated hypothesis extraction). Tool validates investigation existence before update. `investigationId` plumbing from engine → tool result → details → frontend is complete.

2. **DP checklist auto-complete**: Guard condition change from `deep_search` to `hypotheses` phase is logically correct — only block auto-complete when waiting for user hypothesis confirmation.

Minor observation: frontend feedback is sent via `sendMessage()` (chat message to agent), with optimistic UI update via `updateMessageMeta`. If the agent doesn't act on the message, the investigation DB won't be updated but UI shows success. Acceptable trade-off for conversational UX. Non-blocking.

LGTM.

PR #103, opened 2026-03-13
refactor(gateway): merge standalone cron process into gateway

jacoblee-io:
## Summary

Merge the standalone cron process into the gateway, eliminating 17 HTTP internal endpoints, multi-instance coordination, and a separate K8s deployment. The cron service was stateless — all data lives in gateway's DB — so the separate process added unnecessary complexity (heartbeat, dead-instance detection, job claiming/reassignment, HTTP round-trips).

### Solution

- New in-process `CronService` (`src/gateway/cron/cron-service.ts`) wraps the existing `CronScheduler` with direct `configRepo` DB calls. Job execution still goes through `localhost/api/internal/agent-prompt` (same as triggers).
- RPC methods and channel-bridge call `cronService.addOrUpdate()`/`.cancel()` directly instead of HTTP broadcast via `notifyCronService()`.
- Removed multi-instance coordination (`CronCoordinator`, `assignedTo` logic, `cron_instances` table reads).
- Deleted 6 source files, `Dockerfile.cron`, `k8s/cron-deployment.yaml`, 3 helm templates.
- Updated Makefile, helm values, k8s manifests, docs.

Net: **-1100 lines**, one fewer K8s deployment.

### DB schema note

`cron_instances` table and `cron_jobs.assignedTo` column become unused — left in place to avoid migration, can be dropped in a follow-up.

## Test plan

- [x] `npx vitest run` — no new failures
- [x] No remaining imports of deleted modules (`grep` verified)
- [x] Deployed to test cluster — gateway logs show `[cron-service] Loaded 0 active jobs` + `Started`
- [x] Helm upgrade automatically removed `siclaw-cron` deployment
- [ ] Manual: create/pause/delete/rename cron job via UI, verify scheduler picks up changes
- [ ] Manual: wait for a job to fire, verify agent-prompt execution + notification delivery

### DIFF
--- a/src/gateway/cron/cron-service.ts
+++ b/src/gateway/cron/cron-service.ts
@@ -128,14 +128,18 @@ export class CronService {
       const resultText = data.resultText || "";
 
       // Record success
-      await this.configRepo.updateCronJobRun(job.id, "success");
+      try { await this.configRepo.updateCronJobRun(job.id, "success"); } catch (e) {
+        console.warn(`[cron-service] updateCronJobRun failed for ${job.id}:`, e instanceof Error ? e.message : e);
+      }
       console.log(`[cron-service] Job ${job.id} completed in ${data.durationMs}ms, resultText length=${resultText.length}`);
 
       // Write run record + notification
       await this.recordResult(job, "success", resultText, undefined, data.durationMs);
     } catch (err) {
       console.error(`[cron-service] Job ${job.id} failed:`, err);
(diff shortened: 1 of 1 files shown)


jacoblee-io (commented):
Thanks for the review!

- **`src/shared/retry.ts`**: Deleted in 027cfc8 — confirmed no other consumers remain.
- **`docs/design/db-cleanup-spec.md`**: This file doesn't exist on `main` or this branch. Could you double-check which file you meant?

PR #105, opened 2026-03-13
refactor(k8s): replace per-user dynamic PVCs with shared PVC + subPath

jacoblee-io:
## Summary

- Replace per-user dynamically provisioned PVCs with a single shared PVC (`siclaw-data`)
- Gateway creates per-user subdirectories (`users/{userId}/{workspaceId}/`) on the shared volume
- AgentBox pods mount via `subPath` for isolation — no dynamic provisioning or StorageClass required
- Sanitize userId/workspaceId in path segments to prevent traversal and invalid directory names

## Details

**Before**: Each user got a dedicated PVC (`siclaw-user-{userId}`) created on demand via K8s API, requiring a StorageClass that supports dynamic provisioning.

**After**: Operator pre-creates one shared `ReadWriteMany` PVC. Gateway mounts it and creates subdirectories at spawn time; AgentBox pods mount the user-specific subdirectory via `subPath`.

### Changes

- **k8s-spawner.ts**: Remove `ensureUserPvc()`/`userPvcName()`, add `ensureUserDir()`/`sanitizePathSegment()`, use shared PVC `claimName` + `subPath`
- **gateway-main.ts**: Replace `storageClass/accessMode/size` env vars with `claimName/mountPath`
- **Helm values**: Simplify persistence config to `enabled` + `claimName`
- **Helm gateway-deployment**: Mount shared PVC on gateway, pass `SICLAW_PERSISTENCE_MOUNT_PATH`
- **Helm RBAC**: Remove PVC `create` verb (only `get`/`list` needed)
- **siclaw-data-pvc.yaml**: Update to `ReadWriteMany`, 10Gi, updated comments

## Test plan

- [ ] Deploy with `persistence.enabled=true`, verify gateway creates `users/{userId}/{workspaceId}/` on the shared PVC
- [ ] Spawn AgentBox pod, verify it mounts the correct subPath and user data persists across pod restarts
- [ ] Verify special characters in userId/workspaceId are sanitized (no path traversal)
- [ ] Deploy with `persistence.enabled=false` (default), verify emptyDir fallback still works
- [ ] Verify RBAC: gateway no longer needs PVC create permission

### DIFF
--- a/helm/siclaw/templates/NOTES.txt
+++ b/helm/siclaw/templates/NOTES.txt
@@ -1,8 +1,8 @@
 Thank you for installing {{ include "siclaw.fullname" . }}!
 
-{{- if and .Values.agentbox.persistence.enabled (not .Values.agentbox.persistence.storageClass) }}
+{{- if .Values.agentbox.persistence.enabled }}
 
-⚠️  WARNING: agentbox.persistence is enabled but storageClass is empty.
-   The cluster default StorageClass will be used. Set agentbox.persistence.storageClass
-   explicitly for production deployments.
+ℹ️  User data persistence is enabled.
+   Ensure the shared PVC "{{ .Values.agentbox.persistence.claimName }}" exists and supports ReadWriteMany.
+   Gateway will create per-user subdirectories automatically.
 {{- end }}
--- a/helm/siclaw/templates/gateway-deployment.yaml
+++ b/helm/siclaw/templates/gateway-deployment.yaml
@@ -43,12 +43,10 @@ spec:
             {{- if .Values.agentbox.persistence.enabled }}
             - name: SICLAW_PERSISTENCE_ENABLED
               value: "true"
-            - name: SICLAW_PERSISTENCE_STORAGE_CLASS
-              value: {{ .Values.agentbox.persistence.storageClass | quote }}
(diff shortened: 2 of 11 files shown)


jacoblee-io (commented):
Good refactor — dynamic PVC provisioning per user was unnecessarily complex. Shared PVC + subPath is the right call. A few issues below, one of which could cause silent bugs.

jacoblee-io (commented):
Previous comments all addressed — nice. Two small nits remaining, neither is a blocker.

jacoblee-io (commented):
**Re: stale JSDoc** — Fixed, now reads `"/app/.siclaw/user-data"`.

**Re: `users/` vs `user/` prefix** — Intentional. `user/` is the skills PVC directory layout (skills, credentials, kube configs), while `users/` is the shared data PVC layout (memory, sessions). These are different volumes in the actual K8s spawner; the template co-locates them on one PVC for reference only. Renaming `user/` would break existing skills sync paths in `local-spawner.ts` and `resource-handlers.ts`.

jacoblee-io (commented):
All previous comments addressed — JSDoc default fixed, double sanitization cleaned up, mountPath templated via values.yaml. No remaining issues. LGTM, ready to merge.

fred-scitix (approved):

PR #109, opened 2026-03-15
refactor(helm): clean up pseudo-configurable fields and unify image config

jacoblee-io:
## Summary

- **Remove `agentbox.persistence.mountPath`** from Helm values, gateway env vars, and K8sSpawner config — this field was pseudo-configurable (changing it silently breaks the system since the path is hardcoded across AgentBox pod, Dockerfile, entrypoint, and agent framework). Hardcode `/app/.siclaw/user-data` everywhere.
- **Remove `agentbox.image` override block** (registry/repository/tag) from Helm values — both gateway and agentbox images always share the same registry and tag from a single build, so the separate override fields gave false flexibility. `siclaw.agentboxImage` now reuses `siclaw.image` helper.
- **Pin debug pod image** from `busybox:latest` to `busybox:1.36` (current stable), avoiding unnecessary image pulls on every debug pod creation.

## Test plan
- [x] `helm template` renders correctly (verified `SICLAW_AGENTBOX_IMAGE` and `mountPath` values)
- [x] No remaining references to `SICLAW_PERSISTENCE_MOUNT_PATH` or `busybox:latest`
- [x] No remaining references to `.Values.agentbox.image`

### DIFF
--- a/helm/siclaw/templates/_helpers.tpl
+++ b/helm/siclaw/templates/_helpers.tpl
@@ -63,17 +63,8 @@ Usage: {{ include "siclaw.image" (dict "component" "gateway" "ctx" .) }}
 {{- end }}
 
 {{/*
-Build agentbox image string.
-Uses agentbox-specific overrides if set, otherwise falls back to global image settings.
-Supports custom repository name (e.g. "siclaw-agentbox-debug" for Node debug image).
+Build agentbox image string — same registry/tag as gateway, different component name.
 */}}
 {{- define "siclaw.agentboxImage" -}}
-{{- $registry := default .Values.image.registry .Values.agentbox.image.registry -}}
-{{- $repo := default "siclaw-agentbox" .Values.agentbox.image.repository -}}
-{{- $tag := default .Values.image.tag .Values.agentbox.image.tag -}}
-{{- if $registry -}}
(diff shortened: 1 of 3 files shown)


jacoblee-io (commented):
## Review: LGTM

Clean removal of false configurability. All three changes are well-motivated.

### 1. Remove pseudo-configurable `mountPath`

The path `/app/.siclaw/user-data` is hardcoded across Dockerfile.agentbox, entrypoint, and agent framework code — exposing it as a Helm value only invites silent breakage. Hardcoding it in `ensureUserDir()` and the gateway volumeMount is the right call.

Verified: no remaining references to `SICLAW_PERSISTENCE_MOUNT_PATH` in the codebase.

### 2. Unify agentbox image config

`siclaw.agentboxImage` now delegates to `siclaw.image` with `component: "agentbox"`, which produces `{registry}/siclaw-agentbox:{tag}` — identical to the old default. The removed `agentbox.image.{registry,repository,tag}` overrides were never actually used differently from the global `image.*` settings, so this just eliminates dead config surface.

### 3. Pin `busybox:1.36`

Good hygiene — avoids `:latest` drift and unnecessary image pulls on debug pod creation.

No issues found. Ready to merge.

PR #136, opened 2026-03-18
refactor(dp): replace manage_checklist tool with system-event-driven checklist

LikiosSedo:
## Problem

The DP checklist UI was driven by LLM `manage_checklist` tool calls. This is fundamentally fragile:
- **K8s production failure**: Kimi-K2.5 sent `"completed"` instead of `"done"`, schema validation failed 3 times, checklist never updated
- Different models have different tool-calling behavior — UI reliability depended on LLM correctness
- The `deep_search` engine already emits deterministic phase events (`Phase 1/4` ~ `Phase 4/4`), but the frontend wasn't using them to drive the checklist

## Solution

Remove `manage_checklist` tool entirely. Frontend checklist is now driven by the engine's `tool_progress` phase events.

**Backend (3 files):**
- `dp-tools.ts`: Remove `createManageChecklistTool` function, keep all shared types
- `agent-factory.ts`: Remove import and registration
- `deep-investigation.ts` (pi-agent extension): Remove tool registration + `executeManageChecklist` helper, add auto-exit DP mode when `deep_search` completes

**Frontend (1 file):**
- `usePilot.ts`: Delete `manage_checklist` handler, add phase→checklist mapping in `tool_progress` handler (forward-only guard), auto-complete all items + 3s auto-clear when `deep_search` ends

**SDK brain bridge (1 file):**
- `http-server.ts`: Sync phase events to `dpState.checklist` so the auto-continue loop sees correct phase state

**Docs (2 files):**
- `SKILL.md`: Remove all `manage_checklist` references, simplify phase instructions
- `architecture.md`: Remove from tool table

**Backward compat:** Old session history with `manage_checklist` messages are still hidden via `mapMessages` check.

**Net: -99 lines** (90 insertions, 189 deletions)

## Test plan

- [x] `npm test` — 970/970 passed
- [x] `grep -r manage_checklist` — only backward compat + comments remain
- [x] Local E2E: Kimi-K2.5 deep investigation completed without manage_checklist calls, agent finished normally
- [ ] Manual E2E: verify 4 phases update in real-time in Web UI (no page refresh needed)
- [ ] Manual E2E: verify page refresh during Phase 3 restores checklist correctly

### DIFF
--- a/docs/internal/architecture.md
+++ b/docs/internal/architecture.md
@@ -195,7 +195,7 @@ For claude-sdk brain: raw server config is passed to the SDK's native MCP suppor
 4. Loads MCP tools from external servers
 5. Initializes the chosen brain (pi-agent or claude-sdk) with the assembled tools and system prompt
 6. For pi-agent: initializes memory indexer, loads extensions (context pruning, memory flush, deep investigation)
-7. For claude-sdk: adds deep investigation tools (manage_checklist, propose_hypotheses, end_investigation), starts LLM proxy if needed
+7. For claude-sdk: adds deep investigation tools (propose_hypotheses, end_investigation), starts LLM proxy if needed
 
 ## 4. Tool System
 
@@ -233,7 +233,6 @@ All tools are TypeBox `ToolDefinition` objects with structured input schemas and
 | Tool | File | Description |
 |------|------|-------------|
 | `deep_search` | `deep-search/tool.ts` | Structured hypothesis-driven investigation (see Section 5) |
-| `manage_checklist` | `dp-tools.ts` | Update deep investigation phase statuses (claude-sdk only) |
 | `propose_hypotheses` | `dp-tools.ts` | Present hypotheses to user for confirmation (claude-sdk only) |
 | `end_investigation` | `dp-tools.ts` | Early termination of investigation (claude-sdk only) |
 | `manage_schedule` | `manage-schedule.ts` | CRUD for cron schedules. Queries Gateway's internal API |
--- a/skills/core/deep-investigation/SKILL.md
+++ b/skills/core/deep-investigation/SKILL.md
@@ -30,21 +30,23 @@ The following phases are a recommended structure. You may skip or merge phases
 when the situation calls for it — e.g., skip straight to deep_search if the user
 gives a clear direction, or end after triage if the answer is already apparent.
 
+Phase progress is tracked automatically by the system — you do not need to
+manage checklist state manually. The UI updates as the engine progresses
+through each phase.
+
 ### Phase 1: Quick Triage
 
 Gather environment context and confirm the problem exists.
 
-1. Call `manage_checklist` to mark `triage` as `in_progress`.
-2. Run targeted kubectl / diagnostic commands to confirm the symptom.
-3. Call `manage_checklist` to mark `triage` as `done` with a brief summary.
+1. Run targeted kubectl / diagnostic commands to confirm the symptom.
+2. Summarize findings.
 
 ### Phase 2: Propose Hypotheses
 
 Formulate 3-5 ranked hypotheses based on triage findings.
 
-1. Call `manage_checklist` to mark `hypotheses` as `in_progress`.
-2. Call `propose_hypotheses` tool with your formatted hypothesis list.
-3. In DP mode, this is a good point to pause for user input — the user may want
+1. Call `propose_hypotheses` tool with your formatted hypothesis list.
+2. In DP mode, this is a good point to pause for user input — the user may want
    to adjust hypotheses before committing to an expensive deep_search.
    If the user is actively engaged, wait for their confirmation.
    If the user gave a clear directive, you may proceed directly.
@@ -53,17 +55,14 @@ Formulate 3-5 ranked hypotheses based on triage findings.
 
 Validate hypotheses using parallel sub-agents.
 
-1. Call `manage_checklist` to mark `deep_search` as `in_progress`.
-2. Call `deep_search` tool with triageContext and the hypotheses.
-3. When deep_search completes, call `manage_checklist` to mark `deep_search` as `done`.
+1. Call `deep_search` tool with triageContext and the hypotheses.
+2. The system automatically tracks phase progress in the UI.
 
 ### Phase 4: Present Findings
 
 Synthesize the deep_search results into a conclusion.
 
-1. Call `manage_checklist` to mark `conclusion` as `in_progress`.
-2. Write a clear conclusion with root cause analysis and recommendations.
-3. Call `manage_checklist` to mark `conclusion` as `done`.
+1. Write a clear conclusion with root cause analysis and recommendations.
 
 ### Phase 5: Feedback (optional)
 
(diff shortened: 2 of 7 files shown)


LikiosSedo (commented):
# Review Report: refactor(dp): replace manage_checklist tool with system-event-driven checklist

## Summary
Request changes. The core idea (replacing LLM-driven checklist with deterministic engine events) is excellent, but there's a real bug in the SDK brain path where the auto-continue loop exit signal is lost.
6 findings: 1 🔴 critical, 3 🟡 warnings, 2 💬 notes

## 🔴 Must Fix

### [1] SDK brain: `dpState.checklist` never nulled → auto-continue loop won't exit — `http-server.ts` + `dp-tools.ts`
The old `createManageChecklistTool` set `dpState.checklist = null` when conclusion was marked done — this was the exit signal for the SDK brain's auto-continue loop in `claude-sdk-brain.ts`. That tool is removed, but nothing replaces the nulling mechanism. Phase events only mutate item statuses, they never null the checklist. After `deep_search` completes, the auto-continue loop will keep finding non-null checklist items and fire spurious nudges.

The pi-agent path is fine (`deep-investigation.ts` calls `disableDpMode` on tool_result). This only affects SDK brain (claude-sdk).

**Suggestion**: In `http-server.ts`, when Phase 4/4 fires, null `dpState.checklist`. Or detect when `deep_search` tool_result arrives and null it there.

## 🟡 Should Fix

### [2] Feedback prompt fires before agent presents conclusion — `deep-investigation.ts:611-616`
Previously `disableDpMode()` was triggered by `manage_checklist(conclusion=done)`, which happened **after** the agent wrote its conclusion. Now it fires immediately on `deep_search` tool_result — **before** the agent's conclusion message. `disableDpMode` triggers `showFeedbackIfNeeded`, so the "Rate: 👍/👎/❌" prompt flashes in the status bar, then gets cleaned up by `agent_start` a moment later. Minor UX regression.

**Suggestion**: Set a flag on tool_result, consume it in `agent_end` handler to defer `disableDpMode` until after the conclusion message.

### [3] Phase-to-checklist mapping duplicated 3 times — `http-server.ts`, `usePilot.ts` (×2)
The `phaseStr.match(/(\d+)/) → phaseNum - 1 → iterate items` logic appears in: http-server.ts backend sync, usePilot.ts live event handler, and usePilot.ts session-restore path. All must be kept in sync manually.

**Suggestion**: Extract `applyPhaseToChecklist(items, phaseNum)` into `dp-tools.ts` (which already owns `createChecklist`). At minimum, add cross-reference comments.

### [4] `dpTimeout` side effect inside React state updater — `usePilot.ts:471-479`
`dpTimeout(() => resetDpState(), 3000)` is called inside `setDpChecklist(prev => ...)`. State updaters should be pure — React Strict Mode calls them twice, which would schedule duplicate timers. This is a pre-existing pattern (the old `manage_checklist` handler had the same issue), but worth noting since it's being restructured anyway.

**Suggestion**: Move `dpTimeout(...)` call outside the state updater.

## 💬 Notes

### [5] Phase string format is an implicit API
`ProgressEvent.phase` is typed as `string` but parsed via regex (`/(\d+)/`) in 3 places. If the engine ever changes the format, all break silently. Consider adding `phaseNum: number` to the event type so consumers don't parse display strings.

### [6] Checklist item ordering is an implicit contract
`createChecklist()` defines items `[triage, hypotheses, deep_search, conclusion]` at indices 0-3. The `phaseNum - 1` arithmetic couples this ordering to engine phase numbers. Add a comment documenting this constraint.

## ✅ What Looks Good
- Core design decision is sound — replacing LLM-reliability-dependent UI with deterministic engine events eliminates the Kimi-K2.5 class of failures
- Forward-only guards (`status !== 'done' && status !== 'skipped'`) correctly handle pre-triage and skipped-phase scenarios
- Clean deletion: net -99 lines
- Backward compat for old `manage_checklist` messages in `mapMessages` is preserved

jacoblee-io (commented):
# Review Report: refactor(dp): replace manage_checklist tool with system-event-driven checklist

## Summary
Approve with minor fixes. The core direction — replacing LLM-driven checklist with deterministic engine phase events — is architecturally sound and directly addresses the Kimi-K2.5 schema mismatch failure.

5 findings: 1 🔴 critical, 3 🟡 warnings, 1 💬 note

### ⚠️ Process Note
Self-fix pattern detected: commit `8c9f9e0` introduced the refactoring, then `47e0226` fixed SDK checklist null, deferred DP exit, shared helper extraction, and React StrictMode double-fire.

---

## 🔴 Must Fix

### [1] Side effects inside `setMessages` updater — `usePilot.ts`
When `deep_search` completes, `setDpChecklist()`, `setDpFocus()`, and `dpTimeout()` (setTimeout) are called inside the `setMessages(prev => ...)` updater function. React may invoke updaters more than once in StrictMode/concurrent mode — `dpTimeout` will register duplicate timers, causing `resetDpState()` to fire twice. Commit `47e0226` claims to fix this but side effects appear to still be present inside the updater.

**Suggestion**: Extract the DP state updates out of the `setMessages` updater. Check `last.toolName === 'deep_search'` before calling `setMessages`, or use a ref to capture the condition and trigger DP updates separately.

---

## 🟡 Should Fix

### [2] Phase events create spurious DpChecklist in standalone `deep_search` — `usePilot.ts`
The `tool_progress` phase handler calls `createDefaultDpChecklist()` when `prev` is null. When `deep_search` runs outside DP mode (no `/dp` prefix), `dpChecklist` is null — the handler will create one, causing `DpChecklistCard` to appear for standalone `deep_search` runs where it shouldn't.

**Suggestion**: Add `if (!prev) return prev;` guard at the top of the `setDpChecklist` updater in the phase handler, so checklist is only updated when DP mode is already active.

### [3] Phase sync lacks `_promptDone` guard — `http-server.ts:337`
The `deepProgressBufHandler` runs `applyPhaseToChecklist` unconditionally, but the buffer push above it is guarded by `if (!managed._promptDone)`. After prompt completion, stale `deep_search` events could still arrive and mutate `dpState.checklist` on the wrong session.

**Suggestion**: Wrap the `applyPhaseToChecklist` block in the same `if (!managed._promptDone)` guard.

### [4] Null dereference risk casting `ev.phase` — `http-server.ts:341`, `usePilot.ts`
`ev.phase as string` casts an `unknown` value. If the event is malformed and `ev.phase` is `undefined`, `parsePhaseNum(undefined)` calls `.match()` on undefined, throwing `TypeError`.

**Suggestion**: `const phaseStr = typeof ev.phase === "string" ? ev.phase : "";` — `parsePhaseNum("")` safely returns 0.

---

## 💬 Notes

### [5] Self-fix pattern across commits
Commit `8c9f9e0` introduced the refactoring, then `47e0226` fixed 4 issues: SDK checklist null on `deep_search` completion, deferred DP exit timing, duplicated logic extraction, and React StrictMode double-fire. This suggests the intermediate state had bugs that were caught in review.

---

## ✅ What Looks Good
- Replacing LLM-driven `manage_checklist` with deterministic engine phase events eliminates an entire class of model-compatibility bugs
- `applyPhaseToChecklist()` + `parsePhaseNum()` extraction into `dp-tools.ts` as shared helpers is well-structured
- Forward-only guard pattern (not regressing `done`/`skipped` items) is correctly applied in all three phase-mapping locations
- Net -60 lines with improved reliability — good complexity paydown

jacoblee-io (approved):
LGTM — all 4 findings from round 1 addressed in dd4b48f. Clean fix.

PR #138, opened 2026-03-18
fix(tui): clear hypothesis widget before opening select dialog

LikiosSedo:
## Problem

When `propose_hypotheses` runs in TUI mode, it renders a hypothesis widget above the editor via `setWidget()`, then opens a `select()` dialog. The pi-tui framework renders `widgetContainerAbove` and `editorContainer` sequentially — since the widget is never cleared, both render simultaneously, causing the select options to appear duplicated/overlapping on screen.

## Solution

Clear the widget before opening `ctx.ui.select()` or `ctx.ui.input()`, restore it after the dialog closes. The widget will be definitively cleared by the `tool_call` handler when the model proceeds to the next tool.

## Test plan

- [x] `npm test` — 970/970 passed
- [ ] Manual TUI: run `/dp` investigation, verify select dialog renders cleanly without duplicated options

### DIFF
--- a/src/core/extensions/deep-investigation.ts
+++ b/src/core/extensions/deep-investigation.ts
@@ -92,8 +92,10 @@ function parseHypotheses(text: string): Array<{ title: string; confidence?: numb
     return { title, confidence: confMatch ? parseInt(confMatch[1], 10) : undefined };
   }
 
-  // Strategy 1: Structured headers — ## Hypothesis N / ### H1 (supports CJK)
-  const headerPattern = /^#{2,3}\s*(?:Hypothesis|假设|假說|H)\s*\d[^:：\n]*[:：\s]*(.*)/gim;
+  // Strategy 1: Structured headers — ## Hypothesis N / ### H1 / ### #1 (supports CJK)
+  // Requires a number after the keyword to distinguish individual hypotheses
+  // from summary headings like "## 假设列表（按可能性排序）".
+  const headerPattern = /^#{2,3}\s*(?:Hypothesis|假设|假說|H|#)\s*\d[^:：\n]*[:：\s]*(.*)/gim;
   let m: RegExpExecArray | null;
   while ((m = headerPattern.exec(text)) !== null) {
     const titleLine = m[1]?.trim();
--- a/src/core/extensions/deep-investigation.ts
+++ b/src/core/extensions/deep-investigation.ts
@@ -405,13 +405,22 @@ export default function deepInvestigationExtension(api: ExtensionAPI, memoryRef?
       const hasTUI = ctx.hasUI && isThemeUsable(ctx);
 
       if (hasTUI) {
-        // Render hypotheses widget above editor
-        const widgetLines = formatHypothesesWidget(hypothesesText, ctx.ui.theme);
-        ctx.ui.setWidget("dp-hypotheses", widgetLines);
+        // Build a compact text summary of hypotheses for the select title.
+        // We do NOT use setWidget here — the pi-tui framework's spinner
+        // (80ms animation) triggers requestRender() continuously, and having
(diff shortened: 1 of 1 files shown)


jacoblee-io (commented):
# Review Report: fix(tui): clear hypothesis widget before opening select dialog

## Summary
**Approve with suggestions.** The fix correctly addresses the widget/select overlap bug in pi-tui, and the regex addition handles a real LLM output pattern. A few edge cases in the new flow deserve attention.

5 findings: 0 🔴 critical, 3 🟡 warnings, 2 💬 notes

## 🟡 Should Fix

### [1] Widget not restored on abort in "Adjust hypotheses" branch — `src/core/extensions/deep-investigation.ts:~431`
In the "Adjust" path, `setWidget(undefined)` is called before `ctx.ui.input()`, and `setWidget(widgetLines)` restores it after. If `input()` throws (e.g., `signal` aborted), the restore never executes and the widget stays cleared. | **Dimension**: Correctness
**Suggestion**: Wrap in try/finally: `try { feedback = await ctx.ui.input(...) } finally { ctx.ui.setWidget("dp-hypotheses", widgetLines); }`

### [2] Empty hypotheses list produces blank select dialog title — `src/core/extensions/deep-investigation.ts:~414`
If `parseHypotheses()` returns an empty array (free-form LLM text matching no strategy), `hypoLines` is `""` and the select title becomes `"Review Hypotheses\n\n\n"` — the user sees a blank area where hypotheses should be listed. This is a regression from the old flow which always showed the formatted widget. | **Dimension**: Correctness
**Suggestion**: Add a fallback when `parsed.length === 0` — e.g., include a truncated snippet of the raw `hypothesesText` or a placeholder like `"(hypotheses shown above)"`.

### [3] Regex `|#` in header alternation is slightly over-broad — `src/core/extensions/deep-investigation.ts:~100`
Adding bare `#` to `(?:Hypothesis|假设|假說|H|#)` matches any H2/H3 of the form `## #N ...` (e.g., `### #42 Some system event` — a bug tracker reference). The comment says this targets `### #1` style headers, but nothing restricts broader matches. | **Dimension**: Correctness
**Suggestion**: If the intent is only single-digit numbered hypotheses, consider constraining further (e.g., `#(?=[1-9]\b)`). Low risk since the downstream `parseHypothesis()` call further validates, but worth documenting the assumption.

## 💬 Notes

### [4] Widget persists after "Skip to conclusion" — `src/core/extensions/deep-investigation.ts:~427`
`setWidget("dp-hypotheses", widgetLines)` is called unconditionally after the select dialog, including the "Skip to conclusion" path. Cleanup only happens on the next `tool_call` event. If the model responds with plain text (no tool call), the widget stays visible indefinitely.

### [5] `parseHypotheses` has no test coverage — `src/core/extensions/deep-investigation.ts:~83`
`parseHypotheses` was previously just a rendering helper but now drives the user-visible dialog content. The regex change adding `|#` is also untested. Tests for this function would increase confidence in the parsing logic, especially edge cases like the new `### #1` pattern and the summary-heading exclusion.

## ✅ What Looks Good
- The core fix correctly eliminates the widget/select rendering overlap by moving `setWidget()` after the dialog closes, with a clear explanation of the pi-tui spinner/requestRender interaction
- Embedding a compact hypothesis summary directly in the select title is a pragmatic solution that preserves user context without requiring dual-container rendering
- The `.planning/codebase/` documentation files are thorough and well-structured — useful reference for future contributors

jacoblee-io (approved):
Finding [1] addressed in commit abea40d — `ctx.ui.input()` wrapped in try/finally so the hypothesis widget is always restored on abort.

Remaining findings [2]-[5] are YELLOW/NOTE level and non-blocking.

LGTM.

PR #139, opened 2026-03-18
fix(deep-search): catch Phase 1 sub-agent timeout to prevent investigation failure

LikiosSedo:
## Problem

When `deep_search` is called without `triageContext` (autonomous mode), Phase 1 launches a sub-agent to gather environment context. If this sub-agent exceeds the 5-minute timeout (`SUB_AGENT_TIMEOUT_MS`), the error bubbles up uncaught through `investigate()` to the tool layer's catch block, returning:

> Deep search failed: Sub-agent timed out after 5 minutes

This kills the entire investigation even though Phases 2-4 could still proceed with degraded context. The user initially sees streaming progress (hypothesis trees, phase events), but the final stored result is the timeout error — so revisiting the session shows a failed investigation.

**Reproduction**: Call `deep_search` with only `question` + `budget` (no `triageContext`), where the cluster has enough complexity that Phase 1 context gathering takes > 5 minutes.

## Root Cause

`engine.ts` L397: `runSubAgent()` for Phase 1 has no `try/catch`, unlike Phase 3 hypothesis validation (which catches sub-agent errors and marks hypotheses as inconclusive).

## Fix

Wrap Phase 1's `runSubAgent` in `try/catch`. On failure, fall back to using the question as minimal context so Phases 2-4 can still run.

## Test plan

- [x] `npm test` — 1004/1004 passed
- [ ] Manual: run `deep_search` without triageContext on a complex cluster, verify investigation completes even if Phase 1 times out

### DIFF
--- a/src/tools/deep-search/engine.ts
+++ b/src/tools/deep-search/engine.ts
@@ -394,16 +394,25 @@ export async function investigate(
   } else {
     onProgress?.({ type: "phase", phase: "Phase 1/4", detail: "Gathering context..." });
     const contextPrompt = contextGatheringPrompt(question, budget.maxContextCalls, kubeconfigPath);
-    const contextResult = await runSubAgent(
-      contextPrompt,
-      question,
-      budget.maxContextCalls,
-      options,
-      onProgress,
-      forceContextSummaryPrompt(),
-    );
-    globalCallsUsed += contextResult.callsUsed;
-    contextSummary = parseContextSummary(contextResult.textOutput);
+    try {
+      const contextResult = await runSubAgent(
+        contextPrompt,
+        question,
+        budget.maxContextCalls,
+        options,
+        onProgress,
+        forceContextSummaryPrompt(),
+      );
+      globalCallsUsed += contextResult.callsUsed;
+      contextSummary = parseContextSummary(contextResult.textOutput);
+    } catch (err) {
+      // Phase 1 sub-agent timed out or crashed — fall back to question as context
+      // so Phases 2-4 can still proceed with degraded quality.
+      const msg = err instanceof Error ? err.message : String(err);
+      console.warn(`[deep-search] Phase 1 context gathering failed: ${msg}`);
+      onProgress?.({ type: "phase", phase: "Phase 1/4", detail: `Context gathering failed (${msg}), proceeding with question only` });
+      contextSummary = `Context gathering failed. Original question: ${question}`;
+    }
   }
 
   // --- Prior Investigation Retrieval (3-layer: patterns + similar + validated hypotheses) ---
--- a/src/tools/deep-search/engine.ts
+++ b/src/tools/deep-search/engine.ts
@@ -408,9 +408,12 @@ export async function investigate(
     } catch (err) {
       // Phase 1 sub-agent timed out or crashed — fall back to question as context
       // so Phases 2-4 can still proceed with degraded quality.
+      // Charge the full Phase 1 budget conservatively — the sub-agent may have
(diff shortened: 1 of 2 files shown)


jacoblee-io (commented):
# Review Report: fix(deep-search): catch Phase 1 sub-agent timeout to prevent investigation failure

## Summary
**Approve with suggestions.** Both fixes address real bugs — the Phase 1 timeout was a legitimate investigation-killer, and the leaked `setTimeout` was a genuine resource concern. A few edge cases in resource cleanup and budget accounting deserve attention.

5 findings: 0 🔴 critical, 3 🟡 warnings, 2 💬 notes

## 🟡 Should Fix

### [1] `session.dispose()` not called on timeout/error path — `sub-agent.ts:~417`
When `Promise.race` rejects with a timeout, the error re-throws past the try/finally, so `session.dispose()` (which lives after the try block) is never reached. The session object and its internal listeners/handles are left live. With the Phase 1 catch in `engine.ts` swallowing the error, the undisposed session accumulates silently. | **Dimension**: Correctness
**Suggestion**: Move `session.dispose()` into the `finally` block in `runSubAgent()` so it runs on both success and error paths.

### [2] Phase 1 catch block does not account for partial `callsUsed` — `engine.ts:~397`
When `runSubAgent` throws, `globalCallsUsed` is not incremented. If the Phase 1 sub-agent consumed tool calls before timing out, the Phase 3 budget calculation (`budget.maxTotalCalls - globalCallsUsed`) will overcount remaining budget. Compare with Phase 3's catch which always charges at least 1 call. | **Dimension**: Correctness
**Suggestion**: Add `globalCallsUsed += budget.maxContextCalls` (conservative) or `globalCallsUsed += 1` (minimum signal) in the catch block.

### [3] Orphaned `session.prompt()` continues after race timeout — `sub-agent.ts:~406`
When the race timer fires first, `session.prompt()` is still running but nobody calls `session.abort()`. The session continues executing internally (burning CPU, holding resources) despite `sessionFinished = true` guarding events out. | **Dimension**: Correctness
**Suggestion**: Track whether the timeout fired (e.g., `let timedOutByRace = false`) and call `session.abort().catch(() => {})` in the `finally` block before `session.dispose()`.

## 💬 Notes

### [4] Raw error message exposed in progress event — `engine.ts:~409`
The `onProgress` call includes `${msg}` directly from the error. If the error contains internal details (paths, stack traces), these flow to the TUI or WebSocket. Consider truncating: `msg.slice(0, 120)`.

### [5] No test coverage for the Phase 1 timeout fallback path
The new catch block is the core correctness claim of the PR, but no test exercises it. A test mocking `runSubAgent` to throw a timeout and asserting `investigate()` resolves with a degraded `contextSummary` would strengthen confidence.

## ✅ What Looks Good
- The Phase 1 fallback strategy (using the question as minimal context) is a pragmatic degradation that preserves Phases 2-4
- The `raceTimer` cleanup in `sub-agent.ts` correctly prevents the leaked `setTimeout` rejection — `clearTimeout` on an already-fired timer is harmless, so both paths are safe
- Clear, well-structured commit messages with root cause analysis in the PR description

jacoblee-io (approved):
All 4 findings from the previous review have been addressed in commit 39799f7:

1. ✅ `session.dispose()` moved into `finally` — guaranteed cleanup on both success and timeout paths
2. ✅ `globalCallsUsed += budget.maxContextCalls` in Phase 1 catch — conservative budget accounting
3. ✅ `session.abort()` called when `timedOutByRace` — stops orphaned `session.prompt()` from burning resources
4. ✅ `msg.slice(0, 120)` — error message truncated in progress event

LGTM.

PR #142, opened 2026-03-18
feat(feedback): add interactive session feedback system

LikiosSedo:
## Summary

- Adds a system-level interactive feedback mechanism: users click a button in WebUI, Gateway enriches the message with session timeline + feedback protocol instructions, agent guides multi-turn review, structured report saved to DB
- New `feedback_reports` table with full schema parity (SQLite + MySQL + migration + indexes)
- `save_feedback` tool registered for both pi-agent and claude-sdk brain paths
- Feedback API endpoint on both HTTP server (local mode) and HTTPS/mTLS server (K8s mode)
- Frontend feedback button only visible when session has messages (no empty-session feedback)

## Design Decisions

- **System-level UI trigger** — Gateway intercepts `[Feedback]` prefix and enriches with SKILL.md + diagnostic timeline, no changes to `prompt.ts`
- **sessionIdRef closure** — follows existing `run_skill` pattern instead of env vars
- **Dual server registration** — HTTP (local) + HTTPS/mTLS (K8s) for the feedback endpoint
- **Gateway DB storage** — `feedback_reports` table, independent from agent memory system (`userdata/memory/`)

## Data Flow

```
User clicks feedback button → chat.send("[Feedback]")
  → Gateway saves raw message to DB
  → Detects [Feedback] prefix → enriches with SKILL.md + session timeline
  → Agent receives full context, runs 4-phase feedback protocol
  → Agent calls save_feedback → POST /api/internal/feedback → feedback_reports table
```

## Test plan

- [x] `npx tsc --noEmit` passes (backend + frontend)
- [x] `npm test` — all 1004 tests pass
- [ ] Local Gateway startup → feedback button visible in WebUI after sending a message
- [ ] Click feedback button → agent starts review (timeline summary → numbered options)
- [ ] Complete feedback flow → verify `feedback_reports` row in DB
- [ ] Verify button hidden on empty/new sessions

### DIFF
--- a/skills/core/session-feedback/SKILL.md
+++ b/skills/core/session-feedback/SKILL.md
@@ -1,0 +1,70 @@
+---
+name: session-feedback
+description: Interactive session feedback — reviews the diagnostic process with the user, identifies decision points, and saves structured feedback to improve diagnostic capabilities.
+tags: [feedback, meta, improvement]
+---
+
+# Session Feedback Protocol
+
+You are now conducting an interactive feedback review of the current diagnostic session. Follow these phases strictly.
+
+## Phase 1 — Review
+
+Based on the Session Diagnostic Timeline provided, summarize the diagnostic process:
+
+1. List key steps in chronological order
+2. For each step, note:
+   - What was decided / what action was taken
+   - Which tool or skill was used
+   - The outcome (success/error/partial)
+3. Present as a numbered list in the user's language
+
+Keep it concise — focus on decision points, not every single tool call.
+
+## Phase 2 — Interactive Evaluation
+
+Present these options to the user (numbered for easy selection):
+
+1. **Overall direction** — Was the investigation heading in the right direction?
+2. **A specific step** — Was a particular step wrong or suboptimal?
+3. **Missing checks** — Were important diagnostics skipped?
+4. **Conclusion accuracy** — Was the final diagnosis correct?
+5. **All good** — Satisfied with the session, nothing to change
+6. **Free input** — Other feedback
+
+After the user selects an option, drill deeper:
+- What specifically was wrong?
+- What would the ideal action have been?
+- How would you rate it (1-5)?
+
+The user can provide feedback on **multiple aspects** — after finishing one, ask if they want to discuss another.
+
+## Phase 3 — Report Generation
+
+Once the user is done providing feedback, synthesize a structured report:
+
+- **Strengths**: What the agent did well
+- **Improvements**: What should be done differently next time
+- **Decision Points**: Each evaluated step with `wasCorrect`, `comment`, `idealAction`
+- **Tags**: Categorize issues (e.g., `wrong-skill`, `slow-path`, `missing-check`, `correct-diagnosis`, `wrong-order`)
+- **Overall Rating**: 1-5 based on discussion
+
+Present the report to the user for confirmation. If they want changes, adjust accordingly.
+
+## Phase 4 — Save
+
+After user confirms the report, call the `save_feedback` tool with the structured data:
+
+```
+save_feedback({
+  overallRating: <1-5>,
+  summary: "<brief summary>",
+  decisionPoints: "<JSON array>",
+  strengths: "<JSON array>",
+  improvements: "<JSON array>",
+  tags: "<JSON array>",
+  feedbackConversation: "<JSON summary of this dialogue>"
+})
+```
+
+After saving, thank the user and confirm the feedback has been recorded.
--- a/skills/core/session-feedback/SKILL.md
+++ b/skills/core/session-feedback/SKILL.md
@@ -10,9 +10,24 @@ You are conducting an interactive feedback review of the current diagnostic sess
 
 **Language: Always follow the user's language (from their profile or recent messages). All output — phase titles, timeline, options, report — must be in the user's language.**
 
+## Phase 0 — Scope Selection
+
+Scan the conversation history and identify distinct tasks or investigations performed in this session. Summarize each as a one-liner with what was done and the outcome.
+
+- If **1 task**: state what will be reviewed and proceed directly to Phase 1.
+- If **2+ tasks**: present them as numbered options using `- **1.** label` format, most recent first. Let user pick which to review.
+
+Example (3 tasks in session):
+
+- **1.** Node NotReady diagnosis on roce-test — found 3 nodes down, RDMA config mismatch suspected
+- **2.** Pod crash loop analysis on envoy-gateway — identified OOM, suggested resource limits
+- **3.** Review entire session
+
+After user selects (or if only 1 task), proceed to Phase 1 scoped to that task only.
+
 ## Phase 1 — Compressed Timeline with Self-Reflection
 
-Analyze the Session Diagnostic Timeline and present a **compressed timeline** as a regular markdown numbered list (NOT a code block):
+Analyze the selected task's diagnostic steps and present a **compressed timeline** as a regular markdown numbered list (NOT a code block):
 
 **Compression rules:**
 - Group consecutive same-purpose tool calls into one line, marked with `(N steps)`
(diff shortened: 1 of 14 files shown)


jacoblee-io (changes requested):
# Review Report: feat(feedback): add interactive session feedback system

## Summary
**Request changes.** The feature is well-scoped and the data model is solid, but several issues need attention — particularly around the `[Feedback]` prefix interception pattern and frontend edge cases.

13 findings: 3 🔴 critical, 7 🟡 warnings, 3 💬 notes

## 🔴 Must Fix

### [1] `[Feedback]` prefix as implicit RPC protocol — silently replaces user message — `rpc-methods.ts:541`
When enrichment succeeds, `promptText` is entirely replaced by the enriched content; any text the user appended after `[Feedback]` is discarded. Additionally, if a user naturally types a message starting with "[Feedback]", their message is silently intercepted. This is the only server-side prefix interception — other prefixes (`[Deep Investigation]`, `[Skill: name]`) are parsed client-side only, creating asymmetric behavior. | **Dimension**: Architecture + Correctness
**Suggestion**: Consider using a distinct RPC method (e.g., `prompt:feedback`) rather than intercepting message content. At minimum, append any user text after the prefix to the enriched prompt, and document `[Feedback]` as a reserved prefix.

### [2] Unauthenticated feedback endpoint on HTTP server — `server.ts:809`
The HTTP-server feedback endpoint trusts `data.userId` from the request body with no auth check. Unlike the HTTPS/mTLS path (which uses cert identity), anyone who can reach the HTTP port can forge feedback for any user. | **Dimension**: Architecture
**Note**: This follows the same pattern as existing `/api/internal/agent-prompt` and other internal endpoints, which also trust callers on the HTTP path. This is pre-existing architectural debt — not new to this PR — but worth calling out since the feedback endpoint stores user-attributed data.
**Suggestion**: If not fixing for all internal endpoints now, add a comment documenting the local-mode trust assumption.

### [3] Endpoint duplication pattern in server.ts — `server.ts:1075`
The feedback endpoint is copy-pasted for HTTP and HTTPS servers. The HTTP server has 3 internal endpoints not present in HTTPS. This pattern is acknowledged in comments but creates parity drift risk. | **Dimension**: Architecture
**Note**: This is pre-existing debt — the PR correctly follows the existing pattern. A shared `handleInternalRequest()` function would eliminate this class of bug for all future endpoints.
**Suggestion**: Extract shared handler (can be a follow-up, not necessarily blocking this PR).

## 🟡 Should Fix

### [4] Feedback button visible on empty sessions — `InputArea.tsx:127`
The PR description states "feedback button only visible when session has messages", but `InputArea` receives no `messages` prop. The button is accessible on brand-new empty sessions, producing an incoherent feedback review with `_No messages in this session._` as the timeline. | **Dimension**: Correctness
**Suggestion**: Pass `hasMessages={messages.length > 0}` from `PilotArea` and gate the feedback button.

### [5] Session timeline truncates at 200 messages with no indication — `rpc-methods.ts:111`
`buildSessionTimeline` fetches at most 200 messages. For long sessions, earlier messages are silently dropped and step numbering starts from message 201, misleading the feedback review. | **Dimension**: Correctness
**Suggestion**: Prepend `_(showing last 200 of N total messages)_` or fetch the total count.

### [6] `workspaceId` from env var is fragile in local mode — `save-feedback.ts:57`
The tool reads `workspaceId` from `process.env.SICLAW_WORKSPACE_ID`, which is not set per-user in local mode (LocalSpawner, in-process). Feedback records will have no workspace association for local users. | **Dimension**: Architecture
**Suggestion**: Pass `workspaceId` through a ref (like `sessionIdRef`) rather than env var.

### [7] ThinkingIndicator `setTimeout` inside `setInterval` leaks on unmount — `PilotArea.tsx:467`
If the component unmounts in the 300ms window between `setVisible(false)` and the `setTimeout` callback, the timer fires after unmount. | **Dimension**: Correctness
**Suggestion**: Track the `setTimeout` id and clear it in the interval's cleanup: `return () => { clearInterval(interval); clearTimeout(timeoutId); }`.

### [8] Feedback hint timer not cancelled on session switch — `PilotArea.tsx:203`
The 7s hint timer is not cleared when the user switches sessions. If session A shows the hint and the user switches to session B, session A's timer fires and hides session B's hint prematurely. | **Dimension**: Correctness
**Suggestion**: Store the timer id in a ref and clear it in the `sessionKey` effect.

### [9] `hasShownFeedbackHintRef` used before declaration — `PilotArea.tsx:183`
The `sessionKey` effect callback references `hasShownFeedbackHintRef` which is declared after it. This works at runtime (React defers effect execution), but is fragile. | **Dimension**: Correctness
**Suggestion**: Move the `useRef` declaration before the `sessionKey` `useEffect`.

### [10] `readFeedbackSkillContent()` reads file on every feedback request — `rpc-methods.ts:96`
Does a `fs.promises.readFile` on every `[Feedback]` prompt. The content is static and never changes at runtime. | **Dimension**: Architecture
**Suggestion**: Cache in a module-level variable on first read.

## 💬 Notes

### [11] Unsafe type assertion on gateway response — `save-feedback.ts:120`
Response cast as `{ ok: boolean; id: string }` with no runtime check. If the server returns 200 without `id`, the success message reads `"id: undefined"`.

### [12] `[Feedback]` literal not rendered as a UI chip — `InputArea.tsx:131`
Unlike `[Deep Investigation]` and `[Skill: name]` which get chip rendering in the chat, `[Feedback]` messages appear as raw text in the user bubble.

### [13] CHECK constraint asymmetry — `migrate-sqlite.ts:361` vs `schema-sqlite.ts:449`
`migrate-sqlite.ts` enforces `CHECK(overall_rating BETWEEN 1 AND 5)` but `schema-sqlite.ts` has no corresponding constraint. Not a runtime bug (TypeBox validates first), but makes the schema file a misleading source of truth. | **Dimension**: Invariants

## ✅ What Looks Good
- The feedback data model (`decision_points`, `strengths`, `improvements`, `tags`) is well-designed for future diagnostic path learning (IM Phase 2 alignment)
- Full schema parity across SQLite DDL, Drizzle SQLite, and Drizzle MySQL — all 11 columns match correctly
- The `MAX_CONVERSATION_BYTES` guard in `save-feedback.ts` (100KB cap with safe drop instead of truncation) prevents unbounded DB writes

LikiosSedo (commented):
# Review Report: feat(feedback): add interactive session feedback system

## Summary
Approve with fixes. Architecture is sound — feedback enrichment at Gateway layer is correct placement, DB schema has proper 3-file parity. One blocking issue from an untracked stray file, plus several should-fix items.

7 findings: 1 🔴 critical, 4 🟡 warnings, 2 💬 notes

### ⚠️ Process Note
Self-fix pattern detected: 10 commits including 4 fix rounds addressing prior review feedback. Consider squashing before merge.

### ⚠️ Reviewer Note
Correctness Reviewer did not complete (rate limited). Findings below cover Invariants + Architecture only.

## 🔴 Must Fix

### [1] Untracked `snapshot-repo.ts` references non-existent `sessionSnapshots` schema — `src/gateway/db/repositories/snapshot-repo.ts`
This untracked file imports `sessionSnapshots` from `schema.ts`, but no `session_snapshots` table exists in any schema file. Will crash at first call. Appears to be a leftover from a parallel work stream. | **Dimension**: Invariants

**Suggestion**: Either complete the schema (all 3 files) or remove the file before merge.

## 🟡 Should Fix

### [2] `[Feedback]` sentinel persisted as session title — `src/gateway/rpc-methods.ts:425-428`
`updateSessionMeta` sets the session title to `"[Feedback]"`. Sessions appear in the session list with an unhelpful title.

**Suggestion**: Skip `updateSessionMeta` for feedback-sentinel messages, or set a display-friendly title like "Session Feedback".

### [3] Migration DDL has CHECK constraint not in Drizzle schema — `migrate-sqlite.ts` vs `schema-sqlite.ts`
`migrate-sqlite.ts` includes `CHECK(overall_rating BETWEEN 1 AND 5)` but `schema-sqlite.ts` has just `integer("overall_rating")`. Schema and DDL out of sync.

**Suggestion**: Either add `.check(...)` to Drizzle schema or remove `CHECK` from DDL. Keep consistent.

### [4] TypeBox Integer min/max silently dropped for claude-sdk brain — `src/core/tool-adapter.ts`
`save_feedback` uses `Type.Integer({ minimum: 1, maximum: 5 })` for `overallRating`, but the tool-adapter doesn't propagate `minimum`/`maximum` to Zod. SDK sessions won't see the 1-5 constraint.

**Suggestion**: Extend `tool-adapter.ts` Integer case to propagate min/max.

### [5] Plain-HTTP feedback endpoint has no auth comment — `src/gateway/server.ts:867`
Accepts `userId` from request body with no verification. Consistent with existing local-trust pattern (ADR-006), but unlike read-only endpoints this one writes to DB.

**Suggestion**: Add explicit code comment cross-referencing ADR-006.

## 💬 Notes

### [6] `save_feedback` lacks brain-agnostic annotation in `agent-factory.ts`
Adjacent to `investigation_feedback` (pi-agent only) without a comment marking the boundary.

### [7] SKILL.md protocol is well-structured
4-phase design with scope disambiguation is appropriate complexity for multi-task sessions.

## ✅ What Looks Good
- Feedback enrichment correctly placed at Gateway layer (has DB access for timeline)
- DB schema has proper 3-file parity (`schema-sqlite.ts`, `schema-mysql.ts`, `migrate-sqlite.ts`)
- Dual server registration (HTTP + HTTPS) covers both deployment modes
- ESM cleanup: replaced `require("http")` with proper ESM import

jacoblee-io (changes requested):
# Review Report: feat(feedback): add interactive session feedback system

## Summary
Request changes. 9 verified findings: 3 critical, 6 warnings.

### Process Note
This PR has 11 commits, with 7 fixing bugs from earlier commits in the same PR (e64cae7, cdfc030, 9289cbc, 64991b2 all address issues from the initial implementation). Consider squashing fix-only commits.

---

## Must Fix

### [1] Unauthenticated feedback endpoint on public HTTP server

`POST /api/internal/feedback` is registered on the public HTTP server (line 871), which is the same port serving the Web UI. The comment at lines 773-775 explicitly says other `/api/internal/*` endpoints were moved to the mTLS-only HTTPS server "to enforce zero-trust security for AgentBox communication." This endpoint contradicts that pattern.

On the HTTP path, `certIdentity` is always undefined, so `userId` falls back to the POST body (line 82: `certIdentity?.userId || data.userId`), allowing any caller that can reach the HTTP port to write feedback records attributed to any user.

**Fix**: Remove the HTTP-server registration (line 871-877) and keep only the HTTPS/mTLS path. For local mode, validate `req.socket.remoteAddress` is loopback before trusting body-supplied `userId`.

https://github.com/scitix/siclaw/blob/64991b2da80c84373bbd6fd632d3c6460b1d2ee3/src/gateway/server.ts#L770-L777

https://github.com/scitix/siclaw/blob/64991b2da80c84373bbd6fd632d3c6460b1d2ee3/src/gateway/server.ts#L869-L878

### [2] No sessionId-to-userId ownership validation

The feedback endpoint accepts any `sessionId` + `userId` combination without verifying the session belongs to the user (line 82). Even with mTLS providing the correct `userId`, a compromised agent session could write feedback for another user's sessions by supplying a different `sessionId`.

**Fix**: Before saving, query `chatRepo.getSession(sessionId)` and verify `session.userId === userId`.

https://github.com/scitix/siclaw/blob/64991b2da80c84373bbd6fd632d3c6460b1d2ee3/src/gateway/server.ts#L79-L84

### [3] Feedback enrichment silently fails when chatRepo is null

At line 549: `if (message.startsWith("[Feedback]") && chatRepo)`. When `chatRepo` is null (TUI mode, no DB), the enrichment block is skipped and the raw `[Feedback]` sentinel is forwarded to the agent with no skill instructions and no timeline, producing a confused response with no error surfaced to the user.

**Fix**: When feedback is triggered but `chatRepo` is null, return an error or user-visible warning instead of silently sending the raw sentinel string.

https://github.com/scitix/siclaw/blob/64991b2da80c84373bbd6fd632d3c6460b1d2ee3/src/gateway/rpc-methods.ts#L548-L551

---

## Should Fix

### [4] `chipSeq` useRef mutation does not trigger React re-render

`chipSeq` is a `useRef(0)` (line 199) incremented in the chip click handler (line 450) and passed as `draftSeq={chipSeq.current}` to `InputArea` (line 474). Since ref mutations don't trigger re-renders, clicking the same chip twice sets the same `chipDraft` value — React bails out of re-rendering, `InputArea` never receives the updated `draftSeq`, and the `useEffect([draft, draftSeq])` doesn't fire. The code comment says "ensures useEffect fires even for repeated same-value drafts" but this goal is not achieved.

**Fix**: Replace `chipSeq` useRef with `useState<number>` so incrementing triggers a re-render.

https://github.com/scitix/siclaw/blob/64991b2da80c84373bbd6fd632d3c6460b1d2ee3/src/gateway/web/src/pages/Pilot/components/PilotArea.tsx#L198-L200

https://github.com/scitix/siclaw/blob/64991b2da80c84373bbd6fd632d3c6460b1d2ee3/src/gateway/web/src/pages/Pilot/components/PilotArea.tsx#L449-L451

### [5] `[Feedback]` sentinel stored as plain user message, pollutes timeline

The sentinel is stored via `chatRepo.appendMessage()` (line 480) as a regular user message. `incrementMessageCount` (line 485) is called unconditionally, inflating session counts. The title/preview guard at line 487 prevents updating session metadata, but the message and count are already persisted. Additionally, any user can manually type `[Feedback]` to trigger the enrichment — the sentinel is a UI convention with no integrity check.

**Fix**: Either use a separate RPC method instead of a sentinel prefix, or skip `incrementMessageCount` for sentinel messages and store with a metadata flag.

https://github.com/scitix/siclaw/blob/64991b2da80c84373bbd6fd632d3c6460b1d2ee3/src/gateway/rpc-methods.ts#L479-L493

### [6] Single JSON parse error aborts entire feedback save

All JSON field parsing (`decisionPoints`, `strengths`, `improvements`, `tags`, `feedbackConversation`) is in a single `try/catch` (lines 86-89). If any one LLM-generated field has malformed JSON, the entire feedback report is lost.

**Fix**: Parse each field independently so valid fields are saved even when one fails.

https://github.com/scitix/siclaw/blob/64991b2da80c84373bbd6fd632d3c6460b1d2ee3/src/tools/save-feedback.ts#L85-L90

### [7] Message content in timeline not sanitized

`buildSessionTimeline()` includes the first 100 characters of every message (line 125) in the enriched prompt without sanitization. If an earlier message contained an adversarial prompt injection payload, those characters are re-injected into the feedback prompt context.

**Fix**: Strip control characters or limit timeline to tool names and structural markers only.

https://github.com/scitix/siclaw/blob/64991b2da80c84373bbd6fd632d3c6460b1d2ee3/src/gateway/rpc-methods.ts#L124-L126

### [8] Duplicate `gatewayUrl` resolution pattern

Both `save-feedback.ts:60` and `manage-schedule.ts:110` contain identical `cfg.server.gatewayUrl || ...` patterns. CLAUDE.md explicitly flags duplicate helpers as a common review pitfall.

**Fix**: Extract `resolveGatewayUrl(cfg)` into a shared utility.

https://github.com/scitix/siclaw/blob/64991b2da80c84373bbd6fd632d3c6460b1d2ee3/src/tools/save-feedback.ts#L59-L61

https://github.com/scitix/siclaw/blob/64991b2da80c84373bbd6fd632d3c6460b1d2ee3/src/tools/manage-schedule.ts#L109-L111

### [9] Brain type cross-verification not documented

`save_feedback` uses TypeBox schema and is registered for all brain types. CLAUDE.md requires "When adding new tools, test both brain types (TypeBox for pi-agent, Zod/MCP for claude-sdk)." No test or note in the PR confirms claude-sdk verification.

**Fix**: Add a note or test confirming `save_feedback` works under claude-sdk brain.

https://github.com/scitix/siclaw/blob/64991b2da80c84373bbd6fd632d3c6460b1d2ee3/src/tools/save-feedback.ts#L4-L12

---

## What Looks Good
- SQLite DDL parity correct: `feedback_reports` in both `schema-sqlite.ts` and `migrate-sqlite.ts` with indexes
- Skill bundle contract respected: feedback skill in `skills/core/` (baked into image), no `materialize()` calls
- mTLS scope maintained: HTTPS handler correctly uses `certIdentity` from verified mTLS cert
- ESM import fix in `gateway-client.ts` replaces `require("http")` with proper `import`

jacoblee-io (changes requested):
## Re-review after commit 7d11d6e

Findings #3 (TUI guard), #4 (chipSeq useState), and #6 (independent JSON parsing) are correctly fixed. Three fixes verified, all look good.

### Still open (2 critical, 4 warnings)

**Must fix:**

1. **Unauthenticated feedback endpoint on public HTTP server** — `POST /api/internal/feedback` is registered on the HTTP server (line 871), contradicting the zero-trust pattern documented at lines 773-775. On the HTTP path `certIdentity` is always undefined, so `userId` falls back to the POST body (line 82), allowing any caller to write feedback as any user.

https://github.com/scitix/siclaw/blob/7d11d6e46d8614bcc138ab2b3a7fc52f4bc1a8df/src/gateway/server.ts#L869-L878

2. **No sessionId-to-userId ownership check** — The endpoint accepts any `sessionId`+`userId` without verifying the session belongs to the user (line 82). A compromised session could write feedback for another user's sessions.

https://github.com/scitix/siclaw/blob/7d11d6e46d8614bcc138ab2b3a7fc52f4bc1a8df/src/gateway/server.ts#L79-L84

**Should fix:**

3. **`[Feedback]` sentinel stored as plain user message** — Stored via `appendMessage` (line 480) with `incrementMessageCount` (line 485) called unconditionally, inflating session counts. Any user can type `[Feedback]` to trigger enrichment.

https://github.com/scitix/siclaw/blob/7d11d6e46d8614bcc138ab2b3a7fc52f4bc1a8df/src/gateway/rpc-methods.ts#L479-L493

4. **Message content in timeline not sanitized** — `buildSessionTimeline` includes first 100 chars of messages (line 125) in the enriched prompt without sanitization, creating a prompt injection surface.

https://github.com/scitix/siclaw/blob/7d11d6e46d8614bcc138ab2b3a7fc52f4bc1a8df/src/gateway/rpc-methods.ts#L124-L126

5. **Duplicate `gatewayUrl` resolution** — Identical pattern in `save-feedback.ts:60` and `manage-schedule.ts:110`. CLAUDE.md flags duplicate helpers as a review pitfall.

https://github.com/scitix/siclaw/blob/7d11d6e46d8614bcc138ab2b3a7fc52f4bc1a8df/src/tools/save-feedback.ts#L59-L61

6. **Brain type cross-verification** — `save_feedback` registered for all brain types but no test or note confirms claude-sdk verification per CLAUDE.md checklist.

jacoblee-io (approved):
LGTM — critical findings addressed or justified with sound rationale. YELLOW items are non-blocking.

PR #143, opened 2026-03-18
refactor(prompt): reduce system prompt token usage by ~50%

LikiosSedo:
## Problem

`buildSreSystemPrompt()` was ~6200 chars with 13 behavioral rules, many redundant with tool descriptions or system-level enforcement:
- Skill management semantics duplicated in `create_skill`/`update_skill` tool descriptions
- ~22 lines of UI navigation guides (sidebar paths, /setup steps) that an agent doesn't need
- kubectl read-only warnings despite `SAFE_SUBCOMMANDS` enforcing it at system level
- Multi-cluster credential selection flow already in `credential_list` output hints
- Memory tool usage details already in tool descriptions

## Solution

- Remove all instructions that are covered by tool descriptions or system constraints
- Delete entire Environment & Configuration section (replaced by one-line mode label in identity)
- Consolidate overlapping behavioral rules (13 → 7)
- Restructure from `string +=` concatenation to `parts[]` array + helper functions

### What's preserved (safety-critical, no system enforcement)

1. `credential_list` before first kubectl — behavioral guidance
2. NEVER output credential details — model could leak in response text
3. NEVER read credential files — bash tool could bypass
4. Tool output injection defense — only prompt-level defense
5. Credential paste handling

### Result

| Mode | Before | After | Reduction |
|------|--------|-------|-----------|
| CLI | ~6200 chars | ~2900 chars | -53% |
| Web | ~6100 chars | ~2900 chars | -52% |
| Channel | ~6200 chars | ~3200 chars | -48% |

No changes to function signature or callers (`agent-factory.ts` unchanged).


chent1996 (changes requested):
Good refactoring direction — modular structure is cleaner and the `.planning/` cleanup is correct. Compared to PR #159, this PR is better: it retains the critical "Know when to stop" 4-step escalation and "Conclusion first" rules without adding mandatory pre-checks overhead.

However, some deletions remove safety-critical or behaviorally-important content that has no alternative enforcement mechanism.

**Note**: This PR conflicts with PR #159 (both modify `prompt.ts`). Only one should be merged.

### Must restore

1. **Safety: "Default to read-only"** — The new Safety section only keeps tool-output injection defense. "Default to read-only. Never modify cluster state unless explicitly asked" and "Warn about impact before suggesting destructive operations" were deleted. These have **no system-level enforcement** — `SAFE_SUBCOMMANDS` only covers kubectl, not arbitrary bash commands. Without these rules, the agent could suggest `kubectl delete`, `helm uninstall`, or destructive bash scripts without warning.

2. **Credentials: multi-kubeconfig flow** — "If credential_list returns multiple kubeconfigs, present the list (names only) and ask which one to use. Then pass `--kubeconfig=<name>` (credential name, NOT a file path)" was deleted. Without this, the agent won't know the `--kubeconfig=<name>` syntax for multi-cluster environments. This was added to solve a real issue where agents used file paths instead of credential names.

### Should restore

3. **"Trust your tools"** — Without this, agents retry the same kubectl command 3 times hoping for different output. Common LLM behavior pattern that this rule specifically prevented.

4. **"One tool for kubectl"** — Without this, agents may use `pod_exec` or `node_exec` for kubectl operations instead of `bash`. These tools create debug pods, which is far more expensive.

### Acceptable deletions

- Environment & Configuration section (UI navigation guides) — agents don't need sidebar paths
- "Stay on topic" — covered by "Stay focused"
- "Sufficient info → stop" — partially merged into "Know when to stop"
- Writing Memory YYYY-MM-DD convention — acceptable simplification
- Detailed skill management guidance for web mode — covered by tool descriptions

PR #166, opened 2026-03-25
Fix skill diff: baseline tracking + personal fork eye icon

LikiosSedo:
## Problem

1. **Skill space diff showed full file as additions** instead of incremental diff. The old logic looked up a team skill by `dirName` for the baseline — failed for builtin forks, and the fallback hit a path traversal bug (`resolveUnderDir` with relative `skillsDir`).

2. **Personal forked skills had no diff viewer.** No eye icon existed on the skills list page, and no fork baseline was persisted at fork time.

3. **Eye icon showed unconditionally** in skill spaces, even for newly created skills with no meaningful baseline.

## Solution

**Fork-time baseline persistence** (both skillset and personal):
- Save fork source content as `published` tag at fork time, creating an immutable baseline snapshot
- Set `publishedVersion` atomically with the baseline write

**Improved `skill.diff` RPC**:
- Skillset: read own `published` tag first (not team skill by dirName), with fallback chain for pre-fix forks (builtin → team → skillset → personal source)
- Personal fork: new branch mirrors skillset logic — own `published` baseline, same fallback chain
- All branches return `baselineLabel` + `compareLabel` for dynamic frontend subtitles
- Existing publish-review path (non-fork approved skills) now also returns labels

**Frontend**:
- Eye icon visibility unified: `forkedFromId || publishedVersion > 0` (both skill space and personal list)
- Dynamic subtitle from backend labels instead of hardcoded strings
- Personal skills list: new diff modal + handler reusing `DiffViewerModal`

## Test plan

- [ ] Fork a global/builtin skill to skill space → diff shows "No changes detected"
- [ ] Edit the forked skill → diff shows only the delta
- [ ] Fork a global skill to personal → eye icon appears, diff shows "No changes detected"
- [ ] Edit the personal fork → diff shows incremental changes
- [ ] Approve a skill space skill → baseline advances, next diff is incremental from approved version
- [ ] Newly created (non-fork) skill in skill space → no eye icon until first approve
- [ ] Non-fork personal skill → no eye icon until first approve; after approve shows "Published → Draft"
- [ ] Pre-fix forks (no published tag) → fallback to fork source, no errors

### DIFF
--- a/src/gateway/rpc-methods.ts
+++ b/src/gateway/rpc-methods.ts
@@ -2790,10 +2790,16 @@ export function createRpcMethods(
 
       // Save content
       if (skillContentRepo) {
-        await skillContentRepo.save(id, "working", {
+        const forkedFiles = {
           specs: effectiveSpecs,
           scripts: effectiveScripts,
-        });
+        };
+        await skillContentRepo.save(id, "working", forkedFiles);
+        // Persist the fork source as the initial baseline so later diffs show
+        // what changed inside the Skill Space after the fork.
+        await skillContentRepo.save(id, "published", forkedFiles);
+        await skillRepo.update(id, { publishedVersion: inheritVersion });
+        console.log(`[skill.fork] Initialized Skill Space baseline for ${id} from ${sourceId}`);
       }
 
       // Notify all members to reload
@@ -2852,10 +2858,16 @@ export function createRpcMethods(
 
     // ── 6. Save content ──
     if (skillContentRepo) {
-      await skillContentRepo.save(id, "working", {
+      const forkedFiles = {
         specs: effectiveSpecs,
         scripts: effectiveScripts,
-      });
+      };
+      await skillContentRepo.save(id, "working", forkedFiles);
+      // Persist the fork source as the initial baseline so later diffs show
+      // what changed after the fork (same pattern as skillset forks).
+      await skillContentRepo.save(id, "published", forkedFiles);
+      await skillRepo.update(id, { publishedVersion: inheritVersion });
+      console.log(`[skill.fork] Initialized personal baseline for ${id} from ${sourceId}`);
     }
 
     const hasScripts = effectiveScripts && effectiveScripts.length > 0;
@@ -3152,6 +3164,7 @@ export function createRpcMethods(
       oldPrefix: string,
       newPrefix: string,
     ): string {
+      if (!oldFiles && !newFiles) return "Baseline not found — cannot compute diff.";
       const parts: string[] = [];
 
       // SKILL.md diff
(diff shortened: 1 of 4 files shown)


jacoblee-io (commented):
# Review: Fix skill diff baseline tracking + personal fork eye icon

**4 files, 190+/28-, 1 commit** — clean single-commit PR.

## 🟡 Should Fix

### [1] Fallback chain duplicated ~60 lines between skillset and personal fork — `rpc-methods.ts:3199-3260` / `rpc-methods.ts:3270-3330`

The two fallback blocks (builtin check → source lookup → label assignment) are nearly identical. If a new fork source type is added or a bug is found in one, the other will drift.

**Suggestion**: Extract a helper like `resolveBaselineFiles(meta, skillContentRepo, skillRepo, skillWriter): { files, label, prefix }` and call it from both branches.

### [2] Fallback path still passes unvalidated `dirName` to `readSkill` — `rpc-methods.ts:3210,3282`

For pre-fix forks missing a `published` baseline, `dirName` is extracted from `forkedFromId` and passed to `skillWriter.readSkill("builtin", dirName)`. If `forkedFromId` were ever `"builtin:../../something"`, this could read outside `skills/builtin/`. The primary path (reading own `published` tag) eliminates this for new forks, but old forks still hit the fallback.

**Suggestion**: Add a `resolvePathUnderDir` guard (or at minimum an assertion that `dirName` contains no `..`) before calling `readSkill`. Low urgency since `forkedFromId` is written at fork-creation time from validated source data, but defense-in-depth is the project convention.

## 💬 Notes

### [3] `buildFullDiff(null, null)` produces empty output silently — `rpc-methods.ts`

If both baseline and compare files are null after all fallbacks (e.g., fork source deleted, no published tag, and current working/staging both empty), the diff returns an empty string with no explanation. The user sees "No changes detected" which is misleading — there's nothing to compare.

**Suggestion**: Consider returning a distinct message like `"Baseline not found — cannot compute diff"` when `baselineFiles` is null, separate from the genuine "no changes" case.

### [4] `handlePersonalDiff` catch uses `err: any` — `index.tsx:~80`

Minor — matches the existing pattern in `SkillSpaceDetail.tsx` so it's consistent, but `unknown` with narrowing would be stricter TypeScript.

## ✅ What Looks Good
- **Primary fix is correct**: Reading own `published` tag first eliminates the fragile `dirName` lookup and the path traversal bug for new forks
- **Fork-time baseline persistence is atomic with the working copy save** — both `save("working")` and `save("published")` use the same `forkedFiles` object, ensuring the baseline exactly matches the initial fork state
- **Eye icon condition `forkedFromId || publishedVersion > 0`** correctly covers both forks (always have a baseline) and non-fork skills (only after first publish)
- **Dynamic labels from backend** — clean separation of concerns; frontend no longer hardcodes comparison descriptions

jacoblee-io (approved):
LGTM — all review findings addressed. Path traversal guard and buildFullDiff null guard both look correct.

PR #170, opened 2026-03-26
fix(skills): sync skill name between DB and SKILL.md frontmatter

LikiosSedo:
## Problem

Skill rename was broken: the `skill.update` handler treated `name` as immutable, silently ignoring UI renames. Additionally, DB `name` and SKILL.md frontmatter `name:` were never synchronized — they diverged from creation time (DB stored "New Custom Skill" while frontmatter had the template default "new-skill").

This caused:
- Renaming a skill bumped version but didn't change the displayed name
- Creating a new skill after renaming an old one with the same original name failed silently (dirName collision)
- Save errors in the SkillEditor were swallowed (`console.error` only)

## Solution

**Establish DB `name` as authoritative, frontmatter `name:` as a sync target:**

### Backend (`rpc-methods.ts`, `file-writer.ts`)
- Add `setFrontmatterName()` helper to `SkillFileWriter` with newline sanitization to prevent YAML injection
- `skill.create`: sync `params.name` into specs frontmatter before persisting
- `skill.update`: `params.name` is authoritative → synced into specs frontmatter; if no explicit name, extract from frontmatter as fallback for DB name. Handles the case where specs is not provided (patches existing specs in DB)
- `skill.create`: add display name uniqueness check (separate from dirName check) + auto-dedup dirName on collision to handle stale dirNames left after renames

### Frontend (`SkillEditor.tsx`)
- Bidirectional sync: name input change → updates frontmatter `name:` in specs textarea; editing frontmatter `name:` in textarea → updates name input
- New skill template uses `"New Custom Skill"` in frontmatter (was hardcoded `"new-skill"`)
- Save errors shown in a red banner instead of silently swallowed

## Test plan

- [ ] Create a new skill → verify YAML frontmatter shows `name: New Custom Skill` (not `name: new-skill`)
- [ ] Rename skill via name input → verify frontmatter `name:` updates in real-time
- [ ] Edit frontmatter `name:` in YAML editor → verify top name input updates
- [ ] Save renamed skill → refresh page → verify name persists
- [ ] Rename skill A, then create new skill with A's original name → should succeed
- [ ] Try creating a skill with a duplicate display name → should show error banner

### DIFF
--- a/src/gateway/skills/file-writer.ts
+++ b/src/gateway/skills/file-writer.ts
@@ -161,15 +161,30 @@ export class SkillFileWriter {
     return result;
   }
 
+  /** Split specs into { before, yaml, after } around the frontmatter block */
+  private splitFrontmatter(specs: string): { before: string; yaml: string; after: string } | null {
+    const match = specs.match(/^(---\n)([\s\S]*?)(\n---)([\s\S]*)$/);
+    if (!match) return null;
+    return { before: match[1], yaml: match[2], after: match[3] + match[4] };
+  }
+
+  /** Strip YAML quotes from a raw value string */
+  private unquoteYaml(raw: string): string {
+    const v = raw.trim();
+    if (v.startsWith("'") && v.endsWith("'")) return v.slice(1, -1).replace(/''/g, "'");
+    if (v.startsWith('"') && v.endsWith('"')) return v.slice(1, -1);
+    return v;
+  }
+
   /** Parse YAML frontmatter from SKILL.md content */
   parseFrontmatter(specs: string): { name: string; description: string } {
-    const match = specs.match(/^---\n([\s\S]*?)\n---/);
-    if (!match) return { name: "", description: "" };
-    const yaml = match[1];
+    const fm = this.splitFrontmatter(specs);
+    if (!fm) return { name: "", description: "" };
+    const { yaml } = fm;
 
-    // Extract name (always a simple string)
+    // Extract name (may be quoted with single or double quotes)
     const nameMatch = yaml.match(/^name:\s*(.+)$/m);
-    const name = nameMatch ? nameMatch[1].trim() : "";
+    const name = nameMatch ? this.unquoteYaml(nameMatch[1]) : "";
 
     // Extract description — handles both inline and block scalar (>- / >)
     let description = "";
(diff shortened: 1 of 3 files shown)


src/gateway/skills/file-writer.ts
```diff
@@ -194,6 +194,27 @@ export class SkillFileWriter {
     return { name, description };
   }
 
+  /** Replace the `name:` field inside YAML frontmatter, preserving everything else */
+  setFrontmatterName(specs: string, newName: string): string {
+    // Sanitize: strip newlines to prevent YAML injection
+    const safeName = newName.replace(/[\r\n]/g, "").trim();
+    if (!safeName) return specs;
+    const match = specs.match(/^(---\n)([\s\S]*?)(\n---)/);
+    if (!match) {
```
chent1996: **Bug: YAML special characters in skill name produce invalid frontmatter**

`setFrontmatterName` strips newlines but doesn't quote the value. Skill names are user input — if the name contains YAML-sensitive characters, the output is broken:

| Name | Output | Problem |
|---|---|---|
| `My Skill: Advanced` | `name: My Skill: Advanced` | Colon ambiguity — some parsers split at first `: ` |
| `# Overview` | `name: # Overview` | YAML comment — value becomes empty |
| `{production}` | `name: {production}` | YAML flow mapping — parsed as object |
| `It's a test` | `name: It's a test` | Single quote inside unquoted value |

Currently `parseFrontmatter` uses regex (`/^name:\s*(.+)$/m`) so it round-trips fine, but this is fragile — any future use of a real YAML parser (or external tools reading SKILL.md) will break.

**Fix**: Quote the value with single quotes (YAML spec):

```typescript
const safeName = newName.replace(/[\r\n]/g, "").trim();
if (!safeName) return specs;
// Single-quote to prevent YAML special char issues; escape inner quotes
const quoted = `'${safeName.replace(/'/g, "''")}'`;
// then use: `name: ${quoted}`
```

Same issue exists in the frontend `handleNameChange` and specs `onChange` — but since those are just preview and the backend is authoritative, fixing the backend is sufficient.

chent1996 (changes requested):
Good fix — establishing DB `name` as authoritative with frontmatter as sync target is the right design. The bidirectional sync in the editor, error banner, and display-name uniqueness checks are all solid improvements.

One real bug below.

jacoblee-io (commented):
### Review notes (non-blocking)

**1. Divergent frontmatter regex between `parseFrontmatter` and `setFrontmatterName`**

`parseFrontmatter` uses `/^---\n([\s\S]*?)\n---/` (one capture group), while `setFrontmatterName` uses `/^(---\n)([\s\S]*?)(\n---)/` (three capture groups) and reconstructs the string by concatenating `match[1] + updatedYaml + match[3] + specs.slice(match[0].length)`. These two parsing approaches are not the same abstraction — maintaining two separate frontmatter regexes is fragile. Consider extracting a shared helper that returns `{ before, yaml, after }` or similar so both functions use the same split logic.

**2. Frontend `handleNameChange` skips newline sanitization**

Backend `setFrontmatterName` strips `\r\n` to prevent YAML injection, but the frontend counterpart in `handleNameChange` directly interpolates `newName` into the YAML:
```tsx
yaml.replace(/^name:\s*.+$/m, `name: ${newName}`)
```
Not a security issue since `<input>` doesn't accept newlines and the backend re-sanitizes on save, but the asymmetry could bite if this code is later reused with a different input source.

**3. `parseFrontmatter` called twice on the same specs in `skill.create`**

Once during name sync (`skillWriter.parseFrontmatter(syncedSpecs).name`) and again for description extraction (`skillWriter.parseFrontmatter(syncedSpecs).description`). Minor, but could destructure a single call.

jacoblee-io (approved):
LGTM — name sync logic is correct and the DB-as-authority design is sound. Left a few non-blocking nits on regex consolidation and minor asymmetry.

chent1996 (approved):
All previous review findings addressed:

✅ **YAML quoting** — `setFrontmatterName` now single-quotes values (`'name'`), with `''` escaping for inner quotes. `parseFrontmatter` updated with `unquoteYaml()` to strip quotes on read. Round-trip safe for colons, hashes, braces, and quotes.

✅ **Name uniqueness on rename** — `skill.update` now checks for duplicate display names before applying, matching the existing `skill.create` behavior.

✅ **Input trimming** — `params.name` trimmed before use.

✅ **`splitFrontmatter` extracted** — shared helper eliminates the duplicated regex between `parseFrontmatter` and `setFrontmatterName`.

Clean fix, good to merge.

PR #197, opened 2026-04-01
fix(tools): remove kubectl exec from restricted-bash, consolidate to pod_exec/node_exec

chent1996:
## Summary

- Remove `exec` from `SAFE_SUBCOMMANDS` — `kubectl exec` via `restricted-bash` is no longer allowed
- Delete `validateExecCommand()` function and all related tests (kubectl.test.ts, sensitive-path-protection.test.ts)
- Add targeted error message for `kubectl exec` with hint to use `pod_exec`/`node_exec`
- Update agent-facing prompts (restricted-bash description, deep-search sre-knowledge) to guide toward `pod_exec` tool
- Update `dns-debug/SKILL.md` examples from `kubectl exec` to `pod_exec` syntax
- Fix DDL parity: add missing `user_disabled_skill_spaces` table to `migrate-sqlite.ts` DDL_STATEMENTS (introduced by PR #188, caught by ddl-parity test from PR #193)

## Motivation

`kubectl exec` through `restricted-bash` was a redundant execution path. The dedicated `pod_exec` and `node_exec` tools provide strictly better security:
- `pod_exec` uses `execFile` (no shell injection surface) with `blockPipeline: true`
- Both tools have their own `preExecSecurity` pipeline with pod/node name validation
- Consolidating removes a maintenance burden (two separate validation paths for the same operation)

## Test plan

- [x] `npx vitest run src/tools/infra/kubectl.test.ts` — 18 tests pass
- [x] Full test suite — all tests pass (local dep-missing failures are pre-existing)
- [x] DDL parity test passes
- [ ] Verify `kubectl exec` is rejected with actionable hint in restricted-bash
- [ ] Verify `pod_exec` tool still works for pod command execution

### DIFF
--- a/docs/design/invariants.md
+++ b/docs/design/invariants.md
@@ -160,12 +160,12 @@ These are **intentionally NOT in the allowlist** despite being common:
 Allowed subcommands (read-only):
 ```
 get, describe, logs, top, events, api-resources, api-versions,
-cluster-info, config, version, explain, auth, exec
+cluster-info, config, version, explain, auth
 ```
 
 **All write operations are permanently blocked**: `apply`, `create`, `delete`, `patch`, `scale`, `drain`, `cordon`, `edit`, `replace`, `label`, `taint`, `rollout undo`.
 
-`kubectl exec` commands pass through the binary allowlist (Pass 3) for the exec'd command.
+`kubectl exec` is not allowed as a subcommand — use the dedicated `pod_exec` or `node_exec` tools instead.
 
 ### 3.5 Skill Script Exemption
 
--- a/docs/design/security.md
+++ b/docs/design/security.md
@@ -478,23 +478,21 @@ Stripped patterns:
 
 ```
 Allowed: get, describe, logs, top, events, api-resources, api-versions,
-         cluster-info, config, version, explain, auth, exec
+         cluster-info, config, version, explain, auth
 ```
 
 **All write operations permanently blocked**: apply, create, delete, patch, scale, drain,
 cordon, edit, replace, label, taint, rollout undo.
 
-### 8.2 kubectl exec Restrictions
+`kubectl exec` is **not** allowed as a subcommand — use the dedicated `pod_exec` or
+`node_exec` tools instead, which provide their own security validation pipeline.
 
-When `kubectl exec` is used (e.g., `kubectl exec my-pod -- ip addr`), the exec'd command
-passes through the binary allowlist. Only commands in `ALLOWED_COMMANDS` can be exec'd.
-
-### 8.3 kubectl config view --raw Blocked
+### 8.2 kubectl config view --raw Blocked
 
 `kubectl config view --raw` is explicitly blocked — it outputs kubeconfig with embedded
 certificates and tokens in plaintext.
 
-**Source**: `src/tools/infra/command-sets.ts` — `SAFE_SUBCOMMANDS`, `validateExecCommand()`
+**Source**: `src/tools/infra/command-sets.ts` — `SAFE_SUBCOMMANDS`
 
 ---
 
--- a/docs/internal/architecture.md
+++ b/docs/internal/architecture.md
@@ -579,10 +579,10 @@ Every `kubectl` invocation is restricted to read-only subcommands:
 
 ```
 get, describe, logs, top, events, api-resources, api-versions,
-cluster-info, config, version, explain, auth, exec
+cluster-info, config, version, explain, auth
 ```
 
-All write operations — `apply`, `delete`, `patch`, `create`, `scale`, `drain`, `cordon`, `edit`, `replace`, `label`, `taint`, `rollout undo` — are rejected. For `kubectl exec`, the command after `--` goes through the same binary allowlist (Pass 2).
+All write operations — `apply`, `delete`, `patch`, `create`, `scale`, `drain`, `cordon`, `edit`, `replace`, `label`, `taint`, `rollout undo` — are rejected. `kubectl exec` is not allowed — use the dedicated `pod_exec` or `node_exec` tools instead.
 
 **Pass 4 — Per-Command Flag Restrictions** (`validateCommandRestrictions`)
 
--- a/docs/internal/command-whitelist.md
+++ b/docs/internal/command-whitelist.md
@@ -13,8 +13,8 @@ Commands are validated in two layers:
 
 These validations apply to three tools:
 - `restricted-bash` — sandboxed shell execution (also allows `kubectl` and skill scripts)
+- `pod-exec` — command execution inside Kubernetes pods
 - `node-exec` — command execution on Kubernetes nodes via privileged debug pods
-- `kubectl-exec` — command execution inside Kubernetes pods
 
 Additionally, `restricted-bash` validates **shell operators** (see [Shell Operator Restrictions](#shell-operator-restrictions)).
 
@@ -555,11 +555,11 @@ Allowed subcommands (`SAFE_SUBCOMMANDS` in `src/tools/infra/command-sets.ts`):
 
 ```
 get  describe  logs  top  events  api-resources
-api-versions  cluster-info  config  version  explain  auth  exec
+api-versions  cluster-info  config  version  explain  auth
 ```
 
-Note: `exec` is allowed as a kubectl subcommand, but the exec'd command passes
-through the binary allowlist for the applicable context. `kubectl config view --raw`
+Note: `exec` is **not** allowed as a kubectl subcommand — use the dedicated
+`pod_exec` or `node_exec` tools instead. `kubectl config view --raw`
 is explicitly blocked (would expose embedded certificates/tokens).
 
 ---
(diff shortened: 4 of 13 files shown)


LikiosSedo (commented):
## Review Report

**Request changes** — the exec consolidation is architecturally sound, but there are 4 issues to address before merge.

4 findings: 0 🔴 critical, 4 🟡 warnings, 2 💬 notes

## 🟡 Should Fix

### [1] DDL schema mismatch: extra `ON DELETE CASCADE` on `user_id` — `src/gateway/db/migrate-sqlite.ts:196`

The PR adds to `DDL_STATEMENTS`:
```sql
user_id TEXT NOT NULL REFERENCES users(id) ON DELETE CASCADE
```
But the Drizzle schema (`schema-sqlite.ts:365`) defines `user_id` **without** `ON DELETE CASCADE`:
```typescript
userId: text("user_id").notNull().references(() => users.id)  // no onDelete
```
And the existing MIGRATIONS entry (line 492) also has no cascade on `user_id`. This means fresh databases (via `DDL_STATEMENTS`) will have a different FK constraint than migrated databases — a subtle schema drift.

**Suggestion**: Remove `ON DELETE CASCADE` from the `user_id` line to match the Drizzle schema.

### [2] Missing test update: `deep-search.test.ts:63` will break — `src/tools/workflow/deep-search/deep-search.test.ts:63`

This test asserts `expect(semantics).toContain("kubectl exec")`. Since `sre-knowledge.ts` removes "kubectl exec" from `toolSemantics()`, this test will fail after merge.

**Suggestion**: Update to `expect(semantics).toContain("pod_exec")` and add `expect(semantics).not.toContain("kubectl exec")`.

### [3] Dead code in security-pipeline.ts not cleaned up — `src/tools/infra/security-pipeline.ts:128-147`

The `strategy === "auto"` branch specifically detects `kubectl exec` inner commands for output analysis. With `exec` removed from `SAFE_SUBCOMMANDS`, `kubectl exec` can never reach execution via restricted-bash, making this code path unreachable dead code.

**Suggestion**: Remove the `if (sub === "exec")` block (lines 129–146) and update the `analyzeTarget` type/comment to remove the "auto: detect kubectl exec" description. Check if any tests in `security-pipeline.test.ts` cover this path and remove them too.

### [4] Capability gap: `pod_exec` doesn't support piped output — `src/tools/cmd-exec/pod-exec.ts:129`

`pod_exec` uses `blockPipeline: true`, so `kubectl exec pod -- ps aux | grep nginx` style pipelines are lost. The PR's own `sre-knowledge.ts` previously documented piped `kubectl exec` as the canonical SRE pattern. While the agent can work around this (run `pod_exec`, then process output separately), this is a behavioral regression for deep-search sub-agents that relied on single-command piped diagnostics.

**Suggestion**: At minimum, document this trade-off in the PR description. Consider whether `pod_exec` should support a `--pipe` option in a follow-up, or add guidance in `sre-knowledge.ts` for the two-step pattern (pod_exec → bash grep).

## 💬 Notes

### [5] DDL duplicated across DDL_STATEMENTS and MIGRATIONS — `src/gateway/db/migrate-sqlite.ts:196,492`

After this PR, `user_disabled_skill_spaces` will exist in both `DDL_STATEMENTS` and the `MIGRATIONS` array. Both use `CREATE TABLE IF NOT EXISTS` so there's no runtime issue, but it's confusing to have two sources of truth in one file.

### [6] Unrelated DDL fix bundled with kubectl exec removal

Commit `7467be1` (DDL parity fix) is correct and needed, but has zero relation to the kubectl exec change. Consider splitting into its own PR for cleaner history.

## ✅ What Looks Good

- **Security parity confirmed**: `pod_exec` via `preExecSecurity` provides equivalent or stronger protection — full 6-pass pipeline, `CONTAINER_SENSITIVE_PATHS`, and `execFile` (no shell injection surface). No security regression.
- **Thorough doc updates**: All 4 documentation files consistently updated. Section renumbering in security.md handled correctly.
- **Clean test migration**: Removed tests properly replaced with a positive assertion that `exec` is excluded from `SAFE_SUBCOMMANDS`.

chent1996 (commented):
Thanks for the thorough review!

**[1] ✅ Fixed** — removed the extra `ON DELETE CASCADE` on `user_id` to match `schema-sqlite.ts` and the MIGRATIONS entry. See commit 51b7adb.

**[2] Won't break** — `sre-knowledge.ts:21` still contains the string `"(NOT kubectl exec via bash)"`, so `toContain("kubectl exec")` continues to pass. Verified locally.

**[3] Not dead code** — `preExecSecurity()` is called directly by `pod-exec` and `node-exec` without `pipelineValidators` (that's an optional param only `restricted-bash` passes). So the `resolveOutputAction` auto-strategy for kubectl exec is still reachable through those tools — it's valid defense-in-depth, not unreachable dead code.

**[4] Not a regression** — `pod_exec` was already the recommended tool before this PR; the previous `kubectl exec` via restricted-bash had the same limitation (each command in a pipeline is individually validated). The `sre-knowledge.ts` guidance has been updated to direct agents to `pod_exec` with post-processing.

**[5] Acknowledged** — this is the existing migration architecture design (MIGRATIONS for incremental, DDL_STATEMENTS for fresh installs). Not in scope for this PR.

**[6] Acknowledged** — the DDL fix is 6 lines and was needed to unblock CI. Splitting would add merge overhead without meaningful benefit.

jacoblee-io (commented):
## PR Review

### Critical (1)

**`skills/core/dns-debug/SKILL.md` still guides the agent to use `kubectl exec`**

Lines 21, 27, 33, 116 contain `kubectl exec <pod> -n <ns> -- ...` examples. After this PR, these commands will be rejected by `validateKubectlInPipeline`. The agent will follow these instructions, get blocked, and waste turns. Should be updated to use `pod_exec` tool syntax.

---

### Important (2)

**1. Error message is misleading and lacks actionable guidance**

When `kubectl exec` is rejected, the generic check at `restricted-bash.ts:68-72` produces:

```json
{"error": "kubectl subcommand \"exec\" is not allowed in read-only mode.", "allowed": [...]}
```

Two problems:
- "read-only mode" is misleading — `exec` isn't a write operation, it's blocked because dedicated tools exist
- No hint to use `pod_exec`/`node_exec` — the agent may try workarounds instead of switching tools

Suggested: add a targeted check before the generic rejection (same pattern as the `logs` rate-protection check):

```typescript
if (subcommand === "exec") {
  return JSON.stringify({
    error: "kubectl exec is not available through restricted_bash.",
    hint: "Use the pod_exec tool to run commands inside a pod, or node_exec for host-level diagnostics.",
  }, null, 2);
}
```

**2. Dead code in `security-pipeline.ts` — kubectl exec detection now unreachable**

`resolveOutputAction` in `security-pipeline.ts` (~lines 128-147) has a `sub === "exec"` branch that is now unreachable: `validateKubectlInPipeline` rejects `exec` before `resolveOutputAction` is ever called. The test at `security-pipeline.test.ts:72` ("detects kubectl exec inner command") still passes only because it omits the `pipelineValidators` the real restricted-bash tool provides — giving false coverage.

Options: (a) remove the dead branch + test, or (b) add a comment that it's defense-in-depth.

---

### Suggestions (2)

**1. Add behavioral test for exec rejection**

The current test only checks set membership (`SAFE_SUBCOMMANDS.has("exec") === false`). A behavioral test would be stronger:

```typescript
it("rejects kubectl exec (use pod_exec instead)", () => {
  const err = validateKubectlInPipeline(["kubectl exec my-pod -- ip addr"]);
  expect(err).toContain("not allowed");
});
```

This protects against regressions in the subcommand extraction logic.

**2. DDL commit could be noted in PR body**

Commit `7467be18` restores DDL parity for `user_disabled_skill_spaces` — valid fix, but unrelated to exec removal. Worth a line in the PR body.

---

### Positive

- Sound security consolidation — `pod_exec` uses `execFile` (no shell injection surface) with `blockPipeline: true`
- Thorough doc updates across 4 design/internal docs
- Clean removal — no stale `validateExecCommand` references remain in `src/`
- `sre-knowledge.ts` properly redirects deep-search agent to `pod_exec`
- `pod_exec`/`node_exec` already have comprehensive test suites covering all scenarios previously handled by `validateExecCommand`

chent1996 (commented):
Thanks for catching the SKILL.md issue — that was a real gap.

**[Critical] ✅ Fixed** — `dns-debug/SKILL.md` updated from `kubectl exec` to `pod_exec` tool syntax (4 occurrences). See 901d862.

**[Important-1] ✅ Fixed** — Added targeted `exec` check before the generic rejection, with actionable hint pointing to `pod_exec`/`node_exec`. Same pattern as the `logs` rate-protection check.

**[Important-2] Addressed with comment** — Added defense-in-depth comment on `security-pipeline.ts:129`. This code is not dead — `preExecSecurity()` is called by `pod-exec` and `node-exec` without `pipelineValidators` (optional param), so the kubectl exec detection in `resolveOutputAction` remains reachable as an independent security layer.

**[Suggestion-1] ✅ Fixed** — Added behavioral tests in `restricted-bash.test.ts` verifying exec rejection with actionable hint, including namespace flag variant.

**[Suggestion-2]** — Updated PR description to note the DDL parity fix.

jacoblee-io (approved):

PR #205, opened 2026-04-01
feat(cron): persist tool call traces for cron job executions

LikiosSedo:
## Summary

- Extract shared SSE consumer (`sse-consumer.ts`) from `chat.send` inline loop — both pilot and cron paths reuse it for DB persistence
- Each cron execution creates a fresh `source="cron"` session with full tool call + assistant message persistence in `messages` table
- `cron_job_runs.sessionId` links each execution to its audit session
- Cron sessions filtered from pilot session list via `sessions.source` discriminator
- `agent-prompt` endpoint gains `persistToDb` mode for DB persistence with proper abort signal handling

## Motivation

Cron jobs previously discarded all tool call data — the only output was a `resultText` string stored in `cron_job_runs`. There was zero audit trail for what tools were called, what arguments were passed, or what results came back.

The pilot (interactive) path already persists full tool call traces to the `messages` table. This PR makes cron reuse that same infrastructure.

## Design decisions

- **Fresh session per execution** (not persistent/reused) to avoid pi-agent compaction edge cases and JSONL growth
- **Shared SSE consumer** extracted as reusable module — pilot adds WS forwarding + DP tracking via `onEvent` callback, cron just needs the DB persistence + result extraction
- **`sessions.source`** field cleanly separates cron from pilot in all queries
- **`cron_job_runs.sessionId`** (not `cron_jobs.sessionId`) because sessions are per-execution, not per-job

## Schema changes

- `sessions`: add `source VARCHAR NOT NULL DEFAULT 'pilot'`
- `cron_job_runs`: add `session_id VARCHAR`
- Both SQLite (`schema-sqlite.ts`, `migrate-sqlite.ts`) and MySQL (`schema-mysql.ts`, `init-schema.ts`) paths updated

## Test plan

- [x] `npm test` — 1475 tests pass (43 files), zero failures
- [ ] Manual: create cron job → verify session created with `source="cron"` → verify tool calls in `messages` table → verify pilot list excludes cron sessions
- [ ] Manual: verify cron run history still shows status/resultText (existing UI unchanged)

### DIFF
--- a/src/gateway/cron/cron-service.ts
+++ b/src/gateway/cron/cron-service.ts
@@ -113,15 +113,16 @@ export class CronService {
     const workspaceId = current.workspaceId ?? undefined;
     console.log(`[cron-service] Executing job ${job.id} (${job.name}) for user ${job.userId}${workspaceId ? ` ws=${workspaceId}` : ""}`);
 
+    // Fresh session per execution — avoids compaction edge cases and JSONL growth
+    let sessionId: string | undefined;
     try {
-      // Fresh session per execution — avoids compaction edge cases and JSONL growth
       const cronSession = await this.chatRepo.createSession(
         current.userId,
         `Cron: ${current.name}`,
         workspaceId,
         "cron",
       );
-      const sessionId = cronSession.id;
+      sessionId = cronSession.id;
 
       const prompt = buildCronPrompt(current);
 
@@ -171,7 +172,7 @@ export class CronService {
       try { await this.configRepo.updateCronJobRun(job.id, "failure"); } catch (e) {
         console.warn(`[cron-service] updateCronJobRun failed for ${job.id}:`, e instanceof Error ? e.message : e);
       }
-      await this.recordResult(job, "failure", "", err instanceof Error ? err.message : String(err));
+      await this.recordResult(job, "failure", "", err instanceof Error ? err.message : String(err), undefined, sessionId);
     } finally {
       try {
         await this.configRepo.unlockJob(job.id, executionId);
--- a/src/gateway/cron/cron-service.ts
+++ b/src/gateway/cron/cron-service.ts
@@ -10,6 +10,7 @@ import crypto from "node:crypto";
 import { CronScheduler, type CronJobRow } from "../../cron/cron-scheduler.js";
 import type { ConfigRepository } from "../db/repositories/config-repo.js";
 import type { NotificationRepository } from "../db/repositories/notification-repo.js";
+import { ChatRepository } from "../db/repositories/chat-repo.js";
 import { CRON_LIMITS } from "../../cron/cron-limits.js";
 
 const EXECUTION_TIMEOUT_MS = 5 * 60 * 1000; // 5 minutes
@@ -29,6 +30,7 @@ export class CronService {
 
   private configRepo: ConfigRepository;
   private notifRepo: NotificationRepository;
+  private chatRepo: ChatRepository;
   private sendToUser: SendToUserFn;
   private gatewayPort: number;
   private staleLockTimer: NodeJS.Timeout | null = null;
@@ -38,11 +40,13 @@ export class CronService {
   constructor(opts: {
     configRepo: ConfigRepository;
     notifRepo: NotificationRepository;
+    chatRepo: ChatRepository;
     sendToUser: SendToUserFn;
     gatewayPort: number;
   }) {
     this.configRepo = opts.configRepo;
     this.notifRepo = opts.notifRepo;
+    this.chatRepo = opts.chatRepo;
     this.sendToUser = opts.sendToUser;
     this.gatewayPort = opts.gatewayPort;
     this.scheduler = new CronScheduler((job) => this.execute(job));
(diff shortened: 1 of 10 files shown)


LikiosSedo (commented):
## Review: feat(cron): persist tool call traces for cron job executions

This directly addresses the P1 audit finding (no execution audit trail for cron jobs). Architecture is clean — the shared SSE consumer extraction is the right call. A few issues:

### [Bug] task_report detection uses `lastToolName` instead of `toolName` — fragile with parallel tool calls

`sse-consumer.ts` line ~122 (in the `tool_execution_end` block):
```typescript
if (lastToolName === "task_report" && text) {
  taskReportText = text;
}
```

`lastToolName` tracks the last **started** tool, not the currently **ending** tool. With parallel tool calls:
1. `tool_execution_start` for `restricted_bash` → `lastToolName = "restricted_bash"`
2. `tool_execution_start` for `task_report` → `lastToolName = "task_report"`
3. `tool_execution_end` for `restricted_bash` → checks `lastToolName === "task_report"` → **incorrectly captures restricted_bash output as taskReportText**

Fix: use `toolName` (from `evt.toolName`, already extracted in the same block) instead of `lastToolName`:
```typescript
if (toolName === "task_report" && text) {
  taskReportText = text;
}
```

The `turn_end` path (claude-sdk) also uses `lastToolName` — that one is harder to fix since `turn_end` messages may not carry `toolName`. Consider storing a `Set<string>` of active tool names rather than a single `lastToolName`.

In practice cron jobs rarely hit this (task_report is almost always the last tool call), but it's a correctness issue that also affects the pilot path since it now shares this code.

### [Suggestion] Cron session accumulation

Each cron execution creates a new session. With 20 active jobs × hourly = 480 sessions/day per user. The existing purge cycle (180-day soft delete) means up to ~86k cron sessions per user before cleanup.

Options:
- Add a shorter retention for cron sessions (e.g., 30 days)
- Add `source` filter to the session purge logic
- At minimum, document this growth expectation

Not blocking — the current purge will handle it — but worth a TODO.

### [Nit] Type assertion in server.ts

```typescript
resultText = (sseResult as { resultText: string }).resultText;
```

`consumeAgentSse` already returns `SseConsumptionResult` — no cast needed:
```typescript
const sseResult = await Promise.race([...]);
resultText = sseResult.resultText;
```

(May need adjusting the Promise.race return type since `timeout.promise` throws rather than resolving.)

### [Nit] SSE summary log tag change

The pilot path previously logged `[rpc] SSE stream ended...` with event count + duration. That's now replaced by `[sse-consumer]` — fine, but the old log line in the pilot `finally` block was removed (lines 508-509). Intentional? The `[sse-consumer]` summary log covers it, just want to confirm.

### What looks good

- **Schema DDL parity**: both `schema-sqlite.ts`/`schema-mysql.ts` AND both migration files (`migrate-sqlite.ts`/`init-schema.ts`) updated — matches the CLAUDE.md invariant
- **`sessions.source` discriminator**: clean separation, backwards-compatible default `'pilot'`
- **`cron_job_runs.sessionId`**: nullable, correct for old runs. Per-execution (not per-job) — right design
- **`listSessions` filter**: `ne(sessions.source, "cron")` keeps cron out of pilot UI
- **Abort signal plumbing**: timeout → `abortCtrl.abort()` → breaks SSE loop — correct
- **`sessionId` outer scope**: failure path also records the audit session link — good catch in commit 2
- **User message persistence**: cron prompt saved before execution — complete audit trail

**Verdict**: Fix the `lastToolName` bug, rest is LGTM. This is a meaningful improvement to cron observability.

chent1996 (approved):
Well-structured extraction — the SSE consumer cleanly separates DB persistence (shared) from caller-specific logic (WS forwarding, DP tracking) via the `onEvent` callback.

Key verifications:
- ✅ Schema parity: `sessions.source` and `cron_job_runs.session_id` added consistently in MySQL DDL, SQLite DDL, MySQL Drizzle, SQLite Drizzle, and both migrations
- ✅ `listSessions` filters out `source="cron"` — pilot session list stays clean
- ✅ Fresh session per cron execution — avoids compaction edge cases and JSONL growth
- ✅ `task_report` detection uses `toolName` from the `tool_execution_end` event (not `lastToolName`) — correct for parallel tool calls
- ✅ `onEvent` callback fires AFTER DB writes — `dbMessageId` is available for WS forwarding
- ✅ `persistToDb` path has proper abort signal wiring (timeout → `abortCtrl.abort()`)
- ✅ Old `waitForAgentCompletion` preserved for non-persistent path (backward compat)

Good to merge.

chent1996 (commented):
Withdrawing my earlier APPROVE — need to investigate a potential sessionId mismatch issue before final verdict.

chent1996 (commented):
`persistToDb` flag and the `waitForAgentCompletion()` fallback path are unnecessary.

`agent-prompt` has exactly two callers: `cron-service` and `triggers` (gateway-main.ts:126) — both internal. There's no external API compatibility concern.

Keeping two code paths (`consumeAgentSse` vs `waitForAgentCompletion`) adds maintenance burden with no benefit. Triggers also execute commands that should be audit-recorded, so they'd need `persistToDb: true` anyway.

Suggest: always use `consumeAgentSse()` in the `agent-prompt` handler and remove `waitForAgentCompletion()` entirely.

chent1996 (approved):
Second-round fixes address the issues I identified:

✅ **AbortController leak fixed** — `finally { abortCtrl.abort() }` ensures cleanup regardless of success/timeout/error
✅ **Single code path** — `waitForAgentCompletion` deleted, both persistent and non-persistent go through `consumeAgentSse` (`chatRepo: null` skips DB writes)
✅ **Session filter hardened** — `eq(source, 'pilot')` instead of `ne(source, 'cron')` — won't leak future source types
✅ **toolName normalization** — `tool_execution_end` now checks both `evt.toolName` and `evt.name` for claude-sdk compat

Schema parity verified (sessions.source + cron_job_runs.session_id across all 4 schema files + 2 migrations). Good to merge.

PR #215, opened 2026-04-03
fix(web): improve abort UX with visual feedback and prompt restore

LikiosSedo:
## Summary

- Stop button shows spinning loader after first click to indicate abort is in progress
- Button remains clickable (not disabled) so user can retry if first abort didn't reach AgentBox
- On abort, last sent message is restored to the input box for easy re-send or edit

## Motivation

Users click Stop repeatedly because there's no feedback that the abort was sent. The agent may take time to finish a running tool call before it can honor the abort. Without visual feedback, users think the button is broken.

Additionally, after aborting, users typically want to re-send the same prompt (or a modified version). Restoring it to the input box saves the copy-paste step.

## Test plan

- [ ] Manual: send a message, click Stop — verify spinner appears, prompt returns to input box
- [ ] Manual: click Stop multiple times — verify each click sends abort (not blocked)
- [ ] Manual: agent finishes normally — verify button resets to square icon

### DIFF
--- a/src/gateway/web/src/pages/Pilot/components/InputArea.tsx
+++ b/src/gateway/web/src/pages/Pilot/components/InputArea.tsx
@@ -40,6 +40,7 @@ interface InputAreaProps {
 export function InputArea({ onSend, onAbort, disabled, isLoading, contextUsage, isCompacting, pendingMessages, onRemovePending, dpFocus, dpActive, onSetDpActive, hasMessages, draft, draftSeq }: InputAreaProps) {
     const [value, setValue] = useState('');
     const [isFocused, setIsFocused] = useState(false);
+    const [isAborting, setIsAborting] = useState(false);
     const isComposingRef = useRef(false);
     const textareaRef = useRef<HTMLTextAreaElement>(null);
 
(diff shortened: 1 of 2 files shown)


jacoblee-io (commented):
Two suggestions:

**1. `lastSentRef` captures steer and system messages too**

`wrappedSendMessage` records every call into `lastSentRef`, including steer messages (sent while `isLoading`) and the "dig deeper" button's hardcoded text. If user sends prompt → steers mid-run → aborts, the restored text is the steer message, not the original prompt. Aborting after "dig deeper" fills the input box with a long system-generated string.

Suggestion: only record `lastSentRef` on the primary `chat.send` path (when `!isLoading`), not for steers or system-generated prompts.

**2. Minor: `wrappedAbort` doesn't await `abortResponse`**

`abortResponse?.()` is fire-and-forget here, so `chipDraft` is set even if the abort RPC fails. Low impact since abort failure is rare, but worth noting.

Rest looks good — reusing `chipDraft`/`chipSeq` for the restore mechanism is clean, and the `useEffect` reset on `!isLoading` is correct.

jacoblee-io (approved):
LGTM — abort spinner feedback and prompt restore are solid UX wins. Suggestions about lastSentRef scope noted in previous comment, non-blocking.

PR #226, opened 2026-04-08
fix(command-sets): move -r/-R/--recursive block from category to per-command

chent1996:
Closes #218

## Summary
- Remove `-r/-R/--recursive` from `CONTEXT_POLICIES.local.categoryBlockedFlags.text` (was blocking all text commands)
- Add `blockedFlags: ["-r", "-R", "--recursive"]` to `grep`/`egrep`/`fgrep` CommandDef entries
- This unblocks `jq -r`, `sort -r`, `yq -r` in local context while keeping `grep -r` blocked
- Side effect: `grep -r` is now blocked in **all** contexts (previously only local), which is a security tightening

## Test plan
- [x] 292 existing tests pass
- [x] 8 new regression tests covering:
  - `grep -r` blocked in local/node/pod contexts
  - `grep -R` and `--recursive` blocked
  - `egrep/fgrep -r` blocked
  - `jq -r`, `sort -r`, `sort -rn`, `yq -r` allowed in local context
  - `grep` without `-r` allowed in local context

### DIFF
--- a/src/tools/infra/command-sets.test.ts
+++ b/src/tools/infra/command-sets.test.ts
@@ -1381,4 +1381,47 @@ describe("validateCommandRestrictions", () => {
       expect(validateCommandRestrictions("")).toBeNull();
     });
   });
+
+  describe("grep -r blocked per-command, not per-category (#218)", () => {
+    const local = { context: "local", piped: true };
+    const node  = { context: "node" };
+    const pod   = { context: "pod" };
+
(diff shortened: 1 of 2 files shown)


jacoblee-io (approved):
LGTM — clean fix, security tightened.

One minor nit: `categoryBlockedFlags: {}` can be removed entirely since the field is optional. No functional impact either way.

PR #237, opened 2026-04-23
feat(cli): unify TUI with local Portal via read-only snapshot

LikiosSedo:
## Summary

When a local Portal (`http://127.0.0.1:3000`) is running, the TUI now treats
Portal as the **source of truth** for skills, knowledge pages, SSH hosts,
kubeconfigs, and agent metadata — instead of reading from the operator's
on-disk directories. Standalone TUI (no Portal) keeps its legacy filesystem
behaviour unchanged.

This closes the loop started by PR #235 (embedded-SQLite local mode): the
SQLite Portal now actually drives the CLI it ships next to.

### CLI snapshot API

`GET /api/v1/cli-snapshot?agent=<name>` — Portal returns the full set of
resources the CLI needs for one session (skills, knowledge, hosts,
kubeconfigs, plus agent metadata for discovery UX). Core skills are
excluded; only global + personal-skillset + agent-bound content ships.
See `src/portal/cli-snapshot-api.ts`.

### TUI: ephemeral materialization

On startup, `siclaw` probes the local Portal and, on hit, unpacks the
snapshot into `.siclaw/.portal-snapshot/{skills,knowledge,credentials}`.
SIGINT / SIGTERM wipe the directory. Materializers live in `src/lib/`.

### New slash commands

- `/ls [skills|knowledge|mcp|credentials|agents]` — workspace snapshot,
  summary + subcommands for detailed listings
- `/agent` — lists Portal agents + current binding; edit/create is a hint
  linking to Portal Web (TUI stays read-only)
- `/setup` in Portal mode — switches to read-only list view with
  "Open in Portal →" links, so edits don't silently dead-end in the
  ephemeral dirs

### Agent scoping

`siclaw --agent <name>` fetches a per-agent snapshot scoped to that agent's
bound skills / credentials / knowledge / MCP config / model. Omitting
`--agent` returns a global union view.

### First-run wizard redirects to Portal

When Portal is reachable on first run (no `settings.json`), the wizard
prints Portal setup instructions, offers an interactive "open browser now?"
prompt (Y/n), and exits cleanly. Avoids a per-workstation "ghost provider"
in local `settings.json` that never reaches Portal and silently desyncs
across workstations. Standalone TUI without Portal keeps the existing
interactive preset picker.

### Skill filter

With a Portal snapshot active, the user-global skills that
pi-coding-agent's `DefaultResourceLoader` auto-discovers (e.g.
`~/.pi/agent/skills/`) are filtered out — otherwise the agent picks up
tools the Portal operator never opted into. Only skills whose path sits
under the Portal-materialized dir or the repo's `skills/platform/` remain.

## Test plan

- [x] `npm test` — 3128/3129 pass (the single failure is a pre-existing
  full-suite parallel flake in `restricted-bash.test.ts` unrelated to this
  PR; 8/8 pass when the file is run in isolation)
- [x] End-to-end smoke: `siclaw local` → create agent + host via REST API →
  `siclaw agents` lists the new agent → `siclaw --agent <name>` materializes
  skills + credentials into `.portal-snapshot/` → Portal `/settings/models`
  auto-opens in browser when provider is missing
- [x] Web UI smoke: portal-web built, served from Portal at `:3000`,
  login + agent/host CRUD round-trip

## Commits

- `fix(cli): create memory dir before initializing memory indexer`
- `feat(cli): unify TUI with local Portal via read-only snapshot`


chent1996 (changes requested):
# Review — TUI ↔ local Portal unification

Overall: the direction is right. Collapsing the TUI's three config surfaces (`settings.json` / filesystem skills-knowledge-creds / Portal DB) into "Portal owns state, TUI is a read-only viewer" is exactly the clean move out of the settings-drift corner we were heading into. Tests are reasonably covered. `needsSetup()` being rewritten to consult the merged snapshot was the right move and ties the first-run wizard into the Portal path cleanly.

Findings ordered by severity.

## 🔴 Blockers

### 1. `autoSeedAdminIfEmpty()` in `bootstrap-portal.ts` seeds `admin/admin` in prod too

`bootstrap-portal.ts:autoSeedAdminIfEmpty` runs on every `bootstrapPortal` call — which includes **`src/portal-main.ts`** (prod K8s), not just `cli-local.ts`. So a fresh K8s Portal deployment with `SICLAW_ADMIN_PASSWORD` unset boots with default credentials `admin/admin` on a service that's typically Ingress-exposed.

The comment says "matches the behaviour promised by README for local mode", but the code makes no distinction between local and prod. Either:
- gate it on a flag passed from the call site (`enableDefaultAdminSeed: true` only in `cli-local.ts`), or
- refuse to boot in prod without `SICLAW_ADMIN_PASSWORD` set.

The current loud console warning is not a substitute for not doing it in the first place.

### 2. Snapshot endpoint is only guarded by a config flag + admin JWT

`registerCliSnapshotRoute` uses plain `requireAuth(req, jwtSecret)` — *any* valid admin JWT passes. In prod, if someone flips `enableCliSnapshot=true` (for debugging, or by accident in a helm values override), every admin gets a blob that includes all providers' `api_key`, all cluster `kubeconfig` YAML, all host `password` / `private_key`. This is basically a "dump everything" endpoint that bypasses the per-resource binding checks every other adapter route enforces.

Two things would harden it without much cost:

1. **Bind to 127.0.0.1 only.** In the handler, check `req.socket.localAddress` / the request's `Host` header, or better: have `startPortal` optionally register a **separate** loopback-only listener for this route. Then `enableCliSnapshot` toggling in K8s simply has no network surface to leak from.
2. **Use a dedicated internal secret, not the JWT secret.** Have `bootstrapPortal` mint a one-shot `cliSnapshotSecret` into `local-secrets.json`, and verify that via header (same shape as `portalSecret`). Don't let the TUI forge arbitrary-role JWTs using `jwtSecret`. Right now `jwt.sign({sub: "cli-local", role: "admin"}, secrets.jwtSecret)` is a self-issued admin credential that would also authenticate against every OTHER admin route on Portal — that's a broader blast radius than necessary for "read a snapshot."

### 3. Knowledge tar extraction has no path-traversal guard

`portal-knowledge-materializer.ts:41`:
```ts
execFileSync("tar", ["xzf", tmpPath, "-C", outDir], { stdio: "pipe" });
```

A tarball with `../../../../tmp/evil.sh` entries will escape `outDir`. In local mode the DB is yours, so low real risk, but the defensive pattern is free:

```
tar --no-absolute-names --no-same-owner -xzf ...
```

plus a `path.relative(outDir, entry)` post-extraction walk asserting no `..`/absolute results. Same hygiene applies to the skill materializer's `UNSAFE_NAME = /(^\.\.$)|(\/)|(\\)|(\0)/` — it misses leading `..foo`, leading `~`, absolute paths, Windows reserved names. Replace with the `path.resolve` + `path.relative` pattern below; it catches the whole class:

```ts
const resolved = path.resolve(outDir, name);
if (!resolved.startsWith(outDir + path.sep)) continue;
```

## 🟡 Concerns

### 4. Cleanup handlers leak on normal exit + stack on SIGINT

`cli-main.ts` registers three independent `process.on("SIGINT", ...)` handlers (skills, knowledge, credentials cleanup). Node fires them in registration order, but none of them call `process.exit()` — which is actually fine for running all three — but there's no `process.on("exit", ...)` or `beforeExit` hook, so a clean `process.exit(0)` on the happy path leaves `.siclaw/.portal-snapshot/` on disk. `.portal-snapshot/credentials/` contains plaintext SSH keys and kubeconfigs; leaving it behind between sessions is the kind of mistake that gets noticed the first time one lands in a `git add -A`.

Consolidate into one cleanup function, register on `SIGINT` + `SIGTERM` + `exit`.

### 5. `setPortalSnapshot()` is module-level mutable state

`core/config.ts` holds a module-scoped `snapshotOverride` and invalidates the `cached` config when it's set. Works today because `cli-main.ts` calls `setPortalSnapshot()` exactly once before any `loadConfig()`. Fragile if any code path ever calls `loadConfig()` earlier (some future plugin, say) — they'll get a cached no-snapshot config that quietly never gets invalidated.

Worth either (a) making `loadConfig()` take an optional snapshot arg and dropping the module state, or (b) asserting single-set semantics at runtime so a second call throws rather than silently swaps providers out from under in-flight callers.

### 6. Multiple knowledge tars unpack into the same flat dir — silent shadowing

The comment in `portal-knowledge-materializer.ts` acknowledges "later wins" but doesn't log the collision. If Portal has two repos shipping `overview.md`, the agent sees one and the operator has no signal about which. At minimum, list which filenames got overwritten during extraction (walk `outDir` before and after each repo). Better: namespace by repo — `<outDir>/<repoName>/*.md`.

### 7. No size cap on snapshot fetch

`fetch(...)` + base64-decode of `dataBase64` happens fully in memory. A misconfigured / malicious Portal returning a 500 MB tar will OOM the TUI before tar even runs. Cap the response body client-side (`Content-Length` sanity check, or `TransformStream` that aborts past N MB). 50 MB feels generous for a wiki.

### 8. Empty `api_key` provider passes through silently

`cli-snapshot-api.ts:apiKey: p.api_key ?? ""` means a provider row with NULL api_key becomes `apiKey: ""` on the wire. TUI then initialises a provider with no key and fails on first request with an unclear error. Filter on the Portal side: only emit providers with a non-empty `api_key`. Similarly, if `is_default` points at a provider with no key, walk to the next valid one.

### 9. `probeLocalPortal()` is cwd-scoped — document the binding

TUI reads `./.siclaw/local-secrets.json` to decide whether Portal is reachable. So `siclaw` in `~/projectA` binds to the `siclaw local` running from `~/projectA`, but `siclaw` in `~/projectB` will not see `~/projectA`'s Portal even if it's reachable at `127.0.0.1:3000`. This is a reasonable model for isolation, but it means "only one `siclaw local` per cwd" becomes a latent rule. Call it out in the PR description / README, and consider a `SICLAW_PORTAL_SECRETS_PATH` override so a shared workstation Portal can be addressed from multiple cwds without copy-pasting the secrets file.

### 10. No-TTY detection is incomplete

`cli-main.ts:L40`:
```ts
if (isPrintMode || !process.stdin.isTTY) { ... process.exit(1); }
```
Good, catches the common pipe case. But `isTTY` can also be `undefined`. `!process.stdin.isTTY` happens to be `true` when it's `undefined`, so this is fine by coincidence — worth an explicit `process.stdin.isTTY !== true` for readability.

## 🟢 Praise

- **`needsSetup()` rewrite through `loadConfig()`** — exactly right. Makes Portal snapshot the single flag for "am I configured", no special-cases upstream.
- **`/ls` + `/agent` command design** — read-only, consistent style, clear pointer to "edit in Portal Web UI → restart". Resists the temptation to build in-session mutation.
- **`/setup` portal-mode degradation** — swapping write operations for "Open in Portal" links is the right UX: no silent edits into an ephemeral dir, no "why did my change disappear" support tickets.
- **`portalSkillsDir` + `skillsOverride` filter in `agent-factory.ts`** — catches the exact class of bug where `DefaultResourceLoader`'s auto-discovery of `~/.pi/agent/skills/` would contaminate the Portal snapshot. The comment explains the why well.
- **memoryDir mkdir moved to before indexer init** — genuine fix with a clear commit comment. Two lines of deleted code after the rebase, no stub left behind.
- **`cli-agents.ts` as a standalone shell-parseable listing** — clean separation from TUI boot, fits Unix tab-completion use case. Good test coverage too.
- **`systemPromptTemplate: portalSnapshot?.activeAgent?.systemPrompt`** — agent-scoped custom prompts fall through cleanly to the existing template mechanism, no new plumbing.

## Nits

- `parseQuery` in `cli-snapshot-api.ts` re-invents `URLSearchParams` — `new URL(req.url!, "http://x").searchParams` is one line.
- `payload.portalUrl = baseUrl` mutates the server-returned object to add a client-side field — fine, but future readers will wonder why the server type has `portalUrl?`. Wrap in a `PortalSnapshotWithMeta` type instead.
- Skill query uses `org_id = 'default'` hardcoded twice (for the `NOT IN` subquery too). Matches existing pattern, but reinforces the "one org for now" invariant. Worth a line in the design doc if you haven't already.
- `cli-agents.ts` uses top-level `await`; ESM supports it from Node 14.8+ so fine for the declared engine (`>=22.12.0`).

## Tests

- Existing tests are good; what's missing:
  - **Security test** for the snapshot endpoint: unauthenticated request returns 401 and leaks nothing.
  - **Path-traversal test** for each materializer with a malicious payload (`../../etc/hosts` in skill name, tarball with absolute paths).
  - **E2E** `siclaw local` + `siclaw --agent X` — mentioned in the PR description as manual smoke; worth automating even as a shell-level happy-path check.

---

Shipping blocker is #1 (prod `admin/admin` seed); #2 and #3 are important but fixable with small surgical changes. Rest are "before or shortly after merge" cleanups. Net: solid foundation, ship after the three blockers are addressed.

chent1996 (changes requested):
# Review — TUI ↔ local Portal unification

Overall: the direction is right. Collapsing the TUI's three config surfaces (`settings.json` / filesystem skills-knowledge-creds / Portal DB) into "Portal owns state, TUI is a read-only viewer" is exactly the clean move out of the settings-drift corner we were heading into. Tests are reasonably covered. `needsSetup()` being rewritten to consult the merged snapshot was the right move and ties the first-run wizard into the Portal path cleanly.

Findings ordered by severity.

## 🔴 Blockers

### 1. `autoSeedAdminIfEmpty()` in `bootstrap-portal.ts` seeds `admin/admin` in prod too

`bootstrap-portal.ts:autoSeedAdminIfEmpty` runs on every `bootstrapPortal` call — which includes **`src/portal-main.ts`** (prod K8s), not just `cli-local.ts`. So a fresh K8s Portal deployment with `SICLAW_ADMIN_PASSWORD` unset boots with default credentials `admin/admin` on a service that's typically Ingress-exposed.

The comment says "matches the behaviour promised by README for local mode", but the code makes no distinction between local and prod. Either:
- gate it on a flag passed from the call site (`enableDefaultAdminSeed: true` only in `cli-local.ts`), or
- refuse to boot in prod without `SICLAW_ADMIN_PASSWORD` set.

The current loud console warning is not a substitute for not doing it in the first place.

### 2. Snapshot endpoint is only guarded by a config flag + admin JWT

`registerCliSnapshotRoute` uses plain `requireAuth(req, jwtSecret)` — *any* valid admin JWT passes. In prod, if someone flips `enableCliSnapshot=true` (for debugging, or by accident in a helm values override), every admin gets a blob that includes all providers' `api_key`, all cluster `kubeconfig` YAML, all host `password` / `private_key`. This is basically a "dump everything" endpoint that bypasses the per-resource binding checks every other adapter route enforces.

Two things would harden it without much cost:

1. **Bind to 127.0.0.1 only.** In the handler, check `req.socket.localAddress` / the request's `Host` header, or better: have `startPortal` optionally register a **separate** loopback-only listener for this route. Then `enableCliSnapshot` toggling in K8s simply has no network surface to leak from.
2. **Use a dedicated internal secret, not the JWT secret.** Have `bootstrapPortal` mint a one-shot `cliSnapshotSecret` into `local-secrets.json`, and verify that via header (same shape as `portalSecret`). Don't let the TUI forge arbitrary-role JWTs using `jwtSecret`. Right now `jwt.sign({sub: "cli-local", role: "admin"}, secrets.jwtSecret)` is a self-issued admin credential that would also authenticate against every OTHER admin route on Portal — that's a broader blast radius than necessary for "read a snapshot."

### 3. Knowledge tar extraction has no path-traversal guard

`portal-knowledge-materializer.ts:41`:
```ts
execFileSync("tar", ["xzf", tmpPath, "-C", outDir], { stdio: "pipe" });
```

A tarball with `../../../../tmp/evil.sh` entries will escape `outDir`. In local mode the DB is yours, so low real risk, but the defensive pattern is free:

```
tar --no-absolute-names --no-same-owner -xzf ...
```

plus a `path.relative(outDir, entry)` post-extraction walk asserting no `..`/absolute results. Same hygiene applies to the skill materializer's `UNSAFE_NAME = /(^\.\.$)|(\/)|(\\)|(\0)/` — it misses leading `..foo`, leading `~`, absolute paths, Windows reserved names. Replace with the `path.resolve` + `path.relative` pattern below; it catches the whole class:

```ts
const resolved = path.resolve(outDir, name);
if (!resolved.startsWith(outDir + path.sep)) continue;
```

## 🟡 Concerns

### 4. Cleanup handlers leak on normal exit + stack on SIGINT

`cli-main.ts` registers three independent `process.on("SIGINT", ...)` handlers (skills, knowledge, credentials cleanup). Node fires them in registration order, but none of them call `process.exit()` — which is actually fine for running all three — but there's no `process.on("exit", ...)` or `beforeExit` hook, so a clean `process.exit(0)` on the happy path leaves `.siclaw/.portal-snapshot/` on disk. `.portal-snapshot/credentials/` contains plaintext SSH keys and kubeconfigs; leaving it behind between sessions is the kind of mistake that gets noticed the first time one lands in a `git add -A`.

Consolidate into one cleanup function, register on `SIGINT` + `SIGTERM` + `exit`.

### 5. `setPortalSnapshot()` is module-level mutable state

`core/config.ts` holds a module-scoped `snapshotOverride` and invalidates the `cached` config when it's set. Works today because `cli-main.ts` calls `setPortalSnapshot()` exactly once before any `loadConfig()`. Fragile if any code path ever calls `loadConfig()` earlier (some future plugin, say) — they'll get a cached no-snapshot config that quietly never gets invalidated.

Worth either (a) making `loadConfig()` take an optional snapshot arg and dropping the module state, or (b) asserting single-set semantics at runtime so a second call throws rather than silently swaps providers out from under in-flight callers.

### 6. Multiple knowledge tars unpack into the same flat dir — silent shadowing

The comment in `portal-knowledge-materializer.ts` acknowledges "later wins" but doesn't log the collision. If Portal has two repos shipping `overview.md`, the agent sees one and the operator has no signal about which. At minimum, list which filenames got overwritten during extraction (walk `outDir` before and after each repo). Better: namespace by repo — `<outDir>/<repoName>/*.md`.

### 7. No size cap on snapshot fetch

`fetch(...)` + base64-decode of `dataBase64` happens fully in memory. A misconfigured / malicious Portal returning a 500 MB tar will OOM the TUI before tar even runs. Cap the response body client-side (`Content-Length` sanity check, or `TransformStream` that aborts past N MB). 50 MB feels generous for a wiki.

### 8. Empty `api_key` provider passes through silently

`cli-snapshot-api.ts:apiKey: p.api_key ?? ""` means a provider row with NULL api_key becomes `apiKey: ""` on the wire. TUI then initialises a provider with no key and fails on first request with an unclear error. Filter on the Portal side: only emit providers with a non-empty `api_key`. Similarly, if `is_default` points at a provider with no key, walk to the next valid one.

### 9. `probeLocalPortal()` is cwd-scoped — document the binding

TUI reads `./.siclaw/local-secrets.json` to decide whether Portal is reachable. So `siclaw` in `~/projectA` binds to the `siclaw local` running from `~/projectA`, but `siclaw` in `~/projectB` will not see `~/projectA`'s Portal even if it's reachable at `127.0.0.1:3000`. This is a reasonable model for isolation, but it means "only one `siclaw local` per cwd" becomes a latent rule. Call it out in the PR description / README, and consider a `SICLAW_PORTAL_SECRETS_PATH` override so a shared workstation Portal can be addressed from multiple cwds without copy-pasting the secrets file.

### 10. No-TTY detection is incomplete

`cli-main.ts:L40`:
```ts
if (isPrintMode || !process.stdin.isTTY) { ... process.exit(1); }
```
Good, catches the common pipe case. But `isTTY` can also be `undefined`. `!process.stdin.isTTY` happens to be `true` when it's `undefined`, so this is fine by coincidence — worth an explicit `process.stdin.isTTY !== true` for readability.

## 🟢 Praise

- **`needsSetup()` rewrite through `loadConfig()`** — exactly right. Makes Portal snapshot the single flag for "am I configured", no special-cases upstream.
- **`/ls` + `/agent` command design** — read-only, consistent style, clear pointer to "edit in Portal Web UI → restart". Resists the temptation to build in-session mutation.
- **`/setup` portal-mode degradation** — swapping write operations for "Open in Portal" links is the right UX: no silent edits into an ephemeral dir, no "why did my change disappear" support tickets.
- **`portalSkillsDir` + `skillsOverride` filter in `agent-factory.ts`** — catches the exact class of bug where `DefaultResourceLoader`'s auto-discovery of `~/.pi/agent/skills/` would contaminate the Portal snapshot. The comment explains the why well.
- **memoryDir mkdir moved to before indexer init** — genuine fix with a clear commit comment. Two lines of deleted code after the rebase, no stub left behind.
- **`cli-agents.ts` as a standalone shell-parseable listing** — clean separation from TUI boot, fits Unix tab-completion use case. Good test coverage too.
- **`systemPromptTemplate: portalSnapshot?.activeAgent?.systemPrompt`** — agent-scoped custom prompts fall through cleanly to the existing template mechanism, no new plumbing.

## Nits

- `parseQuery` in `cli-snapshot-api.ts` re-invents `URLSearchParams` — `new URL(req.url!, "http://x").searchParams` is one line.
- `payload.portalUrl = baseUrl` mutates the server-returned object to add a client-side field — fine, but future readers will wonder why the server type has `portalUrl?`. Wrap in a `PortalSnapshotWithMeta` type instead.
- Skill query uses `org_id = 'default'` hardcoded twice (for the `NOT IN` subquery too). Matches existing pattern, but reinforces the "one org for now" invariant. Worth a line in the design doc if you haven't already.
- `cli-agents.ts` uses top-level `await`; ESM supports it from Node 14.8+ so fine for the declared engine (`>=22.12.0`).

## Tests

- Existing tests are good; what's missing:
  - **Security test** for the snapshot endpoint: unauthenticated request returns 401 and leaks nothing.
  - **Path-traversal test** for each materializer with a malicious payload (`../../etc/hosts` in skill name, tarball with absolute paths).
  - **E2E** `siclaw local` + `siclaw --agent X` — mentioned in the PR description as manual smoke; worth automating even as a shell-level happy-path check.

---

Shipping blocker is #1 (prod `admin/admin` seed); #2 and #3 are important but fixable with small surgical changes. Rest are "before or shortly after merge" cleanups. Net: solid foundation, ship after the three blockers are addressed.

LikiosSedo:
Round-2 follow-ups pushed as `3dab2a0`. Addressed the cheap-and-real concerns from your review:

## Fixed

### Concern #7 — snapshot fetch size cap
`portal-snapshot-client.ts` now enforces a 50 MB cap on snapshot responses:
- Rejects up-front when `Content-Length` exceeds the cap.
- Streams the body with a running byte counter and aborts mid-read if chunked responses go past the cap (cancels the reader so we never buffer past the threshold).
- Tests cover both the advertised-length and chunked-body paths.

### Concern #8 — empty `api_key` provider filter
Portal-side `WHERE api_key IS NOT NULL AND api_key != ''` in the snapshot provider query. Default-selection logic is unchanged — since filtered providers never land in `modelsByProviderId`, any `is_default` pointing at a filtered provider naturally falls through to the next kept provider's default. Test asserts that a provider with NULL / empty key is omitted and that `default` walks past its would-be `is_default` entry.

### Concern #6 — knowledge collision logging (minimum version)
`materializePortalKnowledge` now diffs `{size, mtimeMs}` stats before and after each repo's extraction. Files whose stat changed (or were newly created) are attributed to the repo that just ran; if a different repo had already written the same relative path, the collision is appended to `result.collisions` and logged via `console.warn`. Namespace-by-repo is still deferred — this is the "at minimum, log the collision" slice you flagged. New tests cover the collision and the disjoint-filenames-no-false-positive cases.

### Concern #10 — explicit `isTTY !== true`
`cli-main.ts` now compares against `true` explicitly instead of relying on `!undefined === true`. One-line readability fix with a comment explaining why.

### Nits
- `parseQuery` rewritten to use `new URL().searchParams`.
- `docs/features/portal-cli-integration.mdx`: corrected the stale "signs a short-lived JWT" step (we moved to header-based auth last round), added a "Cwd-scoped pairing" section explaining the one-Portal-per-cwd model (your concern #9).

## 401-leak test

Added an explicit test asserting that `GET /api/v1/cli-snapshot` without the header returns 401 **and** the response body is `{ error: "Unauthorized" }` only — no provider / credential / api_key content leaks through the error path. Seeded the DB with `SECRET-API-KEY-SHOULD-NEVER-LEAK` and grep-asserted it is not in the response.

## Deliberately deferred (not in this round)

Happy to do these in follow-ups — none are cheap + contained enough to bundle here:

- **#5 `setPortalSnapshot()` module state** — refactor cost exceeds current risk (single-call path). Would take `loadConfig(snapshot?)` threading or a single-set assertion.
- **#6 namespace by repo** — real fix but bigger surgery (TUI `[[page]]` resolution + knowledge Read paths both assume a flat dir). Logging closes the silent-shadowing gap for now.
- **#9 `SICLAW_PORTAL_SECRETS_PATH` env** — pure new-feature; doc note added, implementation can come when someone actually hits the multi-cwd case.
- **Nit: `PortalSnapshotWithMeta` type** — aesthetic.
- **E2E `siclaw local` + `siclaw --agent X` shell test** — worthwhile but wants its own PR with fixtures for the agent-scoping subdir.

## Verification

- `npx tsc --noEmit`: clean.
- `npm test`: 3145 passed / 1 pre-existing `kubectl get pods | grep Error` flake in `restricted-bash.test.ts` (not touched by this PR; last modified in the cleanup commit `7255e23`).

Current PR: 5 commits, same file scope, `3dab2a0` at the tip. Ready for another pass when you have a minute.

chent1996 (approved):
Re-reviewed `74caf7a` + `1a03d10`. All three blockers landed cleanly and several concerns got thoughtful treatment beyond the minimum:

**Blockers — resolved**
- **#1 prod admin seed** — `enableDefaultAdminSeed` opt-in, `cli-local.ts` on, `portal-main.ts` off. Right shape.
- **#2 snapshot auth** — triple gate (flag + loopback-only + dedicated `cliSnapshotSecret` with `timingSafeEqual`); `startPortal` throws at boot if misconfigured; client drops `jsonwebtoken` entirely; back-fill for older secrets files. Over-delivered here.
- **#3 path traversal** — skill `isSafeName` via `path.resolve`/`path.relative` is the right pattern; knowledge materializer's post-extraction `realpathSync` walk + wipe-on-escape is portable between BSD and GNU tar and catches symlink attacks the `--no-absolute-names` flag misses on its own. Nicely done.

**Concerns addressed**
- Consolidated `sweepSnapshots` on `exit` + `SIGINT` + `SIGTERM` — no more plaintext SSH keys outliving the session.
- 50 MB snapshot cap with streaming byte counter + `reader.cancel()` on overflow.
- `api_key IS NOT NULL AND != ''` filter server-side.
- Knowledge-collision stat-diff attribution + `console.warn` — the minimum slice closes the silent-shadow gap without committing to the full namespace-by-repo surgery.
- `isTTY !== true`, `parseQuery` via `URL().searchParams`, cwd-pairing doc note.

**Deferred items** (#5 module state, #6 full namespacing, #9 env override, `PortalSnapshotWithMeta` type, E2E shell test) — all have reasonable rationale; happy for those to land as follow-ups.

**Minor / optional nits for later, not blocking**:
- `Date.now()`-based tmp filenames in knowledge/credential materializers — switching to `crypto.randomBytes(8).toString('hex')` would harden the staging path. Low risk given it's always under your own tmpdir.
- Collision detection keys on `{size, mtimeMs}` — on filesystems with coarse mtime granularity (FAT32, some NFS/container FS) two sub-ms writes with the same size could misattribute. Hypothetical.
- Size cap is client-only; a server-side guard would prevent generating oversized responses in the first place when a new field is added in the future.

Good iteration. 👍

PR #239, opened 2026-04-23
fix(local): wire knowledge sync, task notifications, and schedule card rendering

LikiosSedo:
## Summary

Three independent fixes for local-mode (`siclaw local`) + Portal UX gaps
observed while smoke-testing the multi-agent flow. All three are
orthogonal to PR #237 — they touch files unrelated to that PR's CLI
snapshot work.

### 1. `fix(local)` — Knowledge never reached agents; idle timer killed the process

`LocalSpawner.spawn()` bypasses `agentbox-main.ts`, so the
`syncAllResources()` call that K8s pods run at startup never fires in
local mode. Bound knowledge repos stayed in the DB and the agent's
`Read .siclaw/knowledge/index.md` always returned *Path not found*.

Separately, the 5-min idle self-destruct in `http-server.ts` calls
`process.exit(0)`. In K8s that recycles a pod; in local mode AgentBox
shares the `siclaw local` process with Portal, so the exit silently
takes port 3000 down.

**Fix:** `LocalSpawner.spawn` fire-and-forgets `syncResource("knowledge", ...)`
via the agent's mTLS cert (skills / mcp deliberately skipped to respect
invariant #1). `createHttpServer` gains `{ disableIdleShutdown }`;
LocalSpawner passes it so only K8s keeps the timer.

### 2. `fix(portal)` — Cron runs produced no notifications

Runtime's `TaskCoordinator` emits `task.notify` over WebSocket RPC after
each cron run, but Portal never registered a handler. A parallel HTTP
route `/api/internal/task-notify` existed but nothing called it.
`notifications` stayed empty; the NotificationBell was always dark.

**Fix:** Extract `createTaskNotification()` helper (persist + WS push)
in `notification-api.ts`, route the HTTP handler through it, and
register the missing `task.notify` RPC handler in `adapter.ts`.
`adapter-rpc.test.ts`'s "exactly 40 handlers" assertion bumps to 41.

### 3. `fix(ui)` — `manage_schedule` tool results rendered as raw JSON

`PilotArea` checked `message.toolName === "schedule_preview"` but the
backend tool is `manage_schedule` — the condition never matched, so
`ScheduleCard` never rendered. The tool's `ok()` also emitted only
`{ summary }`, so even a matching card couldn't show action / cron /
status.

**Fix:** `manage_schedule.ts` returns structured `{ action, id, name,
schedule, summary }`. `PilotArea` uses the correct tool name and threads
`agentId` down to `ScheduleCard`, which adds an Open button
deep-linking to `/agents/:agentId/tasks/:taskId` (or `/my-tasks` when
id is unknown).

## Test plan

- [x] `npm run build` + `portal-web && npm run build` — clean
- [x] `npm test -- local-spawner http-server manage-schedule notification adapter-rpc` — **169/169 pass**
- [x] End-to-end on `siclaw local` (port 3000):
  - network-doctor agent reads `.siclaw/knowledge/index.md` and quotes the runbook correctly
  - `siclaw local` process stays up past 5 min idle
  - Firing a scheduled task produces a row in `notifications` and
    `GET /api/v1/notifications` returns `unread: 1`
  - Chat response for "create a task every 5 min" renders ScheduleCard
    with View + Open buttons

### DIFF
--- a/portal-web/src/components/chat/ScheduleCard.tsx
+++ b/portal-web/src/components/chat/ScheduleCard.tsx
@@ -1,4 +1,5 @@
-import { Clock, Check, Eye } from "lucide-react"
+import { Clock, Eye, ExternalLink } from "lucide-react"
+import { Link } from "react-router-dom"
 import { cn } from "./cn"
 import type { PilotMessage } from "./types"
 
@@ -22,9 +23,11 @@ interface ParsedResult {
 export function ScheduleCard({
   message,
   onOpenPanel,
+  agentId,
 }: {
   message: PilotMessage
   onOpenPanel?: (msg: PilotMessage) => void
+  agentId?: string
 }) {
   let parsed: ParsedResult | null = null
   try {
@@ -89,7 +92,7 @@ export function ScheduleCard({
           {actionLabel}
         </span>
 
-        {/* View button */}
+        {/* View button — opens side panel with full details */}
         {onOpenPanel && (
           <button
             onClick={() => onOpenPanel(message)}
@@ -98,6 +101,30 @@ export function ScheduleCard({
             <Eye className="w-3 h-3" />
             View
           </button>
+        )}
+
+        {/* Open in Tasks page — deep link when we have the task id + current agentId. */}
+        {!isDelete && parsed.id && agentId && (
+          <Link
+            to={`/agents/${agentId}/tasks/${parsed.id}`}
+            className="inline-flex items-center gap-1 px-1.5 py-0.5 rounded text-[10px] font-medium text-muted-foreground hover:text-amber-400 hover:bg-amber-500/10 transition-colors"
+          >
+            <ExternalLink className="w-3 h-3" />
+            Open
+          </Link>
+        )}
+
+        {/* Fallback: jump to the tasks list.
+            Fires for delete (the per-task page is gone even though `id` is
+            still in the payload), and whenever id or agentId are missing. */}
+        {(isDelete || !parsed.id || !agentId) && (
+          <Link
+            to="/my-tasks"
+            className="inline-flex items-center gap-1 px-1.5 py-0.5 rounded text-[10px] font-medium text-muted-foreground hover:text-amber-400 hover:bg-amber-500/10 transition-colors"
+          >
+            <ExternalLink className="w-3 h-3" />
+            Tasks
+          </Link>
         )}
       </div>
     </div>
--- a/src/tools/workflow/manage-schedule.ts
+++ b/src/tools/workflow/manage-schedule.ts
@@ -121,26 +121,43 @@ Common cron patterns:
         content: [{ type: "text" as const, text: JSON.stringify({ error: msg }) }],
         details: { error: true },
       });
-      const ok = (summary: string) => ({
-        content: [{ type: "text" as const, text: JSON.stringify({ summary }) }],
+      // Emit the full shape expected by portal-web's ScheduleCard /
+      // SchedulePanel so the UI can render a card (with View button + deep
+      // link) instead of the raw JSON summary.
+      const ok = (summary: string, extra: {
+        action?: string;
+        id?: string;
+        name?: string;
+        newName?: string;
+        schedule?: { name: string; description?: string; schedule: string; status: string };
+      } = {}) => ({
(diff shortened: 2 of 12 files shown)


portal-web/src/components/chat/ScheduleCard.tsx
```diff
@@ -99,6 +102,28 @@ export function ScheduleCard({
...
+
+        {/* Open in Tasks page — deep link when we have the task id + current agentId. */}
+        {!isDelete && parsed.id && agentId && (
+          <Link
+            to={`/agents/${agentId}/tasks/${parsed.id}`}
+            className="inline-flex items-center gap-1 px-1.5 py-0.5 rounded text-[10px] font-medium text-muted-foreground hover:text-amber-400 hover:bg-amber-500/10 transition-colors"
+          >
+            <ExternalLink className="w-3 h-3" />
+            Open
+          </Link>
```
chent1996, line 114: **Bug**: When action=delete and id+agentId are both present, neither button renders. The comment assumes delete has no id, but `manage-schedule.ts` delete branch emits `id: target.id`. Change the fallback condition to `(isDelete || !parsed.id || !agentId)`.

src/tools/workflow/manage-schedule.ts
```diff
@@ -213,15 +255,35 @@ Common cron patterns:
           await client.updateAgentTask(target.id, {
             status: params.action === "pause" ? "paused" : "active",
           });
-          return ok(`${params.action === "pause" ? "Paused" : "Resumed"} scheduled task "${target.name}".`);
+          return ok(
+            `${params.action === "pause" ? "Paused" : "Resumed"} scheduled task "${target.name}".`,
```
chent1996, line 266: **Minor**: `schedule: ""` (empty string) in pause/resume is misleading. The current cron expression isn't known at this point. Either fetch the target task's schedule, or omit the `schedule` field entirely (like the update branch does when `params.schedule` is absent).

chent1996 (changes requested):
Three solid fixes, but `ScheduleCard` has a rendering bug for delete actions that I want addressed before merge.

### Bug: delete action renders no button

```tsx
{!isDelete && parsed.id && agentId && <Link>Open</Link>}
{(!parsed.id || !agentId) && <Link>Tasks</Link>}
```

The comment `// Fallback: jump to the tasks list when id unknown (e.g. delete)` assumes delete responses have no `id`, but `manage-schedule.ts` delete branch actually returns `id: target.id`. When action=delete AND id+agentId are both present:
- First condition: `!isDelete` is false → no Open button
- Second condition: `!parsed.id || !agentId` is false → no Tasks button
- **Neither button renders**

Fix:
```tsx
{!isDelete && parsed.id && agentId && <Link>Open</Link>}
{(isDelete || !parsed.id || !agentId) && <Link>Tasks</Link>}
```

### Minor: pause/resume emits `schedule: ""`

The pause/resume branches set `schedule: { schedule: "", ... }` (empty string) instead of fetching the current cron. `ScheduleCard` will render an empty schedule field. Either fetch the current schedule from the target task, or follow the update-branch pattern and omit the `schedule` field when we don't have it.

### Everything else LGTM

- `disableIdleShutdown` design is correct (K8s default preserved, LocalSpawner explicit)
- Knowledge fire-and-forget sync with skills/mcp skipped per invariant #1
- `createTaskNotification` helper extraction is clean DRY
- `task.notify` RPC handler with param validation, adapter-rpc test count updated
- `agentId` prop chain through AgentChat → PilotArea → MessageItem → ScheduleCard

chent1996 (approved):
Both items from the prior review are addressed:

- ScheduleCard delete: `(isDelete || !parsed.id || !agentId)` now correctly renders the Tasks fallback.
- pause/resume: `resolveId` surfaces `schedule + status` from the existing list call, and `ok()` omits the `schedule` field when not available — same pattern as the update branch.

The new release commit (`portal-web/dist/` in the files allow-list + `prepublishOnly` hook + Web UI/CLI pairing docs) closes the loop so `npm i -g siclaw && siclaw local` actually serves the bundled UI. Docs match the §1.4 Read-Only Snapshot Contract.

One optional follow-up (non-blocking): `prepublishOnly` only fires on `npm publish`, not `npm pack`. If any release flow uses `npm pack`, consider `prepack` for symmetry.

LGTM.

PR #243, opened 2026-04-23
fix(cli): detect knowledge collisions from the archive manifest

LikiosSedo:
## Summary

PR #237 merged with a red CI — `src/lib/portal-knowledge-materializer.test.ts > logs a collision when two repos ship the same filename` was failing on the Linux runner throughout rounds 2 and 3 of the review, passing only on the maintainer's macOS box. This PR fixes the underlying detection bug so main goes back to green.

## Root cause

The collision detector inferred "which repo wrote which file" by diffing `{size, mtime, ino}` on each file pre/post tar extraction. All three dimensions happen to tie on a Linux runner when two small same-size archives are extracted in quick succession:

- **GNU tar** `-x` uses `open(O_TRUNC)` rather than unlink+create, so an overwritten file **keeps its inode**.
- **ext4 mtime granularity** on the runner is 1 second; two extractions completing in the same second tie on `mtimeMs`.
- The test's two payloads are both **14 bytes**, so `size` ties too.

With all three fields equal, the `writtenByThisRepo` check evaluated `false` on the second repo and the collision went unrecorded. My round-3 addition of `ino` was a no-op for the same reason: `O_TRUNC` preserves the inode.

Reproduced in a `linux/amd64 node:24` container — `size/mtime/ino` all identical before/after a same-name extraction over an existing file.

## Fix

Switch to the archive's own manifest. `tar -tzf <archive>` lists every entry in the tarball itself — platform-independent, no dependence on filesystem state. Both BSD and GNU tar support it; the only shape difference is BSD's leading `./` prefix which we strip. `snapshotFileStats` and the stat-diff machinery are removed — the archive list is the authoritative answer.

## Test plan

- [x] macOS (BSD tar 3.5.3) — 8/8 pass
- [x] linux/amd64 (GNU tar 1.34) via `docker run --platform linux/amd64 node:24` — **8/8 pass**, collision test now actually logs `overview.md: repo-a → overwritten by repo-b`

## Notes

Diff is tight: `1 file changed, 31 insertions(+), 50 deletions(-)` — the new path is actually shorter than the stat-diff it replaces. No other scope creep.

### DIFF
--- a/src/lib/portal-knowledge-materializer.ts
+++ b/src/lib/portal-knowledge-materializer.ts
@@ -70,10 +70,14 @@ export function materializePortalKnowledge(
     );
     try {
       fs.writeFileSync(tmpPath, Buffer.from(repo.dataBase64, "base64"));
-      // Snapshot file stats BEFORE extraction so we can tell which entries
-      // this repo actually wrote (either newly created or stat-changed by
-      // the extraction) and attribute collisions correctly.
-      const beforeFiles = snapshotFileStats(outDirResolved);
+      // Read the archive's own manifest rather than diffing pre/post fs stats.
+      // Earlier attempts to diff {size, mtime, ino} were unreliable on Linux:
+      // GNU tar's `-x` uses `open(O_TRUNC)` which preserves the inode, and
+      // two tiny files written in the same second (mtime granularity == 1s
+      // on many container mounts) can tie on all three fields, silently
+      // losing the "which repo wrote this" signal. The archive's own entry
+      // list is the authoritative answer and is platform-independent.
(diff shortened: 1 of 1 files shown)


LikiosSedo:
Closing as duplicate — PR #242 already carries a functionally equivalent fix (`f8cd6f2 fix(portal-knowledge): detect collisions via tar manifest, not stat diff`) for the same root cause (GNU tar `O_TRUNC` preserving inode + second-granularity mtime causing all three stat fields to tie on the Linux runner). Same approach: drop stat-diff, read the archive manifest via `tar -tzf`. Will let #242 land.

PR #246, opened 2026-04-24
fix(core): break core → portal type dependency to unblock agentbox build

LikiosSedo:
## Why

The AgentBox Dockerfile (`Dockerfile.agentbox`) intentionally does **not** copy `src/portal/` into the builder stage — it's a deliberate layering rule (AgentBox runtime ↛ Portal UI/REST). But a recent change added type-only imports from `src/portal/cli-snapshot-api.ts` into `src/core/`:

- `src/core/agent-factory.ts:87,89` — two inline `import(...)` references
- `src/core/extensions/agent.ts:13` — `import type ... from "../../portal/cli-snapshot-api.js"`
- `src/core/extensions/ls.ts:19` — same as above

**GitHub Actions' `CI / Type Check`** runs `tsc` on the full source tree so it sees `src/portal/` and reports no error — that's why this slipped past review.

The **internal siflow image build** runs `npx tsc -p tsconfig.agentbox.json` *inside the Dockerfile builder stage*, which only has the AgentBox-scoped COPY subset. There `src/portal/` does not exist, so tsc fails:

```
src/core/agent-factory.ts(87,34): error TS2307: Cannot find module '../portal/cli-snapshot-api.js'
src/core/agent-factory.ts(89,30): error TS2307: Cannot find module '../portal/cli-snapshot-api.js'
src/core/extensions/agent.ts(13,67): error TS2307: Cannot find module '../../portal/cli-snapshot-api.js'
src/core/extensions/ls.ts(19,67): error TS2307: Cannot find module '../../portal/cli-snapshot-api.js'
```

Our internal CI has been unable to produce `siclaw-agentbox` images on any main SHA since this import was introduced (we spotted it while trying to deploy the latest main — the last working image tag is `main-8c2d9a1`).

## What

All four imports are strictly **type-only** and only reference two interfaces: `CliSnapshotAgentMeta` and `CliSnapshotActiveAgent`. Both are plain POJOs with zero runtime dependencies.

- Create `src/shared/cli-snapshot-types.ts` as the canonical home for the two interfaces (already copied by `Dockerfile.agentbox` and in `tsconfig.agentbox.json`'s include list)
- Re-export them from the original `src/portal/cli-snapshot-api.ts` location so existing external consumers (portal-web, siclaw-tui, etc.) don't need to touch their imports
- Point the three `src/core/*` files at the new shared location

**Zero runtime behavior change**; this is purely a compile-time dependency realignment that restores the intended AgentBox ↛ Portal layering.

## Verification

- `npx tsc -p tsconfig.agentbox.json --noEmit` passes locally (exit 0)
- `grep` confirms no remaining `../portal/` imports under the AgentBox compile scope (`core/`, `agentbox/`, `cron/`, `memory/`, `shared/`, `tools/`, `gateway/agentbox/`, `gateway/security/`, and the four gateway top-level `.ts` files in the Dockerfile)
- External consumers keep their existing import path via the re-export in `src/portal/cli-snapshot-api.ts`

## Follow-up (not in this PR)

GitHub Actions `CI / Type Check` running on the full tree makes this class of bug invisible. Consider adding a CI job that specifically runs `tsc -p tsconfig.agentbox.json` inside a Dockerfile-equivalent COPY subset (or just running a lightweight `docker build --target builder -f Dockerfile.agentbox .` smoke test on PRs) to catch future regressions.

### DIFF
--- a/src/core/agent-factory.ts
+++ b/src/core/agent-factory.ts
@@ -84,9 +84,9 @@ export interface CreateSiclawSessionOpts {
    */
   portalCredentialsDir?: string;
   /** Metadata for all Portal-configured agents (used by /agent + /ls to show list). */
-  portalAvailableAgents?: import("../portal/cli-snapshot-api.js").CliSnapshotAgentMeta[];
+  portalAvailableAgents?: import("../shared/cli-snapshot-types.js").CliSnapshotAgentMeta[];
   /** The Portal agent this session is scoped to, null/undefined = unscoped. */
-  portalActiveAgent?: import("../portal/cli-snapshot-api.js").CliSnapshotActiveAgent | null;
+  portalActiveAgent?: import("../shared/cli-snapshot-types.js").CliSnapshotActiveAgent | null;
   /**
    * Base URL of the live local Portal (e.g. http://127.0.0.1:3000). When set,
    * `/setup` switches to read-only mode + opens Portal Web UI for writes so
--- a/src/core/extensions/agent.ts
+++ b/src/core/extensions/agent.ts
@@ -10,7 +10,7 @@
  */
 
 import type { ExtensionAPI } from "@mariozechner/pi-coding-agent";
-import type { CliSnapshotAgentMeta, CliSnapshotActiveAgent } from "../../portal/cli-snapshot-api.js";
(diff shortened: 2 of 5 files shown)


LikiosSedo:
Branch pushed directly to upstream as `fix/core-portal-layering` for internal DevOps CI to run an image build. Closing this fork-based PR; will reopen from the upstream branch later once CI validates.

PR #247, opened 2026-04-25
refactor(dp): rebuild deep investigation on async delegation

LikiosSedo:
## Summary

Rebuild Deep Investigation as a clean single-agent DP loop: remove the old DP state-machine/tool family, expose same-agent delegation only when DP mode is active, and render delegated work as auditable Portal cards with async parent synthesis.

### Problem

The previous DP flow coupled prompt state, special-purpose tools, SSE progress events, and bespoke UI cards. That made the UX noisy, made Kimi overfit to raw checkpoint mechanics, and left sub-agent work hard to audit or reuse for future multi-agent routing.

### Solution

- Replace the old DP state machine with a prompt extension plus hidden checkpoint markers.
- Add same-agent `delegate_to_agent` and notify-driven `delegate_to_agents` runtime bridges, backed by hidden child chat sessions with lineage metadata.
- Persist delegated capsules and tool traces through the normal chat/audit tables while filtering hidden delegation sessions/events from the ordinary chat list.
- Redesign Portal rendering for delegated investigation cards, partial batch status, persisted cards after refresh, and synthesis notices.
- Add smoke scripts and design notes for the single-agent DP loop and future gateway/portal multi-agent routing.

## Test Plan

- [x] `npm test`
- [x] `npm run build`
- [x] `git diff --check origin/main...HEAD`
- [x] Targeted DP/delegation tests: `npm test -- src/agentbox/session.test.ts src/tools/workflow/delegate-to-agent.test.ts src/tools/workflow/delegate-to-agents.test.ts src/core/extensions/deep-investigation.test.ts`

Manual localhost DP smoke was also exercised during development against the configured local Kimi/runtime setup.


chent1996 (changes requested):
Big refactor — rebuilding DP on async delegation is the right call, and the design doc is excellent. I focused review on the new security/correctness boundaries (delegation persistence, RPC contract, SSE consumer changes, prompt lifecycle). A few issues worth fixing before merge.

### Critical

1. **Session can deadlock if `setModel` throws** — `src/agentbox/http-server.ts:301`

   The reordering moves `managed._promptDone = false` to BEFORE `await managed.brain.setModel(found)` (line 352). The intent is sound (prevent double-submit races) but there is no try/finally to reset `_promptDone` if `setModel` throws. If it does, the route catches the error in the outer wrapper and returns 500, but leaves the session with `_promptDone = false`. Subsequent `/api/prompt` calls return 409 *forever*, because `actuallyFinish()` (which would call `scheduleRelease`) is only ever invoked from the `prompt().then/catch` chain — and that chain is never reached.

   Suggest wrapping the setup block (model registration, setModel, prompt kickoff) in try/catch that calls `actuallyFinish()` (or at least resets `_promptDone = true`) on error.

### Important

2. **`tool_execution_start` now writes a "running" row that has no recovery path if the runtime drops between start and end** — `src/gateway/sse-consumer.ts:236`

   Pre-PR, only the completed row was inserted at `tool_execution_end`. Post-PR a "running" row is persisted eagerly. If the agentbox crashes, gateway restarts, or the SSE stream is severed before `tool_execution_end` fires, the row stays `outcome=null, status="running"` indefinitely. On reload, `toolStatusFromMessage` returns `"running"` and the user sees a forever-spinning tool card.

   Frontend's `isStaleDelegationTool` only covers `delegate_to_agent` / `delegate_to_agents` — every other tool (kubectl, skill scripts, etc.) is unprotected. Consider either a startup sweep that flips orphan running rows to `timed_out`, or extending `isStaleDelegationTool` to all tools with a generous threshold.

3. **Race between agentbox `updateDelegationToolMessage` and SSE consumer's `tool_execution_end`** — `src/agentbox/session.ts` (`persistSnapshot`) vs `src/gateway/sse-consumer.ts:162`

   The agentbox's first `persistSnapshot(false)` runs after the first sub-agent finishes and issues `chat.updateDelegationToolMessage` with `WHERE session_id AND tool_name AND delegation_id = ?`. But `delegation_id` is only stamped onto the parent's `delegate_to_agents` row at `tool_execution_end` in the SSE consumer (extracted from `metadata.delegation_id`).

   If a sub-agent finishes faster than the round-trip {gateway → portal RPC → DB write of the parent row}, the first delegation update silently matches 0 rows. Later snapshots overwrite, so the user typically never sees it — but the first task's progress is dropped on the floor for that brief window. Worth either (a) eagerly stamping `delegation_id` when the start row is created (the executor knows it from `toolCallId`) or (b) routing the `chat.updateDelegationToolMessage` through the parent SSE consumer instead of bypassing it.

### Defense-in-depth

4. **`chat.appendMessage` / `chat.updateMessage` / `chat.updateDelegationToolMessage` don't validate that `params.session_id` belongs to the calling `agentId`** — `src/portal/adapter.ts:1883,1903,1929`

   The RPC handler signature is `(params, agentId) => Promise<any>` but the chat handlers ignore the second arg. A misbehaving runtime can write/update any session's messages. This was already true for `appendMessage`; the two new update handlers extend the surface. `internal-api.ts` `validateDelegationEventActor` does check ownership, but uses the in-process `sessionRegistry` which returns *permit* when the record is missing (e.g. after a runtime restart). Worth tightening at the adapter handlers (verify `chat_sessions.agent_id = agentId`) so the in-memory registry is not the only line of defense.

### Minor

5. **`isStaleDelegationTool` 15-minute threshold for `delegate_to_agents`** — `portal-web/src/hooks/usePilotChat.ts`

   Per-child `DELEGATED_AGENT_MAX_RUNTIME_MS` is 10 min and `DELEGATION_BATCH_GRACE_MS + DELEGATED_AGENT_PARTIAL_STEER_WAIT_MS` adds ~145s. A legitimately long batch could approach the 15 min ceiling and get rendered as `timed_out` while the runtime is still steering for partial. Worth bumping or tying the constant to the runtime values.

jacoblee-io (commented):
## Review

**Intent**: Replace DP state machine + special-purpose tools (`propose_hypotheses`, `deep_search`, `end_investigation`) with a DP-mode flag plus same-agent `delegate_to_agent` / `delegate_to_agents`. Background batches notify the parent session async; rich UI cards persist via DB.

Direction is sound — schema migration is safe (nullable adds, no MySQL/SQLite-incompatible types), the new mTLS endpoint is preserved, the new public `dp-state` route uses `requireAuth` (matches PR #245 JWT-only convention). Findings below are real issues worth fixing before merge.

---

### 🔴 Must-fix

**1. User abort does not propagate to in-flight delegated sub-agents**
`src/agentbox/http-server.ts:715-733`, `src/agentbox/session.ts:421-445`
`POST /api/sessions/:sessionId/abort` only calls `managed.brain.abort()` on the parent. The `_backgroundWorkCount` machinery and `DelegatedAgentControl.forceStop` are never triggered, so each sub-agent keeps spending tool budget until `DELEGATED_AGENT_MAX_RUNTIME_MS` (10 min) — and capsule rows continue to land in chat history after the user aborted.
**Fix**: when the parent is aborted, walk active delegation batches for the session and invoke `forceStop()` on every controller (or `requestPartial` then `forceStop`).

**2. Race window between synthetic delegation prompt and a fresh user prompt**
`src/agentbox/session.ts:712-784` (`waitForParentIdle` + `runSyntheticParentPrompt`)
\`waitForParentIdle\` polls every 1 s and returns when \`_promptDone === true\`. The synthetic path then synchronously sets \`_promptDone = false\` and calls \`brain.prompt(notification)\`. Between the poll's positive observation and the synthetic prompt actually starting, a user POST \`/prompt\` can arrive, pass the same \`_promptDone\` check in \`http-server.ts:287\`, set it to \`false\`, and call \`brain.prompt(text)\`. Two concurrent \`brain.prompt\` invocations on the same session is undefined behavior (depending on pi-agent's internal lock it either throws or interleaves events).
**Fix**: introduce a single mutex (\`_promptInflight: Promise|null\`) that both the synthetic queue and the HTTP entry point must acquire, or route \`runSyntheticParentPrompt\` through a private queued variant of the \`getOrCreate.prompt\` path that itself owns \`_promptDone\` mutation.

**3. `runSyntheticParentPrompt` leaves SSE subscription dangling on synchronous `brain.prompt` rejection**
`src/agentbox/session.ts:802-833`
\`managed._promptDone = false\` is set before \`try { await managed.brain.prompt(...) }\`. If \`prompt()\` throws synchronously, \`catch\` writes a synthesis-failed message and \`finally\` correctly restores \`_promptDone = true\`. However \`_eventBuffer = []\` was already cleared and \`stopManagedBuffer\` ran — fine — but the SSE stream that may have been consuming this session is left without a brain subscription, so the next prompt's events fail to forward to any waiting listeners until they resubscribe. Compounds with #2.
**Fix**: serialize via the same mutex from #2 so this code path can't race the SSE replay machinery.

---

### 🟡 Should-fix

**4. `requiresUserApproval` flag is plumbed but never enforced**
`src/core/tool-registry.ts:175-203`, `src/tools/workflow/delegate-to-agent.ts:155-159`, `src/tools/workflow/delegate-to-agents.ts:147-150`
The registry annotates the resolved \`ToolDefinition\` with \`requiresUserApproval: true\`, and tests assert it (\`delegate-to-agents.test.ts:39\`), but no execution wrapper consumes the flag. The PR header advertises a \"permission-gated agent delegation contract\" — currently the model can call delegate tools without any user gate.
**Fix**: either wire it through the guard pipeline so reviewers can verify the gate exists, or drop the flag (and the test) for now and document the future hook. As written, it is misleading: a future reviewer touching the registry might assume gating is enforced.

**5. `_extraEventBuffer` is unbounded**
`src/agentbox/session.ts:97-104, 1487-1495`
Tool-pushed events (\`sessionEventEmitter\`) are buffered into \`_extraEventBuffer\` whenever \`_extraEventSubs.size === 0\`. If a delegation batch runs 10 minutes with no SSE client connected (e.g. user closed the tab while the parent prompt is still in flight), the buffer grows without limit.
**Fix**: cap the buffer (LRU drop or size limit, mirroring \`_eventBuffer\`'s tie to \`_promptDone\`) and emit a diagnostic when overflowing.

**6. `sessionBelongsToIdentity` is permissive for unknown sessions in cross-process delegation persistence**
`src/gateway/internal-api.ts:46-50`
\`if (!owner) return true;\` — anything not currently in the LRU \`sessionRegistry\` (post-restart, post-LRU-eviction) is trusted blindly. Pre-existing for \`chat.appendMessage\` — not a regression. But this PR expands the surface: \`delegation.update_message\` / \`update_tool_message\` / \`append_event\` can now mutate existing rows by id.
**Fix**: on the update paths, add an extra DB lookup against persisted \`chat_sessions.agent_id\` before trusting the body, or document that runtime must \`rememberSession()\` before delegated background work touches lineage rows.

**7. `delegation.ensure_session` does not cross-check `userId` against parent session's known owner**
`src/gateway/internal-api.ts:39-47`
Same trust model as #6. When \`parentSessionId\` is in the registry, \`sessionRegistry.get(parentSessionId).userId\` is known — the validator should compare it to \`event.userId\` and reject mismatches. Otherwise a buggy AgentBox could write child sessions under the wrong user_id.
**Fix**: cheap 3-line check, tightens the audit story without breaking the contract.

**8. `usePilotChat.send` callback has `messages` in deps and recreates on every render**
`portal-web/src/hooks/usePilotChat.ts:769-823`
\`send\` now reads \`messages\` (for \`hasPendingDelegationSynthesis(messages)\`), so the \`useCallback\` deps include \`messages\` and the function reference changes on every state update. Components downstream that depend on \`send\` will rebuild as well.
**Fix**: use a ref (\`messagesRef.current\`) for the predicate or move the synthesis-pending check into a \`streamingRef\`-style ref updated by the same effect that mutates messages.

**9. Stale doc references to a non-existent `dispatch_subagents` tool**
`src/agentbox/http-server.ts:536`, `src/agentbox/session.ts:98`, `src/agentbox/session.ts:1429`, `src/core/agent-factory.ts:99`
Comments cite \`dispatch_subagents\` to explain \`_extraEventSubs\` / \`sessionEventEmitter\`. There is no such tool in the codebase (confirmed via grep).
**Fix**: replace with the actual user (\`delegate_to_agent\` / \`delegate_to_agents\`), or drop the explanatory aside.

**10. Two polling effects can race on the same session**
`portal-web/src/hooks/usePilotChat.ts:518-566`, `:572-612`
The recovered-streaming polling effect and the async-delegation polling effect both call \`setMessages(annotateDelegationSynthesis(items.map(toPilotMessage)))\` on overlapping intervals when both conditions hold. Each effect overwrites the other's last write, and stop conditions (\`hasRunning\` vs \`hasActiveAsyncDelegationSurface\`) can flap.
**Fix**: unify into a single effect with a shared \"is-something-still-going\" predicate.

---

### 🔵 Notes / nice-to-have

- Migration adds 4 nullable \`CHAR(36)\` columns to \`chat_sessions\` / \`chat_messages\` plus 4 indexes — fine under MySQL + SQLite; uses \`safeAlterTable\` / \`ensureIndex\` per CLAUDE.md.
- The new \`/api/v1/.../dp-state\` route uses \`requireAuth\` and scopes by user/agent; matches PR #245's JWT-only convention.
- \`chat.updateDelegationToolMessage\` matches by \`(session_id, tool_name, delegation_id)\` — \`delegation_id\` is a tool-call UUID; predicate is tight enough, no SQL-injection surface (parameterized).
- Old engine, sub-agent, prompts, schemas, and quality-gate files are deleted cleanly; no live \`src/\` import of \`deep-search/*\` remains. The pre-built \`src/gateway/web/dist/\` bundle still contains references but will be rebuilt on the next CI run.
- \`dp-mode\` JSONL persistence shape change is backward-compat: the reader accepts \`{active}\`, \`{enabled}\`, and \`{dpStatus}\` (\`session.ts:1620-1660\`, \`deep-investigation.ts:937-997\`).
- \`withDelegationStatusNotices\` materializes a synthetic \"Synthesizing…\" assistant message client-side; ID is stable (\`delegation-status-\${delegationId}\`) so React keys reuse correctly.

---

**Summary**: Direction is right and migration / security look fine, but #1 (abort doesn't cascade) and #2 (prompt race) are concurrency-correctness must-fixes; #4 leaves the documented \"permission-gated\" contract unenforced — implement or drop it before merge.

jacoblee-io (approved):

chent1996 (approved):
LGTM

PR #250, opened 2026-04-27
chore(helm): refresh values-standalone and add values-sicore example

jacoblee-io:
## Summary

Two Helm values templates were drifting. \`values-standalone.yaml\` still referenced fields the chart no longer reads after #245 and #248 (\`runtime.replicas\`, \`runtime.jwtSecret\`, \`runtime.runtimeSecret\`, \`portal.runtimeSecret\`), so the deploy snippet in its header would have any operator setting four no-op \`--set\` flags. There was also no template for the runtime-only / SiCore mode that this repo's two staging clusters actually use.

## Changes

- **\`values-standalone.yaml\` rewrite** — drop dead fields, simplify the deploy command to only the values the current chart actually consumes (\`database.url\`, \`runtime.portalSecret\`, \`portal.jwtSecret\`, \`portal.portalSecret\`), enable the bundled mysql by default for the demo path, document the production escape hatch (\`mysql.enabled=false\` + external DB).
- **\`values-sicore.yaml\` (new)** — runtime-only template: \`portal.enabled=false\`, no mysql, \`runtime.serverUrl\` points at sicore-apiserver, \`runtime.portalSecret\` = the adapter_secret SiCore returns at \`POST /api/v1/siclaw/admin/runtimes\`. Documents \`SICLAW_AGENT_ID\` semantics (it goes on the wire as \`X-Agent-Id\` but SiCore reverse-looks-up runtime_id from the secret rather than reading the header).

No code, chart-template, or behaviour changes — values-only.

## Test plan

- [x] \`helm template ./helm/siclaw -f helm/siclaw/values-standalone.yaml --set database.url=mysql://... --set runtime.portalSecret=... --set portal.jwtSecret=... --set portal.portalSecret=...\` renders three Deployments (mysql/runtime/portal), no \`SICLAW_RUNTIME_SECRET\`/\`JWT_SECRET\` on runtime, both Recreate strategies present.
- [x] \`helm template ./helm/siclaw -f helm/siclaw/values-sicore.yaml --set runtime.serverUrl=... --set runtime.portalSecret=... --set runtime.env.SICLAW_AGENT_ID=...\` renders a single runtime Deployment with \`SICLAW_PORTAL_SECRET\` + \`SICLAW_AGENT_ID\`, no Portal/MySQL manifests.
- [ ] Apply the SiCore template to a fresh ns and confirm Runtime phone-homes (already verified live in sicore-test on \`main-c57c65e\`; this PR just makes that values shape canonical).

### DIFF
--- a/helm/siclaw/values-sicore.yaml
+++ b/helm/siclaw/values-sicore.yaml
@@ -1,0 +1,67 @@
+# Siclaw Runtime-only Deployment (SiCore upstream)
+# Deploys ONLY the Runtime — phone-homes to an external SiCore apiserver
+# which acts as the management/adapter layer. No Portal, no DB, no MySQL.
+#
+# Prerequisites:
+#   1. SiCore deployed in the same cluster (or reachable via DNS)
+#   2. Register this Runtime in SiCore admin and capture the returned
+#      adapter_secret + runtime id:
+#        curl -X POST <sicore>/api/v1/siclaw/admin/runtimes \
+#          -H "Authorization: Bearer <admin-jwt>" \
+#          -d '{"id":"<runtime-id>","name":"<display-name>"}'
+#   3. Build and push images:
+#        make docker push REGISTRY=your-registry.com TAG=<tag>
+#
+# Deploy:
+#   helm upgrade --install siclaw ./helm/siclaw \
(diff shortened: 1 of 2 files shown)


jacoblee-io:
Reverting — values files won't be included

PR #256, opened 2026-04-28
fix(runtime): make chat.send ack async to escape sicore's 30s RPC timeout

LikiosSedo:
## Summary

sicore's WS RPC (the channel between sicore-apiserver and the runtime gateway) carries a fixed 30s timeout that applies to every `SendCommand`. The previous `chat.send` handler kept the RPC pending until `agentBoxManager.getOrCreate` plus `client.prompt()` returned, which on a cold-start (image pull, container start, ready probe) routinely exceeds 30s. The result was a spurious `CONNECTION_TIMEOUT` red bubble even though the runtime was healthy: sicore tore down the SSE response when the RPC timed out, the eventual reply still arrived later but had no subscriber to receive it, and the user was left with an error bubble plus a ghost session in the DB whose assistant reply was only visible after a page refresh.

### Problem

- 30s is the hard ceiling sicore applies to all WS RPCs (see `internal/siclaw/ws/connection_map.go:245` in scitix/sicore).
- A cold-start of the agentbox pod regularly hits 5–60s in real clusters (image pull from far registry, scheduler delay on busy nodes, ready-probe stabilisation).
- Bumping that timeout helps generic cases but doesn't fix the fundamental shape — long agentbox spawns will always sit somewhere on the tail.

### Solution

Decouple the RPC ack from the slow work. `chat.send` now returns `{ ok, sessionId }` within milliseconds, and every subsequent step runs in a background async block that streams events back to Portal on the existing `chat.event` WS channel. The 30s RPC timeout no longer covers cold-start; it only has to cover "is the runtime still reachable".

The chat.event stream itself is unchanged — `agent_start` / `agent_end` / `agent_message` / `stream_error` / `prompt_done` carry every observable progress signal exactly as before. There is no new event type, no new client-side state to track, and no new failure mode for callers that aren't aware of this PR. Frontends that already render the existing chat.event stream get the timeout fix transparently.

The 409 "Session is already running" path keeps the steer-fallback behaviour but no longer emits `prompt_done` (the still-running prompt fires its own — an extra one would close the SSE prematurely).

User-message persistence (`ensureChatSession` + `appendMessage` + `incrementMessageCount`) moves inside the async block at its earliest point, so the row exists before any agent events could land while still happening before the slow spawn.

The matching frontend complement (Portal sicore-web) is in [scitix/sicore!301](http://gitlab.scitix-inner.ai/k8s/sicore/-/merge_requests/301): it adds a soft "Still working on it…" reassurance after 10s of continuous Thinking… so legitimate long waits don't look like the agent has hung. The reassurance is purely a function of elapsed time — no protocol coupling.

## Test Plan

- [x] `npx tsc --noEmit -p tsconfig.json` clean.
- [x] `npx vitest run` clean (no new test failures introduced).
- [x] Verified end-to-end on a single-node test cluster paired with the scitix/sicore!301 frontend build:
   - Cold-start (~32s simulated): chat completes after the wait without a CONNECTION_TIMEOUT bubble; a single chat session row is created (no ghost session); the assistant reply streams in normally; "Still working on it…" appears once Thinking… crosses the 10s mark.
   - Hot-path (subsequent message in same session, fast spawn): unchanged perceived latency; no behavioural regression.
   - 409 "Session is already running" steer-fallback: a second concurrent send routes into the running prompt's stream — frontend sees one continuous response.
- [ ] One more pass against a freshly-cycled production-like cluster to confirm no regressions on real cold starts.

### DIFF
--- a/src/gateway/server.ts
+++ b/src/gateway/server.ts
@@ -112,20 +112,12 @@ export async function startRuntime(opts: StartRuntimeOptions): Promise<RuntimeSe
       throw new Error("agentId, userId, and text are required");
     }
 
-    // Get or create AgentBox for this agent — one pod per agent, shared by
-    // all callers. Caller identity travels as sessionId.
-    const handle = await agentBoxManager.getOrCreate(agentId);
-    const client = new AgentBoxClient(handle.endpoint, 30000, agentBoxTlsOptions);
-
-    // Pre-generate a UUID so the AgentBox doesn't fall back to the literal
-    // "default" session id (LocalSpawner behaviour), which would merge every
-    // caller's trace into one chat_sessions row.
+    // Pre-generate a UUID so AgentBox doesn't fall back to the literal
+    // "default" session id (LocalSpawner behaviour), which would merge
+    // every caller's trace into one chat_sessions row.
     const sessionId = incomingSessionId ?? crypto.randomUUID();
-    // Record sessionId → userId so AgentBox → Runtime internal-api callbacks
-    // can recover user attribution for Upstream audit.
     sessionRegistry.remember(sessionId, userId, agentId);
 
-    // Build prompt options from params (Upstream sends full context)
     const modelConfig = params.modelConfig as PromptOptions["modelConfig"];
     const promptOpts: PromptOptions = {
       sessionId,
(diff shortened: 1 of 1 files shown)


chent1996 (approved):
LGTM

jacoblee-io (approved):

PR #257, opened 2026-04-30
feat(runtime): make Runtime restart non-disruptive to active AgentBoxes

chent1996:
Three coordinated fixes that together let a Runtime pod restart leave existing AgentBox pods, in-flight chats, and async callback attribution intact.

cert-manager: derive issuer DN from the loaded CA subject instead of hardcoding "Siclaw Runtime CA". Hardcoding broke chart-managed CAs (Helm genCA only sets CN) — issued certs claimed an issuer that no AgentBox could verify against, surfacing as "unable to verify the first certificate" on every mTLS handshake.

helm chart: persist the mTLS CA in a kubernetes.io/tls Secret. lookup reuses the existing CA on upgrade; first-install genCA writes once. Annotated helm.sh/resource-policy=keep so helm uninstall does not orphan AgentBox client certs. SICLAW_CA_CERT/KEY injected into the Runtime Deployment via secretKeyRef before runtime.env so user overrides still win. Default generateCa=true with a fail-fast helper rejecting ephemeral mode in K8s.

session-registry: on cache miss, call a Portal-side chat.resolveSession RPC and back-fill, so async AgentBox callbacks that arrive after a Runtime restart but before the next chat.send still resolve user attribution. Resolver wired through credential-service and internal- api delegation/task paths.

## Test Plan

- [ ] `npm test` passes

## Architecture Checklist

## General Checklist

### DIFF
--- a/helm/siclaw/templates/runtime-ca-secret.yaml
+++ b/helm/siclaw/templates/runtime-ca-secret.yaml
@@ -1,0 +1,42 @@
+{{- /*
+WARNING: this template uses the `lookup` function to reuse a CA Secret that
+already exists in the cluster. `lookup` is a NO-OP in `helm template`,
+`helm install --dry-run`, and any other render-only context — those paths
+ALWAYS take the `genCA` branch and emit a fresh CA. If you pipe such output
+to `kubectl apply -f -`, you will OVERWRITE the live CA Secret and silently
+invalidate every AgentBox client cert in the cluster, causing mTLS to fail
+until each AgentBox pod is recreated.
+
+Use `helm install` / `helm upgrade` (which run against a real cluster) for
+this chart. Render-then-apply pipelines must inject a stable CA via
+`runtime.tls.caSecret` and set `runtime.tls.generateCa=false`.
+*/}}
+{{- include "siclaw.validateRuntimeTls" . }}
+{{- if eq (include "siclaw.runtimeTls.generateCa" .) "true" }}
+{{- $secretName := include "siclaw.runtimeCaSecretName" . }}
+{{- $existing := lookup "v1" "Secret" .Release.Namespace $secretName }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ $secretName }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "siclaw.labels" (dict "ctx" . "component" "runtime") | nindent 4 }}
+  annotations:
+    # Survive `helm uninstall` so re-installing into the same namespace does
+    # not regenerate the CA and orphan AgentBox client certs from the prior
+    # release.
+    "helm.sh/resource-policy": keep
+type: kubernetes.io/tls
+data:
+{{- if and $existing (index $existing.data "tls.crt") (index $existing.data "tls.key") }}
+  # Reuse the CA already in the cluster so `helm upgrade` keeps the same key
+  # material and existing AgentBox client certs stay valid.
+  tls.crt: {{ index $existing.data "tls.crt" }}
+  tls.key: {{ index $existing.data "tls.key" }}
+{{- else }}
+  {{- $ca := genCA "siclaw-runtime-ca" 3650 }}
+  tls.crt: {{ $ca.Cert | b64enc }}
+  tls.key: {{ $ca.Key  | b64enc }}
+{{- end }}
+{{- end }}
--- a/src/gateway/credential-service.test.ts
+++ b/src/gateway/credential-service.test.ts
@@ -3,6 +3,7 @@ import {
   CredentialService,
   createCredentialService,
   CredentialNotFoundError,
+  SessionOwnershipError,
 } from "./credential-service.js";
 import type { FrontendWsClient } from "./frontend-ws-client.js";
 import type { Identity } from "../shared/credential-types.js";
@@ -153,6 +154,44 @@ describe("CredentialNotFoundError", () => {
     expect(err).toBeInstanceOf(Error);
     expect(err.name).toBe("CredentialNotFoundError");
     expect(err.message).toBe("missing");
+  });
+});
+
+describe("CredentialService cross-agent ownership guard", () => {
+  // All four public methods feed through the same `rpcParams`, so any one
+  // bypassing the guard would leak attribution. Parameterized so a future
+  // refactor cannot accidentally regress only one path.
+  type Method = "listClusters" | "listHosts" | "getClusterCredential" | "getHostCredential";
+  const methods: { name: Method; invoke: (svc: CredentialService, id: Identity) => Promise<unknown> }[] = [
+    { name: "listClusters", invoke: (s, id) => s.listClusters(id) },
+    { name: "listHosts", invoke: (s, id) => s.listHosts(id) },
+    { name: "getClusterCredential", invoke: (s, id) => s.getClusterCredential(id, "c1", "purpose-x") },
+    { name: "getHostCredential", invoke: (s, id) => s.getHostCredential(id, "h1", "purpose-x") },
+  ];
+
+  for (const m of methods) {
+    it(`${m.name} throws SessionOwnershipError when session_id resolves to a different agent`, async () => {
+      const reg = new SessionRegistry();
+      reg.remember("sess-1", "u-foreign", "other");
+      const fakeFrontend = new FakeFrontendClient();
+      const guarded = new CredentialService(fakeFrontend as unknown as FrontendWsClient, reg);
+
+      await expect(m.invoke(guarded, { ...identity, sessionId: "sess-1" }))
+        .rejects.toBeInstanceOf(SessionOwnershipError);
+      // Critical: upstream RPC must not have been invoked under any path.
+      expect(fakeFrontend.calls).toHaveLength(0);
+    });
+  }
+
+  it("does not throw when sessionId is empty (graceful degradation)", async () => {
+    const reg = new SessionRegistry();
+    const fakeFrontend = new FakeFrontendClient();
+    fakeFrontend.responses.set("credential.list:cluster", { clusters: [] });
+    const guarded = new CredentialService(fakeFrontend as unknown as FrontendWsClient, reg);
+
+    await expect(guarded.listClusters({ ...identity, sessionId: "" }))
+      .resolves.toEqual([]);
+    expect(fakeFrontend.calls[0].params.userId).toBe("");
   });
 });
 
--- a/src/gateway/internal-api.ts
+++ b/src/gateway/internal-api.ts
@@ -43,51 +43,82 @@ function sendJson(res: http.ServerResponse, status: number, data: unknown): void
   res.end(JSON.stringify(data));
 }
 
-function sessionBelongsToIdentity(sessionId: string | null | undefined, identity: CertificateIdentity): boolean {
+async function sessionBelongsToIdentity(sessionId: string | null | undefined, identity: CertificateIdentity): Promise<boolean> {
   if (!sessionId) return true;
-  const owner = sessionRegistry.get(sessionId);
+  const owner = await sessionRegistry.get(sessionId);
   return !owner || owner.agentId === identity.agentId;
+}
+
+/**
+ * Resolve sessionId → userId, **enforcing that the session belongs to the
+ * calling cert's agent**. Cross-agent attribution is rejected (`ok: false`)
+ * — without this check, AgentBox A could pass a session_id owned by
+ * AgentBox B and have its task / credential request audited under B's user.
+ *
+ * Unknown sessions degrade gracefully (`ok: true`, `userId: ""`); only an
+ * explicit ownership mismatch trips the gate.
+ */
+async function resolveUserForIdentity(
+  sessionId: string | null | undefined,
+  identity: CertificateIdentity,
+): Promise<{ userId: string; ok: boolean }> {
+  if (!sessionId) return { userId: "", ok: true };
+  const owner = await sessionRegistry.get(sessionId);
+  if (!owner) return { userId: "", ok: true };
+  if (owner.agentId !== identity.agentId) return { userId: "", ok: false };
+  return { userId: owner.userId, ok: true };
 }
 
 function agentMatchesIdentity(agentId: string | null | undefined, identity: CertificateIdentity): boolean {
   return !agentId || agentId === identity.agentId;
 }
 
-function validateDelegationEventActor(
+async function validateDelegationEventActor(
   event: DelegationPersistenceEvent,
   identity: CertificateIdentity,
-): { status: number; error: string } | null {
+): Promise<{ status: number; error: string } | null> {
   switch (event.type) {
     case "delegation.ensure_session": {
       if (!event.userId) return { status: 400, error: "delegation.ensure_session requires userId" };
       if (!agentMatchesIdentity(event.agentId, identity)) return { status: 403, error: "delegation agent mismatch" };
-      if (!sessionBelongsToIdentity(event.sessionId, identity)) return { status: 403, error: "delegation session mismatch" };
-      if (!sessionBelongsToIdentity(event.lineage?.parentSessionId, identity)) return { status: 403, error: "delegation parent session mismatch" };
       if (!agentMatchesIdentity(event.lineage?.parentAgentId, identity)) return { status: 403, error: "delegation parent agent mismatch" };
       if (!agentMatchesIdentity(event.lineage?.targetAgentId, identity)) return { status: 403, error: "delegation target agent mismatch" };
+      // Two ownership checks; each can be a Portal RPC on cache miss. Run
+      // them in parallel — the unhappy path wastes one extra RPC but the
+      // happy path's latency is halved.
+      const [own, parentOwn] = await Promise.all([
+        sessionBelongsToIdentity(event.sessionId, identity),
+        sessionBelongsToIdentity(event.lineage?.parentSessionId, identity),
+      ]);
+      if (!own) return { status: 403, error: "delegation session mismatch" };
+      if (!parentOwn) return { status: 403, error: "delegation parent session mismatch" };
       return null;
     }
     case "delegation.append_message": {
-      if (!sessionBelongsToIdentity(event.message.sessionId, identity)) return { status: 403, error: "delegation session mismatch" };
-      if (!sessionBelongsToIdentity(event.message.parentSessionId, identity)) return { status: 403, error: "delegation parent session mismatch" };
       if (!agentMatchesIdentity(event.message.fromAgentId, identity)) return { status: 403, error: "delegation source agent mismatch" };
       if (!agentMatchesIdentity(event.message.targetAgentId, identity)) return { status: 403, error: "delegation target agent mismatch" };
+      const [own, parentOwn] = await Promise.all([
+        sessionBelongsToIdentity(event.message.sessionId, identity),
+        sessionBelongsToIdentity(event.message.parentSessionId, identity),
+      ]);
+      if (!own) return { status: 403, error: "delegation session mismatch" };
+      if (!parentOwn) return { status: 403, error: "delegation parent session mismatch" };
       return null;
     }
     case "delegation.update_message":
     case "delegation.update_tool_message": {
-      if (!sessionBelongsToIdentity(event.message.sessionId, identity)) return { status: 403, error: "delegation session mismatch" };
+      if (!(await sessionBelongsToIdentity(event.message.sessionId, identity))) return { status: 403, error: "delegation session mismatch" };
       return null;
     }
     case "delegation.append_event": {
       if (!event.event.userId) return { status: 400, error: "delegation.append_event requires userId" };
-      if (!sessionBelongsToIdentity(event.event.parentSessionId, identity)) return { status: 403, error: "delegation parent session mismatch" };
+      if (!(await sessionBelongsToIdentity(event.event.parentSessionId, identity))) return { status: 403, error: "delegation parent session mismatch" };
       if (!agentMatchesIdentity(event.event.parentAgentId, identity)) return { status: 403, error: "delegation parent agent mismatch" };
       if (!agentMatchesIdentity(event.event.targetAgentId, identity)) return { status: 403, error: "delegation target agent mismatch" };
       return null;
     }
     case "delegation.emit_chat_event": {
-      if (!sessionBelongsToIdentity(event.sessionId, identity)) return { status: 403, error: "delegation session mismatch" };
+      if (!(await sessionBelongsToIdentity(event.sessionId, identity))) return { status: 403, error: "delegation session mismatch" };
       return null;
     }
     default:
@@ -253,7 +284,8 @@ export async function handleAgentTasksList(
   try {
     // Session may be threaded via query param for GET; resolve → userId for audit.
     const sessionId = new URL(req.url || "/", "http://_").searchParams.get("session_id") ?? "";
-    const userId = sessionRegistry.resolveUser(sessionId);
+    const { userId, ok } = await resolveUserForIdentity(sessionId, identity);
+    if (!ok) { sendJson(res, 403, { error: "session ownership mismatch" }); return; }
     const data = await frontendClient.request("task.list", {
       agent_id: identity.agentId,
       user_id: userId,
@@ -306,7 +338,8 @@ export async function handleAgentTasksCreate(
     const invalid = validateSchedule(body.schedule);
     if (invalid) { sendJson(res, 400, { error: invalid }); return; }
 
-    const userId = sessionRegistry.resolveUser(body.session_id);
+    const { userId, ok } = await resolveUserForIdentity(body.session_id, identity);
+    if (!ok) { sendJson(res, 403, { error: "session ownership mismatch" }); return; }
     const data = await frontendClient.request("task.create", {
       id: randomUUID(),
       agent_id: identity.agentId,
@@ -345,7 +378,8 @@ export async function handleAgentTasksUpdate(
     }
 
     const sessionId = typeof body.session_id === "string" ? body.session_id : undefined;
-    const userId = sessionRegistry.resolveUser(sessionId);
+    const { userId, ok } = await resolveUserForIdentity(sessionId, identity);
+    if (!ok) { sendJson(res, 403, { error: "session ownership mismatch" }); return; }
     const data = await frontendClient.request("task.update", {
       task_id: taskId,
       agent_id: identity.agentId,
@@ -375,7 +409,8 @@ export async function handleAgentTasksDelete(
 ): Promise<void> {
   try {
     const sessionId = new URL(req.url || "/", "http://_").searchParams.get("session_id") ?? "";
-    const userId = sessionRegistry.resolveUser(sessionId);
+    const { userId, ok } = await resolveUserForIdentity(sessionId, identity);
+    if (!ok) { sendJson(res, 403, { error: "session ownership mismatch" }); return; }
     const data = await frontendClient.request("task.delete", {
       task_id: taskId,
       agent_id: identity.agentId,
@@ -496,7 +531,7 @@ export async function handleDelegationEvents(
 ): Promise<void> {
   try {
     const event = await readJsonBody(req) as DelegationPersistenceEvent;
-    const actorError = validateDelegationEventActor(event, identity);
+    const actorError = await validateDelegationEventActor(event, identity);
     if (actorError) {
       sendJson(res, actorError.status, { error: actorError.error });
       return;
--- a/src/gateway/security/cert-manager.ts
+++ b/src/gateway/security/cert-manager.ts
@@ -47,10 +47,27 @@ export interface CertificateBundle {
 export class CertificateManager {
   private caCert: string;
   private caKey: string;
+  /**
+   * Subject attributes of the loaded CA, in their original forge form.
+   *
+   * Issued certs MUST reference the exact subject of the signing CA — X.509
+   * path validation does a byte-wise DN comparison. We carry the attribute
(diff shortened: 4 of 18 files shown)


src/gateway/security/cert-manager.ts
```diff
@@ -306,6 +315,23 @@ function isIpAddress(s: string): boolean {
...
+  const out: Record<string, string> = {};
+  for (const attr of ca.subject.attributes) {
+    if (attr.name === "commonName" && typeof attr.value === "string") out.CN = attr.value;
+    else if (attr.name === "organizationName" && typeof attr.value === "string") out.O = attr.value;
+    else if (attr.name === "organizationalUnitName" && typeof attr.value === "string") out.OU = attr.value;
+  }
+  if (!out.CN) throw new Error("CA certificate has no Common Name (CN) — cannot derive issuer DN");
+  return out;
+}
+
```
chent1996: **Lossy DN projection only partially fixes the bug.** `subjectAttrsFromPem` carries only CN / O / OU through. `createCertificateStatic` (line 258-263) likewise consumes only those three keys when building `issuerAttrs`.

The new test (cert-manager.test.ts:348) uses a CN-only CA (matches Helm `genCA` output), so this path works for the chart-managed case. But typical real-world CAs — cert-manager, Vault PKI, corporate root CAs — include `C` / `ST` / `L` and frequently other attributes. Against those, the issued cert's issuer DN is a strict subset of the CA's subject DN, byte-wise comparison fails, and X.509 path validation rejects with the same "unable to verify the first certificate" symptom this PR is trying to eliminate.

Fix is to pass the CA's full attribute list through to `createCertificateStatic` instead of projecting onto `{CN, O, OU}`. Either:
- expose the raw forge attribute array on `caIssuer` and let `createCertificateStatic` set it directly via `cert.setIssuer(this.caIssuer)`, or
- extend `subjectAttrsFromPem` + `createCertificateStatic` to handle every common attribute (`countryName`, `stateOrProvinceName`, `localityName`, `serialNumber`, etc.) and fail loudly when an unknown name appears.

Minimum: add a test for a CA with C/ST/L attributes to lock down the contract.

helm/siclaw/templates/runtime-ca-secret.yaml
```diff
@@ -0,0 +1,29 @@
+{{- include "siclaw.validateRuntimeTls" . }}
+{{- if eq (include "siclaw.runtimeTls.generateCa" .) "true" }}
+{{- $secretName := include "siclaw.runtimeCaSecretName" . }}
+{{- $existing := lookup "v1" "Secret" .Release.Namespace $secretName }}
+apiVersion: v1
```
chent1996, line 18: **`lookup` is a no-op during `helm template` and `helm install --dry-run=client`.** In those modes `$existing` is always nil, so the template falls into the `else` branch and emits a freshly `genCA`'d Secret. If anyone in the workflow pipes `helm template … | kubectl apply -f -` (a common pattern for GitOps preview, ArgoCD diff, etc.), the resulting apply will overwrite the existing CA Secret in place — `helm.sh/resource-policy: keep` only guards against `helm uninstall`, not `kubectl apply`.

Suggest:
- Add a NOTES.txt warning that the chart is not safe to drive via `helm template | kubectl apply` — this requires `helm install/upgrade`.
- Or factor the genCA call into an `lookup`-aware helper so `helm template` emits a clearly-marked placeholder (e.g. `tls.crt: <SET-BY-LIVE-INSTALL>`) instead of valid cert material that will silently invalidate AgentBox certs.

src/gateway/session-registry.ts
```diff
@@ -64,6 +82,20 @@ export class SessionRegistry {
...
+    const cached = this.map.get(sessionId);
+    if (cached) {
+      cached.lastSeen = Date.now();
+      return cached;
+    }
+    if (!this.resolver) return undefined;
+    const fetched = await this.resolver(sessionId);
+    if (!fetched) return undefined;
+    this.remember(sessionId, fetched.userId, fetched.agentId);
+    return this.map.get(sessionId);
```
chent1996: **Cache-miss stampede on the same sessionId.** This `lookup` doesn't single-flight. Right after a Runtime restart, dozens of in-flight AgentBox callbacks (delegation events, internal-api calls) can all miss simultaneously for the *same* sessionId, each spawning its own Portal RPC. With N concurrent callers per session × M sessions, that's N×M Portal hits when 1×M would do.

A tiny in-flight map fixes it:
```ts
private inflight = new Map<string, Promise<SessionRecord | undefined>>();

private async lookup(sessionId: string | undefined): Promise<SessionRecord | undefined> {
  if (!sessionId) return undefined;
  const cached = this.map.get(sessionId);
  if (cached) { cached.lastSeen = Date.now(); return cached; }
  if (!this.resolver) return undefined;
  let p = this.inflight.get(sessionId);
  if (!p) {
    p = (async () => {
      try {
        const fetched = await this.resolver!(sessionId);
        if (!fetched) return undefined;
        this.remember(sessionId, fetched.userId, fetched.agentId);
        return this.map.get(sessionId);
      } finally {
        this.inflight.delete(sessionId);
      }
    })();
    this.inflight.set(sessionId, p);
  }
  return p;
}
```
Also useful: a short negative cache (e.g. 5s TTL) so misbehaving clients can't sustain a per-request RPC against unknown sessionIds.

src/gateway/server.ts
```diff
@@ -86,6 +86,23 @@ export async function startRuntime(opts: StartRuntimeOptions): Promise<RuntimeSe
...
+  sessionRegistry.setResolver(async (sessionId) => {
+    try {
+      const data = await frontendClient.request("chat.resolveSession", { session_id: sessionId }) as
+        | { found: false }
+        | { found: true; user_id: string; agent_id: string };
+      if (!data.found) return null;
+      return { userId: data.user_id, agentId: data.agent_id };
+    } catch (err) {
+      console.error("[session-registry] resolveSession RPC failed:", err);
+      return null;
```
chent1996, line 121: **No timeout safety on the cache-miss fallback.** `frontendClient.request` defaults to a 30s timeout (frontend-ws-client.ts:25). Cache-miss paths now sit on the hot path of every `internal-api` and `credential-service` request, so a slow or restarting Portal can stall up to 30s per call. After a Runtime restart that's exactly the worst-case window — every active session's first callback hits a miss before `chat.send` re-warms the LRU.

Consider passing a shorter timeout for `chat.resolveSession` (this is a single-row indexed lookup; 2-3s is generous) and surfacing the timeout as a warn-and-continue rather than an exception bubbling through the whole call chain. The existing `catch` here returns `null`, which is good — but it only fires on RPC errors, not on the 30s wall.

src/portal/adapter.ts
```diff
@@ -1861,6 +1861,16 @@ export function buildAdapterRpcHandlers(): Map<string, (params: any, agentId: st
 
   // --- chat.* ---
 
+  handlers.set("chat.resolveSession", async (params) => {
+    const db = getDb();
+    const [rows] = await db.query(
+      `SELECT user_id, agent_id FROM chat_sessions WHERE id = ? AND deleted_at IS NULL LIMIT 1`,
```
chent1996: **`AND deleted_at IS NULL` may drop attribution for legitimate in-flight work.** If a user soft-deletes a chat session in the Portal UI and an AgentBox callback arrives a few seconds later (delegation message, task completion event), `chat.resolveSession` returns `{found: false}`, the registry caches nothing, and the audit gets an empty `user_id`.

For *attribution* (which is what this RPC is for, per server.ts:88-91), soft-deletion shouldn't matter — the action still happened, the user still owns it. Suggest dropping the predicate for this query specifically, or splitting into a separate handler used only for attribution.

If the predicate is intentional (e.g. you actively want to anonymize post-deletion), please add a comment explaining the policy.

chent1996 (commented):
Three coordinated fixes here, and the overall direction is right (persist the CA, derive issuer from the loaded CA, fall back to Portal for stale cache). The new tests for cert-manager and session-registry add good coverage. Comments below — the most important is the lossy DN projection in `subjectAttrsFromPem`, which only partly fixes the bug it's claiming to fix.

**Should fix before merge:**
- `cert-manager.ts:323` — `subjectAttrsFromPem` only carries CN/O/OU through. Helm `genCA` produces CN-only CAs so it works for the chart-managed path, but cert-manager / Vault PKI / corporate CAs commonly include C/ST/L (and others). Those will still fail X.509 chain validation with the same "unable to verify the first certificate" symptom this PR is meant to fix.
- `runtime-ca-secret.yaml:5` — `lookup` returns nil during `helm template` and `helm install --dry-run=client`, so those paths silently regenerate the CA. If the rendered output ever gets piped to `kubectl apply -f -`, the existing CA Secret is overwritten and all AgentBox client certs are invalidated. Worth a doc warning at minimum.

**Worth addressing:**
- `session-registry.ts:86` — concurrent cache misses for the same sessionId fan out N identical Portal RPCs. Right after a Runtime restart you can have dozens of in-flight callbacks all missing → N parallel calls. Cheap to single-flight.
- `server.ts:93` — the resolver has no timeout protection on top of `FrontendWsClient`'s 30s default. If Portal is briefly slow, every internal-api request that hits a cache miss stalls for up to 30s. After a Runtime restart that means every active session's first callback eats the timeout.
- `adapter.ts:1867` — `chat.resolveSession` filters `deleted_at IS NULL`. If a user soft-deletes a session and an in-flight AgentBox callback arrives, attribution is lost (audit gets empty user_id). Confirm intentional; if not, drop the predicate — soft-deleted sessions still need attribution for past activity.

Nice work overall.

chent1996:
Thanks for the careful review — addressed all five in [1696ae3](https://github.com/scitix/siclaw/commit/1696ae3) (force-pushed onto the existing squash commit; PR is still one commit).

**1. Lossy DN projection** — replaced `subjectAttrsFromPem` with a byte-exact copy of the CA's forge `subject.attributes` array. `createCertificateStatic` now takes an `issuerAttrs` array and calls `cert.setIssuer(attrs)` directly, so C / ST / L / serialNumber / anything else the CA carries flows through unchanged. New regression test loads a full-DN CA (C/ST/L/CN/O/OU) and asserts the issued cert's issuer DN equals the CA's subject DN attribute-for-attribute.

**5. `deleted_at IS NULL`** — dropped. Soft-deleted sessions still need attribution for late callbacks; comment in adapter.ts spells out why.

**4. Resolver timeout** — wrapped the `chat.resolveSession` RPC in a 5s `Promise.race` timeout in `server.ts`. Times out → null → "" userId, matching pre-fallback behaviour. Avoids the 30s `FrontendWsClient` default stalling every cache miss.

**3. Single-flight** — `SessionRegistry` now keeps an `inflight: Map<sid, Promise>`; concurrent misses for the same sessionId piggyback on the in-flight resolver call. Test asserts N parallel `resolveUser("s1")` produce exactly one resolver invocation.

**2. `lookup` nil in render-only mode** — added a prominent comment block at the top of `runtime-ca-secret.yaml`, an inline warning under `runtime.tls.generateCa` in `values.yaml`, and a ⚠️ note in `NOTES.txt` for GitOps users. The chart's behaviour is unchanged; the documentation now makes the failure mode explicit and points users at `caSecret` for pre-rendered pipelines.

PTAL.

chent1996:
Both addressed in [edcea21](https://github.com/scitix/siclaw/commit/edcea21) (amend + force-push, PR remains a single commit).

**P1 — cross-agent attribution gate on task / credential paths**

Added `resolveUserForIdentity(sessionId, identity)` in `internal-api.ts` (mirroring the shape of `sessionBelongsToIdentity`):
- session unknown → `{ userId: "", ok: true }` (graceful, same as before)
- session owned by a different agent → `{ userId: "", ok: false }` → handler emits **403 `session ownership mismatch`** and short-circuits before the upstream RPC fires

All four task handlers (`agent-tasks` list / create / update / delete) now go through it. Upstream RPC is never called with a foreign user_id.

`credential-service.ts:rpcParams` does the equivalent — on mismatch it throws a new `SessionOwnershipError`. `credential-proxy.ts` catches it and emits **HTTP 403** for both `credential.get` and `credential.list` paths. Empty `sessionId` still degrades to `userId: ""`.

Two new tests:
- `internal-api.test.ts` — `task.create` with a session_id resolved to `agent-2` while cert is `agent-1` → expects 403 + zero `task.create` RPC calls.
- `credential-service.test.ts` — `listClusters` under the same mismatch → expects `SessionOwnershipError` + zero `credential.list` RPCs.

**P2 — drop CN-required guard**

Removed the explicit "no Common Name" check from the `CertificateManager` constructor. `caSubjectAttrs` now carries whatever the CA presents — RFC 5280 doesn't require CN in subject, and the issuer DN is byte-copied, so a CN-less CA produces a CN-less issuer DN that still validates. Replaced the throw-test with a positive test: load a CN-less CA (only `O` + `OU`) and assert the issued cert's issuer attributes mirror the CA's subject attributes verbatim.

Tests: 3034 pass (was 3031 — +1 cross-agent task test, +1 cross-agent credential test, +1 graceful-empty-sid test, -1 obsolete CN-required test, +2 new ones for ownership / CN-less CA).

PTAL.

src/gateway/server.ts
```diff
@@ -86,6 +86,34 @@ export async function startRuntime(opts: StartRuntimeOptions): Promise<RuntimeSe
...
+  // internal-api callback for the full FrontendWsClient default (30s). On
+  // timeout we degrade to "" userId, which matches the pre-fallback behaviour.
+  const RESOLVE_SESSION_TIMEOUT_MS = 5000;
+  sessionRegistry.setResolver(async (sessionId) => {
+    try {
+      const rpc = frontendClient.request("chat.resolveSession", { session_id: sessionId });
+      const data = await Promise.race([
+        rpc,
+        new Promise<never>((_, reject) =>
+          setTimeout(() => reject(new Error(`chat.resolveSession timed out after ${RESOLVE_SESSION_TIMEOUT_MS}ms`)), RESOLVE_SESSION_TIMEOUT_MS),
```
jacoblee-io: **Timer leak.** When `rpc` wins the race, the `setTimeout` handle is never cleared — every successful Portal resolve leaves an unref'd 5s timer pending in the event loop. The exact scenario this PR targets (burst of async AgentBox callbacks right after a Runtime restart) is also the case that piles up the most stray timers.

Suggested fix — capture the handle and clear it on success:

```ts
let timer: ReturnType<typeof setTimeout> | undefined;
try {
  const data = await Promise.race([
    rpc,
    new Promise<never>((_, reject) => {
      timer = setTimeout(
        () => reject(new Error(`chat.resolveSession timed out after ${RESOLVE_SESSION_TIMEOUT_MS}ms`)),
        RESOLVE_SESSION_TIMEOUT_MS,
      );
    }),
  ]) as ...;
  ...
} finally {
  if (timer) clearTimeout(timer);
}
```

Or extract a `withTimeout(promise, ms)` helper if this pattern shows up elsewhere.

src/gateway/session-registry.ts
```diff
@@ -64,6 +89,34 @@ export class SessionRegistry {
...
+    // Single-flight: piggyback on any in-flight resolver call for this sid.
+    const inflight = this.inflight.get(sessionId);
+    if (inflight) return inflight;
+
+    const resolver = this.resolver;
+    const promise = (async () => {
+      try {
+        const fetched = await resolver(sessionId);
+        if (!fetched) return undefined;
+        this.remember(sessionId, fetched.userId, fetched.agentId);
```
jacoblee-io, line 134: **`forget()` does not cancel an inflight resolver.** If `forget(sid)` is called while a Portal RPC for that sid is pending, this `remember()` still fires when the resolver settles — silently re-inserting an entry that was explicitly invalidated.

The new `single-flights concurrent misses` test acknowledges this implicitly (`// pending was already settled; new call starts fresh`) by forgetting only after the inflight settles, so the unhappy path is not exercised.

Two possible fixes:
- Clear the inflight entry from `forget()`: `this.inflight.delete(sessionId);` — simplest, awaiters get a fresh lookup next time.
- Tombstone: record forgotten sids during a resolver call and have this callback skip `remember()` if its sid was tombstoned mid-flight.

The first is fine if no other code relies on inflight settling after forget.

src/gateway/session-registry.ts
```diff
@@ -64,6 +89,34 @@ export class SessionRegistry {
...
+
+    // Single-flight: piggyback on any in-flight resolver call for this sid.
+    const inflight = this.inflight.get(sessionId);
+    if (inflight) return inflight;
+
+    const resolver = this.resolver;
+    const promise = (async () => {
+      try {
+        const fetched = await resolver(sessionId);
+        if (!fetched) return undefined;
```
jacoblee-io, line 126: **No negative cache.** A buggy AgentBox stuck retrying with a phantom sessionId (e.g. a stale ref held after a session prune) will hammer `chat.resolveSession` indefinitely — one RPC per callback, or every 5s on timeout. The `returns empty string when resolver reports the session is unknown` test in `session-registry.test.ts` explicitly documents this with `// Negative result is not cached — next miss retries.`

Consider caching `null` for a short window (10–30s) to throttle. Could be a separate `notFound: Map<string, number>` alongside the main map, or a sentinel value on `SessionRecord`. Follow-up is fine.

jacoblee-io (commented):
Overall LGTM — three coordinated fixes hang together well and the test coverage on the new contracts (byte-exact issuer DN, single-flight resolver, cross-agent attribution guard) is strong. Two real bugs and a perf concern worth addressing before merge; suggestions can be a follow-up.

**Important**
1. Timer leak in `server.ts` resolver wrapper — `setTimeout` never cleared on rpc success. Accumulates under the exact post-restart burst this PR targets.
2. `forget()` does not cancel an inflight resolver in `session-registry.ts` — an explicit invalidation can be silently undone by a racing Portal RPC.

**Performance**
3. No negative cache for unknown sessions in `session-registry.ts:lookup` — phantom sessionIds hammer Portal indefinitely.

See inline comments. (1) and (2) feel like blockers for this PR's stated goal; (3) is a follow-up.

src/gateway/internal-api.ts
```diff
@@ -43,51 +43,71 @@ function sendJson(res: http.ServerResponse, status: number, data: unknown): void
...
-): { status: number; error: string } | null {
+): Promise<{ status: number; error: string } | null> {
   switch (event.type) {
     case "delegation.ensure_session": {
       if (!event.userId) return { status: 400, error: "delegation.ensure_session requires userId" };
       if (!agentMatchesIdentity(event.agentId, identity)) return { status: 403, error: "delegation agent mismatch" };
-      if (!sessionBelongsToIdentity(event.sessionId, identity)) return { status: 403, error: "delegation session mismatch" };
-      if (!sessionBelongsToIdentity(event.lineage?.parentSessionId, identity)) return { status: 403, error: "delegation parent session mismatch" };
+      if (!(await sessionBelongsToIdentity(event.sessionId, identity))) return { status: 403, error: "delegation session mismatch" };
+      if (!(await sessionBelongsToIdentity(event.lineage?.parentSessionId, identity))) return { status: 403, error: "delegation parent session mismatch" };
```
jacoblee-io: **Sequential awaits.** For `delegation.ensure_session` (here) and `delegation.append_message` (lines 91-92), up to two `sessionBelongsToIdentity` calls run serially. On cache miss each becomes a Portal RPC, doubling latency on the delegation hot path.

```ts
const [own, parentOwn] = await Promise.all([
  sessionBelongsToIdentity(event.sessionId, identity),
  sessionBelongsToIdentity(event.lineage?.parentSessionId, identity),
]);
if (!own) return { status: 403, error: "delegation session mismatch" };
if (!parentOwn) return { status: 403, error: "delegation parent session mismatch" };
```

Wastes one extra RPC on early mismatches (the unhappy path), but cuts happy-path latency in half. Minor — fine to defer.

src/gateway/credential-service.test.ts
```diff
@@ -156,6 +157,32 @@ describe("CredentialNotFoundError", () => {
...
+
+  it("does not throw when sessionId is empty (graceful degradation)", async () => {
+    const reg = new SessionRegistry();
+    const fakeFrontend = new FakeFrontendClient();
+    fakeFrontend.responses.set("credential.list:cluster", { clusters: [] });
+    const guarded = new CredentialService(fakeFrontend as unknown as FrontendWsClient, reg);
+
+    await expect(guarded.listClusters({ ...identity, sessionId: "" }))
+      .resolves.toEqual([]);
+    expect(fakeFrontend.calls[0].params.userId).toBe("");
```
jacoblee-io, line 194: **Coverage gap.** Only `listClusters` is tested for the cross-agent guard. `listHosts`, `getClusterCredential`, and `getHostCredential` all share the same `rpcParams` path so the guard fires for them too — but the credential-proxy `handleCredentialRequest` 403 mapping at `credential-proxy.ts:108-111` is the more important branch. Worth one test per method, or a single parameterized test, to pin the contract.

src/portal/adapter.ts
```diff
@@ -1861,6 +1861,20 @@ export function buildAdapterRpcHandlers(): Map<string, (params: any, agentId: st
 
   // --- chat.* ---
 
+  handlers.set("chat.resolveSession", async (params) => {
+    const db = getDb();
+    // Intentionally NOT filtering on deleted_at — soft-deleted sessions still
```
jacoblee-io, line 1866: The `Intentionally NOT filtering on deleted_at` comment is exactly the kind of invariant that's easy to silently break in a future cleanup PR. Worth a test in `adapter-rpc.test.ts` that resolves a soft-deleted session and asserts it still returns `found: true` — locks the contract into the test suite.

chent1996:
Thanks @jacoblee-io — both blockers + the two test-coverage notes + the delegation parallelization all addressed in [f68c4f0](https://github.com/scitix/siclaw/commit/f68c4f0) (amend + force-push, single commit). Negative cache (item 3) deferred to follow-up per your suggestion.

**(1) Timer leak in resolver — fixed**

`server.ts` now hoists the `setTimeout` handle out of `Promise.race` and clears it in a `finally`:

```ts
let timer: ReturnType<typeof setTimeout> | undefined;
try {
  const data = await Promise.race([
    rpc,
    new Promise<never>((_, reject) => {
      timer = setTimeout(() => reject(...), RESOLVE_SESSION_TIMEOUT_MS);
    }),
  ]) as ...;
  ...
} finally {
  if (timer) clearTimeout(timer);
}
```

Every successful resolve now cancels its timer instead of leaving 5s of unref'd callbacks pending.

**(2) `forget()` cancels inflight — fixed (tombstone variant)**

I went with the tombstone option rather than the simple `inflight.delete()` because of an asymmetry: callers already awaiting the in-flight promise need to *get* a value (otherwise the AgentBox callback that triggered this lookup loses attribution for an event that is itself authoritative — the session existed at audit time). What must NOT happen is the cache re-inserting after explicit invalidation.

So `forget(sid)`:
- removes from cache (as before)
- tombstones `sid` so the in-flight resolver, when it settles, returns the fetched record to the awaiter but skips `remember()`
- removes the inflight entry so the *next* lookup starts fresh

`tombstones` is a `Set<string>`; entries are cleared by both `lookup()`'s `tombstones.delete(sid)` before launching a resolver and the resolver's `finally`, so no buildup.

New test (`session-registry.test.ts`) exercises the unhappy path: kick off resolver → forget → resolver settles → assert the awaiter sees the value but `peek()` shows the cache stayed empty.

**(4) Sequential awaits in delegation paths — parallelized**

Reordered `delegation.ensure_session` and `delegation.append_message` so the cheap sync `agentMatchesIdentity` checks fire first (early-403 unchanged), then the two `sessionBelongsToIdentity` calls run via `Promise.all`. Happy-path latency halved on cache miss; unhappy path costs one extra RPC on session-mismatch as you noted.

**(5) Credential cross-agent guard test coverage — fixed**

`credential-service.test.ts` is now parameterized across all four methods (`listClusters`, `listHosts`, `getClusterCredential`, `getHostCredential`) — any future refactor that bypasses `rpcParams` on just one path will fail loudly. Each asserts `SessionOwnershipError` thrown AND zero upstream RPC invocations.

`credential-proxy.test.ts` gains the two missing 403 mappings — one for `handleCredentialRequest` (per-host/cluster get), one for `handleCredentialList`. Both assert the exact error body so a future "let's improve the message" change re-runs through CI.

**(6) Soft-delete invariant pinned — fixed**

`adapter-rpc.test.ts` `chat.resolveSession` describe block now has three tests:
- found row → `{found: true, ...}`
- missing row → `{found: false}`
- **soft-deleted row** → still `{found: true}` AND a defense-in-depth assertion that the executed SQL string does not contain `deleted_at`. A future cleanup PR that re-adds the predicate will trip both halves.

**(3) Negative cache — follow-up tracked**

Will open a separate PR. Sketch: `notFound: Map<sid, expiresAt>` with a configurable TTL (suggest 30s) checked at the top of `lookup()`. Want to size the TTL against real prod traffic before committing to a number.

Tests: 3043 → 3045 (was 3034 pre-review). PTAL.

src/gateway/session-registry.ts
```diff
@@ -41,29 +58,89 @@ export class SessionRegistry {
...
+    if (!this.resolver) return undefined;
+
+    // Single-flight: piggyback on any in-flight resolver call for this sid.
+    const inflight = this.inflight.get(sessionId);
+    if (inflight) return inflight;
+
+    const resolver = this.resolver;
+    // Snapshot any pre-existing tombstone so this lookup starts fresh; we
+    // only honor tombstones created while THIS resolver call is in flight.
+    this.tombstones.delete(sessionId);
```
jacoblee-io, line 122: **Race still possible:** the tombstone Set is shared across calls, and this line clears it before the in-flight resolver from a *previous* `lookup()` has had a chance to consult it.

Reproducing sequence:
1. `lookup("s1")` runs. `tombstones.delete("s1")` (no-op). `inflight[s1] = P1`. P1 awaits resolver.
2. `forget("s1")` runs. `map.delete`, `tombstones.add("s1")`, `inflight.delete("s1")`.
3. A new `lookup("s1")` arrives. `inflight.get("s1")` → undefined (forget removed P1's tracking). Falls through. **This line clears the tombstone again.** `inflight[s1] = P2`.
4. P1 settles. Body: `tombstones.has("s1")` → **false** (cleared by step 3). Falls through to `this.remember(...)`. **`forget()` is silently undone.**
5. As a side effect, P1's `finally` then does `this.inflight.delete(sessionId)` — but that key now holds P2. So P2's single-flight tracking is also dropped: any further concurrent lookup for s1 would start a third resolver. Fan-out + lost forget.

The existing test at L103-122 dodges this because it never starts a second lookup between `forget()` and the resolver settle.

**Suggested fix** — per-call generation token instead of a shared Set:

```ts
private nextLookupGen = 0;
private currentGen = new Map<string, number>();

forget(sessionId: string): void {
  this.map.delete(sessionId);
  this.inflight.delete(sessionId);
  this.currentGen.delete(sessionId);  // invalidates ALL in-flight callbacks for this sid
}

private async lookup(sessionId): Promise<...> {
  ...
  const myGen = ++this.nextLookupGen;
  this.currentGen.set(sessionId, myGen);
  const promise = (async () => {
    try {
      const fetched = await resolver(sessionId);
      if (!fetched) return undefined;
      // Re-insert only if our generation is still the active one.
      if (this.currentGen.get(sessionId) !== myGen) {
        return { ...fetched, lastSeen: Date.now() };
      }
      this.remember(sessionId, fetched.userId, fetched.agentId);
      return this.map.get(sessionId);
    } finally {
      if (this.currentGen.get(sessionId) === myGen) {
        this.currentGen.delete(sessionId);
        this.inflight.delete(sessionId);
      }
    }
  })();
  this.inflight.set(sessionId, promise);
  return promise;
}
```

This also fixes the secondary bug where P1's `finally` clobbers P2's tracking — the generation check prevents both issues with one mechanism.

Worth adding a test that explicitly does `lookup → forget → lookup → resolver-settles` to pin this. Currently no test exercises a second lookup before the first settles.

jacoblee-io (commented):
Thanks for the quick turnaround. Verified 5 of 6 are fixed correctly; one has a subtle remaining hole.

**✅ Verified fixed**
- (1) Timer leak — handle hoisted, cleared in `finally`. Correct.
- (4) Sequential awaits — `Promise.all` parallelization in `delegation.ensure_session` / `delegation.append_message`. Correct, well-commented.
- (5) Coverage gap — parameterized test covering all 4 credential-service methods (and asserts `fakeFrontend.calls.length === 0`, which is the right invariant). Nice tightening.
- (6) Soft-deleted session test — `expect(sql).not.toContain("deleted_at")` is exactly the right contract pin.
- (3) Negative cache — deferred to follow-up as agreed.

**⚠️ (2) `forget()` race — fix has a subtle hole**

The tombstone Set is shared across lookups, and `lookup()` unconditionally clears it at L122 before starting a resolver. A 3-step interleave can still undo `forget()`. See inline comment on `session-registry.ts:122` — if a new lookup arrives between `forget()` and the in-flight resolver's settle, the new lookup wipes the tombstone and the in-flight resolver will `remember()` after all. The new test at L103-122 doesn't exercise this sequence (it never starts a second lookup before the first resolver settles).

Not a blocker IMO — `forget()` is rarely called concurrently with active resolver traffic in practice — but worth fixing while the abstraction is fresh.

jacoblee-io (approved):
Approving — the production fix is high-value and well-executed.

The Runtime-restart-breaks-AgentBoxes scenario is solved by three coordinated changes (persistent CA Secret, byte-exact issuer DN, Portal-backed session resolver), and the cross-agent attribution guard closes a real security hole. All 6 review items from the previous round were addressed; 5 are fully correct.

One known follow-up: the new `tombstone` mechanism in `session-registry.ts` still has a race under `lookup → forget → lookup → resolver-settles` interleaving (see inline comment). Impact is limited to cache freshness — the re-inserted record carries correct attribution from Portal, so it does NOT regress the security guarantees this PR adds. Worth a follow-up PR with the negative-cache item.

Test coverage on the new contracts (byte-exact DN, single-flight, cross-agent guard, soft-delete invariant) is strong and parameterized where it matters.

PR #258, opened 2026-04-30
feat(trace): per-prompt trace recorder with pluggable MySQL/SQLite store

a-yeyang:
# PR Description
## What this PR does
Adds a trace recorder to siclaw. Every time a user sends one prompt to the agent, we save a complete record of that single run — so we can replay, debug, and analyse it later without asking the user "what did you click?".
What gets recorded (per prompt, one row + one JSON body)
Identity & context
- id (unique), sessionId, promptIdx, userId / username, mode (web / tui / local), brainType, modelName
Conversation
- The original userMessage
- isInjectedPrompt — was this typed by the user, or auto-generated by a UI button / synthetic capsule? (enum, see injected-prompt-kinds.ts)
Run outcome
- outcome (success / error / cancelled), startedAt, endedAt, durationMs
- stepCount, toolCallCount, tokensTotal, costUsd
- dpStatusEnd — Deep Probe workflow state at flush time
Replay payloads
- bodyJson — full step-by-step trace (every tool call, args, output, timings)
- traceSummary / traceSummaryJson — compact chronological projection
- traceEasy — ultra-thin version (user input + AI text + tool name/output only)
The recorder hooks BrainSession.subscribe() + the diagnostic event bus, bounded by agent_start / agent_end. See trace-recorder.ts and the storage contract in trace-store-types.ts.
## Storage backends
Pluggable via the TraceStore interface:
- SQLite — local default (dev / TUI)
- MySQL — cluster mode
- Composite — dual-write to both
Selected by SICLAW_TRACE_MYSQL_URL; recorder code is unchanged across backends.
## Deployment note
Test environment currently uses an in-cluster MySQL pod (siclaw-trace-db StatefulSet, see scripts/deploy-k8s.sh). This is fine for testing but not recommended for production — pod restarts, no managed backups, no HA.
For production, switch to a managed cloud database (Aliyun RDS for MySQL, AWS RDS, or your company's internal managed MySQL). Only SICLAW_TRACE_MYSQL_URL needs to change — no code edits required.


chent1996 (commented):
Thanks for this — the recorder design is solid (clean `TraceStore` interface, two-phase stub→flush UPSERT, explicit-boundary mode that correctly merges pi-agent's internal retry/compaction cycles into one trace). My review focused on bugs and risks; comments below in priority order.

**Should land before merge:**
- `scripts/images.txt:3` has a tag typo (`1cad1531` vs `1cad153` on the other two lines) that will fail the deploy script's own consistency check on line 168.
- `scripts/images.txt` is per-developer state and probably shouldn't be committed at all — `images.txt.example` already serves as the template, and the README recommends per-developer/per-region copies. As-committed, every contributor's branch will keep clobbering this file.
- Hardcoded MySQL password (`siclawsiclawsiclaw`) committed in `scripts/deploy-k8s.sh:44,57` and the README. Even a test-cluster password is permanent in git history once committed; the same script already does the right thing for `PORTAL_JWT_SECRET` / `PORTAL_SECRET` (read from k8s Secret on subsequent deploys, randomly generate on first). Apply the same pattern to the DB passwords.

**Correctness items worth fixing before prod:**
- Race in TraceRecorder auto-mode `agent_start` handler — second `agent_start` while first `startTrace` is still awaiting `persistInFlightStub` corrupts in-flight state. Latent because explicit mode is dominant, but auto-mode is still wired up and tested.
- `pendingSkillMeta` matching falls back to name-only when `scriptName` is empty.
- `chipMatcher`'s regex-escape pattern is mis-bracketed (`]` not in the char class). Currently harmless because all chip labels are alphanumeric, but breaks the moment anyone adds a label with a regex metachar.
- Auth scope for `/traces` list (always pins `userId`) and `/traces/:id` (allows NULL `user_id` through) are asymmetric. Pick one rule.
- `recordSteerEvent` doesn't populate `trace_summary` / `trace_summary_json` / `trace_easy`, so steer rows are invisible to analytics queries that filter on those columns.
- The `is_injected_prompt` TINYINT→VARCHAR migration in MySQL runs four sequential auto-committing DDLs with no transaction — a process death mid-migration leaves the table half-converted. The SQLite version already handles this correctly with rebuildIsInjectedToText.

**Lower priority:**
- Common API filters (`outcome`, `mode`, `dp_status_end`, `session_id`) lack indexes — flag for a follow-up audit once production traffic is in.
- `SET GLOBAL FOREIGN_KEY_CHECKS=0` in the deploy script's MySQL workaround is server-wide and persistent across connections; the README itself acknowledges the proper fix is reordering the DDL in `migrate.ts`. Worth landing the real fix and dropping the workaround.

Good work overall. Let me know if any of the inline comments need more context.

scripts/images.txt
```diff
@@ -0,0 +1,3 @@
+registry-cn-beijing.siflow.cn/k8s/siclaw-portal:yye-feat-trace-store-v2-1cad153
+registry-cn-beijing.siflow.cn/k8s/siclaw-agentbox:yye-feat-trace-store-v2-1cad153
+registry-cn-beijing.siflow.cn/k8s/siclaw-runtime:yye-feat-trace-store-v2-1cad1531
```
chent1996, line 3: **Typo: trailing `1` on the runtime tag.** Lines 1-2 are `…-1cad153`; line 3 is `…-1cad1531`. The deploy script (`scripts/deploy-k8s.sh:168-170`) aborts with `all three images must share one tag` when this file is fed to it as-is. Either the tag is wrong here, or the file is stale and should be regenerated.

scripts/images.txt
```diff
@@ -0,0 +1,3 @@
+registry-cn-beijing.siflow.cn/k8s/siclaw-portal:yye-feat-trace-store-v2-1cad153
```
chent1996, line 1: **`images.txt` probably shouldn't be committed.** It carries one developer's branch tag (`yye-feat-trace-store-v2-…`); other contributors will keep clobbering it on every PR. The repo already has `scripts/images.txt.example` as the template, and your README explicitly recommends per-developer / per-region copies. Suggest: add `scripts/images.txt` to `.gitignore` and remove the committed instance — the example file is enough.

scripts/deploy-k8s.sh
```diff
@@ -0,0 +1,232 @@
...
+# SICLAW_TRACE_MYSQL_URL env var here — that variable is the *runtime* config
+# consumed by `cli-local.ts` against a kubectl-port-forwarded 127.0.0.1:3307,
+# and people commonly leave it sourced in their shell. If we picked it up at
+# deploy time, the runtime pod would inherit "127.0.0.1:3307" and try to talk
+# to its own loopback (which has nothing listening) — exactly the bug that
+# motivated this comment. Keep the two namespaces separated.
+TRACE_DB_HOST="${TRACE_DB_HOST:-siclaw-trace-db}"
+TRACE_DB_PORT="${TRACE_DB_PORT:-3306}"
+TRACE_DB_USER="${TRACE_DB_USER:-root}"
+TRACE_DB_PASS="${TRACE_DB_PASS:-siclawsiclawsiclaw}"
```
chent1996, line 44: **Don't commit the shared DB password.** `siclawsiclawsiclaw` is now permanent in git history (also at line 57 for the portal DB, and in the README's debug curl). The same script already does the right thing for `PORTAL_JWT_SECRET` / `PORTAL_SECRET` at lines 703-722 (read from k8s Secret on redeploys, generate randomly on first deploy). Apply that pattern to `TRACE_DB_PASS` / `PORTAL_DB_PASS` too — the password should never be a default literal in source.

The README §8 even calls this out as a known issue, so this isn't news, but cleaning it up before merge avoids the rotation cost later.

src/core/trace-recorder.ts
```diff
@@ -0,0 +1,953 @@
...
+    }
+
+    // agent_start: in AUTO mode, opens a new trace. In EXPLICIT mode, internal
+    // cycles (empty-response retry / auto-compaction continuation) are merged
+    // into the current trace by external beginPrompt — do nothing here.
+    //
+    // Fire-and-forget the async startTrace/flush from sync event handlers:
+    // brain.subscribe() callbacks don't await, and we don't want to block
+    // event processing on DB writes. Unhandled rejections must be caught.
+    if (type === "agent_start") {
```
chent1996, line 433: **Race condition in auto-mode.** This handler fires `startTrace(...)` async without awaiting, and `startTrace` itself awaits `persistInFlightStub()` (line 226). If a second `agent_start` arrives while the first call is still in that await, the second call sees `this.active === true` and runs `await this.flush()` (line 197), mutating `this.steps` / `this.skillsUsed` while the first promise is still in flight.

In practice the dominant code path is explicit mode (http-server's `/api/prompt` and `cli-main.ts` both call `beginPrompt`/`endPrompt`), so this is latent rather than active — but auto-mode is wired up and tested, and is what runs when neither caller is in the picture. Suggest a re-entrancy guard:

```ts
private startInFlight: Promise<void> | null = null;

private async startTrace(...) {
  if (this.startInFlight) await this.startInFlight;
  this.startInFlight = (async () => { /* existing body */ })();
  await this.startInFlight;
  this.startInFlight = null;
}
```

src/core/trace-recorder.ts
```diff
@@ -0,0 +1,953 @@
...
+      if (parsed) {
+        ref = { skillName: parsed.skillName, scope: parsed.scope, via: "read" };
+      }
+    }
+
+    if (!ref) return;
+
+    // Merge any buffered diagnostic metadata for Path B (carries authoritative scope).
+    if (ref.via === "local_script") {
+      const idx = this.pendingSkillMeta.findIndex(
```
chent1996, line 604: **Brittle match: empty `scriptName` collapses to name-only.** When `ref.scriptName === ""` (the default in the `local_script` branch above when `args.script` / `args.scriptName` is missing), `!ref!.scriptName` short-circuits and the `findIndex` matches by `skillName` alone. Two concurrent calls to the same skill with different scripts will pick up the wrong scope/outcome/durationMs.

Low probability today (most calls are sequential), but worth tightening — either require `scriptName === undefined` (not falsy-empty) before the loose match, or correlate by `toolCallId` instead of `skillName + scriptName`.

src/core/injected-prompt-kinds.ts
```diff
@@ -0,0 +1,214 @@
...
+/**
+ * Frontend chip clicks are encoded on the wire as `[<label>]\n<fullPrompt>`.
+ * Source of truth for labels: portal-web/src/components/chat/PilotArea.tsx.
+ *
+ * Each chip is registered as its OWN kind (chip_dig_deeper, chip_proceed, …)
+ * so analytics can answer "which button was clicked" directly from the DB.
+ * Helper used by the per-chip matchers below.
+ */
+function chipMatcher(label: string) {
+  const re = new RegExp(`^\\[${label.replace(/[.*+?^${}()|[\\]\\\\]/g, "\\$&")}\\]\\n`);
```
chent1996, line 54: **Regex-escape pattern is mis-bracketed.** As written, `/[.*+?^${}()|[\\]\\\\]/g` parses as: char class = `.*+?^${}()|[\` (note: `]` is NOT in the class — the `]` after `\\` closes it), followed by literal `\\]`. Effective behavior: matches any of those metachars *only when followed by `\\]`*, which never appears in chip labels — so `replace` is a no-op and `label` is passed through unescaped.

This is harmless today because every registered chip label is alphanumeric/space (`Dig deeper`, `Proceed`, `Refine`, `Summarize`, `Adjust`, `Skip`). Becomes a real bug the moment someone adds a label like `[A.B]`, `[Refine+]`, etc.

The canonical pattern is `/[.*+?^${}()|[\]\\]/g`.

src/portal/siclaw-api.ts
```diff
@@ -1812,6 +1818,155 @@ export function registerSiclawRoutes(router: RestRouter, config: SiclawConfig, c
...
+   * Build a TraceListOpts from a parsed query string, applying the auth-scope
+   * rule: non-admins are silently pinned to their own user_id even if they
+   * pass a different `user_id` query param.
+   */
+  function buildTraceListOpts(query: Record<string, string>, auth: AuthContext): TraceListOpts {
+    const isAdmin = auth.role === "admin";
+    const opts: TraceListOpts = {};
+    if (query.user_id) opts.userId = query.user_id;
+    if (query.username) opts.username = query.username;
+    if (!isAdmin) opts.userId = auth.userId;  // hard scope; overrides any caller-supplied user_id
```
chent1996, line 1856: **Auth scope inconsistency between list and get-by-id.**

Here (list): `if (!isAdmin) opts.userId = auth.userId` — non-admin queries are hard-pinned to their own userId, so any trace with `user_id IS NULL` (older / anonymous rows) is filtered out.

At line 1934 (get-by-id): `if (auth.role !== "admin" && record.userId && record.userId !== auth.userId)` — non-admin can read NULL-user_id rows because the `record.userId &&` short-circuits.

Result: an authenticated non-admin can `GET /traces/:id` for an anonymous row but it never appears in their list. Pick one rule:
- If anonymous traces should be readable, list filter should be `user_id = ? OR user_id IS NULL`.
- If they shouldn't, get-by-id should refuse them (drop the `record.userId &&` guard).

src/core/trace-recorder.ts
```diff
@@ -0,0 +1,953 @@
...
+   *   - Calling beginPrompt() mid-run would reset `this.active` and prematurely
+   *     flush the main in-flight trace (cutting it in half).
+   *
+   * So we write a fully-formed standalone row atomically, with its own unique
+   * trace_key (suffixed `-steer-<rand>` to guarantee no collision with main
+   * traces). `outcome='completed'` and `duration_ms=0` because the steer itself
+   * is instantaneous from the user's perspective; what happens afterwards
+   * (resumed agent work) lands in the main trace's final flush.
+   */
+  async recordSteerEvent(text: string): Promise<void> {
```
chent1996, line 255: **Steer rows skip the projection columns.** This path writes `bodyJson` but doesn't compute `traceSummary` / `traceSummaryJson` / `traceEasy` (compare to `flush()` at line ~660 which does). Downstream analytics queries that filter `WHERE trace_easy IS NOT NULL` (the cheap path your README touts as "the analytics team's intended consumption") will silently miss every steer row.

Fix is one line — steers have `steps: []`, so:
```ts
const easy = buildTraceEasy({ userMessage: text, steps: [] }).line;
const summary = buildTraceSummary({ userMessage: text, steps: [] });
```
then pass them in the upsert. Or document explicitly that steers are summary-less and have downstream queries `COALESCE` over `kind`.

src/core/trace-store-mysql.ts
```diff
@@ -0,0 +1,408 @@
...
+      // v5 → v6: is_injected_prompt TINYINT (0/1) → VARCHAR(64) enum.
+      // Run only when the live column type is still numeric. The two-step
+      // dance (translate values → MODIFY column) is needed because MODIFY
+      // alone would coerce 0 → '0' and 1 → '1', losing the meaning.
+      const injectedType = colMap.get("is_injected_prompt") ?? "";
+      if (injectedType && (injectedType === "tinyint" || injectedType.startsWith("int"))) {
+        await conn.query(`ALTER TABLE agent_traces ADD COLUMN is_injected_prompt_new VARCHAR(64) NOT NULL DEFAULT 'none'`);
+        await conn.query(`UPDATE agent_traces SET is_injected_prompt_new = CASE WHEN is_injected_prompt = 1 THEN 'unknown_legacy' ELSE 'none' END`);
+        await conn.query(`ALTER TABLE agent_traces DROP COLUMN is_injected_prompt`);
+        await conn.query(`ALTER TABLE agent_traces CHANGE COLUMN is_injected_prompt_new is_injected_prompt VARCHAR(64) NOT NULL DEFAULT 'none'`);
```
chent1996, line 174: **Non-transactional v5→v6 migration.** Lines 171-174 are four sequential DDLs. MySQL DDL auto-commits, so a process death (or connection drop) between any two leaves the table in a half-converted state — `is_injected_prompt_new` exists alongside the original column with no rollback.

The SQLite variant in `trace-store-sqlite.ts` `rebuildIsInjectedToText()` already handles this correctly: rename original, create v6, copy with the CASE expression, drop old — all inside a `BEGIN ... COMMIT`. Recommend the same shape here:
```sql
CREATE TABLE agent_traces_v6 LIKE … (with the new column)
INSERT INTO agent_traces_v6 SELECT … CASE WHEN … END AS is_injected_prompt …
RENAME TABLE agent_traces TO agent_traces_v5_bak, agent_traces_v6 TO agent_traces;
DROP TABLE agent_traces_v5_bak;
```
Single atomic RENAME swaps both names; the rest is recoverable on retry.

PR #259, opened 2026-04-30
feat(chat): per-message timing badges + Metrics latency breakdown

a-yeyang:
Summary
Adds end-to-end timing telemetry to the chat UI and Metrics dashboard. Every assistant message and tool card now shows compact emoji badges reporting where wall-clock time went. The Metrics page gains a Latency Breakdown card with avg / min / max / p90 across all sampled turns.

All timings use a single server clock (Date.now()) anchored at portal POST receipt, threaded into runtime via the chat.send RPC. Persisted in chat_messages.metadata (JSON) and duration_ms, so reloading a session restores all badges.

Badge definitions
Per assistant message:

⏳ TTFT — turn-start → first model token. Shown only on the first message of a turn (turn-anchor; would otherwise double-count).
💭 Thinking — last input boundary (turn-start or previous tool_execution_end) → first text delta. Suppressed on the first message when it equals TTFT.
✍️ Output — first text delta → message_end. The streaming time of the message body.
Per tool message (bash, kubectl, skill, etc.):

💭 Pre-tool thinking — boundary → tool_execution_start. For one-thinking-many-tools batches only the first tool carries a non-zero value.
⚙️ Duration — tool_execution_start → tool_execution_end.
Invariant: ⏳ + Σ💭 + Σ✍️ + Σ⚙️ ≈ turn_total (no overlap, no missing intervals). Badges are selectable for copy-paste auditing.

### DIFF
--- a/src/gateway/sse-consumer.ts
+++ b/src/gateway/sse-consumer.ts
@@ -44,6 +44,14 @@ export interface ConsumeAgentSseOptions {
   onEvent?: OnEventCallback;
   /** Abort signal — breaks the loop when triggered. */
   signal?: AbortSignal;
+  /**
+   * Optional explicit turn-start anchor (ms epoch). When provided, used as
+   * the basis for ⏳/💭/✍️/turn_total measurements instead of the local
+   * `Date.now()` taken when consumeAgentSse begins iterating. Portal sets
+   * this at POST receipt so the timing covers the portal→runtime RPC hop
+   * the runtime cannot otherwise see.
+   */
+  turnStartTime?: number;
 }
 
 export interface SseConsumptionResult {
@@ -137,9 +145,35 @@ export async function consumeAgentSse(opts: ConsumeAgentSseOptions): Promise<Sse
   const pendingToolInputs = new Map<string, string[]>();
   const pendingToolStartTimes = new Map<string, number[]>();
   const pendingToolMessageIds = new Map<string, string[]>();
+  // Per-tool pre-tool-call thinking time captured at tool_execution_start, to
+  // be merged into the row's metadata at tool_execution_end (so the persisted
+  // metadata survives the round-trip without being clobbered by the
+  // extractPersistableDetails extraction from result.details).
+  const pendingPreThinkingMs = new Map<string, number[]>();
 
   let eventCount = 0;
   const startTime = Date.now();
+
+  // ── Per-turn timing capture (for ⏳ TTFT / 💭 thinking / total) ──
+  // turnStartTime: server-side anchor for the whole user→assistant turn.
+  //   Prefer caller-supplied (portal POST timestamp) over local startTime so
+  //   the portal→runtime RPC hop is included in measurements.
+  // firstTokenTime: first model output of any kind (text or tool call) — TTFT.
+  // lastBoundaryTime: latest moment the model was *given input* — turn start
+  //   initially, then bumped to each tool_execution_end. The gap between
+  //   lastBoundaryTime and the next emission (text_delta or tool_execution_start)
+  //   is what we call "model thinking time" — the part the user previously
+  //   couldn't see for tool-call gaps. Single-clock by design.
+  const turnStartTime = opts.turnStartTime ?? startTime;
+  let firstTokenTime: number | undefined;
+  let lastBoundaryTime = turnStartTime;
+  let assistantMsgFirstTextTime: number | undefined;
+  let pendingThinkingMs: number | undefined;
+  // ttft_ms is a turn-scoped anchor (turnStart → first token of the very
+  // first assistant message). Persisting it on subsequent messages would
+  // make a naive UI sum double-count the same interval N times. Tracked
+  // and only emitted once.
+  let firstAssistantPersisted = false;
 
   for await (const event of client.streamEvents(sessionId)) {
     if (signal?.aborted) break;
@@ -177,9 +211,26 @@ export async function consumeAgentSse(opts: ConsumeAgentSseOptions): Promise<Sse
 
       const toolStartTime = shiftPending(pendingToolStartTimes, toolName);
       const durationMs = toolStartTime != null ? Date.now() - toolStartTime : undefined;
+      const preThinkingMs = shiftPending(pendingPreThinkingMs, toolName);
+      // Surface duration + pre-thinking on the live event for frontend.
+      if (durationMs != null) {
+        (evt as Record<string, unknown>).durationMs = durationMs;
+      }
+      if (preThinkingMs != null) {
+        (evt as Record<string, unknown>).preThinkingMs = preThinkingMs;
+      }
       const toolInput = shiftPending(pendingToolInputs, toolName) || "";
       const existingMessageId = shiftPending(pendingToolMessageIds, toolName);
-      const metadata = extractPersistableDetails(toolResult?.details, redactionConfig);
+      const detailsMeta = extractPersistableDetails(toolResult?.details, redactionConfig);
+      // Merge pre-thinking back in — extractPersistableDetails only looks at
+      // the tool *result*, so we'd otherwise lose what we recorded at
+      // tool_execution_start. We persist the value even when 0/small so the
+      // UI can render a 💭 badge on every tool: a 0ms badge on the 2nd-Nth
+      // tool of a batch makes "one thinking → many tools" auditable.
+      const metadata: Record<string, unknown> | null =
+        preThinkingMs != null
+          ? { ...(detailsMeta ?? {}), pre_thinking_ms: preThinkingMs }
+          : detailsMeta;
       const delegationId = typeof metadata?.delegation_id === "string" ? metadata.delegation_id : null;
 
       if (persist) {
@@ -207,12 +258,29 @@ export async function consumeAgentSse(opts: ConsumeAgentSseOptions): Promise<Sse
       if (toolName === "task_report" && text) {
         taskReportText = text;
       }
+      // Bump the model-input boundary: the model now has the tool result and
+      // any subsequent thinking/text/tool-use is computed from this point.
+      lastBoundaryTime = Date.now();
     }
 
     // ── DB persistence: message_update (accumulate assistant text) ──
     if (eventType === "message_update") {
       const ame = evt.assistantMessageEvent as { type?: string; delta?: string } | undefined;
       if (ame?.type === "text_delta" && ame.delta) {
+        const nowAtDelta = Date.now();
+        if (firstTokenTime === undefined) firstTokenTime = nowAtDelta;
+        if (assistantMsgFirstTextTime === undefined) {
+          assistantMsgFirstTextTime = nowAtDelta;
+          // Boundary-based: covers thinking after the previous tool result
+          // (or from turn start), not just the gap inside one assistant
+          // message. The thinking is fully attributed to *this* text bubble.
+          pendingThinkingMs = nowAtDelta - lastBoundaryTime;
+        }
+        // Bump boundary on every text delta so a tool emitted right after
+        // text doesn't double-count the same thinking interval. The text
+        // bubble already showed 💭 for that gap; the tool's pre-thinking
+        // measures only the (typically tiny) handoff after text emission.
+        lastBoundaryTime = nowAtDelta;
         assistantContent += ame.delta;
         currentMsgText += ame.delta;
       }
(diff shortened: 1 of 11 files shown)


src/gateway/sse-consumer.ts
```diff
@@ -294,6 +382,42 @@ export async function consumeAgentSse(opts: ConsumeAgentSseOptions): Promise<Sse
...
+        //                       otherwise double-count if put on every msg)
+        //   💭 thinking_ms   — boundary → first text_delta (per message)
+        //   ✍️ output_ms    — first text_delta → message_end (per message)
+        //   turn_total_ms    — kept for the last message as audit cross-check
+        const nowAtEnd = Date.now();
+        const timing: Record<string, number> = {
+          turn_total_ms: nowAtEnd - turnStartTime,
+        };
+        if (!firstAssistantPersisted && firstTokenTime !== undefined) {
+          timing.ttft_ms = firstTokenTime - turnStartTime;
```
chent1996: **Cross-process clock drift can produce a negative `ttft_ms`.** `turnStartTime` arrives from Portal (one process / pod) via the `chat.send` RPC; `firstTokenTime` is `Date.now()` from Runtime (another process / pod). K8s nodes share NTP but the worst case is tens of ms in either direction — under regional cluster failover or a node out of NTP sync, the skew can be larger.

The frontend formatter masks the symptom (returns `""` for negative inputs), but the value still lands in `chat_messages.metadata.timing.ttft_ms` with a negative sign and gets fed into the metrics aggregation downstream.

Simplest guard: clamp here before persisting.
```ts
if (firstTokenTime !== undefined) {
  const ttft = firstTokenTime - turnStartTime;
  if (ttft >= 0) timing.ttft_ms = ttft;
  // negative → omit; downstream treats absence as "unknown" (correct).
}
```
Same pattern applies to `output_ms` (line ~414) and the `thinkingMs` calculation. Negative values are never meaningful timings — drop them at the source rather than letting them propagate.

src/gateway/sse-consumer.ts
```diff
@@ -294,6 +382,42 @@ export async function consumeAgentSse(opts: ConsumeAgentSseOptions): Promise<Sse
...
+        }
+        if (pendingThinkingMs !== undefined) {
+          // Suppress thinking_ms on the first assistant message when ttft is
+          // already on the row — they cover the same interval (turnStart →
+          // first text token), and showing both would let a naive sum
+          // double-count that segment. After the first message, ttft is
+          // omitted, so thinking_ms takes over as the boundary-based gap.
+          if (!(timing.ttft_ms !== undefined && Math.abs(pendingThinkingMs - (timing.ttft_ms as number)) < 50)) {
```
chent1996: **Double-negative + magic threshold reduces readability.**
```ts
if (!(timing.ttft_ms !== undefined && Math.abs(pendingThinkingMs - (timing.ttft_ms as number)) < 50)) {
  timing.thinking_ms = pendingThinkingMs;
}
```

Intent (after a few seconds of staring): "only suppress thinking_ms when ttft is set AND the two cover the same interval (within 50ms event-dispatch noise)". A positive form is easier:
```ts
const NOISE_FLOOR_MS = 50;  // approximates inter-event dispatch jitter
const overlapsTtft =
  timing.ttft_ms !== undefined &&
  Math.abs(pendingThinkingMs - timing.ttft_ms) < NOISE_FLOOR_MS;
if (!overlapsTtft) {
  timing.thinking_ms = pendingThinkingMs;
}
```

The `50` constant also has no comment explaining where it came from. Naming + sourcing it makes future tuning safer — if someone bumps it to 100 they'll know what they're trading off.

src/gateway/sse-consumer.ts
```diff
@@ -245,7 +332,8 @@ export async function consumeAgentSse(opts: ConsumeAgentSseOptions): Promise<Sse
           durationMs: null,
           metadata: {
             status: "running",
-            started_at: new Date().toISOString(),
+            started_at: new Date(nowAtStart).toISOString(),
+            pre_thinking_ms: preThinkingMs,
```
chent1996: **`pre_thinking_ms` is now persisted on every tool row.** Even a 2ms value lands in `chat_messages.metadata` for every tool call. The earlier behavior was that metadata could collapse to `null` when there were no other persistable fields (the test diff at sse-consumer.test.ts confirms this — previously `expect(toolRow.metadata).toBeNull()`, now `expect(toolRow.metadata).toEqual({ pre_thinking_ms: ... })`).

The payload increase is tiny per row, but it's a one-way schema decision: once production rows carry this field, downstream consumers (analytics, replays) will start depending on it. Worth a comment in the row telling future-you why a 2ms value is durable telemetry rather than diagnostic noise — your PR description's "one-thinking-many-tools auditing" reasoning belongs at the persistence point, not just at the badge-rendering point.

src/portal/siclaw-api.ts
```diff
@@ -2368,6 +2390,70 @@ export function registerSiclawRoutes(router: RestRouter, config: SiclawConfig, c
...
+
+    // Assistant rows: pull metadata JSON, parse client-side, harvest ttft/thinking.
+    const aParams: unknown[] = [cutoff];
+    let assistantSql = `SELECT m.metadata
+      FROM chat_messages m
+      JOIN chat_sessions s ON m.session_id = s.id
+      WHERE m.role = 'assistant' AND m.created_at >= ? AND m.metadata IS NOT NULL
+        AND (s.origin IS NULL OR s.origin NOT IN ('task', 'delegation'))`;
+    if (userFilter) { assistantSql += " AND s.user_id = ?"; aParams.push(userFilter); }
+    const [aRows] = await db.query(assistantSql, aParams) as [Array<{ metadata: unknown }>, unknown];
```
chent1996: **No `LIMIT`, no streaming, no SQL-side aggregation — full-window assistant metadata loaded into memory.** This SELECT pulls every assistant message in the period and parses each `metadata` JSON in Node. The inline comment estimates ~thousands of rows, but production tenants attach multi-KB metadata per assistant turn (timing + tool details + delegation state), and 30 days × heavy users gets large fast.

Under load this is an OOM risk for the Portal process — a single slow `/metrics/timing` request can hold tens to hundreds of MB while it parses and reduces. There's no concurrency cap on the route either.

Minimum hardening:
```ts
assistantSql += " ORDER BY m.created_at DESC LIMIT 50000";
```
plus a 503 response when the result hits the cap (so callers know they're getting a sampled view). Better long-term: a nightly job that materialises `metrics_timing_daily` (or `agent_traces.tokens_total`-style derived columns) so the endpoint just SELECTs pre-summarised rows.

src/portal/siclaw-api.ts
```diff
@@ -2368,6 +2390,70 @@ export function registerSiclawRoutes(router: RestRouter, config: SiclawConfig, c
...
+
+    const ttftValues: number[] = [];
+    const thinkingValues: number[] = [];
+    for (const r of aRows) {
+      const meta = safeParseJson<Record<string, unknown> | null>(r.metadata, null);
+      const timing = meta?.timing as Record<string, unknown> | undefined;
+      if (!timing) continue;
+      if (typeof timing.ttft_ms === "number") ttftValues.push(timing.ttft_ms);
+      if (typeof timing.thinking_ms === "number") thinkingValues.push(timing.thinking_ms);
```
chent1996: **No negative-value filter on `ttftValues` / `thinkingValues`.** The bash branch a few lines below (line 2448) correctly does `.filter((n) => Number.isFinite(n) && n >= 0)`. This branch pushes raw `timing.ttft_ms` / `timing.thinking_ms` directly.

Negative timings are produced by cross-process clock drift (see comment on `sse-consumer.ts:400`). Even after that source is fixed, historical rows already in production may carry negatives from before the fix. Apply the same filter here for consistency:
```ts
if (typeof timing.ttft_ms === "number" && timing.ttft_ms >= 0) ttftValues.push(timing.ttft_ms);
if (typeof timing.thinking_ms === "number" && timing.thinking_ms >= 0) thinkingValues.push(timing.thinking_ms);
```

src/portal/siclaw-api.ts
```diff
@@ -2368,6 +2390,70 @@ export function registerSiclawRoutes(router: RestRouter, config: SiclawConfig, c
...
+    }
+
+    // Bash tool rows: duration_ms is its own column — no JSON parsing needed.
+    const bParams: unknown[] = [cutoff];
+    let bashSql = `SELECT m.duration_ms AS durationMs
+      FROM chat_messages m
+      JOIN chat_sessions s ON m.session_id = s.id
+      WHERE m.role = 'tool' AND m.tool_name = 'bash' AND m.duration_ms IS NOT NULL
```
chent1996: **Card is labelled "Bash" but several tool kinds are persisted with `duration_ms`.** `tool_name = 'bash'` is technically correct (verified `restricted-bash.ts:230` registers as `'bash'`), but:
- `script-exec`, `local_script`, `read`, `write`, `kubectl`, `delegate_to_agent`, etc. all have `duration_ms` set by sse-consumer.
- The dashboard reader sees a `Latency Breakdown` card labelled "Bash" and reasonably assumes "all tools".

Two options, either is fine:
1. Rename the card row to "Bash (`tool_name='bash'`)" so the scope is explicit, OR
2. Drop the `tool_name` predicate — aggregate `duration_ms` across all tool rows, and add a `byTool` breakdown if you want per-tool detail later.

Option 2 is closer to what "latency breakdown" usually means in observability dashboards.

chent1996 (commented):
Nice feature — the single-clock design, the non-overlapping invariant (⏳ + Σ💭 + Σ✍️ + Σ⚙️ ≈ turn_total), and the careful TTFT-vs-thinking dedup are all the right calls. Test coverage is solid (the one-thinking-many-tools case especially). Backward compat is graceful — old rows render unchanged.

Two real concerns and a few smaller items below.

**Worth fixing before merge:**
- **Cross-process clock drift creates negative timings, and aggregation doesn't filter them.** Portal stamps `turnStartMs = Date.now()` in one process; Runtime computes `firstTokenTime - turnStartMs` in another. K8s pods don't guarantee NTP-tight clocks. If Portal's clock is even slightly ahead of Runtime's, ttft is negative. The frontend formatter handles negatives (returns empty string), but the metrics SQL aggregation pushes raw values into `ttftValues` / `thinkingValues` with no filter — the bash branch correctly filters `n >= 0`, the assistant branch doesn't. Negative outliers will silently pull averages down.
- **`/metrics/timing` pulls every assistant `metadata` JSON for the whole window into Node memory and aggregates client-side.** The comment estimates ~thousands of rows but a busy production tenant will accumulate many more over a 30-day window, with multi-KB metadata each. No `LIMIT`, no streaming, no SQL-side aggregation. This is an OOM risk on the Portal process under load. Suggest at minimum a hard `LIMIT 50000` + a 503 response when hit, or a periodic background job that materializes a `metrics_timing_daily` table.

**Smaller stuff:**
- The `Latency Breakdown` card hardcodes `tool_name = 'bash'`. It's the right tool name (verified `restricted-bash.ts:230` registers as `'bash'`) — but `script-exec` / `local_script` / `read` / `write` / `kubectl` etc. all have `duration_ms` too, and the card label is just "Bash". Either rename + scope the card explicitly to bash, or include other tool kinds.
- `pre_thinking_ms` is now persisted on every tool row even when the value is 0 / a few ms. Previously some metadata could collapse to NULL. Tiny payload change but a one-way schema decision — worth documenting why this is durable telemetry and not a debug aid.
- The TTFT-vs-thinking dedup uses a `Math.abs(...) < 50` magic threshold inside a double-negative — readable on the second pass, less so on the first. A short comment naming the unit ("50ms event-dispatch noise floor") and flipping the condition to a positive form would help.

Good work overall.

a-yeyang:
fixed these tiny problems

chent1996 (approved):
LGTM — all 6 items from my prior review have been addressed thoroughly:

- **Negative-timing source clamp**: `nonNegative()` helper applied at every cross-process delta (ttft, thinking, output, turn_total, pre_thinking). Negatives are dropped at the source rather than persisted as garbage; the comment correctly notes that downstream absence = "unknown", which is the right semantics.
- **`/metrics/timing` row cap**: `ROW_LIMIT = 50_000` with the `LIMIT N+1` truncation-detect trick and a `truncated: true` response flag the dashboard surfaces as a "sampled" badge. The TODO for a materialised `metrics_timing_daily` table is the right long-term direction.
- **Aggregation negative filter**: `>= 0` guards added to ttft/thinking pushes; matches the existing tool-branch filter and keeps historical (pre-clamp) rows from poisoning averages.
- **`Latency Breakdown` scope**: dropped the `tool_name = 'bash'` predicate, switched to per-tool bucketing sorted by invocation count DESC, with a Top 3/5/10 client-side slicer in the dashboard. Better than my original suggestion.
- **`pre_thinking_ms` persistence**: now conditional on the `nonNegative()` clamp returning a value, with a clear comment explaining why this is durable telemetry rather than debug noise.
- **`NOISE_FLOOR_MS` constant**: extracted with detailed comment, condition flipped to a positive form (`overlapsTtft`). Much easier to read.

Parallelising the two SELECTs with `Promise.all` is a nice extra — keeps `/metrics/live` and `/metrics/summary` responsive while the timing query is in flight.

One small heads-up (not blocking): the branch picked up commits from PR #260 ("fix sun/moon" dark-mode theme). If PR #260 was independently merged to main, those changes will collapse harmlessly on merge; if not, this PR is shipping two features at once, worth a one-line note in the PR description.

PR #261, opened 2026-05-08
Truncate chat session title fields before persistence

LikiosSedo:
## Summary
- clamp chat session titles and previews to their MySQL column limits before persistence
- reuse the same normalization for adapter RPC, internal adapter route, and portal chat session routes
- add unit coverage for long title/preview handling

## Test plan
- npm test -- src/portal/chat-session-fields.test.ts src/portal/adapter-rpc.test.ts
- npm run build

## Context
Long prompts were being used as chat session titles. MySQL rejects values longer than chat_sessions.title VARCHAR(255), which can surface as `Data too long for column 'title' at row 1` and interrupt chat sends even though the actual message content can be stored separately.

### DIFF
--- a/src/portal/chat-session-fields.ts
+++ b/src/portal/chat-session-fields.ts
@@ -1,0 +1,28 @@
+const CHAT_SESSION_TITLE_MAX_CHARS = 255;
+const CHAT_SESSION_PREVIEW_MAX_CHARS = 500;
+
+function truncateField(value: string, maxChars: number): string {
+  if (value.length <= maxChars) return value;
+
+  let truncated = "";
+  for (const char of value) {
+    if (truncated.length + char.length > maxChars) break;
+    truncated += char;
+  }
+  return truncated;
+}
+
+export function normalizeChatSessionTitle(title: unknown): string {
+  const value = typeof title === "string" && title.length > 0 ? title : "New Session";
+  return truncateField(value, CHAT_SESSION_TITLE_MAX_CHARS);
+}
+
+export function truncateChatSessionTitle(title: unknown): string | null {
+  if (typeof title !== "string") return null;
+  return truncateField(title, CHAT_SESSION_TITLE_MAX_CHARS);
+}
+
+export function normalizeChatSessionPreview(preview: unknown): string | null {
+  if (typeof preview !== "string" || preview.length === 0) return null;
+  return truncateField(preview, CHAT_SESSION_PREVIEW_MAX_CHARS);
+}
--- a/src/portal/siclaw-api.ts
+++ b/src/portal/siclaw-api.ts
@@ -39,6 +39,11 @@ import { evaluateScriptsAI } from "../gateway/skills/ai-security-reviewer.js";
 import { parseFrontmatter } from "../gateway/skills/builtin-sync.js";
 import { validateSchedule } from "../cron/cron-limits.js";
 import { validateKnowledgePackage } from "../shared/knowledge-package.js";
+import {
+  normalizeChatSessionPreview,
+  normalizeChatSessionTitle,
+  truncateChatSessionTitle,
+} from "./chat-session-fields.js";
 
 /** Trace viewer message limit — matches siclaw_main.cron-limits.MAX_TRACE_MESSAGES */
 const MAX_TRACE_MESSAGES = 200;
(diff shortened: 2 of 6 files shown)


src/portal/siclaw-api.ts
```diff
@@ -1334,7 +1335,7 @@ export function registerSiclawRoutes(router: RestRouter, config: SiclawConfig, c
 
     const fields: string[] = [];
     const values: unknown[] = [];
-    if ("title" in body) { fields.push("title = ?"); values.push(body.title); }
+    if ("title" in body) { fields.push("title = ?"); values.push(normalizeChatSessionTitle(body.title)); }
```
chent1996: **Wrong semantics for UPDATE.** `normalizeChatSessionTitle` was designed for INSERT-time defaulting — empty/null → `"New Session"`. Reusing it here means a user calling `PATCH {title: ""}` (intent: clear the title) gets silently rewritten to `"New Session"`. Since the schema declares `title VARCHAR(255)` *nullable* (migrate.ts:279), clearing should be allowed.

Split into two helpers:
```ts
// For INSERT: applies default + truncates
export function normalizeChatSessionTitle(title: unknown): string {
  const value = typeof title === "string" && title.length > 0 ? title : "New Session";
  return truncateField(value, 255);
}

// For UPDATE: truncates only, preserves empty/null intent
export function truncateChatSessionTitle(title: unknown): string | null {
  if (typeof title !== "string") return null;
  return truncateField(title, 255);
}
```
Then the PATCH at this line uses `truncateChatSessionTitle`, INSERT paths keep using `normalizeChatSessionTitle`.

src/portal/chat-session-fields.ts
```diff
@@ -0,0 +1,16 @@
+const CHAT_SESSION_TITLE_MAX_CHARS = 255;
+const CHAT_SESSION_PREVIEW_MAX_CHARS = 500;
+
+function truncateField(value: string, maxChars: number): string {
```
chent1996, line 4: **`String.prototype.slice` operates on UTF-16 code units, not Unicode code points.** If the 255th code unit happens to be the high half of a surrogate pair — e.g. an emoji `👋` (U+1F44B) or a supplementary-plane CJK char — the slice splits the pair and the result contains an unpaired surrogate (0xD800–0xDFFF). MySQL `utf8mb4` *might* store it (it accepts 3-byte CESU-8-style fragments) or *might* reject depending on `sql_mode`/`SET NAMES`, and downstream consumers (JSON.stringify, frontend rendering) get mojibake.

For SRE prompts this is a real edge case — users do paste error messages with emoji status indicators (✅ ❌ 🚨), and a 256-char paste cut at exactly the wrong byte breaks.

Safer:
```ts
function truncateField(value: string, maxChars: number): string {
  if (value.length <= maxChars) return value;
  // Walk by code points; stop before exceeding maxChars in code-unit length.
  let out = "";
  for (const ch of value) {
    if (out.length + ch.length > maxChars) break;
    out += ch;
  }
  return out;
}
```
The `for…of` iterates code points (handles surrogate pairs as one unit). Slightly slower but rounds to a clean character boundary. Worth it for a 4 KB-ish max overhead per call.

chent1996 (commented):
Solid bug fix — centralising the truncate/default logic into one module is cleaner than the inline `truncForVarchar` from PR #258 (which seems to have been lost in the squash anyway, since I can't find it on current main). The 255/500 limits match `migrate.ts:279-280` exactly, and coverage is complete — I checked every `INSERT/UPDATE chat_sessions` write path: 4 in total (REST + RPC in adapter.ts, INSERT + PATCH in siclaw-api.ts), all covered.

One real bug + a couple of edge-case suggestions below.

**Real bug:**
- The PATCH endpoint at `siclaw-api.ts:1338` uses `normalizeChatSessionTitle` for an UPDATE, but that function conflates create-time defaulting with update-time normalisation. A user sending `PATCH {title: ""}` to clear the title gets silently rewritten to `"New Session"` — they can never blank the title via the API. Likely needs two separate helpers (one with default for INSERT, one without for UPDATE).

**Smaller stuff:**
- `value.slice(0, 255)` slices on UTF-16 code units, not Unicode code points. If char #255 happens to be the high half of a surrogate pair (emoji or rare CJK), the slice produces an unpaired surrogate. Edge-case-y for SRE prompts but real if someone pastes a kubectl error containing emoji.
- Silent truncation in the PATCH path has no user feedback. Saving 300 chars and getting 255 back without a warning is friendlier than a 500 but worse than "hey, your title was shortened". Consider returning the truncated value in the response body so the frontend can compare and warn — or validate length client-side.
- (Out of scope but worth noting) The underlying product issue is that the user's first prompt becomes the title, often producing ugly mid-sentence truncations in the sidebar ("帮我查一下 nginx pod 为什么 crash 我看 kubectl describe 显示 ImagePullBack…"). The mainstream fix is an LLM-generated short title; for a bug fix that's the right deferral, but worth filing as a follow-up.

chent1996 (approved):
LGTM

PR #262, opened 2026-05-11
feat(chat): add copy buttons for user messages, tool commands, and tool outputs

a-yeyang:
Adds one-click copy buttons on user messages, agent responses, tool commands, and tool outputs in the chat UI. 

Users no longer have to drag-select text to grab a kubectl command, error trace, or agent reply — saves clicks and prevents copy errors. Aligns Siclaw's chat UX with mainstream AI chat products.

### DIFF
--- a/portal-web/src/components/chat/PilotArea.tsx
+++ b/portal-web/src/components/chat/PilotArea.tsx
@@ -1,4 +1,4 @@
-import { useRef, useEffect, useState, useCallback, useMemo, type MouseEvent } from "react"
+import { useRef, useEffect, useState, useCallback, useMemo } from "react"
 import {
   Terminal,
   User,
@@ -310,8 +310,15 @@ export function PilotArea({
     userScrolledAwayRef.current = distanceFromBottom > 300
   }, [])
 
+  const visibleForCopy = useMemo(() => messages.filter((m) => !m.hidden), [messages])
+
   return (
-    <div className="flex-1 flex flex-col h-full bg-card">
+    <div className="flex-1 flex flex-col h-full bg-card relative">
+      {visibleForCopy.length > 0 && (
+        <div className="absolute top-2 left-3 z-20">
+          <CopySessionButton messages={visibleForCopy} />
+        </div>
+      )}
       <div ref={scrollContainerRef} className="flex-1 overflow-y-auto px-4 lg:px-8 py-8" onScroll={handleScroll}>
         <div className="max-w-5xl mx-auto space-y-8">
           {isLoadingHistory ? (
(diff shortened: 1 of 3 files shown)


portal-web/src/components/chat/PilotArea.tsx
```diff
@@ -1156,6 +1156,38 @@ function DelegationStatusNotice({ content }: { content: string }) {
...
+  className,
+}: {
+  text: string
+  title?: string
+  className?: string
+}) {
+  const [copied, setCopied] = useState(false)
+  const handleCopy = (e: React.MouseEvent) => {
+    e.stopPropagation()
+    navigator.clipboard.writeText(text).then(() => {
```
chent1996: **`navigator.clipboard.writeText` can throw or reject — no handling here.**

- On non-secure origins (plain HTTP, not localhost) `navigator.clipboard` is `undefined`. The `.writeText()` call throws a `TypeError` synchronously, the `.then()` chain never runs, and the user sees nothing happen. Your own `scripts/README.md` documents accessing Portal via `http://<DEV-MACHINE-IP>:3003` — that path is not a secure context.
- Even on a secure context, the promise can reject (permissions denied, document not focused). No `.catch()`, so the error becomes an unhandled rejection.
- The same bug exists in `CopyableMessage.handleCopy` at line 1195 — fix both.

Minimum:
```ts
const handleCopy = async (e: React.MouseEvent) => {
  e.stopPropagation()
  try {
    if (!navigator.clipboard) throw new Error("Clipboard API unavailable (HTTP origin?)")
    await navigator.clipboard.writeText(text)
    setCopied(true)
    setTimeout(() => setCopied(false), 2000)
  } catch (err) {
    console.warn("[copy] failed:", err)
    // Optional: toast/inline hint so the user knows it didn't work
  }
}
```

portal-web/src/components/chat/PilotArea.tsx
```diff
@@ -1527,78 +1562,86 @@ function ToolItem({ message }: { message: PilotMessage }) {
...
-          <ChevronRight
-            className={cn("w-3.5 h-3.5 text-muted-foreground/70 transition-transform shrink-0", isOpen && "rotate-90")}
-          />
-          <Terminal className="w-4 h-4 text-muted-foreground shrink-0" />
-          <span className="font-mono text-xs font-semibold text-foreground shrink-0">{message.toolName}</span>
+      <div className="group/tool bg-card border border-border rounded-lg shadow-sm shadow-black/10 overflow-hidden">
+        <div className="flex items-center gap-2 w-full px-4 py-2 bg-secondary border-b border-border min-w-0">
+          <button
+            type="button"
+            onClick={() => setExpanded(!expanded)}
```
chent1996: **Tool row click-to-expand scope shrank.** The previous structure wrapped the entire header row in one `<button>`, so clicking *anywhere* on the row (command text, empty space, status icons) toggled expand. After this refactor only the chevron + Terminal icon + tool name area triggers expand — that's a much smaller hit target.

The motivation (allowing text selection on `toolInput` via `select-text` + `stopPropagation`) is good. But the side effect is a workflow regression for SREs scanning many tool rows: previously a sloppy click anywhere on the row worked.

Two options:
1. Keep the small expand button but also wrap the *empty trailing space* (the `flex-1` gap between command and status icons) as another expand-trigger affordance.
2. Restore the outer `<button>` but use `onMouseDown={e => e.stopPropagation()}` selectively on the toolInput span — same selection benefit, restores click-anywhere-to-expand.

Option 2 is closer to the original behavior.

portal-web/src/components/chat/PilotArea.tsx
```diff
@@ -1176,15 +1208,18 @@ function CopyableMessage({ isUser, content }: { isUser: boolean; content: string
...
-      )}
+      <button
+        onClick={handleCopy}
+        className={cn(
+          "absolute top-2 right-2 opacity-0 group-hover:opacity-100 transition-opacity p-1 rounded-md",
+          isUser
+            ? "text-white/60 hover:text-white hover:bg-white/15"
+            : "text-muted-foreground/50 hover:text-foreground hover:bg-secondary",
+        )}
+        title="Copy"
```
chent1996: **Title `"Copy"` is less honest than the previous `"Copy markdown"`.** Assistant messages still copy `content` (raw markdown source) — `**bold**`, `[link](url)`, code fence backticks, etc. Users seeing just "Copy" might paste into Slack expecting rendered text and be surprised.

Either keep the original `"Copy markdown"` label (already accurate, mirrors GitHub's convention), or add a render-then-copy path that walks the rendered DOM and serializes to plain text.

portal-web/src/components/chat/PilotArea.tsx
```diff
@@ -1156,6 +1156,38 @@ function DelegationStatusNotice({ content }: { content: string }) {
...
+  className?: string
+}) {
+  const [copied, setCopied] = useState(false)
+  const handleCopy = (e: React.MouseEvent) => {
+    e.stopPropagation()
+    navigator.clipboard.writeText(text).then(() => {
+      setCopied(true)
+      setTimeout(() => setCopied(false), 2000)
+    })
+  }
```
chent1996, line 1267: **Re-click during the 2s feedback window leaves dangling timers.** A user who clicks Copy twice in quick succession spawns two `setTimeout`s; the first one fires 2s after the *first* click, clearing the green check before the user expects. Tiny issue but easy to fix with a ref:

```ts
const timerRef = useRef<ReturnType<typeof setTimeout> | null>(null)
const handleCopy = async (e: React.MouseEvent) => {
  // ... clipboard call ...
  setCopied(true)
  if (timerRef.current) clearTimeout(timerRef.current)
  timerRef.current = setTimeout(() => setCopied(false), 2000)
}
```

chent1996 (commented):
Small, focused PR — the new `CopyIconButton` is a clean reusable component, and adding copy to user messages / tool commands / tool outputs is a real UX win. A few issues worth addressing:

**Real concerns:**
- **`navigator.clipboard.writeText` is unguarded.** No try/catch, no fallback, no error toast. The Clipboard API is `undefined` on plain HTTP origins (and gated by user permission on some browsers); on those paths the user clicks Copy and gets silently nothing. Internal-cluster Siclaw access via `kubectl port-forward http://localhost:3003` actually does have access (localhost is treated as secure), but accessing via `http://<dev-ip>:3003` (also documented in `scripts/README.md`) does not.
- **Tool row click-to-expand scope shrank.** Previously the whole header row was the expand affordance — clicking the command text or the empty space toggled. Now only the chevron + icon + tool-name area triggers expand; the command text has `select-text cursor-text` + `stopPropagation`. Allowing text selection is good, but losing whole-row click is a workflow regression for SREs scanning many tool rows.
- **Title changed `"Copy markdown" → "Copy"` on assistant messages, but the content is still raw markdown source.** Users seeing "Copy" might expect plain text and get back `**bold**` / `[link](url)` syntax. Either keep the more honest label, or actually render-then-copy.

**Smaller stuff:**
- Dead import: `type MouseEvent` at line 1 is no longer referenced (the old IIFE in `ToolItem` that used it was refactored away).
- `package.json` adds `build:web` / `build:all` scripts that are orthogonal to the copy-button feature. Harmless but ideally a separate PR.
- Copy-feedback timeout uses `setTimeout` without tracking the handle. Re-clicking before 2s elapses creates a second timeout that resets the visual state mid-flash. `useRef` to clear the pending timeout would smooth this out.

_(One more nit not anchorable to the diff: `type MouseEvent` at line 1 is now a dead import — its only consumer was the old IIFE that this PR refactored away.)_

chent1996 (approved):
LGTM — every item from my prior review was addressed properly, and the bonus accessibility/UX work is appreciated:

- **Clipboard guard**: `copyTextToClipboard()` tries the async API in secure contexts, then falls back to the legacy textarea + `execCommand` path. Returns a boolean so callers skip state updates on failure. Exactly the right pattern for the HTTP-origin case the README documents.
- **Tool row click-to-expand restored**: outer wrapper became `<div role="button" tabIndex={0}>` covering the whole row again, with `onMouseDown`/`onClick stopPropagation` on the toolInput span to keep text selection. Bonus: `aria-expanded` + Enter/Space keyboard handler — that's a real accessibility upgrade, not just a fix.
- **`title="Copy markdown"`** kept on assistant messages, `"Copy"` on user messages where content is already plain. Honest labels.
- **Dead `MouseEvent` import** removed.
- **`useCopyFeedback`** custom hook with `useRef`-tracked timer + cleanup-on-unmount `useEffect`. Cleanly reused by all three copy components.

The new `CopySessionButton` + `serializeSessionToText` is a sensible addition — plain-text serialization is portable, hidden-message + delegation-notice filtering is correct, and reusing the same `useCopyFeedback` / `copyTextToClipboard` infra keeps the surface small.

A few non-blocking observations (file as follow-ups if you care):
- `CopySessionButton` is `position: absolute top-2 left-3` — worth a quick browser check that it doesn't overlap the first message at the top of long sessions.
- When both clipboard paths fail (rare), the button stays in idle state. A 2-second red-error variant would close the loop.
- Tool serialization uses `$ ` prefix uniformly; reads like `[read]\n$ {"path":"/etc/hosts"}` for non-shell tools. Cosmetic only.

PR #264, opened 2026-05-11
Support zip/tar skill pack upload with upsert semantics

chent1996:
## Summary

Make Portal's admin skill pack upload work end-to-end: drop the missing `unzip` OS dependency, accept tar/tar.gz alongside zip, switch the upload to upsert semantics so a small test pack no longer wipes every other builtin, and fix the preview UI that was rendering blank rows because of a front/back contract mismatch.

### Problem

- Upload was broken in production: `skill-import.ts` shelled out to `unzip`, but the Portal `node:22-slim` image doesn't ship `unzip`, so every upload returned `unzip: not found`.
- The endpoint only accepted `.zip` even though tar packs are the de-facto standard for skill bundles elsewhere.
- `executeImport` did a full three-way sync, so uploading a one-skill test pack would delete every other builtin and unbind them from agents.
- Preview cards displayed `Added: 1` but the row underneath was blank — the frontend `DiffEntry` type expected objects with `name` while the backend was returning bare strings, so `item.name` rendered as `undefined`.

### Solution

- Replace `execSync('unzip')` with `adm-zip`, add `tar` for tar/tar.gz, dispatch by magic bytes — no OS binary needed.
- Guard zip entries against `..` traversal; tar's defaults already sanitize.
- Give `executeImport` a required `mode: "sync" | "upsert"` parameter with no default. Admin pack upload uses `upsert`; `init` / `rollback` / `bootstrap` keep `sync` because those operations are state replacements by definition.
- Reshape `ImportDiff.added/updated/unchanged` to `{ name, description }[]` so the frontend can render the existing description column. History rows still persist `string[]` of names via `.map(d => d.name)`, so older snapshots and the history list keep working.
- Frontend: accept `.zip / .tar / .tar.gz / .tgz` in the file picker, pick a matching Content-Type, drop the never-sent `total_*` fields and read `.length` off the arrays, reword the confirm dialog.

## Test Plan

- [x] `npm test` — 3036 passing (added 8 new tests in `skill-import.test.ts`: tar + tar.gz happy paths, traversal rejection, garbage-buffer rejection, gzip-non-tar rejection, upsert no-delete invariant)
- [x] `cd portal-web && npx vite build` — clean
- [ ] Manual: upload `hello-zip-test.zip`, `hello-tar-test.tar`, `hello-tar-test.tar.gz` against a Portal pointed at a non-empty DB; verify preview shows skill names + descriptions, and that existing builtins are untouched after Confirm Import.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,6 +15,7 @@
         "@mariozechner/pi-coding-agent": "^0.55.3",
         "@modelcontextprotocol/sdk": "^1.27.1",
         "@sinclair/typebox": "^0.34.0",
+        "adm-zip": "^0.5.17",
         "diff": "^8.0.3",
         "js-yaml": "^4.1.1",
         "jsonwebtoken": "^9.0.3",
@@ -24,12 +25,14 @@
         "prom-client": "^15.1.3",
         "sqlite-vec": "^0.1.7-alpha.2",
         "ssh2": "^1.17.0",
+        "tar": "^7.5.15",
         "ws": "^8.18.0"
       },
       "bin": {
         "siclaw": "siclaw.mjs"
       },
       "devDependencies": {
+        "@types/adm-zip": "^0.5.8",
         "@types/diff": "^7.0.2",
         "@types/express": "^5.0.6",
         "@types/js-yaml": "^4.0.9",
@@ -1480,6 +1483,18 @@
       },
       "engines": {
         "node": ">=12"
+      }
+    },
+    "node_modules/@isaacs/fs-minipass": {
+      "version": "4.0.1",
+      "resolved": "https://registry.npmjs.org/@isaacs/fs-minipass/-/fs-minipass-4.0.1.tgz",
+      "integrity": "sha512-wgm9Ehl2jpeqP3zw/7mo3kRHFp5MEDhqAdwy1fTGkHAwnkGOVsgpvQhL8B5n1qlb01jV3n/bI0ZfZp5lWA1k4w==",
+      "license": "ISC",
+      "dependencies": {
+        "minipass": "^7.0.4"
+      },
+      "engines": {
+        "node": ">=18.0.0"
       }
     },
     "node_modules/@jridgewell/sourcemap-codec": {
@@ -3223,6 +3238,16 @@
       "integrity": "sha512-C5Mc6rdnsaJDjO3UpGW/CQTHtCKaYlScZTly4JIu97Jxo/odCiH0ITnDXSJPTOrEKk/ycSZ0AOgTmkDtkOsvIA==",
       "license": "MIT"
     },
+    "node_modules/@types/adm-zip": {
+      "version": "0.5.8",
+      "resolved": "https://registry.npmjs.org/@types/adm-zip/-/adm-zip-0.5.8.tgz",
+      "integrity": "sha512-RVVH7QvZYbN+ihqZ4kX/dMiowf6o+Jk1fNwiSdx0NahBJLU787zkULhGhJM8mf/obmLGmgdMM0bXsQTmyfbR7Q==",
+      "dev": true,
+      "license": "MIT",
+      "dependencies": {
+        "@types/node": "*"
+      }
+    },
     "node_modules/@types/body-parser": {
       "version": "1.19.6",
       "resolved": "https://registry.npmjs.org/@types/body-parser/-/body-parser-1.19.6.tgz",
@@ -3681,6 +3706,15 @@
       },
       "engines": {
         "node": ">= 0.6"
+      }
+    },
+    "node_modules/adm-zip": {
+      "version": "0.5.17",
+      "resolved": "https://registry.npmjs.org/adm-zip/-/adm-zip-0.5.17.tgz",
+      "integrity": "sha512-+Ut8d9LLqwEvHHJl1+PIHqoyDxFgVN847JTVM3Izi3xHDWPE4UtzzXysMZQs64DMcrJfBeS/uoEP4AD3HQHnQQ==",
+      "license": "MIT",
+      "engines": {
+        "node": ">=12.0"
       }
     },
     "node_modules/agent-base": {
@@ -4240,6 +4274,15 @@
       "license": "MIT",
       "engines": {
         "node": ">= 16"
+      }
+    },
+    "node_modules/chownr": {
+      "version": "3.0.0",
+      "resolved": "https://registry.npmjs.org/chownr/-/chownr-3.0.0.tgz",
+      "integrity": "sha512-+IxzY9BZOQd/XuYPRmrvEVjF/nqj5kgT4kEq7VofrDoM1MxoRjEWkrCC3EtLi59TVawxTAn+orJwFQcrqEN1+g==",
+      "license": "BlueOak-1.0.0",
+      "engines": {
+        "node": ">=18"
       }
     },
     "node_modules/cli-highlight": {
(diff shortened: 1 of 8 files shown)


jacoblee-io:
## Code Review

Three parallel reviews run (code quality, test coverage, error handling). Verified the load-bearing claims against the actual files. **Verdict: core refactor is correct and well-tested — merge after fixing one frontend bug and tightening test/error-handling coverage.**

---

### Critical — Must fix this PR

**1. Frontend history list never renders counts** — `portal-web/src/pages/SkillImport.tsx:23-29, 384-391`

`ImportVersion` types `added/updated/deleted` as `number`, but `GET /skills/import/history` (`siclaw-api.ts:1067-1071`) decodes those columns as `string[]`. The render guards `{v.added > 0 && …}` evaluate `["x"] > 0` → `NaN` (falsy), so the badges never appear; if they ever did they'd print `+x,y,z added` literally. This is pre-existing on `main`, but the PR description specifically reassures that the legacy `string[]` shape is preserved *so the history list keeps working* — which is moot when the UI never read it correctly. Two-line fix:

~~~ts
added: string[]; updated: string[]; deleted: string[];
// {v.added.length > 0 && <span>+{v.added.length} added</span>}
~~~

---

### Important — Worth addressing in this PR

**2. `err.message` echoed to client on 500/400** — `siclaw-api.ts:1001, 1022, 1052, 1106`

Leaks absolute `/tmp/skill-import-<uuid>/...` paths from `extractTar` / `parseSkillsDir`, plus mysql2 error text (SQL fragment, table names). Admin-only mitigates impact, but the parse-failure path is reached pre-validation. Replace with a generic message + structured `logError(err, { user, archiveSize, contentType })`.

**3. Missing integration tests for the new endpoint behaviour**

Unit tests on `executeImport` / `parseSkillPack` are solid (18/18 pass). But `src/portal/siclaw-api.skills.test.ts` covers neither (a) the dry-run `deleted: []` zero-out at `siclaw-api.ts:1011` nor (b) that the upload route passes `mode: "upsert"`. A typo-level regression (dropping `mode: "upsert"` or the `...diff, deleted: []` spread) would slip through. Suggested cases:
- `POST /skills/import dry_run=true returns deleted:[] even when DB has builtins absent from pack`
- `POST /skills/import (non-dry-run) calls executeImport with mode=upsert`
- `POST /skills/import/rollback calls executeImport with mode=sync`

**4. Tar slip negative test is missing**

Zip traversal is covered (`skill-import.test.ts:351-362`). Equivalent for tar (`..` segment, absolute path, or symlink with `..` target) is absent. node-tar v7 rejects these by default — verified — but a regression test protects against a future dep swap or option change. Also: `"rejects gzip buffers that are not actually tar inside"` asserts only `.rejects.toThrow()` with no message — fragile.

**5. Post-commit ordering is brittle** — `skill-import.ts:316-354`

`conn.commit()` at line 316 is followed by the `notifyAgentReload` loop (319-323) and the `skill_import_history` INSERT (338), all still inside the same try/catch that calls `conn.rollback()` on error:
- A throw from `notifyAgentReload` triggers `rollback()` on an already-committed connection and skips the history INSERT — DB mutated, no audit row, 500 returned.
- A throw from the history INSERT itself leaves the DB updated with no rollback point.
- Separately, `notifySkillAgents` is **not currently wired** into the ctx that `server.ts:183-187` passes to `registerSiclawRoutes` (no `notifySkillAgents` field). So `ctx?.notifySkillAgents?.()` is a silent no-op in prod — admins importing skills don't notify bound agents. Pre-existing on `main`, but this PR's new endpoint inherits the gap.

Not strictly an in-PR regression, but the diff exercises this path more. Either (a) move the notify loop strictly outside the try/catch with per-call error swallowing, or (b) wire `notifySkillAgents` in `server.ts` and add a `logError` on throw.

---

### Nice-to-have

- **No upload-size cap** on the import route (`siclaw-api.ts:966-968`). Admin-only mitigates abuse, but a zip bomb or malformed tar can fill `/tmp` plus RAM. Soft cap before `Buffer.concat`.
- **`/tmp` is hard-coded** at `skill-import.ts:104`. Prefer `os.tmpdir()` so test asserts and containers with moved tmp don't break.
- **`fs.rmSync` in `finally`** at `skill-import.ts:127` can mask the original error on EBUSY/EACCES. Wrap in `try/catch { logError }`.
- **`computeImportDiff` does an N+1** `agent_skills` query per deleted builtin (`skill-import.ts:183`). Cheap to batch into one `IN (...)`.

---

### Verified non-issues

- Zip-slip name-based check is correct for adm-zip ^0.5.17 — backslash normalisation handles the Windows-mix case.
- node-tar ^7.5.15 default `preservePaths: false` strips absolute paths and rejects `..`-traversal and traversing symlinks — the PR comment is accurate.
- `incomingNames` set vs. `builtinMap` iteration at `skill-import.ts:155, 180` is correct.
- Frontend upload error handling (`SkillImport.tsx:119-122, 155-158`) is thorough — toast on every failure path.

---

### Strengths

- Magic-byte format detection is cleaner than relying on filename/Content-Type.
- `ImportMode` has no default — every caller must declare intent. Good API design.
- Upsert no-delete test asserts the **negative** SQL surface (`expect(sqls).not.toContain("DELETE FROM skills …")`) — exactly the right shape.
- Killing `execSync('unzip')` removes both an OS dep and a command-injection surface.

---

### Recommended action

1. Fix finding **#1** (frontend history typing) — two-line change, same PR.
2. Add the three integration tests in **#3** — cheap, protect the new wire.
3. Add a tar-traversal regression test (**#4**).
4. Scrub `err.message` in 500 responses (**#2**) — log full, return generic.
5. Decide whether to fix the agent-notify wiring (**#5**) here or in a follow-up; either way, a one-line `logError` on notify throw is a free safety win.

jacoblee-io:
## Follow-up review on `1234228` — LGTM

Re-ran the suite against the fix commit: **48/48 tests pass, tsc clean.**

### All previous findings addressed

| # | Issue | Fix |
|---|-------|-----|
| 1 | Frontend history list never renders counts | `ImportVersion` now `string[]`, render uses `.length` (`SkillImport.tsx:27-31, 386-394`) |
| 2 | `err.message` echoed to client | Internals go to `console.error`, response is a generic string (`siclaw-api.ts:344-347, 374-376, 411-413, 461-463`) |
| 3 | Missing integration tests | Three new tests cover dry-run `deleted:[]`, upload→`mode=upsert`, rollback→`mode=sync` (`siclaw-api.skills.test.ts:454-538`) |
| 4 | Tar-slip negative test | Tar header is forged with `../evil.txt` + recomputed checksum so `filter` actually fires (`skill-import.test.ts:443-477`) |
| 5 | Notify loop inside transaction try/catch | Moved outside `try/finally`; each callback wrapped in its own try/swallow + log (`skill-import.ts:344-355`) |
| nit | Hardcoded `/tmp` | `os.tmpdir()` (`skill-import.ts:124`); tests updated accordingly |

### Bonus: routing bug caught and fixed

The author independently caught that `/skills/import/rollback` was being shadowed by `POST /skills/:id/rollback` with `id="import"` (first-match-wins routing). The entire import block was moved above the parameterized routes (`siclaw-api.ts:299-465`, was 1123-1278). I missed this in the first pass — good catch.

### Notable design touches

- **`extractTar` filter is defense-in-depth done right** (`skill-import.ts:91-115`): throwing from `tar`'s `filter` callback gets swallowed by its internal chain, so the fix records `unsafeEntry` and throws *after* extraction completes. The comment explains exactly why, which is what saves the next reader from "re-simplifying" it back into a broken throw.
- **The integration tests mock `skill-import.js` rather than the DB** (`siclaw-api.skills.test.ts:26-30`). Right call — endpoint tests should verify wiring, not re-test `executeImport`'s internals.

### Still deferred (pre-existing, fine for a follow-up)

- `notifySkillAgents` not wired into the ctx in `server.ts:183-187` — agent reload after admin import is currently a silent no-op in prod. The route now calls `console.error` on a thrown notify, so observability is at least non-zero.
- `bootstrap-portal.ts:44` `autoInitBuiltinSkillsIfEmpty` has no try/catch — a transient DB error during one-shot init still takes down Portal boot.

Both are pre-existing on `main` and out of scope for this PR.

**Ready to merge.**

jacoblee-io (approved):
LGTM — all review findings addressed in 1234228, 48/48 tests pass, tsc clean. See follow-up comment for details.

PR #265, opened 2026-05-13
Feat/create chart mcp

a-yeyang:
Add a create-chart MCP server exposing a render_chart tool that emits a fenced ```chart JSON block for pie/bar/line charts from structured data (status counts, category summaries, time-series samples). Portal-web's Markdown renderer detects the block and delegates to a new ChartRenderer that draws inline SVG. The server is bundled into Dockerfile.agentbox and registered in mcp/MCP_LIST.txt. Includes input-schema validation, artifact persistence under .siclaw/user-data/tool-results/create-chart/, and handler unit tests.


portal-web/src/components/chat/ChartRenderer.tsx
```diff
@@ -0,0 +1,655 @@
...
+  )
+}
+
+export function tryParseChartSpec(raw: string): ChartSpec | null {
+  try {
+    const obj = JSON.parse(raw)
+    if (!obj || typeof obj !== "object") return null
+    if (obj.type !== "pie" && obj.type !== "bar" && obj.type !== "line") return null
+    if (!obj.data || typeof obj.data !== "object") return null
+    return obj as ChartSpec
```
chent1996: **Shallow validation lets malformed agent JSON crash the bubble.** Only `type` and `data` (object existence) are checked, then cast to `ChartSpec`. Downstream:
- `BarChart` does `series.forEach(s => s.values.forEach(...))` — fails if `values` isn't an array
- `LineChart` does `series.flatMap(s => s.points)` — fails if `points` is missing or non-array
- `PieChart` does `slices.reduce(...)` — fails if `slices` isn't an array

A buggy or jailbroken model emits `{"type":"bar","data":{"categories":["a"],"series":[{"name":"x","values":"oops"}]}}` → `values.forEach is not a function` → React unwinds the entire chat bubble. No ErrorBoundary in portal-web (grepped) so the crash propagates up to the message renderer.

Mirror the backend `validate()` logic here, or wrap the call site in an ErrorBoundary. Backend validation can't be relied on because the spec round-trips through the LLM (model may rewrite values between MCP response and final reply).

Minimum: tighten this function to assert each variant's nested shape and return `null` when wrong — `Markdown.tsx:75` already falls back to `<ChartParseError>` for null, which gives the user a real diagnostic.

mcp/create-chart/src/handler.ts
```diff
@@ -0,0 +1,171 @@
...
+      const item = s as { name?: unknown; values?: unknown };
+      if (!Array.isArray(item.values)) {
+        throw new Error(`render_chart: bar series[${i}].values must be an array`);
+      }
+      return {
+        name: String(item.name ?? `series ${i}`),
+        values: item.values.map((v) => (typeof v === "number" ? v : Number(v))),
+      };
+    });
+    return { type, data: { categories, series }, ...common };
```
chent1996, line 156: **Bar chart: no validation that `values.length === categories.length`.** The frontend `BarChart` indexes `s.values[gi] ?? 0` (silent zero for missing) and ignores trailing values without warning. A model emitting 3 categories and 5 values produces a chart that looks plausible but silently drops data — a hard-to-debug class of bug.

Suggest:
```ts
const series = d.series.map((s, i) => {
  const item = s as { name?: unknown; values?: unknown };
  if (!Array.isArray(item.values)) {
    throw new Error(`render_chart: bar series[${i}].values must be an array`);
  }
  if (item.values.length !== categories.length) {
    throw new Error(
      `render_chart: bar series[${i}].values length (${item.values.length}) must equal categories length (${categories.length})`,
    );
  }
  return { ... };
});
```
Line series should also validate that `points` x-values are monotonically increasing for line charts (if you want to be strict), or at least dedupe — but length mismatch in bar is the highest-impact issue.

mcp/create-chart/src/handler.ts
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  const result: RenderChartResult = {
+    chart_id: id,
+    type: args.type,
+    spec_path: specPath ?? "",
+    svg_path: specPath ?? "",
+    bytes: Buffer.byteLength(spec, "utf8"),
+    markdown_embed: markdownEmbed,
+    markdown_embed_raw: markdownEmbed,
```
chent1996: **`markdown_embed` and `markdown_embed_raw` are literally the same string.** Both `=` to `markdownEmbed`. If the intent was that one should be JSON-escaped and the other raw, that's not what's happening — both will be JSON-escaped exactly the same way by the MCP transport.

If they're intentionally identical, drop one — the duplicate field is more likely to confuse the LLM (which picks whichever name looks 'safer') than to help it. The `READY_TO_PASTE:` text section in the response body is already the canonical "paste this verbatim" form.

portal-web/src/components/chat/ChartRenderer.tsx
```diff
@@ -0,0 +1,655 @@
...
+  const canvas = document.createElement("canvas")
+  canvas.width = Math.round(w * scale)
+  canvas.height = Math.round(h * scale)
+  const ctx = canvas.getContext("2d")
+  if (!ctx) throw new Error("canvas 2d context unavailable")
+  // Fill opaque white so PNGs pasted into WeChat/QQ don't show a transparent
+  // halo when the theme background was dark — chart-bg already paints, but
+  // anti-aliased edges can leak through without this base.
+  ctx.fillStyle = "#ffffff"
+  ctx.fillRect(0, 0, canvas.width, canvas.height)
```
chent1996, line 544: **Forced white fill creates a dark-mode-PNG hazard.** Trace:
1. User in dark mode → on-screen chart has slate-900 bg + light-gray text (via `inlineComputedStyles` copying live colors onto the SVG clone)
2. PNG export: canvas filled with `#ffffff`, then SVG rasterized on top
3. The SVG's own `<rect class="chart-bg">` should paint slate-900 over the white — restoring readability

But `inlineComputedStyles` only inlines `INLINEABLE_PROPS = ["fill", "stroke", "stroke-width", "font-family", "font-size", "font-weight"]` — if the chart-bg rect's `fill` doesn't make it through (e.g., the live computed style returns a CSS variable that doesn't resolve out-of-document, or a future class adds `background-color` instead of `fill`), the bg rect renders transparent → final PNG is light-gray text on white = unreadable.

The comment says the white fill is to prevent a "transparent halo" — a real concern — but the current implementation makes the failure mode worse than the failure it prevents. Two safer options:
1. Sample the SVG's chart-bg `getComputedStyle().fill` and use that as the canvas fill — falls back to white only when the read fails.
2. Force light theme on the clone before rasterisation (replace `chart-bg` fill with `#fff`, `chart-tick` fill with dark gray, etc.). Trade-off: PNG is always light-themed regardless of on-screen theme. Probably the right call for share-friendly output (WeChat doesn't have dark mode for embedded images either).

chent1996 (commented):
Nice clean architecture for a non-trivial feature — single MCP tool emits a fenced ```chart JSON block, frontend detects and renders as inline SVG with hover-to-PNG export. The contract between MCP and frontend is sharp (types mirror each other, tryParseChartSpec is the bridge). Tailwind dark: variants for theme is elegant.

The big risk is **trust boundary at `tryParseChartSpec`** — anything the LLM emits flows through it, and the validation is too shallow.

**Worth fixing before merge:**
- **Frontend validation gap can crash the chat bubble.** `tryParseChartSpec` only verifies top-level shape (`type`, `data` exists). It hands an `as ChartSpec` cast to `ChartRenderer`, which then does `s.values.forEach`, `series.flatMap`, etc. — a malformed agent JSON crashes the React render and unwinds the whole chat bubble. No ErrorBoundary in portal-web (I grepped). Either deep-validate in `tryParseChartSpec` or add an ErrorBoundary around `<ChartRenderer>`. Deep validation is preferable because it gives the user a useful error (`ChartParseError` already exists right below).
- **Bar chart: no check that `values.length === categories.length`.** Mismatched lengths render as silent visual chaos — `s.values[gi] ?? 0` substitutes zero for missing values, and extra values are silently ignored. Easy to assert in `validate()`.
- **PNG export's `#ffffff` white fill backfires in dark mode if the SVG's bg rect doesn't paint over it.** The chart-bg rect normally covers it, but `inlineComputedStyles` running on a detached clone can miss some properties. If it ever misses, dark-theme text (light gray) ends up rendered on white background — completely unreadable. Consider forcing light theme on the cloned SVG before rasterisation: the on-screen view stays theme-aware, the exported PNG is consistently light-themed for sharing.

**Smaller stuff:**
- `markdown_embed` and `markdown_embed_raw` in the tool result are literally the same string. If one is supposed to be the JSON-escaped form, that's not what's happening; if they're intentionally identical, drop one — agents reading the response will pick whichever looks correct and the dual fields just add confusion.
- The MCP description carries a lot of LLM-discipline text ("Do not rewrite, escape, quote, or wrap…"). That's understandable LLM-prompting tax, but ironically `READY_TO_PASTE:` itself is a string the model has to recognise and not include in its reply. Worth testing what happens when a model literally pastes `READY_TO_PASTE:\n\`\`\`chart...` into its message — the renderer's regex match on `language-chart` should still work because `language-chart` is the code fence class, not the surrounding text, but I'd verify.
- Each MCP server gets its own `node_modules` in the agentbox image (the Dockerfile loop runs `npm ci` per MCP). The MCP SDK plus deps is ~80 MB; once MCP_LIST.txt grows past a few entries this is significant image bloat. Worth a follow-up to share a workspace-level `node_modules`.
- Hardcoded Chinese UI strings ("已复制图片", "下载失败") inside the renderer component are consistent with the rest of the codebase, so OK — flagging only because if i18n is on the roadmap, this is one more callsite.

LikiosSedo:
I took a pass over this PR. Overall, I think the direction is right: making charting a standalone stdio `mcp-create-chart` is a better boundary than coupling it to any specific diagnosis MCP, and the Portal contract around ```chart fenced JSON blocks is clear. The `READY_TO_PASTE` envelope also addresses the real failure mode we saw where the model copied `markdown_embed` out of JSON and accidentally over-escaped the chart spec.

Before merge, I would treat these as merge blockers:

1. `tryParseChartSpec` only does shallow validation and then casts to `ChartSpec`. The chart block comes from the model’s final answer, so it should be treated as untrusted input. Malformed specs such as non-array `bar.series[].values`, missing `line.series[].points`, or non-array `pie.slices` can crash during React render. I would mirror the backend `validate()` shape checks in the frontend and return `null` so the existing `ChartParseError` path handles it.
2. The backend bar validator should require `values.length === categories.length`. Otherwise missing values silently render as zero and extra values are silently dropped, which can produce a plausible but incorrect chart.
3. `markdown_embed` and `markdown_embed_raw` are currently identical. I would drop `markdown_embed_raw`; `READY_TO_PASTE` should be the canonical paste target, and duplicate fields make the model more likely to choose the wrong thing.

The dark-mode PNG export concern is worth a follow-up, but the validation issues are correctness / trust-boundary problems and should be fixed before merging.

LikiosSedo:
Thanks for the update — the latest commit fixes most of the previous blockers: `markdown_embed_raw` is gone, backend bar series length is checked, the PNG dark-mode fallback is safer, and the streaming loading state is a nice touch.

I still see two issues worth fixing before merge:

1. Frontend validation is deeper now, but it still does not fully validate or normalize the values that the renderer assumes. For example, this passes `tryParseChartSpec` today:

```json
{"type":"bar","data":{"categories":[1],"series":[{"name":"s","values":[2]}]}}
```

Then `BarChart` calls `approxTextWidth(c, ...)`, and `approxTextWidth` does `for (const ch of text)`, which throws because `c` is a number. Similarly, missing/non-string `series[].name` can pass validation and later break the legend path. Line series has the same `name` issue. Since chart specs are model-emitted, `tryParseChartSpec` should either strictly validate `categories[]`, `series[].name`, `values[]`, `points[].x/y`, etc., or normalize them before returning `ChartSpec`.

2. Backend bar value coercion can still turn bad values into `NaN`, and `JSON.stringify` serializes `NaN` as `null`. For example, a value of `"oops"` becomes `Number("oops")`, then the emitted chart JSON contains `null`, which can silently produce a wrong/incomplete chart. After coercion, please check `Number.isFinite(n)` and throw when conversion fails.

After those two are tightened, I think the main correctness/trust-boundary concerns are addressed.

LikiosSedo (approved):

PR #271, opened 2026-05-19
Fix chart renderer label clipping and preview

LikiosSedo:
## Summary

- reserve left-side plot padding for rotated bar chart x-axis labels so long first labels are not clipped
- add a lightweight larger chart preview dialog that reuses the existing renderer and keeps copy/download/log controls available
- cover the rotated-label clipping case with a focused chart-utils regression test

## Validation

- `npm test -- portal-web/src/components/chat/chart-utils.test.ts portal-web/src/components/chat/ChartRenderer.test.ts`
- `npm run build` from `portal-web`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,3 +30,18 @@ jobs:
           cache: "npm"
       - run: npm ci
       - run: npm test
+
+  portal-web-test:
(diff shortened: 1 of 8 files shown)


chent1996 (approved):
LGTM

PR #272, opened 2026-05-19
fix: derive developer-role support for compatible models

LikiosSedo:
## Summary
- Add a shared model compatibility helper for default provider/model capabilities.
- Keep developer-role messages enabled for the official OpenAI API.
- Disable developer-role messages for OpenAI-compatible non-OpenAI gateways such as `https://api.example.com/model-api`, preventing 422 errors from providers that reject `role: "developer"`.
- Reuse the same helper in Portal CLI snapshots, provider presets, and `/setup` model creation.

## Verification
- `npm test -- src/core/model-compat.test.ts src/portal/cli-snapshot-api.test.ts`

## Notes
- This replaces the temporary inner build-time patch with a source-level Siclaw fix.

### DIFF
--- a/src/core/extensions/setup.ts
+++ b/src/core/extensions/setup.ts
@@ -40,6 +40,7 @@ import {
   type SiclawConfig,
 } from "../config.js";
 import { PRESETS } from "../provider-presets.js";
+import { defaultProviderModelCompat } from "../model-compat.js";
 
 // ---------------------------------------------------------------------------
 // Credential type labels
@@ -779,8 +780,6 @@ async function handleAddModel(
   const existingProvider = entries.find(([n]) => n === providerName)?.[1];
   if (!existingProvider) return;
 
-  const isAnthropic = existingProvider.api === "anthropic";
-
   const newModel: ProviderConfig["models"][number] = {
     id: modelId,
     name: modelName || modelId,
@@ -789,11 +788,10 @@ async function handleAddModel(
     cost: { input: 0, output: 0, cacheRead: 0, cacheWrite: 0 },
     contextWindow,
     maxTokens,
-    compat: {
-      supportsDeveloperRole: !isAnthropic,
-      supportsUsageInStreaming: true,
-      maxTokensField: "max_tokens",
-    },
+    compat: defaultProviderModelCompat({
+      api: existingProvider.api,
+      baseUrl: existingProvider.baseUrl,
+    }),
   };
 
   // Write to config
--- a/src/core/model-compat.test.ts
+++ b/src/core/model-compat.test.ts
@@ -1,0 +1,25 @@
+import { describe, expect, it } from "vitest";
+import { defaultProviderModelCompat } from "./model-compat.js";
+
+describe("defaultProviderModelCompat", () => {
+  it("keeps developer-role messages for the official OpenAI API", () => {
+    expect(defaultProviderModelCompat({
+      api: "openai-completions",
+      baseUrl: "https://api.openai.com/v1",
+    }).supportsDeveloperRole).toBe(true);
+  });
+
+  it("disables developer-role messages for OpenAI-compatible gateways", () => {
+    expect(defaultProviderModelCompat({
+      api: "openai-completions",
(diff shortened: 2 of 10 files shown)


jacoblee-io (approved):
## Review

LGTM — the headline fix (`Compatible` preset's `supportsDeveloperRole` flipping `true → false`) is correct, every production write-site for `compat` was migrated to the new helper, and the auth gates on `cli-snapshot-api.ts` remain intact (payload-only edit, still read-only).

No blocking issues. Suggestions below are non-blocking — feel free to take them in a follow-up PR.

### 🟡 Worth considering before / right after merge

**1. Possible Azure OpenAI regression — `src/core/model-compat.ts:13`**
The strict `hostname === "api.openai.com"` check drops `supportsDeveloperRole` for any Azure deployment (`*.openai.azure.com`), even though Azure's o-series models do accept `role: "developer"` on recent API versions. If anyone on the team is wired up to Azure for reasoning models, they'll silently lose the flag. Two paths:
- Broaden the check to also match `*.openai.azure.com`, or
- Add a per-provider override field on `ProviderModelCompat`.
If no one uses Azure today, a one-line code comment recording the intentional scope is enough.

**2. Silent degradation when user mistypes Base URL**
`isOfficialOpenAIBaseUrl` catches all `URL` parse errors and returns `false`. Combined with `src/core/extensions/setup.ts` (`baseUrl = entered.trim()` with no validation), a user who types `api.openai.com` without a scheme thinks they configured the official OpenAI provider but silently gets `supportsDeveloperRole: false` — no warning, no log.
Suggested follow-up: wire the existing `validateHttpUrl` pattern (`src/portal/siclaw-api.ts:2657-2667`) into `handleAddProvider`, and add a debug log when the helper's catch fires so misconfig surfaces in support sessions. Keep the catch in the helper itself — it's invoked everywhere and must not throw.

**3. Two un-pinned behavior changes**
- The Compatible preset flipping `true → false` is the headline behavior of this PR, but no test asserts the literal value of `PRESETS.find(p => p.name === "compatible").models[0].compat.supportsDeveloperRole`. A future refactor (e.g. setting the preset baseUrl to a placeholder) could silently re-break the fix.
- `setup.ts handleAddModel` flipped from `!isAnthropic` to URL-derived. This is the originally-broken user-facing flow and has zero direct coverage. Adapter-RPC tests cover the data-flow side, but not this UI path. Consider extracting the small `newModel` build into a pure helper for testability.

### 🟢 Nice to have

- `src/core/model-compat.test.ts` only covers 3 happy paths. Add edge cases: empty/null `baseUrl`, malformed `baseUrl` (no scheme), uppercase URL (`API.OPENAI.COM`), `api: null`, unknown api type, and `api: "openai"` vs `"openai-completions"` (both currently return `true` — pin whichever is intended).
- `model-compat.ts:17-19`: `Required<Pick<ProviderModelCompat, …>>` is over-specified. All call sites spread into a `compat:` field typed `ProviderModelCompat?`, so a plain `ProviderModelCompat` return type type-checks identically and ages better.
- New tests use `toMatchObject({ supportsDeveloperRole: false, … })`. Consider adding an explicit `expect(model.compat).toBeDefined()` so the "field exists at all" invariant is pinned.
- The two REST adapter call sites at `adapter.ts:453, 1042` aren't pinned for `compat` (only the two RPC handlers are). A `.toMatchObject({ compat: { supportsDeveloperRole: false } })` assertion in the compatible-provider case of the matching REST tests would close the gap.

### ✅ Verified clean

- `git grep "compat: {"` shows no orphan hand-coded constructors in production code — only type-signature occurrences remain.
- Null DB columns (`p.api_type`, `p.base_url`) are handled correctly by `api ?? ""` and the nullable `baseUrl` branch.
- The new `try/catch` in `isOfficialOpenAIBaseUrl` is the only error suppression introduced.
- `cli-snapshot-api.ts` edit is payload-only — the three auth gates (`enableCliSnapshot` flag, loopback origin, dedicated `X-Siclaw-Cli-Snapshot-Secret` header) are untouched; endpoint remains read-only. The matrix-required `cli-snapshot-api.test.ts` was extended and passes.

PR #273, opened 2026-05-19
fix: derive AgentBox gateway URL from runtime hostname

LikiosSedo:
## Summary

Fix K8s AgentBox pods using a hard-coded runtime service DNS name when they bootstrap their internal gateway connection.

The K8s spawner previously injected:

```ts
SICLAW_GATEWAY_URL=https://siclaw-runtime.<namespace>.svc.cluster.local:3002
```

That only works when the Helm release fullname is exactly `siclaw`. Releases such as `siclaw-debug` create services like `siclaw-debug-runtime`, so AgentBox cannot resolve the gateway and fails to load settings, MCP servers, skills, and other runtime-managed resources.

## Changes

- Add a small `gatewayUrl()` helper in `K8sSpawner`.
- Preserve explicit `SICLAW_GATEWAY_INTERNAL_URL` as the highest-priority override.
- Otherwise derive the URL from the already configured `SICLAW_GATEWAY_HOSTNAME` and `SICLAW_INTERNAL_PORT`.
- Keep the old `siclaw-runtime.<namespace>` URL as a final fallback.
- Add tests for hostname-derived URLs and explicit override precedence.

## Validation

```bash
npm test -- src/gateway/agentbox/k8s-spawner.test.ts
npm run build
```

Both passed locally.

### DIFF
--- a/src/gateway/agentbox/k8s-spawner.test.ts
+++ b/src/gateway/agentbox/k8s-spawner.test.ts
@@ -79,6 +79,12 @@ const deletePodImpl = { set fn(f: (a: any) => Promise<any>) { g.__k8sImpls.delet
 const createSecretImpl = { set fn(f: (a: any) => Promise<any>) { g.__k8sImpls.createNamespacedSecret = f; }, get fn() { return g.__k8sImpls.createNamespacedSecret; } };
 const listPodImpl = { set fn(f: (a: any) => Promise<any>) { g.__k8sImpls.listNamespacedPod = f; }, get fn() { return g.__k8sImpls.listNamespacedPod; } };
 
+const originalGatewayEnv = {
+  SICLAW_GATEWAY_INTERNAL_URL: process.env.SICLAW_GATEWAY_INTERNAL_URL,
+  SICLAW_GATEWAY_HOSTNAME: process.env.SICLAW_GATEWAY_HOSTNAME,
+  SICLAW_INTERNAL_PORT: process.env.SICLAW_INTERNAL_PORT,
+};
+
 // Import SUT after mocks.
 import { K8sSpawner } from "./k8s-spawner.js";
 
@@ -100,6 +106,13 @@ function resetCalls() {
 
 beforeEach(() => {
   resetCalls();
+  for (const key of Object.keys(originalGatewayEnv) as Array<keyof typeof originalGatewayEnv>) {
+    if (originalGatewayEnv[key] === undefined) {
+      delete process.env[key];
+    } else {
+      process.env[key] = originalGatewayEnv[key];
+    }
+  }
   // Reset impls to defaults
   g.__k8sImpls.readNamespacedPod = async () => { throw Object.assign(new Error("not found"), { code: 404 }); };
   g.__k8sImpls.createNamespacedPod = async () => ({});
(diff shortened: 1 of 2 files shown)


jacoblee-io (approved):
## Review

LGTM — the bug is real (`SICLAW_GATEWAY_HOSTNAME` from the Helm template embeds `{{ .Release.Name }}-runtime`, so any non-`siclaw` release name produces a Service the hardcoded URL can't resolve), the priority order in `gatewayUrl()` is correct, and the mTLS contract is preserved because both the connect-URL here and the gateway server cert's SAN in `src/gateway/server.ts:130` derive from the same `SICLAW_GATEWAY_HOSTNAME` env var.

No blocking issues. Suggestions below.

### 🟡 Worth considering before merge

**1. The fallback branch has no test — `src/gateway/agentbox/k8s-spawner.test.ts`**
The new tests pin two of the three branches in `gatewayUrl()` (HOSTNAME branch, INTERNAL_URL override). The PR description explicitly says "Keep the old `siclaw-runtime.<namespace>` URL as a final fallback" but nothing pins this — `siclaw-runtime.` only appears inside the two new tests, never as the asserted value. A future refactor could silently drop the `else` and break non-Helm K8s installs (static `k8s/gateway-deployment.yaml` path) without any test failing.

Suggested addition (same shape as the two new tests, without setting any env vars):

```ts
it("falls back to siclaw-runtime.<namespace> when no gateway env vars are set", async () => {
  // (the beforeEach already unsets all three SICLAW_GATEWAY_* envs)
  const cm = new FakeCertManager();
  const s = new K8sSpawner({ namespace: "siclaw" });
  s.setCertManager(cm as any);
  // ...same readPodImpl + await s.spawn(...)
  expect(env).toContainEqual({
    name: "SICLAW_GATEWAY_URL",
    value: "https://siclaw-runtime.siclaw.svc.cluster.local:3002",
  });
});
```

**2. Env restoration is asymmetric — `k8s-spawner.test.ts:82-115`**
Reset happens in `beforeEach` only; there's no `afterAll` (or `afterEach`) mirror. After the last test in this suite runs, `SICLAW_GATEWAY_HOSTNAME` / `SICLAW_INTERNAL_PORT` / `SICLAW_GATEWAY_INTERNAL_URL` stay mutated and can leak into whatever test file the vitest worker picks up next. The canonical pattern in this repo is in `src/agentbox/http-server.test.ts:185-205` (snapshot + restore in `afterEach`). Adding an `afterAll` that runs the same restore loop would align with that pattern and future-proof the suite.

### 🟢 Nice to have

- **HOSTNAME is not validated** (`k8s-spawner.ts:72-82`). If the env value already contains a port (`siclaw-runtime.ns.svc:3002`) the helper produces `https://siclaw-runtime.ns.svc:3002:3002`; an empty/whitespace value yields `https://:3002`. The Helm template never emits these so it's unlikely in practice, but a one-line `host?.trim()` + `!host.includes(":")` guard would harden the contract.
- **mTLS footgun on `SICLAW_GATEWAY_INTERNAL_URL`** — the override branch does NOT update the gateway server cert's SAN (the SAN comes from `SICLAW_GATEWAY_HOSTNAME` in `server.ts:130`). If an operator points the override at a hostname not in the cert's SAN list, the TLS handshake will fail. Worth a one-line comment above `gatewayUrl()` recording this constraint so the next reader doesn't trip over it.
- **Port-parsing style mismatch** — `k8s-spawner.ts:78` reads `SICLAW_INTERNAL_PORT` as a raw string, while `src/cli-local.ts:19` and `src/gateway/config.ts:21` use `parseInt(..., 10)`. A trailing-space or non-numeric value is accepted here but the gateway binds the parsed integer, so URL and listener can diverge. Low likelihood, but aligning the three sites would prevent a future support headache.

### ✅ Verified clean

- Priority order is correct: explicit override → derived from HOSTNAME → hardcoded fallback.
- mTLS contract preserved on the common path: both AgentBox's connect-URL and the gateway server cert SAN derive from the same `SICLAW_GATEWAY_HOSTNAME`.
- Legacy `siclaw` release path still works: the second branch fires (HOSTNAME is always set by Helm) and produces the same URL as the old hardcode.
- Static `k8s/gateway-deployment.yaml` installs are compatible (HOSTNAME is hardcoded to `siclaw-runtime.siclaw.svc.cluster.local` there, hits the second branch).
- `toContainEqual({ name, value })` pins both the env-var name and exact value strictly — would catch value drift or array-order regressions.

chent1996 (approved):
Clean, minimal fix for the AgentBox bootstrap URL bug. Confirmed the new priority order (`SICLAW_GATEWAY_INTERNAL_URL` → `SICLAW_GATEWAY_HOSTNAME[:SICLAW_INTERNAL_PORT]` → legacy `siclaw-runtime.<ns>` fallback) lines up with what the Helm chart already injects:

- `helm/siclaw/templates/gateway-deployment.yaml:50-51` sets `SICLAW_GATEWAY_HOSTNAME` to `{{ include "siclaw.fullname" . }}-runtime.<ns>.svc.cluster.local`, so non-default release names (e.g. `siclaw-debug`) now resolve correctly.
- `helm/siclaw/templates/gateway-service.yaml:15-18` exposes Service port `3002` mapped to container `https-internal:3002`, matching the runtime's `SICLAW_INTERNAL_PORT=3002` env, so the derived URL is correct end-to-end.
- The legacy `k8s/gateway-deployment.yaml:53-54` also sets `SICLAW_GATEWAY_HOSTNAME`, so the bare-fallback branch should never trigger in the supported deployment paths.

Test setup is correct: env snapshot is captured at module load and restored in `beforeEach`, so test ordering can't leak `SICLAW_GATEWAY_HOSTNAME` into the existing tests in this file (which rely on the legacy fallback).

This also closes the loose end I had tracked from the MCP-NFS refactor (`SICLAW_GATEWAY_INTERNAL_URL` not being wired up in the gateway deployment — now obviated by the new `SICLAW_GATEWAY_HOSTNAME` branch).

Non-blocking observations (skip or follow up later, none should hold this PR):

1. The new branch does naive string concat: `https://${hostname}:${port}`. If anyone ever sets `SICLAW_GATEWAY_HOSTNAME` with a scheme prefix or trailing `:port`, the result becomes malformed. Today the Helm template emits a bare DNS name so this is fine — but a one-line guard (`hostname.replace(/^https?:\/\//, '').replace(/:\d+$/, '')`) would make this immune to user-supplied env values without changing the happy path.

2. The legacy `siclaw-runtime.${namespace}.svc.cluster.local:3002` fallback still carries the original bug for non-default release names. Given both Helm and the legacy raw manifests now set `SICLAW_GATEWAY_HOSTNAME`, this fallback is effectively dead code. Worth considering removing it (or replacing with a fail-fast `throw`) in a follow-up so a future regression where the env var goes missing surfaces immediately instead of silently mis-resolving DNS.

3. Test coverage gap: no test exercises the no-env-set fallback path. Adding one would lock in the legacy behavior if you keep the fallback, or make it easy to assert a throw if you remove it later. Optional.

PR #274, opened 2026-05-19
Add Mermaid diagram rendering controls

LikiosSedo:
## Summary
- Add native Mermaid rendering for flowchart/graph, sequenceDiagram, and timeline fenced blocks in chat.
- Reuse shared SVG export helpers so charts and Mermaid diagrams share PNG copy/download behavior without changing the chart MCP contract.
- Keep streaming Mermaid blocks in a stable loading state, then render once complete, and include Mermaid images in rich message/session copy.
- Document the Mermaid rendering boundary and update prompt guidance for when agents should use Mermaid.

## Validation
- `npm test -- src/components/chat/Markdown.test.tsx src/components/chat/ChartRenderer.test.ts src/components/chat/chart-utils.test.ts`
- `npx tsc --noEmit`
- `npm run build`
- `git diff --check`
- Deployed test image to `sdliu-siclaw` and verified the Mermaid toolbar/rendered SVG via the local portal page.

### DIFF
--- a/portal-web/src/components/chat/Markdown.test.tsx
+++ b/portal-web/src/components/chat/Markdown.test.tsx
@@ -51,6 +51,17 @@ describe("Markdown Mermaid fences", () => {
     })
   })
 
+  it("repairs leaked stream content prefixes inside Mermaid blocks", () => {
+    expect(validateMermaidSource("flowchart TD\n  178-content: A --> B")).toMatchObject({
+      ok: true,
+      source: "flowchart TD\nA --> B",
+    })
+    expect(validateMermaidSource("sequenceDiagram\n  9-text: A->>B: hello")).toMatchObject({
+      ok: true,
+      source: "sequenceDiagram\nA->>B: hello",
+    })
+  })
+
   it("rejects unsupported diagram families and init directives", () => {
     expect(validateMermaidSource("classDiagram\n  A <|-- B")).toMatchObject({
       ok: false,
--- a/portal-web/src/components/chat/Markdown.test.tsx
+++ b/portal-web/src/components/chat/Markdown.test.tsx
@@ -32,6 +32,15 @@ describe("Markdown Mermaid fences", () => {
     expect(html).not.toContain("language-mermaid")
   })
 
+  it("keeps incomplete mermaid fences in loading state while streaming", () => {
+    const html = renderToStaticMarkup(
+      <Markdown isStreaming>{"```mermaid\nflowchart TD\n  A -->"}</Markdown>,
+    )
+
+    expect(html).toContain("Rendering diagram")
+    expect(html).not.toContain("Failed to render Mermaid diagram")
+  })
+
   it("accepts the initial diagram set", () => {
     expect(validateMermaidSource("flowchart TD\n  A --> B")).toMatchObject({
       ok: true,
--- a/portal-web/src/components/chat/Markdown.test.tsx
+++ b/portal-web/src/components/chat/Markdown.test.tsx
@@ -1,6 +1,7 @@
 import { renderToStaticMarkup } from "react-dom/server"
 import { describe, expect, it } from "vitest"
 import { Markdown } from "./Markdown"
+import { countMermaidEdges, validateMermaidSource } from "./MermaidRenderer"
 
 const incompleteChart = '```chart\n{"type":"bar","data":\n```'
 
@@ -17,5 +18,49 @@ describe("Markdown chart fences", () => {
 
     expect(html).toContain("Chart output incomplete")
     expect(html).not.toContain("Generating chart")
+  })
+})
+
+describe("Markdown Mermaid fences", () => {
+  it("routes mermaid fences to the diagram renderer shell", () => {
+    const html = renderToStaticMarkup(
+      <Markdown>{"```mermaid\nflowchart TD\n  A --> B\n```"}</Markdown>,
+    )
+
+    expect(html).toContain("mermaid-host")
+    expect(html).toContain("Rendering diagram")
+    expect(html).not.toContain("language-mermaid")
+  })
+
+  it("accepts the initial diagram set", () => {
+    expect(validateMermaidSource("flowchart TD\n  A --> B")).toMatchObject({
+      ok: true,
+      kind: "flowchart",
+    })
+    expect(validateMermaidSource("graph LR\n  A --> B")).toMatchObject({
+      ok: true,
+      kind: "flowchart",
+    })
+    expect(validateMermaidSource("sequenceDiagram\n  A->>B: hello")).toMatchObject({
+      ok: true,
+      kind: "sequence",
+    })
+    expect(validateMermaidSource("timeline\n  title Task lifecycle\n  Created")).toMatchObject({
+      ok: true,
+      kind: "timeline",
+    })
+  })
+
+  it("rejects unsupported diagram families and init directives", () => {
+    expect(validateMermaidSource("classDiagram\n  A <|-- B")).toMatchObject({
+      ok: false,
+    })
+    expect(validateMermaidSource("%%{init: {'securityLevel': 'loose'}}%%\nflowchart TD\nA-->B")).toMatchObject({
+      ok: false,
+    })
+  })
+
+  it("counts Mermaid edges for complexity limits", () => {
+    expect(countMermaidEdges("flowchart TD\n  A --> B\n  B --- C\n  C ==> D")).toBe(3)
   })
 })
--- a/portal-web/src/components/chat/PilotArea.tsx
+++ b/portal-web/src/components/chat/PilotArea.tsx
@@ -1195,12 +1195,13 @@ function escapeHtml(text: string): string {
 // whose source failed validation.
 //
 // KNOWN LIMITATION — this PNG↔fence mapping is best-effort, not exact. It
-// assumes `tryParseChartSpec` succeeding here implies the fence rendered a real
-// `.chart-host svg` in the DOM. That can drift if a fence parsed but the chart
-// did not actually mount (e.g. it was still in the streaming/loading state, or
-// future render-gating changes), in which case later images attach to the
-// wrong message. Acceptable for a copy-to-clipboard convenience; if it ever
-// needs to be exact, walk per-message DOM nodes instead of re-parsing text.
+// assumes `tryParseChartSpec` / `validateMermaidSource` succeeding here implies
+// the fence rendered a real `.chart-host svg` / `.mermaid-host svg` in the DOM.
+// That can drift if a fence parsed but the visual did not actually mount (e.g.
+// it was still in the streaming/loading state, or future render-gating changes),
+// in which case later images attach to the wrong message. Acceptable for a
+// copy-to-clipboard convenience; if it ever needs to be exact, walk per-message
+// DOM nodes instead of re-parsing text.
 function serializeSessionToHtml(messages: PilotMessage[], visualUrls: string[]): string {
   const fenceRe = /```(chart|mermaid)\s*([\s\S]*?)```/g
   let visualIdx = 0
@@ -1225,6 +1226,8 @@ function serializeSessionToHtml(messages: PilotMessage[], visualUrls: string[]):
         const parsed = lang === "chart" ? Boolean(tryParseChartSpec(raw)) : validateMermaidSource(raw).ok
         if (parsed && visualIdx < visualUrls.length) {
           html += `<p><img src="${visualUrls[visualIdx++]}" alt="${lang}" style="max-width:100%;height:auto"/></p>`
+        } else if (lang === "mermaid") {
+          html += "<p>[diagram]</p>"
         } else {
           html += `<p>[${lang}]</p>`
         }
--- a/portal-web/src/components/chat/PilotArea.tsx
+++ b/portal-web/src/components/chat/PilotArea.tsx
@@ -21,7 +21,9 @@ import {
 } from "lucide-react"
 import { cn } from "./cn"
 import { Markdown } from "./Markdown"
-import { svgChartToPngDataUrl, tryParseChartSpec } from "./ChartRenderer"
+import { tryParseChartSpec } from "./ChartRenderer"
+import { validateMermaidSource } from "./MermaidRenderer"
+import { svgToPngDataUrl } from "./svg-export"
 import { useCopyFeedback } from "./clipboard"
 import { InputArea } from "./InputArea"
 import { SkillCard } from "./SkillCard"
@@ -1166,7 +1168,7 @@ function serializeSessionToText(messages: PilotMessage[]): string {
     if (m.role === "user") {
       lines.push(`You:\n${m.content.trim()}`)
     } else if (m.role === "assistant") {
-      const body = stripChartFences(m.content ?? "")
+      const body = stripVisualizationFences(m.content ?? "")
       if (body) lines.push(`Assistant:\n${body}`)
     } else if (m.role === "tool") {
       const name = m.toolName ?? "tool"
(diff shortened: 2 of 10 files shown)


jacoblee-io (approved):
## Review

Functional design is sound — Mermaid client-side rendering with `securityLevel: "strict"`, three-layer validation (size/edge caps, `%%{...}` directive ban, family whitelist), and `dangerouslySetInnerHTML` gated by all of the above forms a reasonable XSS trust chain. The dynamic `await import("mermaid")` keeps the cold-start cost off the hot path. Approving on the strength of the design, but flagging items worth addressing before / right after merge.

### 🔴 Must-address before merge

**1. `src/core/prompt.ts` is gated by CLAUDE.md — "REQUIRES HUMAN APPROVAL — describe intent and wait for OK before editing".** The 8 added lines under "Visual Output" (`prompt.ts:114-121`) are content-correct (the supported families and the `123-content:` stream-prefix repair both match `validateMermaidSource` / `normalizeMermaidSource` in the renderer), but the gate's process step (explicit maintainer ack on the PR thread) is missing. Treat this approval as the recorded ack for those 8 lines specifically — but consider splitting future prompt edits into their own PR so the gate is exercised cleanly.

**2. Three load-bearing rationale comments dropped in the `svg-export.ts` extraction.** The pre-PR `ChartRenderer.tsx` carried multi-line WHY comments above `svgToPngBlob`, `inlineComputedStyles`, and `svgChartToPngDataUrl`. They explained:
- Why `.chart-bg` live-fill sampling exists (avoid the light-grey-on-white failure mode when `inlineComputedStyles` misses the bg fill). **This warning is exactly what Mermaid SVGs will now hit silently** — no `.chart-bg` → fallback to `#ffffff`.
- Why `removeAttribute("class")` is intentional (colors are now inlined; classes would bloat output and require Tailwind in consumer context).
- Why `svgChartToPngDataUrl` exists (rich-clipboard text/html payload, otherwise the chart bubble would copy as raw JSON).

Restore them in `svg-export.ts` so a future reader doesn't strip "dead code" or get bitten by the white-bg fallback.

### 🟡 Important

**3. `INLINEABLE_PROPS` allowlist is chart-shaped, undersized for Mermaid.** `svg-export.ts:4` lists only `[fill, stroke, stroke-width, font-family, font-size, font-weight]`. Mermaid SVGs rely on `stroke-dasharray` (dashed edges, dotted self-loops), `marker-end` / `marker-start` (arrowheads), and `opacity`. Combined with `tgt.removeAttribute("class")`, the serialized clone loses those declarations. On-screen rendering still uses the in-document stylesheet — but PNG copy/download of any sequenceDiagram or styled flowchart will be **missing arrowheads and dashed lines**. Recommend adding at minimum `stroke-dasharray`, `stroke-linecap`, `stroke-linejoin`, `marker-end`, `marker-start`, `marker-mid`, `opacity`.

**4. `svgBackground()` falls back to white-only for Mermaid.** `svg-export.ts:30-34` queries `.chart-bg`; Mermaid SVGs have none. Mermaid is configured with `theme: "base"` and a white background (`mermaidConfig()` at `MermaidRenderer.tsx:101`), so the PNG content is readable, but dark-theme users get a light-themed PNG that visibly mismatches their UI. Acceptable trade-off if intentional — but should be documented next to `svgBackground()`, or expose a `bgSelector` parameter so callers declare their own background source.

**5. Test coverage gap on PilotArea — the PR's marketed feature is untested.**
- `stripVisualizationFences()` (the `[chart]` vs `[diagram]` placeholder logic) has no test.
- `serializeSessionToHtml()`'s `(chart|mermaid)` regex + per-fence PNG mapping has no test.
- The dual-selector `'.chart-host svg[role="img"], .mermaid-host svg[role="img"]'` is load-bearing for rich copy; not pinned.
- No streaming-state pin for Mermaid (matching the existing chart streaming test would be ~10 lines).

A regression that emits raw Mermaid source into the copied text, or attaches the wrong PNG to the wrong fence, would not be caught.

**6. Error-handling inconsistencies that mislead users.**
- `MermaidRenderer.tsx:242, 255` (`onCopyImage`, `onDownload`): bare `catch { flash("err", "Copy failed") }` — no `console.warn(err)`. Same repo's `PilotArea.tsx:1271, 1377` already use `console.warn("[copy]...", err)` — match that pattern.
- `svg-export.ts:101` (`copyBlobToClipboard`): returns `false` for both "feature unsupported" and "permission denied". Caller surfaces *"Clipboard image copy not supported — downloaded PNG instead"*, which lies when the cause is actually permission denial. Either bubble the error or return a discriminated reason.
- `MermaidRenderer.tsx:285`: parser-reject message is `"Mermaid parser rejected the diagram syntax."` with no line info because of `suppressErrors: true`. Drop the flag (or call `mermaid.parse` without it inside the try) so the `ParseError.hash.line` reaches the user.

**7. `MermaidRenderer.tsx` is 428 lines with effectively zero WHY comments.** At minimum add one-liners on:
- `mermaidConfig()` — `securityLevel: "strict"` is the load-bearing XSS defence given the `dangerouslySetInnerHTML` mount below; this is the single most important security claim in the file.
- `validateMermaidSource()` — summarize the layered strategy (size/line/edge caps → directive ban → whitelist) and call out that `%%{` is rejected so a response cannot weaken the renderer's security configuration.
- `decorateMermaidSvg()` — justify regex-over-DOM (SVG already passed `mermaid.render` in strict mode + validation; this is decoration of a trusted string).
- The `cancelled` token — name the race (streaming completes → async dynamic import resolves after a newer effect already replaced state).

### 🟢 Nice to have

- `docs/design/chart-rendering.md:206-208` says unsupported Mermaid families "fall through to `<pre>`" — they actually render an inline `MermaidError` shell with collapsible source. Tighten the wording.
- Placeholder string inconsistency: `stripVisualizationFences` emits `[diagram]`, but `serializeSessionToHtml` uses `[${lang}]` → `[mermaid]`. Pick one.
- `KNOWN LIMITATION` comment at `PilotArea.tsx:1195-1206` still references `tryParseChartSpec` only — generalize to also mention `validateMermaidSource`.
- `decorateMermaidSvg`, `safeDownloadName`, and the `validateMermaidSource` over-limit branches are all good candidates for cheap unit tests.
- `ChartRenderer.test.ts` / `chart-utils.test.ts` are byte-identical to `main` after the helper extraction. The moved code is only covered transitively through ChartRenderer rendering paths — consider a small `svg-export.test.ts` to pin the extracted helpers directly.

### ✅ Verified clean

- `decorateMermaidSvg` regex is safe under the current trust chain: only rewrites the top-level `<svg>` opening tag, and the attribute pattern `\srole=(["']).*?\1` is anchored to whitespace so it won't match `role=` inside an attribute value.
- Stream-prefix repair regex `/^\s*\d+-(?:content|text):\s*/gm` is anchored to line start, so a legitimate node label like `A["42-content: foo"]` is not stripped.
- Dynamic import (`await import("mermaid")`) is correctly placed inside the render effect — Mermaid only loads when a fence renders. Bundle impact (~1106 lockfile lines) is acceptable for the feature value.
- Streaming `cancelled` token is checked at both `setState` sites (lines 287, 289). Cleanup is synchronous; cancelled async work wastes CPU but cannot corrupt state.
- Prompt content matches code exactly: claimed supported families = `validateMermaidSource` whitelist; `123-content:` claim = `normalizeMermaidSource` regex.
- `PilotArea.tsx:1342-1348` doc comment was correctly updated from "raw chart JSON spec" to "raw chart JSON spec or Mermaid source".

src/core/prompt.ts
```diff
@@ -111,6 +111,14 @@ const DEFAULT_TEMPLATE = `You are Siclaw, a personal SRE AI assistant. You help
...
 
+## Visual Output
+
+- You may use Mermaid diagrams as a native response format when the user asks to draw/diagram a flow, sequence, lifecycle, timeline, topology, or dependency chain, or when a compact diagram clearly makes an SRE explanation easier to verify.
+- Supported Mermaid forms are \`flowchart\` / \`graph\`, \`sequenceDiagram\`, and \`timeline\`. Keep diagrams small and readable; prefer roughly 5-12 nodes/events and avoid decorative detail.
+- Use \`flowchart\` for cause/effect, decision, dependency, or remediation flows; \`sequenceDiagram\` for request paths and cross-component call order; \`timeline\` for incidents, task lifecycles, and investigation progress.
+- Inside Mermaid fences, output only Mermaid syntax. Do not add line numbers, event labels, or stream prefixes such as \`123-content:\`.
+- Do not force a diagram into simple answers. If exact times or relationships are unknown, label them as unknown/approx instead of inventing precision.
+
```
jacoblee-io, line 121: **🔴 CLAUDE.md gate** — this file is marked "REQUIRES HUMAN APPROVAL — describe intent and wait for OK before editing". The 8 added lines are content-correct (claims match `validateMermaidSource` and `normalizeMermaidSource` exactly), but the gate's process step was skipped. Approving the PR records the ack for these 8 lines specifically; please split future prompt edits into their own PR so the gate is exercised cleanly.

portal-web/src/components/chat/svg-export.ts
```diff
@@ -0,0 +1,109 @@
+// Browser helpers shared by chat visualisations that render as inline SVG.
+// They intentionally know nothing about ChartSpec or Mermaid syntax.
+
+const INLINEABLE_PROPS = ["fill", "stroke", "stroke-width", "font-family", "font-size", "font-weight"] as const
```
jacoblee-io: **🟡 Undersized for Mermaid.** Mermaid SVGs rely on `stroke-dasharray` (dashed edges, dotted self-loops), `marker-end` / `marker-start` (arrowheads), and `opacity`. Combined with the `removeAttribute("class")` below, the PNG copy/download path will silently drop arrowheads and dashed lines from sequenceDiagram and styled flowchart exports. On-screen rendering is unaffected (CSS still applies in-document).

Recommend extending to at least:
```ts
const INLINEABLE_PROPS = [
  "fill", "stroke", "stroke-width",
  "stroke-dasharray", "stroke-linecap", "stroke-linejoin",
  "marker-end", "marker-start", "marker-mid",
  "opacity",
  "font-family", "font-size", "font-weight",
] as const
```

portal-web/src/components/chat/svg-export.ts
```diff
@@ -0,0 +1,109 @@
...
+    width: rect.width || svg.clientWidth || 900,
+    height: rect.height || svg.clientHeight || 520,
+  }
+}
+
+function svgBackground(svg: SVGSVGElement): string {
+  const bgRect = svg.querySelector<SVGRectElement>(".chart-bg")
+  const liveBg = bgRect ? window.getComputedStyle(bgRect).fill : ""
+  return liveBg && liveBg !== "none" && liveBg !== "transparent" ? liveBg : "#ffffff"
+}
```
jacoblee-io, line 59: **🔴 Rationale comment lost in the move.** The pre-PR ChartRenderer had a 7-line block above this lookup explaining why `.chart-bg` live-fill sampling exists:

> *"Sample chart-bg's live fill BEFORE cloning — this is the same color the user sees on-screen (light: white, dark: slate-900). Used as the canvas base so if inlineComputedStyles ever fails to carry the bg fill onto the detached clone (CSS var that doesn't resolve out-of-document, future refactor swapping `<rect>` for CSS background, etc.) we still rasterise on the right base color — light-grey text on white was the worst-case failure mode the previous hardcoded #ffffff fallback would produce."*

This warning is **exactly** the failure mode Mermaid SVGs will now hit silently — no `.chart-bg` element → fallback to `#ffffff` on every Mermaid PNG export. Dark-theme users will get a light-themed PNG that mismatches their UI. Either restore the comment + document the Mermaid trade-off, or accept a `bgSelector` parameter so each caller declares its own background source.

portal-web/src/components/chat/svg-export.ts
```diff
@@ -0,0 +1,109 @@
...
+  for (let i = 0; i < srcAll.length; i++) {
+    const cs = window.getComputedStyle(srcAll[i])
+    const tgt = dstAll[i]
+    for (const prop of INLINEABLE_PROPS) {
+      const v = cs.getPropertyValue(prop)
+      if (v) tgt.style.setProperty(prop, v)
+    }
+    tgt.removeAttribute("class")
+  }
+}
```
jacoblee-io, line 38: **🔴 Rationale comment lost.** The pre-PR ChartRenderer had:

> *"Strip class attrs — colors are now inlined, classes would just bloat the serialised output and require Tailwind in the consumer's context."*

Without this, `tgt.removeAttribute("class")` reads as dead code and a future reader may delete it. Please restore.

portal-web/src/components/chat/svg-export.ts
jacoblee-io: ✅ Resolved by `ba193dd` — `INLINEABLE_PROPS` extended to include `stroke-dasharray`, `stroke-linecap`, `stroke-linejoin`, `marker-end`, `marker-start`, `marker-mid`, `opacity`. Mermaid arrows / dashed edges will now survive PNG export. Thanks for the quick turnaround.

portal-web/src/components/chat/svg-export.ts
jacoblee-io: ✅ Resolved by `ba193dd` — rationale restored above `svgBackground`, and explicitly notes that Mermaid SVGs render on an intentional white base from `mermaidConfig()`. Closes the loop nicely.

portal-web/src/components/chat/svg-export.ts
jacoblee-io: ✅ Resolved by `ba193dd` — class-strip rationale restored inside `inlineComputedStyles`.

src/core/prompt.ts
jacoblee-io: ℹ️ No code change needed here — flagged this as a process note. The 8 added lines are content-correct, and this approval records the ack. Future prompt edits should go through their own PR so the gate is exercised independently.

portal-web/src/components/chat/Markdown.test.tsx
```diff
@@ -19,3 +20,58 @@ describe("Markdown chart fences", () => {
...
+describe("Markdown Mermaid fences", () => {
+  it("routes mermaid fences to the diagram renderer shell", () => {
+    const html = renderToStaticMarkup(
+      <Markdown>{"```mermaid\nflowchart TD\n  A --> B\n```"}</Markdown>,
+    )
+
+    expect(html).toContain("mermaid-host")
+    expect(html).toContain("Rendering diagram")
+    expect(html).not.toContain("language-mermaid")
+  })
```
jacoblee-io, line 33: 🟡 **Remaining gap from round 1.** This `describe` block does not pin the "Keep streaming Mermaid blocks in a stable loading state" claim that the PR description markets. The chart equivalent is right above (lines 9-21). A 5-line mirror would close it:

```ts
it("keeps incomplete mermaid fences in loading state while streaming", () => {
  const html = renderToStaticMarkup(
    <Markdown isStreaming>{"```mermaid\nflowchart TD\n  A -->"}</Markdown>
  )
  expect(html).toContain("Rendering diagram")
})
```

Optional — fine as a follow-up if PilotArea tests are also going into a separate PR.

portal-web/src/components/chat/PilotArea.tsx
```diff
@@ -1219,11 +1221,15 @@ function serializeSessionToHtml(messages: PilotMessage[], chartUrls: string[]):
...
         const before = body.slice(last, match.index).trim()
         if (before) html += `<p>${escapeHtml(before).replace(/\n/g, "<br/>")}</p>`
-        const parsed = tryParseChartSpec(match[1].trim())
-        if (parsed && chartIdx < chartUrls.length) {
-          html += `<p><img src="${chartUrls[chartIdx++]}" alt="chart" style="max-width:100%;height:auto"/></p>`
+        const lang = match[1]
+        const raw = match[2].trim()
+        const parsed = lang === "chart" ? Boolean(tryParseChartSpec(raw)) : validateMermaidSource(raw).ok
+        if (parsed && visualIdx < visualUrls.length) {
+          html += `<p><img src="${visualUrls[visualIdx++]}" alt="${lang}" style="max-width:100%;height:auto"/></p>`
```
jacoblee-io, line 1228: 🟡 **Remaining gap from round 1.** `stripVisualizationFences`, `serializeSessionToHtml` (this loop), and `copyBubbleWithCharts` were all modified to handle the mermaid path but `PilotArea.test.ts` still does not exist. The dual-selector `.chart-host, .mermaid-host` and the `(chart|mermaid)` regex + per-fence PNG mapping are load-bearing for the PR's rich-copy feature.

Fine as a follow-up PR (new test file is more work than fits this PR's scope). Suggested coverage:
- `stripVisualizationFences` emits `[diagram]` for `mermaid` and `[chart]` for `chart`
- `serializeSessionToHtml` maps two interleaved chart+mermaid fences to PNGs in DOM order
- `.chart-host svg, .mermaid-host svg` querySelector contract is preserved

PR #275, opened 2026-05-20
Fix empty streamed tool call arguments

LikiosSedo:
## Summary

- Preserve parsed tool-call arguments from streamed deltas when the provider leaves final tool call arguments empty.
- Avoid clearing partial/message arguments to `{}` after a failed repair attempt, while still dropping unsafe cached repair candidates.
- Add regression coverage for both the stream wrapper and output guard paths.

## Root Cause

Some OpenAI-compatible streaming providers can emit valid JSON tool-call argument text in `toolcall_delta`, but leave the final `toolCall.arguments` object empty. The previous malformed-argument repair logic only cached arguments through the malformed JSON repair path, and it could also clear message arguments when a repair attempt did not apply. That made valid streamed arguments disappear before tool execution, producing empty tool calls such as `bash {}`.

## Validation

- `npx vitest run src/core/stream-wrappers.test.ts`
- Previously run on this branch before deployment:
  - `npm run build`
  - `npm test`
  - DeepSeek V4 Pro live smoke against the local Siclaw provider config
- Deployed test images to `sdliu-siclaw` for manual validation:
  - `registry-cn-shanghai.siflow.cn/k8s/siclaw-portal:toolcall-fix-bd942f8-20260520133305`
  - `registry-cn-shanghai.siflow.cn/k8s/siclaw-runtime:toolcall-fix-bd942f8-20260520133305`
  - `registry-cn-shanghai.siflow.cn/k8s/siclaw-agentbox:toolcall-fix-bd942f8-20260520133305`

## Notes

The runtime Dockerfile on current `main` is `Dockerfile.runtime`, not the older `Dockerfile.gateway`.

### DIFF
--- a/src/core/stream-wrappers.ts
+++ b/src/core/stream-wrappers.ts
@@ -214,6 +214,29 @@ type ToolCallArgumentRepair = {
   trailingSuffix: string;
 };
 
+function parseToolCallArgumentObject(raw: string): Record<string, unknown> | undefined {
+  if (!raw.trim()) return undefined;
+  try {
+    const parsed = JSON.parse(raw) as unknown;
+    return parsed && typeof parsed === "object" && !Array.isArray(parsed)
+      ? parsed as Record<string, unknown>
+      : undefined;
+  } catch {
+    return undefined;
+  }
+}
+
+function hasUnsafeTrailingToolCallArgumentSuffix(raw: string): boolean {
+  const jsonPrefix = extractBalancedJsonPrefix(raw);
+  if (!jsonPrefix) return false;
+  const suffix = raw.slice(raw.indexOf(jsonPrefix) + jsonPrefix.length).trim();
+  if (!suffix) return false;
+  return (
+    suffix.length > MAX_TOOLCALL_REPAIR_TRAILING_CHARS ||
+    !TOOLCALL_REPAIR_ALLOWED_TRAILING_RE.test(suffix)
+  );
+}
+
 /**
  * Try to parse and repair malformed tool call arguments.
  * Returns undefined if the JSON is already valid or cannot be repaired.
@@ -245,6 +268,12 @@ export function tryParseMalformedToolCallArguments(raw: string): ToolCallArgumen
   }
 }
 
+function shouldReplaceToolCallArguments(value: unknown): boolean {
+  if (value == null) return true;
+  if (typeof value !== "object" || Array.isArray(value)) return false;
+  return Object.keys(value).length === 0;
+}
+
 function repairToolCallArgumentsInMessage(
   message: unknown,
   contentIndex: number,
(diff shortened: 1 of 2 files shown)


src/core/stream-wrappers.ts
```diff
@@ -245,6 +268,12 @@ export function tryParseMalformedToolCallArguments(raw: string): ToolCallArgumen
   }
 }
 
+function shouldReplaceToolCallArguments(value: unknown): boolean {
+  if (value == null) return true;
+  if (typeof value !== "object" || Array.isArray(value)) return false;
+  return Object.keys(value).length === 0;
```
chent1996, line 274: `shouldReplaceToolCallArguments` only treats `null`/`undefined`/`{}` as replaceable. If any OpenAI-compatible provider ever represents empty args as an empty **string** (some send `arguments` as a JSON-encoded string rather than an object), `typeof "" !== "object"` returns `false` here and the empty call would slip through unrepaired. Today this layer normalizes args to objects (events use `arguments: {}`), so this is likely unreachable — but if string-form args are possible upstream, consider also treating `value === ""` as replaceable. Non-blocking.

src/core/stream-wrappers.ts
```diff
@@ -333,22 +362,28 @@ function wrapStreamRepairMalformedToolCallArguments(stream: any): any {
               return result;
             }
             partialJsonByIndex.set(event.contentIndex, nextPartialJson);
+            if (hasUnsafeTrailingToolCallArgumentSuffix(nextPartialJson)) {
```
chent1996, line 365: Minor / FYI: when this deletes the cached repair mid-stream, any args already written into `event.message` from an earlier valid parse are **not** cleared (the old `clearToolCallArgumentsInMessage` is gone). I traced this and it's benign under append-only streaming — the retained value is the valid prefix and `shouldReplaceToolCallArguments` prevents it from being clobbered later — so cache and message can momentarily disagree by design. Just flagging so it's intentional, not accidental. No change needed.

src/core/stream-wrappers.ts
```diff
@@ -359,9 +394,7 @@ function wrapStreamRepairMalformedToolCallArguments(stream: any): any {
           ) {
             const repairedArgs = repairedArgsByIndex.get(event.contentIndex);
             if (repairedArgs) {
-              if (event.toolCall && typeof event.toolCall === "object") {
-                (event.toolCall as { arguments?: unknown }).arguments = repairedArgs;
-              }
+              repairToolCallArguments(event.toolCall, repairedArgs);
```
chent1996, line 397: This exact end-handler logic is duplicated in `createRepairMalformedArgsGuard` (~line 530), and **only the guard is registered** in `guard-pipeline.ts:84`. This `wrapStreamRepairMalformedToolCallArguments` path (and `wrapStreamFnRepairMalformedToolCallArguments`) is dead production code, kept alive only by its tests. Updating both in lockstep is correct, but the duplication is a drift hazard — see the summary. Suggest deleting the dead wrappers or extracting a shared helper in a fast follow.

chent1996 (approved):
**Correct fix for a real stability bug.** I traced both the bug and the new logic and it holds up.

**Root cause confirmed.** Two independent defects combined to produce `bash {}`:
1. The repair-failed `else` branch called `clearToolCallArgumentsInMessage(...)` on `event.partial`/`event.message`, which are the **live** streamed message objects. If a provider had already written valid `arguments` directly into that object, the unconditional clear wiped them to `{}`. Removing the clear is the right fix.
2. `tryParseMalformedToolCallArguments` returns `undefined` when the accumulated JSON is *already valid* (it only handles malformed input), so valid streamed args were never cached and could not be restored when the final `toolCall.arguments` came back empty. The new `parseToolCallArgumentObject` fast-path closes that gap.

**No regressions found.** I worked through append-only streaming scenarios (nested objects mid-stream, trailing whitespace, trailing junk > tolerance, corrected-then-valid). Because streaming is append-only, the buffer first becomes a complete valid object only at the natural close, so `toolCall` / `partial` / `message` cannot realistically diverge. The new `shouldReplaceToolCallArguments` gate (only fill empty/null, never clobber provider-populated args) is the correct conservative contract, and `hasUnsafeTrailingToolCallArgumentSuffix` preserves the prior drop-on-junk safety using the same constants/RE as the existing malformed path. Consistent.

**Main thing to address (see inline on line 397): duplicated logic, half of it dead code.** `wrapStreamRepairMalformedToolCallArguments` / `wrapStreamFnRepairMalformedToolCallArguments` (and the Trim pair) are not wired into anything — only the `createRepairMalformedArgsGuard` OutputGuard is registered in `guard-pipeline.ts:84`. This PR had to apply the identical change twice, once to dead code. That is a real drift hazard: a future fix can touch one copy, miss the other, and the dead path's tests still pass while production silently diverges. Recommend either deleting the unused `wrapStreamFn*` wrappers + their tests, or extracting the shared delta/end processing into one helper both call. Pre-existing, so not blocking — but this PR touches exactly that duplicated code, so it is a natural moment to collapse it.

Two minor notes inline (non-blocking).

Note: I could not run the suite in this review checkout (no `node_modules`; repo pins vitest ^3 while the new `vitest.config.ts` needs the pinned install). Reviewed by logic; relying on the author's reported `npm test` + live DeepSeek V4 smoke.

PR #277, opened 2026-05-20
fix(test): eliminate flaky kubectl timeout in restricted-bash test

a-yeyang:
The "allows kubectl pipelines" test cases called tool.execute(), which spawns a real subprocess. Blocked-command tests return before exec() and are fast; 
allowed-command tests reach exec() and run kubectl get pods -n default, which hangs on CI (no cluster reachable) until vitest's 5 s timeout fires.

Fix: replace tool.execute() in the allows suite with synchronous preExecSecurity() calls — the test intent is "security validator does not block this command", 
not "kubectl succeeds".


chent1996 (approved):
LGTM

PR #282, opened 2026-05-21
Fix/streaming cache stuck

a-yeyang:
fix(streaming): keep Chat mounted so SSE stream survives navigation

Navigating away from /chat unmounted the component, destroying the usePilotChat hook and its SSE stream. On return, the Thinking spinner was gone.

Fix: render <Chat /> permanently in Layout with CSS hidden/block toggle. The stream and all hook state now survive sidebar navigation intact.

### DIFF
--- a/portal-web/src/hooks/usePilotChat.ts
+++ b/portal-web/src/hooks/usePilotChat.ts
@@ -1160,6 +1160,16 @@ export function usePilotChat({ agentId, sessionId }: UsePilotChatOptions): UsePi
           }
         } finally {
           if (abortControllerRef.current === controller) abortControllerRef.current = null
+          // Unconditionally mark this session as no longer streaming in the cache.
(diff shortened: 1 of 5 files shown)


LikiosSedo:
One concern before merging: keeping `<Chat />` always mounted in `Layout` currently mounts it even when the user never visits `/chat`. Because `Chat` loads `/agents`, auto-selects the first agent, and then mounts `AgentChat`, non-chat pages can now trigger hidden chat/session/message requests, recovered-stream polling, and possibly chat-related error toasts in unrelated screens.

The keep-alive direction makes sense for preserving an active SSE stream across sidebar navigation, but I think it should be lazy: mount Chat persistently only after the user has entered `/chat` once, or only while there is an active chat/session to preserve. That keeps the stream-survives-navigation fix without making hidden chat state part of every page load.

LikiosSedo (approved):

PR #288, opened 2026-05-25
Add portal attachment OCR parsing

LikiosSedo:
## Summary
- Add image/PDF attachment upload and preview support in the Portal chat input and user bubbles.
- Preprocess supported attachments in the Portal chat gateway through an independent PaddleOCR backend service, then append extracted OCR evidence to the agent prompt.
- Add a Helm-managed OCR Deployment/Service with mobile PP-OCRv5 defaults, Portal env wiring, and design/backend docs.

## Notes
- This PR is based directly on `origin/main` and intentionally does not include PR286 / `feat/codex-style-steer` changes.
- This is not an MCP flow; raw image/PDF bytes stop at Portal -> OCR backend, and Runtime/AgentBox receive text evidence only.
- OCR backend accepts only `file_base64` sources to avoid URL/path read behavior.

## Validation
- `npm test -- src/portal/chat-gateway.test.ts`
- `npm run build -- --pretty false`
- `npm run build` in `portal-web/`
- `python3 -m py_compile services/ocr-backend/server.py services/ocr-backend/parse_once.py`
- `helm lint helm/siclaw`
- `helm template siclaw helm/siclaw --namespace sdliu-siclaw`
- `helm template siclaw helm/siclaw --namespace sdliu-siclaw -f helm/siclaw/values-standalone.yaml`
- `git diff --check`

### DIFF
--- a/helm/siclaw/values.yaml
+++ b/helm/siclaw/values.yaml
@@ -23,6 +23,7 @@ database:
 # first, so multiple replicas of the same release would continuously fight.
 # Scale horizontally by deploying multiple releases with distinct identities.
 runtime:
+  enabled: true
   portalSecret: ""      # Runtime → Portal/Upstream API auth (X-Auth-Token on phone-home WS)
   # Where Runtime gets Agent/Credential info from:
   #   Standalone mode: http://siclaw-portal:3003
@@ -76,6 +77,68 @@ portal:
       cpu: "500m"
       memory: "512Mi"
   env: {}
+
+# -- OCR backend for Portal image/PDF attachment preprocessing
+ocr:
+  enabled: true
+  replicas: 1
+  # Set when OCR is provided by another service. If enabled=false and this is
+  # empty, Portal treats OCR as unavailable and keeps chat sends recoverable.
+  externalUrl: ""
+  timeoutMs: 120000
+  image:
+    # Empty repository uses image.registry + image.tag as siclaw-ocr:<tag>.
+    repository: ""
+    tag: ""
+    pullPolicy: ""
+  service:
+    port: 8088
+  device: "cpu"
+  preload: true
+  enableStructure: false
+  # Per-Pod admitted in-flight OCR requests. PaddleOCR inference is serialized
+  # inside each backend process; requests above this limit receive HTTP 503.
+  maxConcurrency: 1
+  # Per-attachment OCR text characters that Portal may inject into the prompt.
+  maxEvidenceTextChars: 32768
+  # Total OCR evidence characters that Portal may inject for one chat turn.
+  maxTotalEvidenceTextChars: 65536
+  # Hard process-level watchdog for a single OCR request. If PaddleOCR hangs past
+  # this window, the backend exits and Kubernetes restarts the container.
+  hardTimeoutMs: 150000
+  # PDF inputs are capped before PaddleOCR runs so a long paper does not occupy
+  # OCR workers and delay screenshot/table requests. Set 0 to disable.
+  maxPdfPages: 10
+  # Decoded image pixel cap before OCR runs. Set 0 to disable.
+  maxImagePixels: 50000000
+  textDetectionModelName: "PP-OCRv5_mobile_det"
+  textRecognitionModelName: "PP-OCRv5_mobile_rec"
+  textDetLimitSideLen: ""
+  maxRequestBytes: "10485760"
+  cache:
+    # Cache PaddleOCR/PaddleX model downloads across container restarts in the
+    # same Pod. This does not store user uploads or OCR text.
+    enabled: true
+  resources:
+    requests:
+      cpu: "2"
+      memory: "4Gi"
+    limits:
+      cpu: "4"
+      memory: "8Gi"
+  env: {}
+  networkPolicy:
+    enabled: false
+    # Extra Kubernetes NetworkPolicy from entries for OCR-only/shared-service
+    # deployments. Example:
+    # ingressFrom:
+    #   - namespaceSelector:
+    #       matchLabels:
+    #         kubernetes.io/metadata.name: sicore
+    #     podSelector:
+    #       matchLabels:
+    #         app.kubernetes.io/name: sicore-apiserver
+    ingressFrom: []
 
 # -- AgentBox (spawned at runtime by K8s spawner)
 agentbox:
--- a/services/ocr-backend/server.py
+++ b/services/ocr-backend/server.py
@@ -1,0 +1,711 @@
+#!/usr/bin/env python3
+"""Siclaw attachment OCR backend.
+
+This service intentionally uses only the Python standard library for HTTP so the
+production image stays small and easy to replace. Portal sends pasted image/PDF
+attachments here and forwards the extracted text evidence to the Siclaw agent.
+"""
+
+from __future__ import annotations
+
+import base64
+import binascii
+import json
+import os
+import tempfile
+import threading
+import time
+import traceback
+import uuid
+import warnings
+from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer
+from pathlib import Path
+from typing import Any
+from urllib.parse import urlparse
+
+
+OCR_PIPELINE: Any | None = None
+STRUCTURE_PIPELINE: Any | None = None
+OCR_LOCK = threading.Lock()
+STRUCTURE_LOCK = threading.Lock()
+REQUEST_SEMAPHORE: threading.BoundedSemaphore | None = None
+REQUEST_SEMAPHORE_LOCK = threading.Lock()
+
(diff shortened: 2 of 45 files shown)


src/portal/chat-gateway.ts
```diff
@@ -23,6 +24,199 @@ import {
...
+  if (normalizedAttachments.length === 0) return text;
+
+  const sections: string[] = [];
+  for (const attachment of normalizedAttachments) {
+    const mimeType = normalizeMimeType(attachment.mimeType ?? attachment.mime_type, attachment.filename);
+    const kindHint = ocrKindHint(text, attachment, mimeType);
+    const started = Date.now();
+    const requestId = crypto.randomUUID();
+    try {
+      const res = await fetch(ocrBackendUrl(), {
```
chent1996: **OCR backend has no auth.** This POST goes to `http://siclaw-<release>-ocr-backend:8088/parse` (ClusterIP Service, no TLS, no shared secret, no JWT). Any pod in the namespace — including an agentbox compromised through a malicious skill or a misbehaving sidecar — can fan out OCR requests against this CPU/memory-heavy backend (PaddleOCR + 4–8 GiB memory limit, concurrency=1). Even with 503+Retry-After, an attacker can DoS the OCR service for legitimate Portal users.

Aligns with the cluster's internal-trust posture today, but the design doc explicitly invites reuse from other gateways ("Sicore can call a shared OCR service…"), so the moment that happens the lack of auth becomes more material. Suggest a shared-secret header similar to `cliSnapshotSecret` in `cli-snapshot-api.ts` — cheap to add now, awkward to retrofit later. Non-blocking.

src/portal/chat-gateway.ts
```diff
@@ -23,6 +24,199 @@ import {
...
+      sections.push([
+        `### ${attachment.filename}`,
+        `kind: ${String(evidence.kind ?? "unknown")}`,
+        `route: ${String(evidence.route ?? "unknown")}`,
+        `language: ${String(evidence.language ?? "unknown")}`,
+        `confidence: ${String(evidence.confidence ?? "unknown")}`,
+        `elapsed_ms: ${elapsedMs}`,
+        warnings.length ? `warnings: ${warnings.join("; ")}` : "",
+        "",
+        String(evidence.text ?? "").trim() || "(no OCR text extracted)",
```
chent1996: **No bound on OCR text size injected into the prompt.** A scanned multi-page PDF can produce tens of KB of recognized text, multiplied by up to 4 attachments per turn. This goes straight into the user prompt sent to the model — counts against the context window, billed tokens, and the existing context-budget guard (`tool-result-context-guard.ts`) which targets tool results, not prompt text.

Worth either: (a) capping `evidence.text` length per attachment here (e.g., truncate at N KiB with a `truncated` warning), or (b) routing OCR text through the same context budget the rest of the pipeline uses. Non-blocking but easy to forget once shipped.

helm/siclaw/values.yaml
```diff
@@ -77,6 +77,46 @@ portal:
...
+    repository: ""
+    tag: ""
+    pullPolicy: ""
+  service:
+    port: 8088
+  device: "cpu"
+  preload: true
+  enableStructure: false
+  # Per-Pod OCR parse concurrency. Requests above this limit receive HTTP 503.
+  maxConcurrency: 1
```
chent1996, line 101: `maxConcurrency` in `values.yaml` reads as "this many parallel OCR decodes per Pod," but `OCR_LOCK` / `STRUCTURE_LOCK` in `services/ocr-backend/server.py:406` globally serialize every `ocr.predict`/`pipeline.predict` call (PaddleOCR is not thread-safe, so the lock is correct). With the text route, `maxConcurrency > 1` only lets more requests queue on the lock — no actual parallelism, just more memory pressure and longer tail latencies.

Either tighten the doc to say "max in-flight requests admitted before returning 503," or drop the knob until a non-serialized engine is wired in. Minor.

services/ocr-backend/server.py
```diff
@@ -0,0 +1,594 @@
...
+    source_type = source.get("type")
+    if source_type != "file_base64":
+        raise ValueError(f"unsupported source.type: {source_type!r}")
+
+    data = source.get("data")
+    filename = str(source.get("filename") or "upload.bin")
+    if not isinstance(data, str) or not data:
+        raise ValueError("source.data is required")
+    safe_name = Path(filename).name or "upload.bin"
+    path = tmp / safe_name
```
chent1996: Defense-in-depth: `Path(filename).name` strips POSIX path separators, so on Linux a malicious `filename="../../etc/passwd"` resolves to just `passwd` — safe today. But `filename=".."` gives `safe_name=".."`, so `path = tmp / ".."` would write to tmp's parent if it weren't for `write_bytes` failing on a directory. The current safety is by accident, not by design.

Add an explicit guard, e.g.:
```python
resolved = (tmp / safe_name).resolve()
if not resolved.is_relative_to(tmp.resolve()):
    raise ValueError("unsafe filename")
```
Locks in the intent against future refactors. Minor / non-blocking.

chent1996 (approved):
**Substantial, well-scoped feature.** End-to-end image/PDF attachment support via OCR preprocessing in Portal, with a clean architectural split: raw bytes never reach Runtime/AgentBox — only extracted text crosses that boundary. Design doc, Helm wiring, frontend size mirrors, and graceful failure-tolerance (OCR error doesn't kill the chat send) are all in good shape.

Defenses I checked:
- Size limits stacked at three layers (browser 6MiB → Portal 8MiB base64/attachment → OCR backend `PADDLEOCR_MAX_REQUEST_BYTES` 10MiB). Math checks out: one 6MiB raw image fits comfortably under each cap.
- `parseChatRequestBody` correctly returns 413 on oversize and `requireAuth` gates body parsing, so unauth requests can't trigger large allocations or OCR fan-out.
- OCR call happens AFTER auth + agent-connected + model-bound checks in `/chat/send` — wasted work is bounded. (See an asymmetry note on `/chat/steer` below.)
- Hard-timeout watchdog (`RequestWatchdog` → `os._exit(124)`) is the right answer for pathological PaddleOCR hangs given Python's no-clean-thread-kill story.
- 503+`Retry-After:1` when the concurrency semaphore is full — Portal falls back to an "OCR failed" note so chat stays recoverable. Good UX choice.
- Frontend `stripAttachmentOcrEvidence` keeps the user bubble showing the original prompt, while the LLM sees the augmented text. Symmetric and tested.

Four inline observations below (none blocking, but #1 and #2 are worth a follow-up):

1. OCR backend is unauthenticated — any pod in the namespace can POST attachments. Consistent with the cluster's internal-trust model but worth a shared-secret header.
2. OCR text appended to the prompt is unbounded — no per-attachment text cap or per-turn token budget. A scanned PDF can emit tens of KB of text × 4 attachments.
3. `OCR_LOCK` serializes `ocr.predict`, so `maxConcurrency > 1` doesn't actually parallelize the text route. Either document this or rename for clarity.
4. `materialize_source` path-traversal hardening is correct-by-accident on Linux; explicit `is_relative_to(tmp)` check would lock in the intent.

Two minor nits (no inline, just noting): `normalizeChatAttachments` runs twice on the happy `/chat/send` path (line 329 then again inside `appendOcrEvidence`), and `/chat/steer` runs OCR before any `isConnected` check — `/chat/send` checks first. Aligning the order avoids wasted CPU/latency when runtime is offline.

Note: I could not run `npm test`/`pytest`/`helm template` locally in this review checkout (no `node_modules`, no Python deps installed). Reviewed by reading. Trusting the author's reported validation matrix in the PR description.

jacoblee-io (approved):
Approved. The updated revision addresses every blocking and important item from the prior review:

- Removed Chinese string literals from production code; OCR `kind_hint` no longer classifies by prompt keywords (mime/extension only).
- OCR calls now run in parallel; per-attachment errors, malformed responses, and timeouts are logged with request_id/elapsed_ms and surfaced as recoverable evidence notes.
- Added decoded-image pixel cap (SICLAW_OCR_MAX_IMAGE_PIXELS) to prevent decompression-bomb OOM; bad base64 now returns 400.
- Hard-timeout watchdog constraint documented and all profiles pinned to maxConcurrency=1.
- Added OCR NetworkPolicy template (Portal-only ingress); best-effort structure fallbacks now log and re-raise MemoryError.
- Frontend attachment/steer errors routed through toast and a shared failure handler instead of being injected into the prompt.

Two non-blocking notes for a follow-up:
1. ocr.networkPolicy.enabled defaults to false, so in the default standalone profile the unauthenticated OCR backend is still reachable by any in-cluster pod. Consider defaulting it true when portal.enabled, or calling it out in NOTES.txt.
2. ATTACHMENT_ONLY_PROMPT is duplicated in attachment-prompt.ts and chat-gateway.ts and must stay byte-identical for steer-echo matching; nothing enforces that across the two codebases.

Please confirm CI is green on the squashed commit before merging.

PR #289, opened 2026-05-28
fix(memory): disable Siclaw memory by default

LikiosSedo:
## Summary
- add `SICLAW_MEMORY_ENABLED` as a global runtime/agentbox memory kill switch
- disable memory tools, memory prompt injection, PROFILE updates, and session memory auto-save when the flag is off
- default the Helm runtime deployment to `SICLAW_MEMORY_ENABLED=false` and pass the flag through to spawned AgentBox pods
- make `agent.clearMemory` remove the complete agent `memory/` directory, including `.memory.db`, `PROFILE.md`, topics, investigations, and historical markdown files

## Verification
- `npm run build`
- `npm test -- src/core/prompt.test.ts src/tools/query/memory-search.test.ts src/tools/query/memory-get.test.ts src/gateway/memory-cleanup.test.ts src/agentbox/session.test.ts src/gateway/agentbox/k8s-spawner.test.ts src/agentbox/http-server.test.ts`
- `helm template siclaw ./helm/siclaw >/tmp/siclaw-memory-mr-helm.yaml && rg -n "SICLAW_MEMORY_ENABLED|value: \"false\"" /tmp/siclaw-memory-mr-helm.yaml`
- `npm test`

### DIFF
--- a/src/core/prompt.ts
+++ b/src/core/prompt.ts
@@ -27,12 +27,15 @@ export function buildSreSystemPrompt(mode?: "cli" | "web" | "channel" | "task", 
   const modeLabel = MODE_LABELS[mode ?? "cli"] ?? "Web UI";
   const settingsPath = mode === "cli" ? "`/setup`" : "sidebar **Settings**";
   const credentialsPath = mode === "cli" ? "`/setup` → Credentials" : "**Settings → Credentials**";
+  const memoryEnabled = isMemoryEnabled();
 
   // Variable substitution
   let prompt = template
     .replace(/\{\{mode\}\}/g, modeLabel)
     .replace(/\{\{settingsPath\}\}/g, settingsPath)
-    .replace(/\{\{credentialsPath\}\}/g, credentialsPath);
+    .replace(/\{\{credentialsPath\}\}/g, credentialsPath)
+    .replace(/\{\{memoryIntro\}\}/g, memoryEnabled ? MEMORY_INTRO : "")
+    .replace(/\{\{memorySection\}\}/g, memoryEnabled ? MEMORY_SECTION : "");
 
   // Mode-conditional blocks: strip the non-matching mode block
   const keepMode = mode === "web" ? "web" : "cli";
@@ -49,12 +52,6 @@ export function buildSreSystemPrompt(mode?: "cli" | "web" | "channel" | "task", 
 
   // Append hardcoded safety section — NOT overridable by agent templates
   prompt += SAFETY_SECTION(credentialsPath);
-
-  if (!isMemoryEnabled()) {
-    prompt = prompt
-      .replace(" You remember context from previous sessions and grow more helpful over time.", "")
-      .replace(/\n### Memory — Search On Demand\n\nUse `memory_search`[\s\S]*?(?=\n## Environment & Configuration)/, "");
-  }
 
   return prompt;
 }
@@ -95,7 +92,15 @@ Respond in the user's language. \`[System: respond in X]\` overrides to language
 // ---------------------------------------------------------------------------
 // Bundled default template — overridable via agent settings
 // ---------------------------------------------------------------------------
-const DEFAULT_TEMPLATE = `You are Siclaw, a personal SRE AI assistant. You help your user manage and troubleshoot their infrastructure — Kubernetes clusters, cloud resources, and DevOps workflows. You are competent, direct, and warm. You remember context from previous sessions and grow more helpful over time.
+const MEMORY_INTRO = " You remember context from previous sessions and grow more helpful over time.";
+
+const MEMORY_SECTION = `
+
+### Memory — Search On Demand
+
+Use \`memory_search\` **on demand** when symptoms suggest a previously-seen issue — search for past investigations, what was tried, what the root cause was. Use \`memory_get\` to pull details when a match looks relevant. Don't search reflexively — search purposefully.`;
+
+const DEFAULT_TEMPLATE = `You are Siclaw, a personal SRE AI assistant. You help your user manage and troubleshoot their infrastructure — Kubernetes clusters, cloud resources, and DevOps workflows. You are competent, direct, and warm.{{memoryIntro}}
 
 ## Core Behavior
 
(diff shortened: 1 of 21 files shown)


src/gateway/memory-cleanup.ts
```diff
@@ -0,0 +1,35 @@
...
+export function clearAgentMemory(agentId: string, userDataBase = "/app/.siclaw/user-data"): { memoryDir: string; deletedFiles: number } {
+  const base = path.resolve(userDataBase);
+  const memoryDir = path.resolve(base, "agents", sanitizeAgentId(agentId), "memory");
+
+  if (!memoryDir.startsWith(base + path.sep)) {
+    throw new Error(`Refusing to clear memory outside user data base: ${memoryDir}`);
+  }
+
+  const deletedFiles = countFiles(memoryDir);
+  fs.rmSync(memoryDir, { recursive: true, force: true });
```
jacoblee-io, line 33: **Coordination gap on a live pod.** This now removes the entire `memory/` directory (including `.memory.db` and the dir itself), whereas the previous implementation deliberately preserved `.memory.db` and `PROFILE.md` and deleted only content `.md` files.

The clear-memory endpoint (`chat-gateway.ts` -> `server.ts` `agent.clearMemory`) sends the RPC without restarting the pod, then notifies the live AgentBox via `resetMemory()`. But the enabled-path `resetMemory()` in `session.ts` only calls `sync()` + `clearInvestigations()` on the existing shared indexer; it never closes/recreates the indexer, and `createMemoryIndexer()` does not mkdir. So on a running pod the shared `_sharedMemoryIndexer` is left pointing at a deleted directory and DB file: `sync()` scans a path that no longer exists and `clearInvestigations()` writes to an unlinked inode.

The old code preserved `.memory.db` precisely so the live indexer stayed valid and `sync()` could reconcile the deleted content files. Now only one side of that contract changed.

Suggestion: on the enabled path, after deletion fully tear down and recreate the indexer (close -> recreate dir + skeleton `PROFILE.md` -> new indexer), or restart the pod on clear, or keep the directory and reset DB contents instead of `rm -rf`. Could you confirm the intended behavior when clear-memory runs against a live pod?

src/gateway/memory-cleanup.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export function clearAgentMemory(agentId: string, userDataBase = "/app/.siclaw/user-data"): { memoryDir: string; deletedFiles: number } {
+  const base = path.resolve(userDataBase);
+  const memoryDir = path.resolve(base, "agents", sanitizeAgentId(agentId), "memory");
+
+  if (!memoryDir.startsWith(base + path.sep)) {
+    throw new Error(`Refusing to clear memory outside user data base: ${memoryDir}`);
+  }
+
+  const deletedFiles = countFiles(memoryDir);
```
jacoblee-io, line 32: Nit / semantics change: `countFiles` counts all files recursively (including `.memory.db`, `PROFILE.md`, and SQLite `-wal`/`-shm`), whereas the previous code counted only `.md` files. If the Portal surfaces this number as "deleted N files", the reported value's meaning changes. Cosmetic, just flagging.

src/core/prompt.ts
```diff
@@ -48,6 +50,12 @@ export function buildSreSystemPrompt(mode?: "cli" | "web" | "channel" | "task",
   // Append hardcoded safety section — NOT overridable by agent templates
   prompt += SAFETY_SECTION(credentialsPath);
 
+  if (!isMemoryEnabled()) {
+    prompt = prompt
+      .replace(" You remember context from previous sessions and grow more helpful over time.", "")
+      .replace(/\n### Memory — Search On Demand\n\nUse `memory_search`[\s\S]*?(?=\n## Environment & Configuration)/, "");
```
jacoblee-io: This post-hoc string surgery is fragile: it depends on three literals staying exactly in sync - the full sentence above, the `### Memory - Search On Demand` heading, and the lookahead requiring the very next section to be `## Environment & Configuration`. If the wording or section order changes, this `replace` silently no-ops (memory instructions leak into the prompt even when disabled), or over-deletes if a section is inserted in between. `prompt.test.ts` guards the current text, which helps, but consider assembling the prompt conditionally (or asserting the replace matched) rather than removing text afterward.

Process note: `CLAUDE.md` marks `src/core/prompt.ts` as requiring human approval to modify - worth calling out explicitly in the PR description.

src/gateway/agentbox/k8s-spawner.ts
```diff
@@ -172,6 +172,9 @@ export class K8sSpawner implements BoxSpawner {
       { name: "SICLAW_GATEWAY_URL", value: this.gatewayUrl(namespace) },
       { name: "SICLAW_AGENT_ID", value: agentId },
     ];
+    if (process.env.SICLAW_MEMORY_ENABLED !== undefined) {
```
jacoblee-io, line 175: Nit: this only propagates the flag when the gateway itself has `SICLAW_MEMORY_ENABLED` set. The Helm template always injects it, so production is fine, but a gateway started without the variable (manual/dev) will spawn pods that fall back to the enabled default. Worth a doc note so the default is not surprising.

jacoblee-io (commented):
Reviewed against head 5d9a3ed. Overall the kill-switch is wired through consistently (agent-factory, session, cli-main, prompt, tool availability, Helm) with solid defense-in-depth: file tools are blocked from touching the memory dir when disabled, and `buildAppendSystemPrompt(memoryDir: string | null)` handles the null case correctly (no `null/PROFILE.md` leak).

One higher-priority item to confirm before merge: the `agent.clearMemory` change deletes the entire memory dir including `.memory.db`, but the live-pod `resetMemory()` only re-syncs the existing indexer rather than recreating it (see inline comment). Plus a fragility note on the prompt regex and a couple of nits.

jacoblee-io (approved):
Approving — both of the main findings from my earlier review are addressed:

- **Live-pod indexer coordination (a5dbd0b2):** `resetMemory()` now closes the stale indexer, nulls it, and rebuilds via the new shared `createSharedMemoryIndexer()` helper (mkdir → create → sync → startWatching). This correctly handles the Gateway deleting the whole `memory/` dir incl. `.memory.db`, and the contract is now consistent on both sides. Good test coverage on close/rebuild/sync/watch.
- **Prompt assembly (b2b1f9fb):** the fragile post-hoc regex is replaced with `{{memoryIntro}}` / `{{memorySection}}` template placeholders substituted conditionally — no longer dependent on heading text or section ordering. Tests assert the placeholders never leak and `## Environment & Configuration` is always present.

Non-blocking follow-up (your call): the old `resetMemory` kept the same indexer instance alive specifically because active sessions hold references to it via tool closures. The rebuild now swaps the instance, so a session that was already active before a mid-session clear-memory could still hold the closed indexer in its `memory_search`/`memory_get` closures and error if invoked afterward. Narrow window, currently untested — worth confirming whether active-session references should be refreshed or whether this edge is acceptable.

LGTM.

PR #290, opened 2026-05-29
Sub-agents + Tasks-v2 plan ledger (background gated, fan-out concurrency capped)

jacoblee-io:
## Summary

Brings the agent core in line with Claude Code's latest sub-agent + Tasks-v2 design, on top of the `pi-coding-agent` 0.55→0.73 upgrade (native parallel tool execution). One cohesive feature across backend, Portal UI, Helm, and docs.

- **Sub-agents (`spawn_subagent`)** — delegate one bounded task to an isolated, one-level-deep child agent; the model fans out by emitting N calls in a turn (pi runs them concurrently). Replaces the legacy `delegate_to_agent(s)` batch tools and their dead runtime.
- **Tasks-v2 plan ledger** (`task_create/update/list/get`) — incremental, id-based, `blockedBy` dependency graph, completion discipline. Durable across turns (stable `taskListId` = sessionId + PV snapshot/rehydrate + close cleanup) and auto-clears on completion (CC V2 `resetTaskList` parity).
- **System prompt** aligned with CC's style (plan proactively for complex work; fan out across independent targets; no recursion).
- **Portal Web UI** — floating plan panel, live sub-agent execution cards, chat hygiene (task events hidden from the transcript).

## Background jobs: built but gated OFF

The Job runtime (`startBackgroundSubagent`, `subagentJobs`, `job_stop`, Jobs bar, session-release gating, child-session persistence) is complete and kept intact, but **`RUN_IN_BACKGROUND_ENABLED = false`** until a completion notification back to the parent model exists (today the child summary is dropped). The flag gates three model-facing surfaces at once (the `run_in_background` param, the `job_stop` tool, and the prompt guidance); flip it once the notification lands. See design §7.

## Fan-out concurrency cap

pi runs a tool-call batch via unbounded `Promise.all`, and each `spawn_subagent` is a full in-process child agent (own LLM stream + kubectl/bash) sharing one AgentBox pod — so a wide fan-out would start N child agents + N LLM streams at once (provider 429s, pod pressure).

- A per-pod `ConcurrencyLimiter` (FIFO semaphore) wraps the foreground spawn executor; excess spawns **queue** rather than reject — the model still emits N calls and gets N reports, they just don't all run at once.
- Configurable via `SICLAW_SUBAGENT_CONCURRENCY` (default 5); the AgentBox doesn't inherit runtime env, so `k8s-spawner` forwards it onto the pod. Helm: `runtime.subagentConcurrency` (default 5).
- **Queued-state UI**: pi paints the whole batch as "running" at once, so queued children would falsely spin. The executor emits a `queued` progress (then `running` on slot acquisition); the Sub-agent card shows a Clock icon + "Queued" badge until it starts. Verified live (cap=2, 5-way fan-out → 2 running, 3 queued, strict serialized hand-off).

## Follow-up hardening (latest 8 commits)

Iterative cleanup after live testing on `siclaw-inner`, each a single-theme commit:

- **Unify k8s tools on a `cluster` parameter** — `bash`, `node_exec`, `pod_exec`, `*_script`, `resolve_pod_netns` now take a `cluster` name (from `cluster_list`) and inject the kubeconfig automatically; inline `--kubeconfig` is rejected. Removes the dead `ensureKubeconfigsForCommand` / `resolveKubeconfigByName` path.
- **Replace dead `knowledge_search` with wiki catalog injection** — the unused tool + its `knowledgeIndexer` wiring (session, tool-registry, `gateway-client.searchKnowledge`, local-spawner) is gone; the wiki's `index.md` catalog is injected into the system prompt instead, so the agent Reads only the pages it needs on demand.
- **Align system prompt with CC** — top-level sections promoted to CC's single-hash headings; Core Behavior / Safety / Communicating tightened toward CC's lean style (collaborator framing, trust-results-diagnose-failures, untrusted-tool-output + system-reminder + no-fabricated-links safety, narrate-as-you-work).
- **Disable memory by default** — `isMemoryEnabled()` defaults to `false` when the env is unset (local/TUI/tests), keeping `memory_search`/`memory_get` and all memory-facing prompt text out unless opted in. Helm already drives the env from `runtime.memory.enabled`, so prod behaviour is unchanged.
- **Fix cluster credential endpoints** — `credential.get` / `credential-request` now return `debug_image` + `is_production`/`api_server`/`description` metadata, so the debug pod uses the configured registry image instead of falling back to docker.io busybox.
- **Frontend: keep a streaming assistant turn in one bubble** — `text → tool → text` no longer splits into two bubbles (which previously broke markdown tables until refresh).
- **CC-v2 task tool descriptions** + **drop keyword→action mapping in manage-schedule** (no keyword/regex intent classification).

## Testing

- Backend `npm test` + `tsc --noEmit` clean; frontend `tsc --noEmit` clean. Follow-up commits keep the affected suites green (prompt / config / overview-generator / restricted-bash / kubeconfig / memory / gateway-client / local-spawner / session / task — 479 passing in the targeted run).
- `helm template` verified: default renders `SICLAW_SUBAGENT_CONCURRENCY=5`, override renders the set value, empty omits it.
- Deployed to the `siclaw-inner` test namespace and exercised end-to-end (plan, fan-out, concurrency cap queueing). Memory-off confirmed live (`Memory disabled by SICLAW_MEMORY_ENABLED` in the agentbox log; no `memory_search`/`memory_get` registered).

## Notes for reviewers

- `src/core/prompt.ts` is human-approval-gated. Beyond removing the dead "Background work" bullet, this PR also aligns it with CC's lean section style and single-hash headings (see "Align system prompt with CC" above) — changes were made with explicit approval.
- Design contracts/rationale live in `docs/design/2026-05-29-subagents-background-task-ledger.md` (§3 concurrency, §7 background status).


jacoblee-io:
## Review — Sub-agents + Tasks-v2 plan ledger

Reviewed the new core (`spawn-subagent.ts`, `task-tools.ts`, `job-stop.ts`, `subagent-registry.ts`, `task-ledger.ts`, `concurrency-limiter.ts`, `task-events.ts`) plus the `agent-factory.ts` / `tool-registry.ts` / `session.ts` changes, across four lenses (general review, error handling, type design, test coverage).

**Verdict:** solid, well-tested feature work, no crash-level bugs. Project rules all pass (no Chinese in code, ESM/named-exports, no keyword intent classification, security pipeline intact, limiter releases in `finally`, AbortSignal forwarded). Two real correctness issues and one documented-but-unenforced security surface below.

---

### 🔴 Critical / Important

**1. Shared-ledger auto-clear can wipe the parent's live plan mid-fan-out** — `src/agentbox/session.ts` (`scheduleLedgerAutoClear`) + `src/core/task-ledger.ts:102-112`

Parent and sub-agents share one ledger (`taskListId == parent sessionId`). When `allCompleted()` is true a 5s timer calls `ledger.clear()`, wiping all tasks. It is mostly contained (child sessions are created without `sessionEventEmitter`, so child `task_update`s don't drive the timer; a new pending task cancels it). Residual real risk: if the parent marks its last task `completed` while a sub-agent is still running, the timer fires, `allCompleted()` is still true (child completions are invisible to it), and the plan is cleared out from under the still-running child and the parent's next turn → silent loss of the live plan.
**Fix:** don't auto-clear while any child shares the `taskListId` and is in flight — gate the timer callback on the limiter's `activeCount` / `subagentJobs`.

**2. Sub-agent tool policy is declared but never enforced (security surface)** — `src/core/subagent-registry.ts:19-30`

`SubagentType.tools` / `disallowedTools` / `readOnly` and `SUBAGENT_ALWAYS_DENIED_TOOLS` have zero runtime consumers. `runSpawnedSubagent` reads only `type.systemPromptAddendum`. Recursion is correctly prevented — but structurally, by omitting `spawnSubagentExecutor` from the child (so the `available` guard hides `spawn_subagent`), not by the deny-list. Consequences:
- The comments ("the executor denies spawn_subagent … see SUBAGENT_ALWAYS_DENIED_TOOLS") describe a mechanism that doesn't exist — misleading.
- A future custom subagent type setting `tools: [...]` would silently get the full toolset. An unenforced safety field invites false trust.
- `subagent-registry.test.ts` asserts on the dead constant → false-confidence test that passes even if recursion protection breaks.

**Fix:** either wire the deny-list / tool-policy into the child's `allowedTools`, or remove the unused fields + constant and correct the comments to "recursion is prevented by withholding the executor." Then add a test asserting the child's actual tool list excludes `spawn_subagent`.

**3. Child completion status is weakly visible to the parent model** — `src/tools/workflow/spawn-subagent.ts:156-172` + `runSpawnedSubagent`

`status` is in `content` (good), but the model-facing text is `summary`. Two gaps:
- If the child's `prompt()` resolves with a plausible-but-half-done answer (model gave up, didn't throw), `status:"done"` + a confident partial `summary` → the parent treats unverified findings as authoritative. Mitigations only cover the throw/abort paths.
- `finalText` is captured only from `message_end`. The main SSE consumer also handles `turn_end` and a `task_report` toolResult terminal; the child subscriber handles neither — so a child that ends via `turn_end`/`task_report` is falsely reported as `failed: (sub-agent produced no output)`.

**Fix:** mirror the SSE consumer's terminal-detection set in the child subscriber; surface ledger-completion state (did the child's owned tasks actually complete?) alongside `status` instead of relying on `status:"done"` meaning "verified."

---

### 🟡 Suggestions (non-blocking)

- **`model` param is a silent no-op** — `spawn_subagent` accepts/advertises a per-call `model` override but `runSpawnedSubagent` ignores it. Drop from the schema or implement.
- **Unexpected child failures aren't logged** — the generic `catch` in `runSpawnedSubagent` sets a one-line `summary` but never logs the error/stack, and `finalText || …` can discard the real error message. Add a `console.warn`.
- **"Terminal event ALWAYS" isn't always** — `enqueuePersist` is gated on `persistTrace`, which flips false on the first persistence error, so the terminal status event (last in the queue) is the most likely to be skipped → UI shows a child stuck "running" forever. Route the terminal event off the `persistTrace` gate.
- **Ledger snapshot writes aren't serialized/atomic** — fire-and-forget `writeFile` to `.plan-ledger.json` on every task event; wide fan-out can interleave/truncate. Use temp-write + rename (or the existing transcript-persist queue).
- **`SpawnSubagentResult` & `TaskEvent` are flat optional-bags, not discriminated unions** — illegal states are representable (`status:"launched"` carries a `summary:"launched"` sentinel; `done` can carry `jobId`). The lone consumer already branches on `status`; converting to a union is low-cost and removes the sentinels.
- **Validation/test gaps** — `task_create` doesn't validate an empty subject (inconsistent with `spawn_subagent`); no test for self-blocking (`addBlockedBy:["1"]` on task 1) or `blockedBy` referential integrity on create.
- **Concurrency limiter has no dequeue-on-abort** — a queued child aborted while waiting still acquires a slot and spins up a throwaway child session (briefly shows "running"). No leak/hang, just wasted work + misleading UI. Consider short-circuiting on `signal.aborted` before `lim.run`, or making the limiter abort-aware.

---

### Test coverage

New leaf-unit tests are exemplary (limiter, ledger, task tools, registry — 39 tests pass). The gap is the orchestration layer: `runSpawnedSubagent` / `createSpawnSubagentExecutor` / `scheduleLedgerAutoClear` / `persistLedgerSnapshot` have no direct coverage. Highest-value missing tests: (1) child tool list excludes `spawn_subagent`; (2) child throw → parent `failed`, not fake success; (3) abort → child killed + partial surfaced; (4) shared-ledger auto-clear timing. The ~347 deleted lines in `session.test.ts` are not a regression — they tested the removed `delegate_to_agent(s)` tools.

---

### prompt.ts (human-approval-gated)

`src/core/prompt.ts` is marked **REQUIRES HUMAN APPROVAL** in CLAUDE.md; the PR body states changes were approved — surfacing, not blocking. Changes: full `DEFAULT_TEMPLATE` rewrite (`##`→`#` headings, merged the old "Understand Before Acting / skill hard-gate" into a leaner section, new "Multi-step Work & Sub-agents" section, expanded Safety), and memory text made conditional via `{{memoryIntro}}`/`{{memorySection}}` gated on `isMemoryEnabled()`. One item worth a human eye: the previously mandatory step-by-step skill-check gate was softened to "prefer a matching skill" — a deliberate relaxation of a hard gate.

---

**Recommended action:** address #1 (plan data-loss) and #2 (wire-or-retract the tool policy + fix the false-confidence test and misleading comments) before merge; treat #3 and the logging / terminal-event items as fast follow.

<sub>🤖 Generated with multi-agent review (Claude Code)</sub>

jacoblee-io:
### Follow-up: Deep Investigation (DP) mode impact

Checked whether this PR breaks DP mode. **It does not** — DP works correctly. Summary for the record:

**DP wiring is fully intact.** Activation (`[Deep Investigation]` prefix / `/dp` / Ctrl+I), exit (`[DP_EXIT]`), the `MutableDpStateRef` → `dpStateRef` → agentbox (`session.ts`) → `http-server.ts` → `client.getDpState()` → frontend (`usePilotChat`) state chain, `session_start` restore (incl. legacy shapes), context filter, and the Hypothesis Checkpoint barrier are all unchanged. Test suites pass: `deep-investigation.test.ts` (21) + `session.test.ts` (31) + `http-server.test.ts` (29) + `client.test.ts` (23) = **104 passed**.

**What the PR changed in DP (intentional, safe):** it removed only the dead delegation-synthesis logic that depended on the now-deleted `delegate_to_agent(s)` tools — `DELEGATION_BATCH_COMPLETE_MARKER`, `isDelegationBatchCompleteNotification`, `hasPendingDelegationSynthesis`, and the "delegation synthesis barrier". The DP activation prompt's parallel-checks guidance was rewritten from `delegate_to_agent(agent_id="self")` / `delegate_to_agents` to `spawn_subagent` fan-out (semantically equivalent), and the step list went 11 → 9.

**Caveat — DP's parallel-investigation quality now rides on the sub-agent path**, so two findings from the main review have an *indirect* DP impact:
- **#1 (shared-ledger auto-clear):** a DP investigation that builds a plan with `task_create` and fans out sub-agents can have its plan cleared mid-flight if the parent marks its last task `completed` while a child is still running.
- **#3 (child completion status weakly visible):** DP step 5 tells the model to synthesize sub-agent evidence back into its answer, but if a child ends via `turn_end`/`task_report` it can be misreported as `failed: (sub-agent produced no output)` and the parent loses that evidence — degrading the parallel investigation.

Single-agent DP (no fan-out) is unaffected either way. Fixing #1/#3 is what makes DP's parallel mode reliable.

LikiosSedo (commented):
Two follow-ups after `33fa8b5` (hiding the `task_*` tools from sub-agents is the right call — keeps the plan parent-owned and kills the "mutation with no SSE emitter → invisible UI" path cleanly). Both are minor / non-blocking.

**1. The design doc now contradicts the code — please update the contract.**

`33fa8b5` makes the plan parent-owned and removes a sub-agent's ability to touch the ledger, but `docs/design/2026-05-29-subagents-background-task-ledger.md` still states the opposite in two places:

- §5 (~L102): "The parent **and the sub-agents it spawns share the same ledger**. This is what lets a sub-agent own and complete a task."
- §8 (~L229): "As each child returns, mark its task `completed` (**the child** or the parent)…"

A sub-agent no longer has any `task_*` tool, so it can neither own nor complete a task. Suggest rewording to something like: *"the plan is parent-owned; sub-agents don't read or write the ledger — the parent marks tasks complete as children report back."*

(Same theme, tiny: the new comment at `session.ts` L559-561 says "the child sees the plan for context", but `task_list`/`task_get` are hidden from sub-agents too, so the child can't actually see the plan — and `taskListId` is still passed to the child session but no longer has a consumer there, i.e. it's now vestigial.)

**2. The auto-clear deferral guard is now stale — its premise no longer holds.**

In `src/agentbox/session.ts`, `scheduleLedgerAutoClear` (~L494-498):

```ts
// its task completions are invisible to allCompleted() (children have no
// sessionEventEmitter), so clearing now would wipe the parent's live plan
// out from under the in-flight child. Re-arm and retry once they finish.
if (this.subagentLimiter.activeCount > 0) { /* re-arm */ }
```

After `33fa8b5` a sub-agent can't mutate the ledger at all, so "a child completing/creating a task mid-clear" can no longer happen — the stated reason is gone. What's left is an unintended cross-session side effect: `subagentLimiter` is **pod-wide**, so session A's completed plan won't auto-clear while an *unrelated* session B still has a sub-agent running. Suggest either dropping the guard, or scoping the "active spawn" count per `taskListId` and refreshing the comment.

LikiosSedo (approved):

PR #293, opened 2026-06-01
feat: multi-level SSH jump host (ProxyJump) support

jacoblee-io:
## Summary

Adds standard OpenSSH ProxyJump support so `host_exec` / `host_script` can reach
hosts behind one or more bastions, and exposes the capability in the Portal
(host config + connection test). Bastions are modeled the ssh_config way — a
host's `jump_host_id` self-reference — keeping the inventory platform-neutral.
A target authenticates either directly (password / key) or in **managed** mode,
where its private key lives on the bastion ("management server") and is sourced
at dial time.

### Solution
- **Schema:** `hosts.jump_host_id` (self-ref, depth ≤3) + `passphrase`; `HostMeta.jump_host`.
- **Execution:** broker-free `ssh-dial` dials the chain via ssh2 `forwardOut`+`sock`
  (end-to-end encrypted, reverse teardown); `acquireSshTarget` recurses on the
  jump-host name (cycle-guarded). Single-hop behaviour is unchanged.
- **Managed target auth:** `auth_type="managed"` — the target carries no secret
  of its own; its private key is discovered on the preceding hop (the bastion /
  management server) at dial time. Requires a jump host; the target's own
  `username` still applies. Rationale in `docs/design/ssh-jump-host.md` / ADR-013.
- **Neutral boundaries:** `jump_host_id` → bastion NAME at the WS-RPC / REST /
  cli-snapshot boundaries. Binding a target transitively authorizes transit
  through its bastion chain (the agent never receives a bastion's secrets).
  Product-specific naming was genericized to "management server".
- **Portal:** jump-chain validation on write; a real `/test` that dials the chain
  and runs `echo ok`; a per-host **Test Connection** button on the host list; and
  connection testing from unsaved form data. On edit, blank credential fields keep
  the saved value, and the connection test falls back to the stored secret instead
  of failing on a blank password.

## Test Plan
- [x] Backend + portal-web type-check clean (`tsc`)
- [x] Unit coverage — `ssh-dial` multi-hop / teardown / failure; recursive
      `acquireSshTarget` (cycle / depth / passphrase); managed target auth;
      `validateJumpChain`; transitive authorization; cli-snapshot jump expansion
- [x] Portal host-api — connection test from form data; managed-requires-jump;
      **edit falls back to the stored secret**; a new form (no id) still requires
      a credential
- [x] Real cluster smoke — 3 cross-node sshd pods; `dialSshChain` ran
      outer→inner→target with the target IP unreachable directly, proving the chain
- [ ] Manual — bind a host with a jump in the Portal and run `host_exec` through
      it; test a managed target via its bastion


LikiosSedo (commented):
Reviewed the ProxyJump implementation end-to-end (ssh-dial / ssh-client / adapter / host-api / migrate) — solid work: consistent depth+cycle guards in all four places (target + 3 bastions), reverse teardown, broker-free execution layer, `requireAdmin` on every Portal route, managed-requires-jump enforced at write/boundary/acquire/dial. Backend `tsc` + full suite (3159) and portal-web build are green locally. A few points, ordered by severity — none are hard blockers.

### 🟠 The ADR/doc security framing overstates bastion-secret isolation (please reword)

ADR-013 says transitive authorization grants *"network transit through its bastions **(not their secrets)**"*, and `docs/design/ssh-jump-host.md` says *"The agent never receives a bastion's credential material."*

That's not what the code does for an **explicit-credential bastion**. `credential.get` for a bastion returns `buildHostSshCredential` (`adapter.ts` ~L79–91), which **includes the bastion's `private_key`/`password` files**, and the broker materializes them onto the agentbox (0600). It *has* to — the agentbox authenticates to every hop, so it must hold each bastion's credential. So binding a (possibly low-trust / test) target `T` transitively causes the agentbox to be handed bastion `B`'s full long-lived credential. Only **managed** targets keep a key exclusively on the bastion.

To be clear, this is **not a code over-grant** — I checked: the materialized bastion key is only usable for bound targets' chains (`host_exec` targets must be bound / in `host_list`, so the agent can't point it at an arbitrary host). It's a **threat-model-doc accuracy** issue, which matters because operators reason about blast radius from this text. Suggest:
- Reword to: *explicit-credential bastions' keys/passwords are materialized onto the agentbox (0600) like any bound host; only `managed`-target keys stay solely on the bastion.*
- State the blast radius explicitly: **binding a target hands that agentbox the credentials of every bastion in its chain.** Combined with "`is_production` is enforced only on the entry host," binding a *test* target whose chain includes a *prod* bastion materializes that prod bastion's key onto a test agentbox — worth calling out as the intended trust model (or reconsidering).

### 🟡 WS-RPC `credential.get` doesn't enforce `is_production` matching (pre-existing, now also gates jump transit)

- HTTP path (`adapter.ts` ~L339–352): `prodMatch=true` — the binding query joins `a.is_production = h.is_production` and `isJumpOfBoundHost(..., true)`.
- WS path (~L1863–1872): plain `agent_hosts WHERE agent_id AND host_id` + `isJumpOfBoundHost(..., false)` — **no `is_production` check** (the comment at L117–118 acknowledges it; the cluster branch is the same, so it's pre-existing).

This PR didn't introduce the asymmetry, but the new transitive jump authorization now inherits it. If the WS mirror is the production path (K8s agentbox → Runtime), the cross-env guard is absent in prod. Could you confirm this is intended rather than a latent gap?

### 🟡 / nits

- **`fetchManagedKeyFromJump` (`ssh-dial.ts` ~L109) has no output cap** — it accumulates the bastion's `cat` output unbounded (`runCommand` caps at 10 MB; this helper doesn't). Low risk since you own the bastion, but a small cap (e.g. 64 KB) would be tidy.
- **Managed mode takes the first readable `~/.ssh/id_*`** in a fixed `ed25519>rsa>ecdsa>dsa` order (documented), so a bastion with multiple keys where the intended one isn't first fails auth with a generic message. Fixed precedence isn't configurable — fine, just noting.
- **Transient client leak on abort/timeout during a managed key fetch**: if the chain aborts/times out while `fetchManagedKeyFromJump` is in flight, its `.then(dialNext)` can `connectHop` a new client *after* `closeAll()` already ran, so that client isn't torn down by `closeAll` (it self-closes on its own `readyTimeout`). Bounded/self-healing, very minor.

jacoblee-io:
Thanks for the thorough review @LikiosSedo — addressed in `f3de1f2` (rebased onto Jacob's `e41dd5f`):

**🟠 Doc framing (bastion-secret isolation) — fixed.** You're right, the "agent never receives a bastion's credential material" wording was wrong for explicit-credential bastions. Reworded ADR-013 and `ssh-jump-host.md` to state it accurately:
- The agentbox authenticates every hop, so an **explicit-credential bastion's key/password IS materialized onto the agentbox (0600)** like any bound host — only a **managed** target's key stays solely on the bastion.
- Spelled out the blast radius: **binding a target hands that agentbox the credentials of every explicit bastion in its chain**, and since `is_production` is enforced only on the entry host, binding a *test* target whose chain includes a *prod* bastion materializes that prod key onto a test agentbox. Documented as the intended trust model (keep chains within one tier, or use `managed`), and kept your note that the materialized key can't be pointed at an arbitrary host (targets must be bound).

**🟡 WS-RPC `credential.get` `is_production` gap — confirmed pre-existing, documented.** Correct that the WS path (and the cluster branch) don't join `is_production`; the new transitive jump check mirrors each path's existing policy rather than introducing a new asymmetry. I did not change the cross-cutting binding semantics in this PR (shared with clusters); added an explicit ⚠️ in ADR-013 consequences flagging it as a separate follow-up. Happy to file that as its own issue.

**🟡 nit — `fetchManagedKeyFromJump` output cap — fixed.** Now capped at 256 KB (rejects with a clear error if exceeded).

**🟡 nit — transient client on abort/timeout during managed fetch — fixed.** The fetch continuation now checks `settled` before `connectHop`, so no untracked client is spawned after `closeAll()`.

**🟡 nit — fixed `id_*` precedence — acknowledged, left as-is** (documented order, matches the upstream behaviour we're mirroring).

Full suite + `tsc` (backend & portal-web) green.

LikiosSedo (approved):
Approving — all review findings are resolved in `f3de1f2`:

- 🟠 **Doc accuracy**: ADR-013 / `ssh-jump-host.md` now correctly state that an explicit-credential bastion's key/password is materialized onto the agentbox (0600) like any bound host, and only a `managed` target's key stays solely on the bastion. The blast radius (binding a test target whose chain includes a prod bastion pulls that prod key onto a test agentbox) and the mitigating bound-target constraint are both spelled out. Accurate now.
- 🟡 **WS-RPC `is_production` gap**: reasonably recorded as a tracked follow-up (pre-existing, shared with the cluster branch — out of scope for this PR).
- **Managed key fetch** capped at 256 KB; the **post-teardown client spawn** is guarded by the `settled` check on the managed-fetch continuation.

Verified locally on `f3de1f2`: backend `tsc` clean, full suite (3159) green, portal-web build clean, ssh-dial/ssh-client suites pass. Implementation is solid — consistent depth+cycle guards (target + 3 bastions) across all four enforcement points, reverse teardown, broker-free execution layer, `requireAdmin` on every Portal route, managed-requires-jump enforced at write/boundary/acquire/dial.

Non-blocking follow-up: the two hardening changes (256 KB cap, settled-guard on the managed-fetch continuation) ship without dedicated tests — worth backfilling when convenient.

PR #297, opened 2026-06-02
feat(host): jump_chain login + host_list query (full host SSH capability)

jacoblee-io:
## Why
The agent SSH path resolved the jump chain by **name-recursion**: the adapter shipped only `metadata.jump_host` (a bastion *name*) and the Runtime re-called `credential.get(name)` per hop. This caused the deployed `Broker.ensureHost(<id>) completed but host not in registry` on multi-tenant sicore, and left siclaw Portal with a real bug — the explicit branch **silently direct-connected** when `jump_host_id` was set but unresolvable (dangling).

Design: `docs/design/ssh-jump-host.md` §3–§6.

## What
The Portal adapter now **pre-resolves the whole bastion chain by `jump_host_id`** and ships it as an ordered `jump_chain` `[outermost…nearest]` in the `credential.get` response; the Runtime consumes it directly (zero recursion) and falls back to the legacy `metadata.jump_host` when absent.

- **types** (`shared/credential-types.ts`): `ChainHop`/`ChainHopMeta` + `credential.jump_chain`.
- **portal** (`portal/host-api.ts`, `portal/adapter.ts`): shared `walkJumpChainRows` + `chainHopFromRow`; `buildHostSshCredential` emits `jump_chain`, **dual-emits** `metadata.jump_host` for not-yet-migrated Runtimes, and **fails closed** on a dangling/cyclic/over-deep jump, a managed bastion, or a credential-less bastion (closes the silent direct-connect).
- **broker** (`agentbox/credential-broker.ts`): materialize each hop under **isolated** per-hop paths (kept out of the target's `filePaths`), record the chain on the registry entry, unlink hop files on evict; `ensureHost` name-fallback; managed precondition accepts `jump_chain` OR `metadata.jump_host`.
- **runtime** (`tools/infra/ssh-client.ts`): nests the chain onto `target.jumpHost`; `ssh-dial.ts` unchanged.

## Compatibility
Backward compatible — dual-emit + Runtime fallback, no flag-day. Secrets boundary unchanged: `host_list` stays metadata-only; the chain (with decrypted bastion keys) rides only on `credential.get` (materialized to 0600, never logged).

## Scope / follow-ups (out of scope here)
- sicore Go adapter emitting `jump_chain` (separate internal repo).
- Cleanup: delete name-recursion + transitive authz once all Runtimes upgrade (keep the name-fallback).
- `host_list` `query` + pagination (orthogonal track).

## Test
- `npx tsc --noEmit` clean; `npm test` → **3195 passed / 2 skipped**.
- Added: broker materialization/eviction/fallback, ssh-client chain consumption (incl. "no recursion" + "prefers chain over a stale name"), adapter 2-hop emit + dangling fail-closed, `chainHopFromRow` invariants ③④.
- Live cluster e2e (jump pods) pending a deploy from this branch.


jacoblee-io:
Folded in **#298 (host_list query)** so this PR is the complete host story — **discovery + connection**:
- **Connection (jump_chain)**: server pre-resolves the bastion chain; Runtime consumes it (no name-recursion); fail-closed; managed last-hop preserved.
- **Discovery (host_list query)**: `host_list` gains `query`+`limit`+`cursor` (server-side filter over name/ip/description, IPv4-exact, `total`/`next_cursor`), folded into `credential.list(kind=host)`; metadata-only.

Combined: `tsc --noEmit` clean; `npm test` 3209 passed / 2 skipped. credential-types.ts overlap resolved (ChainHop + HostListResult both kept).

LikiosSedo:
I think the earlier jump-chain issues are now fixed in this head (`ensureHostForTool` returns the resolved registry entry, and `walkJumpChainRows` now matches the write-time depth cap). One remaining issue before merge:

**[P2] `host_list` exposes `id` as a selection handle, but `credential.get` still only resolves hosts by name.**

The new host search path returns `HostMeta.id` and `host_list` includes it in the tool output, while `HostMeta` documents it as a stable selection handle. However both Portal credential-get paths still do `SELECT ... FROM hosts WHERE name = ?` for `source_id`:

- HTTP adapter route: `src/portal/adapter.ts`, host credential-request branch
- WS-RPC adapter handler: `src/portal/adapter.ts`, `credential.get` host branch

So if a model/user takes the `id` returned by `host_list` and passes it to `host_exec` / `host_script`, the Runtime/Broker side is now capable of id->name fallback, but the standalone Portal adapter will return `Host not found` before the broker ever sees the resolved credential.

Suggested fix: make Portal host credential lookup accept `source_id` by either `name` or `id` (then continue binding checks against the resolved `host.id`), and add an adapter test covering: `credential.list(kind=host, query=...)` returns `id`, then `credential.get(source=host, source_id=<that id>)` returns the same host credential.

LikiosSedo (approved):
Approved after re-reviewing latest head 8e84daac5baab1c5d2bd860df061656097be80df. The previous id/name contract issue is fixed: host_exec and host_script now prefer the host_list id, while the jump_chain runtime path still consumes the resolved HostLocalInfo and avoids per-hop recursion.\n\nValidation run locally in a detached review worktree:\n- git diff --check origin/main...HEAD\n- npx vitest run src/portal/host-api.test.ts src/portal/adapter-rpc.test.ts src/agentbox/credential-broker.test.ts src/tools/infra/ssh-client.test.ts src/tools/query/host-list.test.ts src/gateway/credential-proxy.test.ts src/tools/cmd-exec/host-exec.test.ts src/tools/script-exec/host-script.test.ts\n\nAll passed: 8 test files, 250 tests.

PR #298, opened 2026-06-02
feat(tools): host_list query + pagination

jacoblee-io:
## What
`host_list` gains an optional `query` (+ `limit`/`cursor`) for **server-side, agent-scoped** filtering over name/ip/description — so the agent can *find* a host instead of dumping every bound host into context.

- **No separate tool or RPC.** Per design §12.2 the query **folds into the existing `credential.list(kind=host)`**; there is no `host_search` tool, and the path does not use the (pre-existing, untouched) `credential.hostSearch` RPC.
- **IPv4-exact smart-match** (so `10.0.0.5` doesn't substring-hit `10.0.0.51`); otherwise substring.
- **Pagination**: default limit 20 (max 100), opaque `cursor`, response `total` + `next_cursor`.
- **`id`** surfaced as a stable selection handle; **metadata only — never secrets**.
- **No-arg is capped too**: host_list always goes through the filtered path (blank query = browse), so it's no longer an unbounded dump.
- **Contract-safe**: the filtered path (`broker.queryHosts`) bypasses the broker registry — no reconcile — preserving `reconcileFullList`'s full-snapshot requirement; the full-list path (no query) is untouched and still feeds the broker.

Implements `docs/design/agent-host-jump-chain.md` §12 (siclaw v1: query + pagination; filters / facets / by-id deferred).

## Chain
`host_list` → `broker.queryHosts` → `transport.queryHosts` → `POST /api/internal/credential-list {query,limit,cursor}` → proxy → `service.queryHosts` → `credential.list` (kind=host, query).

## Test
- `npx tsc --noEmit` clean; `npm test` → **3191 passed / 2 skipped**.
- credential.list host-query (filter / paginate / IP-smart / limit-clamp / total), proxy query-vs-list dispatch, broker passthrough (no reconcile), tool query / browse / paging.

### DIFF
--- a/src/agentbox/credential-broker.test.ts
+++ b/src/agentbox/credential-broker.test.ts
@@ -8,6 +8,7 @@ import type {
   ClusterMeta,
   HostMeta,
   CredentialPayload,
+  HostListResult,
 } from "./credential-transport.js";
 
 /**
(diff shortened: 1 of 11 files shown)


jacoblee-io:
Folded into #297: cherry-picked the host_list query commit onto feat/ssh-jump-chain, so the full host capability — discovery (host_list `query`+pagination) and connection (jump_chain login) — ships as one PR. No code lost; resolved the credential-types.ts overlap (ChainHop + HostListResult both kept).

PR #299, opened 2026-06-03
Keep Siclaw chat stable when agent selector expands

LikiosSedo:
## Summary

- Keep the Siclaw chat agent rail at a fixed narrow width.
- Render the expanded agent selector as an overlay instead of resizing the chat layout.
- Collapse the overlay after selecting an agent or clicking the blank area outside the selector.
- Add a compact hover/focus preview for collapsed rail agent icons, matching Sicore's lighter preview affordance.

## Why

Expanding or collapsing the agent selector changed the flex width of the chat area, which caused the message column and reply box to shift horizontally. Sicore's Siclaw integration keeps the chat stable by treating the selector as a rail/overlay style control, and also exposes enough agent detail from the collapsed rail without forcing the full list open.

## Testing

- `npm run build` in `portal-web`
- `git diff --check`
- Local visual QA with Vite + mock API + headless Chrome: verified the compact preview appears on rail hover, the reply box and session controls keep the same x coordinates before/after expansion, and a blank-area click collapses the overlay.

### DIFF
--- a/portal-web/src/pages/Chat.tsx
+++ b/portal-web/src/pages/Chat.tsx
@@ -58,6 +58,7 @@ export function Chat() {
 
   const handleSelectAgent = (agentId: string) => {
     setSelectedAgentId(agentId)
+    setAgentSelectorCollapsed(true)
     if (location.pathname.startsWith("/chat")) {
       setSearchParams({ agent: agentId })
     }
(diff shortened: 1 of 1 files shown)


jacoblee-io (commented):
Reviewed the diff — solid, well-scoped fix. The fixed-width rail + absolutely-positioned overlay correctly eliminates the chat layout shift (chat area is now `flex-1 min-w-0` and no longer reflows when the selector expands). Three collapse paths (select agent / backdrop click / close button) are all wired up, and the collapsed-rail hover/focus preview is clamped to the viewport with `pointer-events-none`. 👍

A few non-blocking nits:

1. **No `Escape` to close the overlay.** It only collapses via outside-click / close button. Esc-to-close is a common expectation for an overlay/popover — worth a small keydown handler.
2. **`h-[calc(100%-45px)]` is a fragile magic number.** The overlay panel isn't using flex layout, so the scroll area hardcodes the header height (the header is `px-3 py-2`, not guaranteed to be exactly 45px → risk of a few px clip/gap). Suggest making the panel `flex flex-col` and the scroll area `flex-1` (same robust pattern the collapsed rail already uses) instead of the calc.
3. **`220` is hardcoded in two places** (`w-[220px]` panel + `left-[220px]` backdrop). A shared constant avoids them drifting apart.
4. **Preview position isn't recomputed on rail scroll / window resize** — minor; a stale tooltip position if the rail scrolls while hovering.

None of these block merge. I'd suggest folding in at least #1 and #2; the rest are optional.

jacoblee-io (approved):

PR #300, opened 2026-06-04
feat(chat): scroll-to-select copy, image-safe copy & Markdown export

a-yeyang:
Adds a select mode to the chat for exporting parts of a conversation.
Toggle it, then scroll — every message that enters the viewport is
auto-checked as you go (wheel or scrollbar, either direction), so you
build a selection without clicking. A floating toolbar offers All, Clear,
Copy, and Download; per-row checkboxes fine-tune.

Copy is also fixed: it now walks the rendered DOM, rasterizing
chart/Mermaid SVGs to PNG and keeping real  elements, so charts,
diagrams, generated images, and attachments copy as pictures instead of
raw markdown/base64. Download exports the selected messages as a Markdown
file, with images and mermaid preserved so they render in a viewer.

### DIFF
--- a/portal-web/src/components/chat/PilotArea.tsx
+++ b/portal-web/src/components/chat/PilotArea.tsx
@@ -1,4 +1,4 @@
-import { useRef, useEffect, useState, useCallback, useMemo, useLayoutEffect, Fragment } from "react"
+import { useRef, useEffect, useState, useCallback, useMemo, useLayoutEffect } from "react"
 import type { KeyboardEvent } from "react"
 import { formatToolInput } from "../../hooks/usePilotChat"
 import {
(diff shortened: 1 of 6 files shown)


LikiosSedo:
Thanks for the PR. I reviewed the current head (`b47b38d`) against `main`.

The select/copy direction looks reasonable overall: the selection state is nicely isolated in a pure model, and walking the rendered DOM for rich copy is a better approach than trying to re-parse message markdown and remap rendered SVGs later. The added portal tests pass locally, and `portal-web` builds successfully.

One thing looks blocking before merge:

- The PR title/body promises Markdown download/export, but the implementation currently only adds `All`, `Clear`, `Copy`, and exit controls in selection mode. I do not see a `Download` control or download handler. Also, the new transcript serializer intentionally replaces chart/Mermaid/image markdown with `[chart]`, `[diagram]`, and `[image]`, so it does not currently produce a Markdown file with images/Mermaid preserved as described.

Please either add the actual Markdown download/export path, or narrow the PR title/body to the implemented select + copy behavior.

Non-blocking follow-up: `handleScroll` currently scans every `[data-msg-id]` and calls `getBoundingClientRect()` on each scroll event. For long chat histories this can become expensive. Consider an `IntersectionObserver`-based visible set, or at least throttling the scan with `requestAnimationFrame`.

LikiosSedo:
Re-reviewed the updated head (`7c23e153`). The previous blocking issue is addressed: selection mode now has a Markdown download action, and `serializeMessagesToMarkdown` preserves normal message Markdown such as inline `` images and Mermaid fences.

One remaining scope point to clarify:

- The Markdown exporter only serializes `message.content`. It does not include `message.attachments` for user-uploaded/pasted files. In the current Siclaw OCR flow those attachments are transient: the frontend keeps them in React state for the current view, Portal sends them to OCR, and Runtime/history persistence only stores the OCR/text prompt path. After refresh, the original attachment preview is not restored from history. So this is probably not a blocker if the intended export scope is “persisted chat content plus markdown images/Mermaid already in content”.
- If the intended export scope is “everything currently visible before refresh”, including a user-pasted screenshot tile that is still present in `message.attachments`, then the exporter should add those attachments as Markdown image data URLs and cover that with a test.

Non-blocking follow-up still applies: `handleScroll` scans every `[data-msg-id]` and reads layout on each scroll event. For very long chat histories, consider `IntersectionObserver` or `requestAnimationFrame` throttling.

LikiosSedo (approved):

PR #303, opened 2026-06-07
feat(tools): background execution for host_exec + all script tools (+ sub-agent completion fixes)

jacoblee-io:
## What

Adds `run_in_background` to **`host_exec` + all four script tools** (`local_script` / `host_script` / `pod_script` / `node_script`), completing background-execution coverage across the tool family (previously only `bash` / `node_exec` / `pod_exec` / `spawn_subagent` supported it). Also fixes several background-completion bugs surfaced by full-chain testing.

## Why

Long-running work (SSH host capture/perf, skill orchestration, soak/perftest) lives mostly on host/script tools — exactly the tools that lacked background support. The model had to block the whole turn on them.

## How

**New background backends** (`eef4ca2`)
- `host_exec` / `host_script` run via **ssh2 in-process**, so the existing subprocess-spawn background runner couldn't drive them — added a **streaming** ssh path (`runCommandStream` in `ssh-dial.ts`, `sshExecStream` in `ssh-client.ts`) and a `streamFactory` mode in `background-bash-runner.ts`.
- `node_script` / `pod_script` pipe the script via **stdin** to `kubectl exec` — added stdin support to the runner.
- `local_script` reuses the argv subprocess path.
- All background paths go through `preExecSecurity` (whitelist) first, reject non-line-safe (JSON) sanitizers, and don't widen the read-only/whitelist boundary. Gated by `BACKGROUND_BASH_ENABLED`.

**Sub-agent background completion fixes** (the UI was visibly wrong)
- `455d2fb` — the model spawned a bogus *"Wait for notification"* sub-agent to "wait"; launch guidance now forbids polling/sleeping/spawning a waiter — end the turn, report on notification (or use the foreground form if you need the result inline).
- `2776b46` — on a completion notification the model re-dispatched the same job; notification guidance now states the result is final and must not be re-run.
- `5718805` — **core**: the synthetic-turn persist guard kept a reaction only when it made a tool call (`turnHadTool`). A sub-agent's result is **inline in the notification summary**, so the model reports it as pure text (no tool call) and the answer was being dropped. A notification without `output_file` now allows a text-only report to persist (bash/exec, which carry `output_file`, keep the strict guard). `46f54ce`/`9b322b4` add the live background indicator + completion fold for the sub-agent card.

**Pre-existing bug found via testing** (`0c9c251`)
- `pod_script` built `kubectl … -i exec <pod>` — `-i` (an `exec` flag) **before** the `exec` subcommand → `Error: flags cannot be placed before plugin name`. Every `pod_script` call failed (exit 1), in both foreground and the new background path. Fixed to `… exec -i <pod> …` (matches `node_script`/`node_exec`). The background completion pipeline is what surfaced it — it correctly reported the failure.

**Test fixture** (`09aaf02`) — `exec-smoke` skill (side-effect-free `sleep N; echo MARKER`) to drive deterministic full-chain smokes across every backend.

## Verification (test env, siclaw-inner :test)

Full-chain background (launch → `exec_job_event` terminal status → model follow-up report), all **PASS**:

| Tool | Backend | Result |
|---|---|---|
| spawn_subagent | inline result | ✅ ×10 — no "wait" sub-agent, answer always reported |
| bash | subprocess | ✅ completed |
| host_exec | ssh2 in-process stream | ✅ completed |
| local_script | subprocess | ✅ completed |
| host_script | ssh2 stream | ✅ completed |
| pod_script | kubectl exec | ✅ completed (after the `-i`/`exec` fix) |
| node_script | debug-pod + nsenter | ✅ completed |

- `npm test`: 167 files / 3323 tests green (+ new regression tests for the persist guard and the `pod_script` argv order).
- Real-browser (CDP) smokes for the exec box fold + the sub-agent card indicator/fold.

## Known residual (model variance, not a correctness issue)

On the deliberately-artificial prompt "use a sub-agent to compute a trivial sum in the background", the model still **occasionally** spawns one extra *identical* compute sub-agent on the completion notification (~40% of runs). The answer is always correct and the bogus "wait" sub-agent is gone; guidance discourages it but can't fully eliminate the variance, and there's no clean non-hacky code-level guard for sequential post-completion re-dispatch.

## Notes for review
- `exec-smoke` is a smoke fixture baked into the core image; drop it if undesired.
- `src/core/prompt.ts` was **not** touched.

### DIFF
--- a/docs/design/tools.md
+++ b/docs/design/tools.md
@@ -161,8 +161,10 @@ Step 8: formatExecOutput()
 
 | Tool | Method | How |
 |------|--------|-----|
-| `node_script` | base64 inject | Encode script → echo + base64 -d inside nsenter. Supports `netns` param for pod network namespace. |
+| `node_script` | stdin pipe | `kubectl exec -i` → `nsenter … sh -c '<reader>'` with the script piped to stdin (`sh -s` / `python3 -`). Supports `netns` param for pod network namespace. |
 | `pod_script` | stdin pipe | `kubectl exec -i` with script piped to stdin |
+| `host_script` | stdin pipe | ssh2 channel with the script piped to the remote `sh -s` / `python3 -` |
+| `local_script` | argv | interpreter + scriptPath (+ args) spawned directly in the AgentBox |
 
 ---
 
@@ -177,7 +179,7 @@ The most complex tool. Handles full shell pipelines with:
 - Production mode: `sudo -E -u sandbox` user isolation
 - Skill script detection bypass (`isSkillScript`)
 - 3-layer output sanitization (see §6.2)
-- Optional `run_in_background` (see §9) — exposed only when a `backgroundBashExecutor` is injected
+- Optional `run_in_background` (see §9) — exposed only when a `backgroundExecExecutor` is injected
 
 **kubectl access**: There is no dedicated kubectl tool — all kubectl commands go
 through `restricted_bash` with pipeline validation. `validateKubectlInPipeline()`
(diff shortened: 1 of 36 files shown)


jacoblee-io:
## Review fixes (all addressed + verified in test env)

Thanks for the review — triaged each point. Summary:

### 🔴 P0 — fixed
- **#1 `runCommandStream` no `'error'` handler** → a channel error now resolves `done` as a failure (no uncaught throw, no hang, chain torn down). Same listener added to the foreground `runCommand` (ssh-client + Portal `/test`). `e4d6a9d`
- **#3 stdin EPIPE** → `'error'` listeners on `stream.stdin` / `child.stdin` (the async EPIPE a `try/catch` can't catch). `e4d6a9d`
- **#2 job_stop didn't kill the remote process** → `host_exec`/`host_script` now run under `setsid` and record the session id; job_stop kills the whole remote **session** over a fresh ssh connection. First did process-group kill, but an e2e test showed **2 survivors**: GNU `timeout` puts its child in its own process group, so a group-kill missed it. Switched to `pkill -s <sid>` (group-kill fallback) — re-tested: **0 survivors**. `e4d6a9d` + `d1e8039`

### 🟠 P1 — fixed
- **#4 `pod_script` `-i` before `exec`** → already fixed before the review in `0c9c251` (`exec -i pod`); verified.
- **#6 `allowTextOnlyPersist` per-batch** → changed `batch.some` → `batch.every(!outputFile)`: a mixed batch keeps the strict guard, so a shell job's pure-ack can't ride a sub-agent's inline-result allowance. `a410d3e`
- **#7 job_stop lost during dial** → runner honours a prior "stopped" status the moment streams go live; `runCommandStream` closes immediately if its signal is already aborted. `e4d6a9d`

### 🟡 P2
- **#8 timed_out/partial shown green** → `annotateSubagentCompletions` now maps them to their own statuses; `statusTone` already renders both amber. `a410d3e`
- **#5 `pod_script` needs `timeout` in the pod / orphan-on-stop** — documented limitation (distroless pods); same pattern as `node_script`. Not changed this round.
- **#9 / #10** — acknowledged: the fold self-heals via the active-bg poller; the "no re-dispatch" guard is guidance-level (no clean code-level circuit-breaker for post-completion re-dispatch).

### Scope
- **`growth/` removed** from the branch (accidental `git add -A`; now untracked). `105ea96`

### Verification (test env)
- `npm test`: 167 files / 3329 tests green (+ new regression tests: ssh-dial error/EPIPE/abort, host_exec setsid+session-kill wiring, mixed-batch persist drop).
- Full-chain background re-run after the hardening — bash / host_exec / 4 script tools all `completed`, no regression.
- New e2e `host-stop-orphan.mjs`: job_stop → **0 remote survivors**.

jacoblee-io:
## Follow-ups optimized (3rd-review's 5 open items)

- **#5 nesting fragility** → extracted the setsid-wrap + session-kill + pidfile logic into a shared `bg-session.ts` helper (host_exec / host_script / pod_script all use it), with unit tests. The tricky quoting now lives in one covered place. `4d94059`
- **#3 pod_script orphan-on-stop** → brought to parity with host/node: the in-pod command runs under `setsid`, and job_stop kubectl-exec's a session-kill into the pod. e2e `pod-stop-orphan.mjs`: **0 survivors**. `87a397e` `c67daa6`
- **#9 card stuck "Running…" on a silent completion** → emit a targeted `subagent_done` chat event on background sub-agent completion (mirrors `exec_job_done`); the frontend folds the card in place, no refetch — works regardless of page/stream state. `486dbf8`

Deliberately left as-is (sound by design):
- **#4** (`every` drops a sub-agent's prose in the rare sub-agent+shell same-window batch) — partitioning would cost a second synthetic turn for a rare case; the result still shows via the card fold. Kept `every`.
- **#10** (no code circuit-breaker for re-dispatch) — guidance-level by design; a description-match guard would be brittle.

### Verification (test env)
- `npm test` 168 files / 3334 green; portal-web 102 green (+ bg-session, pod_script onAbort, subagent_done emit tests).
- e2e: HOST job_stop reap **0 survivors**, POD job_stop reap **0 survivors**.
- Regression: bash / host_exec / all 4 script tools background full-chain still `completed`.

nightmeng (approved):

PR #304, opened 2026-06-07
feat(tools): one-step pod-netns execution (node_exec/script + host_exec/script, kubectl & ssh)

jacoblee-io:
## What

Adds a **one-step** way to run host tools inside a **pod's network namespace** — pass `pod` (+ `namespace`) directly to `node_exec` / `node_script` (kubectl) or `host_exec` / `host_script` (ssh), and the tool resolves the netns and runs `… ip netns exec <netns> <cmd>`. The two-step `resolve_pod_netns` → `…(netns=)` path is retained for reuse/advanced use.

```
node_exec:  pod="rdma-a", namespace="rdma-test", command="show_gids"            # kubectl (node resolved too)
host_exec:  host="<node>", pod="rdma-a", namespace="rdma-test", command="show_gids"  # ssh (host = the node, root)
# same for node_script / host_script
```

## Why

RDMA/RoCE connectivity tests need tools (`show_gids`, `ib_write_bw`, perftest) that pods usually lack. The fix is to enter the **node's host namespaces** and re-enter **only the pod's net namespace** → host tools against the pod's network. Previously this worked only via kubectl, as a two-step (`resolve_pod_netns` then `node_exec(netns=)`), and not at all over SSH.

## How

- **New `src/tools/infra/pod-netns-resolve.ts`** — shared `validateNetnsName`, `buildCrictlNetnsScript`, and two resolvers:
  - `resolvePodNetnsViaKubectl` (crictl in a privileged debug pod via `nsenter -t 1`)
  - `resolvePodNetnsViaSsh` (crictl directly on the node — SSH already lands in the host namespaces, so **no `nsenter`**)
  - `resolve_pod_netns` refactored onto it (no behavior change).
- **node_exec / node_script** — `node` now optional; when `pod` is given the tool resolves `{node, netns}` internally and reuses the existing netns-exec path (literal, validated netns — no fragile inline resolver in the background/setsid stack).
- **host_exec / host_script** (new netns capability) — when `pod` is given, resolve the netns over SSH and wrap `ip netns exec <netns> sh -c '<cmd>'` (host_exec) / `ip netns exec <netns> bash -s …` (host_script); foreground + background (setsid + session-kill unchanged).

**Security / invariants:** the netns / `nsenter` / `ip netns exec` prefix is always tool-constructed (never raw model input); the inner command still passes the same whitelist (`preExecSecurity`); `pod`/`namespace`/`container` are validated; the netns name (from crictl) is regex-checked. No new capability vs the existing `node_exec(netns)` — same mechanism, now one step + over SSH. host_* requires a **root** SSH credential on the node (crictl + `ip netns exec` need CAP_SYS_ADMIN). Assumes the CNI creates `/var/run/netns/<name>`.

## Verification

- `npx tsc --noEmit` clean; `npm test` **169 files / 3346 tests** green (+ new: `pod-netns-resolve` resolver tests, and pod-path tests on all four tools — netns wrapping, invalid-name rejection, no-pod-unchanged).
- **End-to-end in the test cluster** (pod `managed-target` on node `nodepool-061`), all PASS:

| tool × transport × mode | result |
|---|---|
| node_exec(pod=) kubectl fg | host tool saw the pod's `eth0,lo` (== `pod_exec`) |
| node_script(pod=) kubectl fg | skill script ran in the pod netns |
| node_exec(pod=) kubectl **bg** | completed + reported |
| host_exec(pod=) **ssh** fg | saw pod's `eth0,lo`; node itself has 40 ifaces → entered the pod netns over SSH |
| host_script(pod=) **ssh** fg | skill script ran in the pod netns |
| host_exec(pod=) **ssh** bg | completed + reported (bg + setsid + session-kill in netns over SSH) |

Note: real RDMA tools (`show_gids`/perftest) weren't exercised — the test nodes have no RDMA hardware; `ip -br link` was used as the netns proxy (same "host tool in pod netns" mechanism).

### DIFF
--- a/docs/design/tools.md
+++ b/docs/design/tools.md
@@ -133,10 +133,22 @@ Step 10: formatExecOutput() / processToolOutput() → infra/exec-utils.ts / tool
 | Pipeline support | Yes (pipes, `&&`, `;`) | No (`blockPipeline: true`) |
 | Execution | `runInDebugPod` | `execFileAsync("kubectl")` |
 
-**Pod network namespace diagnostics**: To run host-level tools in a pod's
-network namespace, use `resolve_pod_netns` (in `query/`) to get the netns name,
-then call `node_exec` with the `netns` parameter. This replaces the former
-`pod_nsenter_exec` tool.
+**Pod network namespace diagnostics** (host tools against a pod's network — RDMA/RoCE
+on a pod that lacks `show_gids`/perftest): enter the pod's **net** namespace while keeping
+the host's mount namespace, i.e. `… ip netns exec <netns> <host-tool>`. Two transports,
+one mechanism (`src/tools/infra/pod-netns-resolve.ts` resolves pod → netns):
+
+- **One step (preferred)** — pass `pod` (+ `namespace`) directly; the tool resolves the netns
+  internally and runs the command inside it:
+  - kubectl: `node_exec(pod, namespace, command)` / `node_script(pod, …)` — node is resolved too.
+  - ssh: `host_exec(host=<node>, pod, namespace, command)` / `host_script(…)` — `host` is the node;
+    its SSH credential must be **root** (crictl + `ip netns exec` need CAP_SYS_ADMIN).
+- **Advanced / reuse** — `resolve_pod_netns(pod, namespace)` → `{node, netns}`, then
+  `node_exec(node, netns, command)`; lets one resolution serve many commands.
+
+The netns/nsenter/`ip netns exec` prefix is always **tool-constructed** (never raw model input):
+the inner command goes through the same whitelist; the netns name (from crictl) is regex-validated.
+Assumes the CNI creates `/var/run/netns/<name>`. Replaces the former `pod_nsenter_exec` tool.
 
 ---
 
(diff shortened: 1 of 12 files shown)


jacoblee-io:
## P0 fixed — namespace/container now validated at the choke point (`c634a6c`)

You're right, and thanks for catching it. `namespace` (and `container`) reached `buildCrictlNetnsScript`'s double-quoted shell string unvalidated → root command injection on both transports, bypassing the whitelist.

Fix: validate **pod + namespace + container** inside `resolvePodNetnsViaKubectl` / `resolvePodNetnsViaSsh` (`validatePodTarget`) — the single choke every resolver and tool passes through, so no current or future caller can reach the crictl script with an unsafe value. Added `validateNamespace` / `validateContainerName` (RFC-1123 label) to `exec-utils`, and corrected the comment that falsely claimed the caller validated namespace.

Regression tests added:
- `validateNamespace` rejects the injection payload (`default" ; id ; echo "`), shell metachars, uppercase, dots, >63 chars.
- `resolvePodNetnsViaKubectl` / `resolvePodNetnsViaSsh` reject a shell-injecting namespace **before** any debug-pod / SSH exec runs (asserts `runInDebugPod` / `sshExec` are never called).

`npx tsc --noEmit` clean; `npm test` 169 files / 3349 tests green.

nightmeng (approved):

PR #306, opened 2026-06-08
fix(bg): don't wait for a paired server's completion — carve a generic server/client exception

jacoblee-io:
## What

Carves a **generic paired server/client exception** into the `run_in_background` guidance so the model stops waiting for a blocking background *server's* completion before launching its *client*. Stated generically — these are general diagnostic tools, not RDMA tools.

## Why

The shared launch guidance said, absolutely: *"END YOUR TURN NOW — do NOT call any other tool yet, and do NOT wait/sleep"* (added to stop the model spawning "wait" sub-agents / polling). For a **paired** test that rule deadlocks: a server/listener with no peer arg blocks until a client connects, but its completion notification only fires when it **exits** — so a model that "ends the turn and waits" never launches the client, and the server times out (`exit 124`). First hit in two-node RDMA perftest runs, but the failure mode is general (any server/listener→client, or capture→traffic flow).

## How

Keep the default (end turn; don't poll/sleep/spawn-a-waiter), but add an **EXCEPTION**: when the background command is one **side** of a protocol that only makes progress once its counterpart runs, do NOT wait — run the counterpart now, then read the `output_file` once the test finishes. Stated generically, with no RDMA-specific wording:

- `background-launch.ts` — the shared launched-result message (read on every background launch; the primary lever).
- `node_exec` / `host_exec` / `pod_exec` and `node_script` / `host_script` / `pod_script` / `local_script` — replaced the contradictory absolute *"do NOT call any tool"* with the default + server/listener exception, so **every** background-capable tool's param description now agrees with the shared launch result.
- Re-illustrated `run_in_background` with **whitelist-only, non-RDMA** examples: an open-ended `tcpdump` capture (start in background, stop with `job_stop`) on `node_exec`/`host_exec`, and a paired capture/traffic example on `node_exec` (`tcpdump -c 5` in the background, then `curl` to generate the packets it's waiting for). `nc`/`netcat`/`socat`/`iperf` are intentionally excluded from the command whitelist, so they can't appear in a runnable example; the packet count is kept low enough that one request trips it.

The command whitelist (incl. the RDMA/perftest groups) and the perftest tuning-flag note are left intact — those are capability/security facts, not "how to run RDMA". `prompt.ts` is **untouched** (it carries no background guidance). Added `background-launch.test.ts` asserting both the default and the (generic) exception wording.

## Verification

- `npx tsc --noEmit` clean.
- Background suites green: `background-launch`, `node-pod-background`, `host-exec-background`, `script-background`.
- Wording-only change to tool/param descriptions — no runtime, `job_stop`, or sanitization behavior changed.

### DIFF
--- a/src/tools/cmd-exec/background-launch.test.ts
+++ b/src/tools/cmd-exec/background-launch.test.ts
@@ -1,0 +1,33 @@
+import { describe, it, expect } from "vitest";
+import { backgroundLaunchedResult, backgroundNotLineSafeError } from "./background-launch.js";
+
+describe("backgroundLaunchedResult message", () => {
+  const msg = () => {
+    const r = backgroundLaunchedResult("functions.node_exec:0", "/o/file.output", "Running on the node in the background.");
+    return JSON.parse(r.content[0].text).message as string;
+  };
+
+  it("keeps the default: end the turn, don't poll/sleep/spawn-a-waiter", () => {
+    const m = msg();
+    expect(m).toMatch(/END YOUR TURN/);
+    expect(m).toMatch(/do NOT read this file, poll, sleep, or spawn a sub-agent/i);
+  });
+
+  it("carries the paired server/client EXCEPTION (so a perftest server isn't waited on → no deadlock)", () => {
+    const m = msg();
+    expect(m).toMatch(/EXCEPTION/);
+    expect(m).toMatch(/immediately run the counterpart/i);
+    expect(m).toMatch(/deadlock/i);
+    expect(m.toLowerCase()).toContain("client"); // the counterpart to run now
+  });
+
+  it("still hides the internal handles from the user", () => {
+    expect(msg()).toMatch(/do NOT show them to the user/);
+  });
+
+  it("backgroundNotLineSafeError stays a blocked structural-redaction rejection", () => {
+    const r = backgroundNotLineSafeError();
+    expect((r.details as any).blocked).toBe(true);
+    expect((r.details as any).reason).toBe("background_not_line_safe");
+  });
+});
--- a/src/tools/cmd-exec/background-launch.ts
+++ b/src/tools/cmd-exec/background-launch.ts
@@ -43,11 +43,16 @@ export function backgroundLaunchedResult(
       task_id: jobId,
       output_file: outputFile,
       message:
-        `${runningWhere} END YOUR TURN NOW — do NOT read this file or call any other tool yet, and do NOT ` +
-        "wait/sleep. You will be notified automatically when it completes; read the output_file ONLY after " +
-        "that notification arrives. To stop it early, use job_stop. NOTE: task_id and output_file are internal " +
-        "handles for YOUR use only (job_stop / reading the file later) — do NOT show them to the user; just tell " +
-        "the user in plain language what is running and that you'll report back when it finishes.",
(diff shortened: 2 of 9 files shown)


LikiosSedo:
Reviewed current head `92f89eda` against `main` (`2f76f5a`). No blocking issues found.

This change looks reasonable to me: the old launch guidance was correct for independent long-running background jobs, but it creates a real deadlock for paired server/client or capture/traffic workflows because the completion notification only arrives after the blocking side exits. Keeping the default "end the turn / do not poll" rule while carving out the paired-orchestration exception matches the runtime contract.

I also checked the upstream/downstream path: the shared `backgroundLaunchedResult()` message is consumed by the background exec/script family through `spawnBackgroundBash`, then completion flows through `JobRegistry` and `<task_notification>` before the model reads the sanitized `output_file`. This PR does not change the runtime/job_stop/sanitization behavior, so the risk is mainly wording/agent-behavior risk, not execution risk.

Non-blocking suggestions:

1. Please update the PR body: it says the `node_exec` RDMA perftest example was updated, but the actual code replaces that with a generic `tcpdump` + `curl` paired capture/traffic example. The code direction looks intentional; the description is just stale.
2. Follow-up cleanup worth considering: `bash`, `pod_exec`, and the `*_script` tools still have older schema text that says not to call any tool after a background launch, while they also reuse the new shared launch result. Not a blocker because the launch result is the decisive instruction, but aligning those descriptions would reduce future prompt ambiguity.

Local verification I ran:

- `npm test -- src/tools/cmd-exec/background-launch.test.ts src/tools/cmd-exec/restricted-bash-background.test.ts src/tools/cmd-exec/node-pod-background.test.ts src/tools/cmd-exec/host-exec-background.test.ts src/tools/infra/command-sets.test.ts`
- `npx tsc --noEmit`

LikiosSedo (approved):
Re-reviewed the updated head `fdc869d` after the follow-up commit. Approved.

The new commit addresses the main wording mismatch I called out earlier: `pod_exec` and the `*_script` background-capable tools now use the same default + paired-test exception as the shared `backgroundLaunchedResult()` path. That makes the model contract consistent for the important diagnostic tools: default is still to end the turn and avoid polling/sleeping, while paired/orchestrated tests can immediately launch the counterpart to avoid deadlocking a blocking server/listener or capture.

I also rechecked the downstream path: this remains wording/model-behavior only. The background launch still flows through the existing executor, `JobRegistry`, completion notification, `output_file`, `job_stop`, and sanitization path; no runtime/job lifecycle behavior is changed.

One small non-blocking cleanup remains: `bash` / `restricted-bash` still has the older pre-call schema wording that says not to call `read` or any other tool after a background launch, while it returns the same updated shared `backgroundLaunchedResult()` after launch. I don't think this blocks the PR because the exec/script diagnostic paths are now aligned and are the main paired-test surface, but it would be nice to align `restricted-bash` too or narrow the PR body from "every background-capable tool".

Local verification on this head:
- `npm test -- src/tools/cmd-exec/background-launch.test.ts src/tools/cmd-exec/restricted-bash-background.test.ts src/tools/cmd-exec/node-pod-background.test.ts src/tools/cmd-exec/host-exec-background.test.ts src/tools/script-exec/script-background.test.ts src/tools/infra/command-sets.test.ts`
- `npx tsc --noEmit`

PR #308, opened 2026-06-08
feat: add pi-agent 0.78.1 model fallback routing

LikiosSedo:
## Summary

Upgrades Siclaw to `@earendil-works/pi-coding-agent@0.78.1` and builds the model fallback routing layer on top of that baseline.

This PR includes both the compatibility upgrade and the runtime/Portal fallback implementation: ordered model candidates, failure classification, cooldowns, routing state persistence, AgentBox smoke coverage, and compact chat UI metadata for the model actually used on each reply.

## Commit Layout

- `chore: upgrade pi-agent to 0.78.1`
- `feat: add model fallback routing`
- `feat: show model route metadata in chat`

## Changes

- Replace the old `@mariozechner/pi-*` packages with `@earendil-works/pi-coding-agent@0.78.1` and aligned `@earendil-works/pi-agent-core` / `@earendil-works/pi-tui` imports.
- Raise the Node runtime baseline to `>=22.19.0` and pin Portal/Runtime/AgentBox Docker bases to Node 22.19 slim images.
- Add ordered fallback routing in AgentBox / Runtime with configurable candidates, per-failure fallback decisions, cooldowns, retry-after hint handling, and persisted route state.
- Preserve live streaming and token/cost stats while routed attempts buffer failed-attempt events and replay only the winning response.
- Surface `model_route_*` events through the gateway/SSE path and persist fallback metadata on assistant/tool messages.
- Add Portal agent settings for fallback order and compact chat model indicators, including a lightweight fallback explanation icon.
- Add AgentBox smoke and focused unit coverage for routing defaults, failure classification, cooldown behavior, model switching, recovery, synthetic turns, and UI metadata.

## Review Hardening

- Use escaped route candidate keys so provider/model ids containing `/` do not collide.
- Classify quota/billing 403s before auth failures and distinguish transport aborts from user stops.
- Treat missing/stale fallback candidates as skipped or exhausted attempts instead of hard-aborting the request.
- Route background/synthetic parent turns through the same session model-routing runner when a multi-candidate policy is active.
- Persist route metadata for routed tool-only turns, not only assistant text replies.
- Harden Portal agent updates so an explicit `model_routing: undefined` internal payload does not bind `undefined` into SQL.

## Compatibility Boundary

The public HTTP/SSE surface remains compatible:

- `/api/prompt`
- `/api/stream/:sessionId`
- `/api/models`
- `/api/sessions/:id/model`
- `createSiclawSession`
- `PiAgentBrain`
- dynamic `registerProvider()`
- request/session-level `setModel()` semantics

Manual user model selection remains strict and clears route cooldowns for that session.

## Validation

Current head `1a8f6f5`:

- `npx vitest run src/core/model-routing.test.ts src/agentbox/notify-parent.test.ts src/agentbox/http-server.test.ts src/agentbox/session.test.ts src/portal/agent-api.test.ts src/gateway/sse-consumer.test.ts` (`191 passed`)
- `npm run build`
- `npx tsc -p tsconfig.agentbox.json`
- `npm run build:web`
- `git diff --check origin/main...HEAD`

Earlier compatibility/deployment checks for this feature line also covered Runtime/AgentBox/Portal Docker builds, Node 22.19 in-container verification, and standalone/integrated prompt smoke paths.

## Related Work

The Sicore-side bridge/configuration MR is separate. This PR provides the Siclaw runtime and standalone Portal capability that Sicore consumes.


jacoblee-io:
## Review

Heads-up on scope first: this isn't just a dependency bump — it bundles the pi-agent 0.78.1 upgrade (scope rename `@mariozechner` → `@earendil-works`, lockfile +4124 lines, ~248 mechanical import rewrites) **with a ~1800-line net-new model-routing feature** (`model-routing.ts` +1022, `model-routing-config.ts`, AgentSettings UI, a migration, agentbox run-loop changes). The two are functionally independent (routing only leans on the existing `BrainSession` interface plus one optional `onResponse` tap), so I'd strongly suggest splitting into PR-1 (upgrade + mechanical renames + node bump, independently revertable) and PR-2 (routing on top) — otherwise a prod regression in 0.78.1 can't be reverted without losing the feature and vice versa, and the lockfile churn buries the feature diff.

### Correctness

1. **`candidateKey` uses an ambiguous `/` separator** — `core/model-routing.ts:207` returns `` `${provider}/${modelId}` ``. Model ids commonly contain `/` (e.g. `meta-llama/Llama-3-70B`), so `{provider:"a", modelId:"b/c"}` and `{provider:"a/b", modelId:"c"}` both key to `a/b/c` → distinct candidates collide in dedup + cooldown state (one candidate's cooldown suppresses the other). The frontend uses a different separator (`::`, `AgentSettings.tsx:68`), so the two layers also disagree on candidate identity. Use a non-collidable key (JSON or `\x00` join) and share one definition.

2. **403 is classified as `auth` before the billing check** — `model-routing.ts:343` matches `\b403\b|forbidden` → `auth` *before* `isBillingLike()` runs. A "403 … quota/usage limit exceeded" is tagged `auth` (in `DEFAULT_NO_FALLBACK_ON`), so the routed turn hard-fails instead of failing over. Run the billing check before the 403→auth rule.

3. **`\baborted\b` matches transport errors** — `model-routing.ts:329`: "connection aborted" / "request aborted" → `user_abort` (no-fallback). A transient network give-up is treated as a deliberate user cancel and skips the healthy fallback.

4. **A misconfigured fallback candidate hard-aborts the turn** — `model-routing.ts:571` + `model-routing-config.ts:30`. A fallback whose provider row is missing from the DB hydrates with no `modelConfig` → a `setup`/`format_error` candidate; the loop then `throw`s instead of skipping it. A mistyped/stale fallback model id thus turns a *recoverable* primary `rate_limit` into a hard turn failure — worse than having no routing configured. Skip unusable candidates and fall through to the exhausted result.

5. **Background / synthetic turns bypass routing** — `agentbox/session.ts:609`: `runSyntheticPrompt` (background-job completion reactions, inline sub-agent results) calls `brain.prompt()` directly, not `runPromptWithModelRouting`. When the primary is down, the live HTTP turn fails over but a background completion turn hits the dead primary and the model's reaction is silently dropped — the "silent disappearance" the `canPersist` warning was added to prevent.

6. **Checkpoint rewind can leave a dangling `tool_use`** — `pi-agent-brain.ts:244`: `restorePromptCheckpoint` rewrites pi-agent's internal `state.messages` to discard a failed attempt. If that attempt emitted a partial assistant `tool_use` before failing, the fallback model can see an unmatched `tool_use` → an unmatched-tool-result error on the retry. Worth a test that fails loudly if the `branch`/`getLeafId`/`resetLeaf` internals change.

7. **A user-pinned model silently disables fallback** — `http-server.ts:361`: `userSelectedModelStrict` short-circuits `routeEnabled`. After an operator pins a model (`POST …/model`), if that provider starts returning 529/overloaded, no fallback is attempted — routing is off exactly when it's wanted, with no UI signal. Consider surfacing this as explicit state rather than an implicit kill-switch in the `.model-route-state.json` sidecar.

8. **Route audit metadata dropped on tool-only turns** — `sse-consumer.ts:638`: `model_route` metadata is attached only in the assistant-text persist branch (gated on non-empty `assistantContent`). A routed turn that ends in pure tool calls loses the persisted record of which model actually served it (the live notice bubble still fires, but history loses the audit row).

### Robustness / altitude

9. **INSERT/UPDATE asymmetry** — `agent-api.ts:169`: the UPDATE path pushes `encodeModelRoutingForDb(body.model_routing)` without the `?? null` guard the INSERT path uses; `encodeModelRoutingForDb(undefined)` returns `undefined`, and binding `undefined` throws in node:sqlite. Latent (the UI always sends an object or null) but a real robustness gap for direct PATCH callers.

10. **Routing orchestration lives in the agentbox HTTP handler, not core** — the whether-to-route gate, event buffering, and SSE bridging are hand-assembled inline in `http-server.ts` (~358-588), so the TUI / `tui-background-host` path and synthetic turns inherit no routing (see #5). The one non-mechanical upgrade touchpoint is also a `(session as unknown as {agent?:{onResponse?}}).agent.onResponse` monkey-patch in `pi-agent-brain.ts:219` — the `as unknown` cast defeats the compiler, so a future bump that renames it breaks routing silently (degrades to errorMessage-only classification, no signal). A thin documented adapter + a test would catch that.

### Also worth a look (didn't make the top 10)
- `server_error` is matched before `model_not_found`, so "503 … model not exists" gets the short cooldown instead of the 10-min one (`model-routing.ts`).
- `loadProviderConfigs` re-fetches the primary provider config per prompt and runs 2×K queries in a serial `await` loop on the per-prompt chat path — batch with `WHERE name IN (…)` / `provider_id IN (…)` and pass the already-loaded primary in (the snapshot resolver already proves this is free).
- Provider-config construction (the `{name, baseUrl, apiKey, models:[…]}` shape) is duplicated 4× (adapter ×3 + `loadProviderConfigs`) and `resolveAgentModelRouting` / `resolveSnapshotModelRouting` are near-identical — extract a shared `buildModelConfigForProvider` + one hydrate helper.

### Verified clean
- `prompt.ts` is untouched (no human-approval gate tripped).
- The migration stores `model_routing` as `TEXT` (MySQL + SQLite compatible, no `TIMESTAMP(3)`/`ON UPDATE`/JSON column — invariants §5 OK), with a matching `migrate-sqlite` assertion.
- The `@mariozechner` → `@earendil-works` rename is applied consistently across the ~30 tool files (no leftovers).

Direction is sound — these are mostly failure-classification edge cases, the routing-scope gaps (synthetic turns, user-pin, TUI), and the PR-cohesion concern, rather than the core engine being wrong.

jacoblee-io:
## Re-review (after the rebase + restructure)

Nice turnaround — the branch is now rebased on latest `main` (incl. #309), split into 3 readable commits (`chore: upgrade pi-agent` / `feat: add model fallback routing` / `feat: show model route metadata`), and the substantive findings from the last pass are addressed. Confirmed fixed:

- **candidateKey collision** — now `` `${encodeURIComponent(provider)}/${encodeURIComponent(modelId)}` `` (model-routing.ts:207), so a `/` inside a model id (e.g. `meta-llama/Llama-3`) no longer collides. ✅
- **403 vs billing ordering** — `isBillingLike()` now runs before the `403→auth` rule, so a quota/credit 403 falls over instead of being tagged no-fallback `auth`. ✅
- **`aborted` over-match** — split into `userAbortPattern` vs `transportAbortPattern`; "connection/request/socket aborted" no longer classifies as `user_abort`. ✅
- **Hard-throw on a misconfigured fallback** — the throw is now gated to `source === "prompt_error"` only; a `setup`/missing-provider candidate is skipped and the chain exhausts gracefully instead of aborting the turn. ✅
- **Synthetic / background turns** — `runSyntheticPrompt` now goes through `runPromptWithModelRouting` (session.ts:721), so background-completion and inline sub-agent turns inherit fallback. ✅
- **Route metadata on tool-only turns** — `attachModelRouteMetadata` is now applied to tool rows and tool-start metadata, not just the assistant-text branch. ✅
- **agent-api UPDATE undefined bind** — guarded with `if (modelRouting !== undefined)` before pushing. ✅
- **Cooldown ordering** — `model_not_found` is now matched before `server_error`, so "503 … model not exists" gets the 10-min cooldown, not the 2-min one. ✅
- **Provider re-fetch** — `loadProviderConfigs` now dedupes providers (`[...new Set(...)]`), killing the per-candidate refetch. ✅

Plus a solid test addition (model-routing.test.ts ~41 cases, http-server/notify-parent/pi-agent-brain/session tests).

### Small items remaining (none blocking)

1. **`loadProviderConfigs` is still a serial per-provider loop** (model-routing-config.ts:90) — two `await` queries per provider in sequence. Deduped now, but with K distinct candidate providers it's still 2K serial round-trips on the per-prompt path. Could collapse to one `WHERE name IN (…)` + one `WHERE provider_id IN (…)`. Minor.

2. **Multi-hop "success" context** (sse-consumer.ts `modelRouteSuccessMetadata`) — `switched_from_*` is attached only when `latestSwitch.to_candidate_key === candidateKey`. For A→B→C the success on C correctly shows B, but a success landing on a candidate that isn't the last switch target attaches no failure context. Cosmetic (badge loses the "why").

3. **Checkpoint rewind + tool_use pairing** (pi-agent-brain.ts:254 `restorePromptCheckpoint` via `branch`/`resetLeaf`) — this is presumably covered by the existing `repairToolUsePairingGuard` on the next attempt, but a targeted test (a failed attempt emits a partial assistant `tool_use`, the fallback retries and must not see an unmatched tool_use) would lock that contract down, since it now depends on private pi-agent leaf internals reached via the `as unknown` cast.

4. **User-pin suppresses fallback** (`shouldUseModelRouteRunner`) — now an extracted, tested helper (good). Just confirming the intended UX: when an operator pins a model and that provider later starts failing, routing stays off for the session with no UI signal. Fine if intended; a small "fallback paused (pinned)" hint would close the gap.

Direction and execution both look good now — from my side this is close to mergeable once the above are triaged (most are optional/minor).

jacoblee-io (approved):
Approving — the substantive findings from the prior passes are all addressed (candidateKey collision, 403-vs-billing ordering, `aborted` over-match, hard-throw on misconfigured fallback, synthetic/background turns now routed, route metadata on tool-only turns, UPDATE undefined-bind, cooldown ordering, provider re-fetch dedup), and there's solid test coverage. Nicely rebased on main and split into readable commits.

Remaining items are non-blocking and fine as follow-ups:
- `loadProviderConfigs` serial per-provider queries → could batch with `IN (…)`.
- multi-hop `switched_from` success context (cosmetic badge gap).
- user-pin suppressing fallback with no UI signal (confirm it's intended).

One I'd suggest knocking out soon (the only potential correctness gap left): a targeted test for `restorePromptCheckpoint` + tool_use pairing — a failed attempt that emitted a partial assistant `tool_use`, then a fallback retry, must not leave an unmatched tool_use for the next model. Presumably covered by `repairToolUsePairingGuard`, but worth locking down since it now leans on private pi-agent leaf internals.

Please confirm CI is green before merge (the review here was static). LGTM. 🚀

PR #309, opened 2026-06-08
fix(stop): make the Stop button truly stop everything a session runs (bg jobs + fg exec + UI)

jacoblee-io:
Two commits.

## 1) fix(stop): make the Stop button truly stop everything a session runs

Clicking the dialog Stop should halt the live turn, its background jobs, and any foreground exec — and reflect that in the UI.

- **Background jobs**: `/abort` only called `brain.abort()` (live turn + foreground tools/sub-agents); `run_in_background` jobs are decoupled. Added `stopSessionJobs()` (called from `/abort`) which calls `JobRegistry.stopJob` on every running job.
- **Terminal stop**: user-stopped jobs are marked `suppressNotifyTurn` — the card folds to "stopped" but the model is NOT woken by a synthetic turn to react to its own cancellation (the "comes back to life after Stop"). The model's own `job_stop` is unchanged.
- **Foreground orphan**: `node_exec`/`host_exec` foreground only killed the local kubectl/ssh client, which does not propagate to a host-namespace (`nsenter -t 1`) or non-PTY ssh process. They now run as a `setsid` + `timeout` killable session and reap the remote group over a FRESH connection on abort (`killRemoteSessionViaKubectl`/`ViaSsh`); node_exec foreground PINS the debug pod so it cannot be idle-evicted mid-exec; node_exec background is unified onto the same `pkill -s` helpers. `pod_exec` threads `signal` and returns a clean `Aborted.`.
- **Stuck UI**: `sse-consumer` finalizes un-ended tool rows as stopped (preserving toolName/toolInput) and persists partial assistant text on abort; `persistSyntheticMessage` now writes a terminal `outcome` — a successful tool in a background-completion turn was stored with `outcome=null`, which the frontend rendered as a forever "running"/Thinking… card on refresh. The frontend maps `metadata.status` stopped to `"aborted"`, guards the wholesale-replace refetch pollers, and no longer bounces the sent message back into the input.

## 2) fix(ui): show resolved host name in host_exec/host_script cards (not the id)

The model passes an opaque host id, and `formatToolInput` had no `host_exec` case, so the card showed a bare UUID and hid the command. `acquireSshTarget` now carries `HostMeta.name` on `SshTarget.name`; the tools put it in result details as `host_label` (persisted to tool metadata); `formatToolInput` reads it and renders `<name> $ <command>` like node_exec, falling back to the id. Model behavior is unchanged (still uses the unambiguous id).

## Tests
tsc (root + portal-web) clean; `npm test` green (171 files / 3365 tests) plus portal-web vitest green. Deployed to siclaw-inner and verified.

### DIFF
--- a/portal-web/src/components/chat/PilotArea.tsx
+++ b/portal-web/src/components/chat/PilotArea.tsx
@@ -275,26 +275,17 @@ export function PilotArea({
     setEditingDraft("")
   }, [sessionKey])
 
-  const lastSentRef = useRef<string | null>(null)
   const wrappedSendMessage = useCallback(
     (text: string, attachments?: ChatAttachment[]) => {
-      if (!isLoading) lastSentRef.current = text
       sendMessage(text, attachments)
     },
-    [sendMessage, isLoading],
+    [sendMessage],
   )
+  // Stop just stops: abort the running turn and leave the input alone. We intentionally do NOT
+  // restore the sent message back into the input box — the turn has usually already been
+  // partly processed (tools ran), so re-filling the input reads as the message "bouncing back".
   const wrappedAbort = useCallback(() => {
     abortResponse?.()
-    if (lastSentRef.current) {
-      // Strip the injected markers + fullPrompt so only what the user actually
-      // typed returns to the input. Same parsers used to render past user
-      // bubbles cleanly, so round-trip is consistent.
-      const { text: afterDp } = parseDeepInvestigation(lastSentRef.current)
-      const { text: userTyped } = parseActionChipMarker(afterDp)
-      setChipSeq((s) => s + 1)
-      setChipDraft(userTyped)
-      lastSentRef.current = null
-    }
   }, [abortResponse])
 
   const scrollToBottom = useCallback((smooth = true) => {
@@ -359,11 +350,6 @@ export function PilotArea({
     const text = editingDraft.trim()
     if (!text || isLoading) return
     wrappedSendMessage(text)
-    // wrappedSendMessage records lastSentRef so wrappedAbort can restore the
-    // draft if the user aborts a normal send. For edit-resends the message is
-    // already visible in the conversation, so we don't want abort to re-fill
-    // the input with it.
-    lastSentRef.current = null
     setEditingMessageId(null)
     setEditingDraft("")
   }, [editingDraft, isLoading, wrappedSendMessage])
--- a/portal-web/src/hooks/format-host-tool.test.ts
+++ b/portal-web/src/hooks/format-host-tool.test.ts
@@ -1,0 +1,28 @@
+import { describe, it, expect } from "vitest"
+import { formatToolInput } from "./usePilotChat"
+
+// host_exec/host_script: the model passes an opaque host id; the backend resolves the friendly
+// name into metadata.host_label. The card must read `<name> $ <command>` like node_exec.
+describe("formatToolInput — host_exec / host_script", () => {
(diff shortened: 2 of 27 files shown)


LikiosSedo:
Thanks for pushing this. The overall direction makes sense to me: stopping session background jobs from `/abort`, suppressing model wake-up for user-stopped jobs, finalizing aborted tool rows, and adding killable foreground exec sessions are all the right fixes for the “Stop comes back to life / card stays running” failure mode.

I think there are a couple of gaps worth fixing before merge:

1. `host_exec` foreground abort can still surface as an SSH failure instead of a clean stop. In `src/tools/cmd-exec/host-exec.ts`, the `sshExec()` call is wrapped in a `catch` that immediately returns `reason: "ssh_exec_failed"`. But the real SSH path rejects with `Error("Aborted")` when the AbortSignal fires (`ssh-dial.ts` / `ssh-client.test.ts` already cover that behavior), so the later `if (signal?.aborted) return "Aborted."` branch is unreachable for the real abort-reject path. The current test mocks `sshExec` as resolving after abort, so it misses this. Please check `signal?.aborted` inside the catch before returning `ssh_exec_failed`, and add a test where the mocked `sshExec` rejects after the controller aborts.

2. The PR title/body says Stop truly stops everything a session runs, but the script tools are not covered to the same level. `host_script` foreground still directly calls `sshExec(..., signal)` without the new `setsid + pgid + fresh SSH reap` wrapper, and it has the same abort-before-catch issue. `node_script` foreground still uses `runInDebugPod(..., signal)` without the foreground pgid/reap/pin logic added to `node_exec`. Either include script tools in this stop-hardening pass, or narrow the PR claim and add a follow-up issue/PR for scripts.

3. Smaller UI gap: the host friendly name appears to work on history reload, but live cards are likely still created with the raw host id. `tool_execution_start` calls `formatToolInput(toolName, args)` before `host_label` exists, and `tool_execution_end` stores `toolDetails/metadata` but does not recompute `toolInput`. Please recompute the live tool header when result metadata arrives, or include metadata in the live start event if available.

I verified the focused test/build surface locally and it passes: backend related Vitest tests, root `npm run build`, portal hook tests, and portal `npm run build`. These issues are mostly coverage/edge-path gaps rather than direction problems.

jacoblee-io:
Thanks for the careful review — all three addressed in `f41ba5a`:

1. **host_exec abort → ssh_exec_failed**: you're right, the SSH path rejects with `Error("Aborted")` so the post-try `signal.aborted` branch was unreachable. Now both `host_exec` and `host_script` foreground check `signal.aborted` *inside* the catch and return a clean `Aborted.`. Added tests where the mocked `sshExec` **rejects** after the controller aborts (not just resolves).

2. **Scripts not hardened**: `node_script` and `host_script` foreground now run as a killable `setsid` + `timeout` session and reap the remote group over a FRESH connection on abort (`killRemoteSessionViaKubectl` / `ViaSsh`); `node_script` foreground also pins the debug pod. `node_script` background is unified onto the shared bg-session helpers (`pkill -s`) and its inline `spawn`/`randomBytes` machinery removed; `host_script` background `onAbort` DRY-ed onto the shared ssh reaper. So the "stops everything" claim now holds for the script tools too.

3. **Live host card showed the raw id**: the `tool_execution_end` live handler now recomputes `toolInput` from the arriving result metadata (`host_label`), so the card flips id → friendly name live without waiting for a refetch.

`tsc` (root + portal-web) clean; full `npm test` green (171 files / 3367 tests) + portal-web vitest (109). Re-deployed to siclaw-inner.

LikiosSedo (approved):

PR #310, opened 2026-06-09
Refine chat selection export controls

LikiosSedo:
## Summary
- replace scroll-to-select with a single sticky Select following boundary
- remove All/Clear from the selection toolbar and make the boundary toggle the current range
- export one file per download: HTML when the rendered selection contains visuals, Markdown otherwise

## Validation
- npx vitest run src/components/chat/selection-model.test.ts src/components/chat/transcript.test.ts
- npx tsc --noEmit --pretty false
- npm run build

### DIFF
--- a/portal-web/src/components/chat/PilotArea.tsx
+++ b/portal-web/src/components/chat/PilotArea.tsx
@@ -34,9 +34,8 @@ import { downloadBlob } from "./svg-export"
 import { serializeMessagesToText, serializeMessagesToMarkdown, stripVisualizationFences, stripImageData } from "./transcript"
 import {
   EMPTY_SELECTION,
-  selectVisible,
+  toggleFollowing,
   toggleMessage,
-  selectAll as selectAllMessages,
   selectedIds as computeSelectedIds,
   type SelectionState,
 } from "./selection-model"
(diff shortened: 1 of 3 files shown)


jacoblee-io (approved):
## Review — LGTM ✅

Clean, well-scoped refactor. The scroll/IntersectionObserver auto-select is fully replaced by an explicit "Select following" boundary, and the change is internally consistent: removed exports (`selectVisible`, `selectAll`), their imports, UI buttons (All/Clear), tests, and comments are all updated together. The pure model in `selection-model.ts` is correctly clamped (`selectFollowing` handles out-of-range start indices) and `toggleFollowing` clears on re-click — both covered by tests.

The download change (one file: HTML when `hasVisual`, Markdown otherwise, with a try/catch fallback to Markdown on rich-export failure) matches the PR description and reuses the existing `buildCopyHtml` `hasVisual` signal. Good.

### Non-blocking nits

1. **Redundant wrapper `<div key={msg.id}>` in the message map** (`PilotArea.tsx`). The sticky boundary is rendered once *outside* the `.map()`, so each row no longer needs an extra outer wrapper — the `key` and `data-msg-id` could sit on the same div as before. The extra nesting is harmless (`space-y-8` still applies) but appears vestigial from an earlier iteration. Worth flattening for clarity.

2. **`getBoundaryStartIndex` silent fallback to 0** (`PilotArea.tsx`). When `findIndex` returns `-1` (id not found in `selectableMessages`), it falls back to `0`, which selects *everything* rather than nothing. This shouldn't happen in practice (rows are derived from `selectableMessages`), but a stale-DOM edge case would surprise the user by selecting the whole transcript. Consider falling back to `selectableMessages.length` (select nothing) instead, matching the "no row below boundary" case.

Neither blocks merge.

PR #311, opened 2026-06-09
feat(bg-jobs): status-aware background output (task_output tool + eager file + GC)

jacoblee-io:
## Problem
Reading a background job's `output_file` blindly returned a hard ENOENT while the job had produced nothing yet (e.g. an `ib_write_bw` server backgrounded and blocked waiting for a client). The model mis-read "no output file" as the job having failed or vanished.

## Changes
- **Eager-create** the output file at launch (`DiskTaskOutput.ensureCreated`) → a read before any output yields an empty file, not ENOENT.
- **New `task_output(task_id[, tail_lines])` tool**: consults the JobRegistry (new `JobRegistry.snapshot()` + `taskOutputReader` bridge injected by the agentbox session manager and the TUI host) plus the on-disk file, reporting running/completed/failed/stopped with the output. Running → PARTIAL output + "wait for the completion notice"; re-snapshots after the read so a job that finishes mid-read reports terminal. Gated to when a background mode is on AND the reader is injected.
- **Bounded read**: `readTaskOutput` reads only the last ~2MB (tail) / ~8MB (full) from the file END (no OOM / V8 max-string), trailing-newline aware, `exists:false` for a missing file, and opens with `O_NOFOLLOW` (symmetric with the append side).
- Point the launch result, the completion notification, and the eight exec/script tool descriptions at `task_output(task_id)` instead of reading `output_file`.
- **GC**: `sweepStaleTaskOutputs` reclaims `*.output` >24h, throttled once/hour, protecting every file with a LIVE writer via a process-wide guard (registry-independent → a silent long-running job, even one owned by another session/agent sharing the dir in local mode, is never swept); guard released on settle, with a fallback release if setup throws before settle() is wired. TUI shutdown reclaims its jobs' files.

## Tests
disk-output (ensureCreated / bounded+tail read / large output / sweep + live-writer protection incl. cross-registry), task_output (running/silent/completed/failed/stopped/unknown/no-reader), setup-throw guard-release. `tsc` clean; `npm test` green (174 files, 3389).

### DIFF
--- a/src/agentbox/session.ts
+++ b/src/agentbox/session.ts
@@ -26,6 +26,7 @@ import type {
   SpawnSubagentStatus,
   JobStopExecutor,
   BackgroundExecExecutor,
+  TaskOutputReader,
   AgentMode,
 } from "../core/tool-registry.js";
 import { getSubagentType, DEFAULT_SUBAGENT_TYPE, getSubagentConcurrency, getSubagentMaxRuntimeMs, getBackgroundBashConcurrency } from "../core/subagent-registry.js";
@@ -368,6 +369,11 @@ export class AgentBoxSessionManager {
   private createJobStopExecutor(): JobStopExecutor {
     // Shared stop logic lives on JobRegistry (same as the TUI path).
     return async (jobId) => this.jobs.stopJob(jobId);
+  }
+
+  private createTaskOutputReader(): TaskOutputReader {
+    // Snapshot the job's live status so task_output can report running/terminal (same as TUI).
+    return (jobId) => this.jobs.snapshot(jobId);
   }
 
   /**
@@ -1427,6 +1433,7 @@ export class AgentBoxSessionManager {
       spawnSubagentExecutor: this.createSpawnSubagentExecutor(),
       jobStopExecutor: this.createJobStopExecutor(),
       backgroundExecExecutor: this.createBackgroundExecExecutor(),
+      taskOutputReader: this.createTaskOutputReader(),
     });
 
     // Populate sessionIdRef so skill_call events can associate with this session
--- a/src/cli-main.ts
+++ b/src/cli-main.ts
@@ -221,6 +221,7 @@ const buildSiclawOpts = (sm: SessionManager) => ({
   // stay TUI-unavailable; that needs the agentbox child-session machinery.)
   backgroundExecExecutor: tuiBackgroundHost.createBackgroundExecExecutor(),
   jobStopExecutor: tuiBackgroundHost.createJobStopExecutor(),
+  taskOutputReader: tuiBackgroundHost.createTaskOutputReader(),
 });
 
 const { brain, session, services, extensionsResult, modelFallbackMessage, customTools, skillsDirs, memoryIndexer, mcpManager } =
--- a/src/core/agent-factory.ts
+++ b/src/core/agent-factory.ts
@@ -115,6 +115,8 @@ export interface CreateSiclawSessionOpts {
   jobStopExecutor?: import("./tool-registry.js").JobStopExecutor;
   /** Runtime bridge that launches a background bash command. Injected by agentbox / TUI host. */
   backgroundExecExecutor?: import("./tool-registry.js").BackgroundExecExecutor;
+  /** Runtime bridge that reads a background job's live status. Injected by agentbox / TUI host. */
+  taskOutputReader?: import("./tool-registry.js").TaskOutputReader;
 }
 
 export interface SiclawSessionResult {
@@ -393,6 +395,7 @@ export async function createSiclawSession(
       spawnSubagentExecutor: opts?.spawnSubagentExecutor,
       jobStopExecutor: opts?.jobStopExecutor,
       backgroundExecExecutor: opts?.backgroundExecExecutor,
+      taskOutputReader: opts?.taskOutputReader,
     },
     allowedTools,
     activeMode: opts?.activeMode ?? "normal",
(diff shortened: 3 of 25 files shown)


LikiosSedo:
One issue to fix before merge:

The PR correctly moves background-output consumption to the status-aware `task_output(task_id)` path, but a couple of model-visible examples still instruct the model to read the raw `output_file` directly:

- `src/tools/cmd-exec/node-exec.ts` examples around the background `tcpdump` flow still say to `read the output_file` / `read its output_file`.
- `src/tools/cmd-exec/host-exec.ts` has the same stale wording in the background `tcpdump` example.

Those examples are in exactly the paired/open-ended capture scenarios where the model is likely to copy the instruction verbatim, bypassing the new status-aware tool, exit-code reporting, and bounded read behavior. Please update those examples to say `task_output(task_id)` instead, and consider adding a small text-regression test so future tool-description edits do not reintroduce raw `output_file` guidance.

jacoblee-io:
Fixed in `0a0b29c`. Thanks for catching it.

- `src/tools/cmd-exec/node-exec.ts` — both background `tcpdump` examples (the open-ended capture and the paired capture→traffic flow, step 3) now say `task_output(task_id)` instead of `read the/its output_file`.
- `src/tools/cmd-exec/host-exec.ts` — the background `tcpdump` example updated the same way.

Added a text-regression test `src/tools/cmd-exec/background-output-guidance.test.ts` that scans all 8 background-capable tools (node/host/pod/restricted-bash exec + node/host/pod/local script) and fails if any reintroduces `read the/its/your output_file` guidance, plus asserts each still references `task_output`. The legit references (`a task_id and output_file`, `not the raw output_file`) are unaffected — the pattern only matches "read … output_file".

`tsc` clean; `npm test` green (176 files, 3472).

LikiosSedo (commented):
## Review

Read the full diff and ran the affected tests locally. Solid, well-reasoned work — the lifecycle edge cases (guard release on all exit paths, cross-registry live-writer protection, post-read re-snapshot, bounded tail read, symmetric `O_NOFOLLOW`) are all handled carefully. Approvable; the items below are non-blocking.

### Verification
- `disk-output.test.ts`, `background-bash-runner.test.ts`, `background-launch.test.ts` → **29 tests pass** locally (covers the hard logic: tail/trailing-newline read, sweep + live-writer protection incl. cross-registry, setup-throw guard release).
- Confirmed injection symmetry: both runtimes that inject `backgroundExecExecutor` also inject `taskOutputReader` (agentbox `session.ts:1436`, TUI `cli-main.ts`), so the tool gating never tells the model to use a tool that isn't registered.
- Confirmed `JobRegistry.delete()` has no callers, so a finished job stays in the registry → the post-read re-snapshot in `task_output` correctly reports a job that finished mid-read as terminal (not a false `found:false` → stale `running`).

### Findings (non-blocking)

**1. [low] Consistency: 3 spots still point the model at `output_file` instead of `task_output(task_id)`** — the stated motivation is to stop the model from blindly reading `output_file`, but these were missed:
- `src/tools/cmd-exec/host-exec.ts:83` — tcpdump example: "…stop it later with job_stop, then **read the output_file**"
- `src/tools/cmd-exec/node-exec.ts:102` — same tcpdump example
- `src/tools/cmd-exec/node-exec.ts:106` — "…then **read its output_file**"
- `src/tools/cmd-exec/restricted-bash.ts:409` — stale code comment: "The model reads progress via `read` on output_file" (internal, lowest priority)

These example lines now contradict the updated main descriptions. Worth a quick sweep for consistency.

**2. [edge/optional] Newline-less large output reads back empty** — in `readTaskOutput`, when `truncated` is set, the (possibly partial) first line is dropped via `indexOf("\n")`. If the read window (last 2MB tail / 8MB full) contains **no** newline at all — e.g. a pathological `base64 /dev/urandom`-style stream, which the writer force-flushes in 1MB chunks *without* inserting newlines — `nl === -1` collapses `content` to `""`. So a large newline-less output reports `truncated:true` with empty `output`, even for `tail_lines:0`. Real command output virtually always has newlines, so impact is low, but strictly it's full data loss on read for that case. Could fall back to returning the raw last-cap bytes when no newline is found.

Nice work overall.

jacoblee-io:
Thanks for the thorough pass. Both follow-ups addressed in `ed7ee52`:

**1. [low] consistency** — the 3 example lines (host-exec:83, node-exec:102/106) were already fixed in `0a0b29c`. Now also fixed the remaining stale **code comment** at `restricted-bash.ts:409` ("reads progress via `read` on output_file" → "reads progress via task_output(task_id)"). Sweep is clean — no `read … output_file` guidance left in any of the 8 background tools.

**2. [edge] newline-less large output** — fixed in `readTaskOutput`: when the read window contains no newline at all, we now keep the raw last-cap bytes instead of collapsing to `""` (only drop the partial first line when a newline actually exists). Added a regression test (`disk-output.test.ts` → a ~3MB newline-less blob now reads back the raw tail, `truncated:true`, non-empty).

`tsc` clean; `npm test` green (176 files, 3473).

LikiosSedo (commented):
Re-reviewed at `ed7ee52`. Both points addressed — LGTM. ✅

- **Doc consistency**: all 4 spots now point at `task_output(task_id)` (host-exec tcpdump example, node-exec ×2, restricted-bash comment). `git grep` confirms no `read the output_file` residue left.
- **Newline-less large output**: `readTaskOutput` now keeps the raw last-cap bytes when the window has no `\n` instead of collapsing to `""`. Verified the tail branch is also correct (single-element split → returned as-is with `truncated:true`), so no empty-output data loss for the pathological blob case.

Nice, thanks for the quick turnaround.

LikiosSedo (approved):

PR #312, opened 2026-06-09
Guard model fallback replay and compact preflight

LikiosSedo:
## Summary
- prevent automatic model-route prompt replay after a failed attempt has already executed tools
- add fallback context preflight so smaller-context candidates compact before prompting and fail closed when they still cannot fit
- cover the routing, PiAgentBrain, and AgentBox HTTP paths with focused tests

## Why
When a fallback attempt fails after tool execution, replaying the same user prompt on another model can duplicate external side effects. Also, if 1M-context models are unavailable and the route falls back to a 256K/128K model, the runtime should compact proactively instead of blindly hitting context overflow.

## Validation
- `npx vitest run src/core/brains/pi-agent-brain.test.ts src/core/model-routing.test.ts src/agentbox/http-server.test.ts`
- `npx vitest run src/core/compaction.test.ts src/core/extensions/compaction-safeguard.test.ts`
- `npx tsc -p tsconfig.agentbox.json`
- `npx tsc --noEmit --pretty false`
- `git diff --check`

### DIFF
--- a/src/agentbox/http-server.test.ts
+++ b/src/agentbox/http-server.test.ts
@@ -431,6 +431,63 @@ describe("http-server — model routing", () => {
     expect(sm.persistModelRouteState).toHaveBeenCalledWith("route-fallback", s.modelRouteState);
   });
 
+  it("does not fallback by replaying the prompt after a tool has executed", async () => {
+    const s = await sm.getOrCreate("route-tool-side-effect");
+    const seenModels: string[] = [];
+    s.brain.prompt.mockImplementation(async () => {
+      const model = s.brain.getModel();
+      seenModels.push(`${model.provider}/${model.id}`);
+      s.brain.emitter.emit("event", {
+        type: "tool_execution_start",
+        toolCallId: "call_1",
+        toolName: "read",
+        args: { path: "README.md" },
+      });
+      s.brain.emitter.emit("event", {
+        type: "tool_execution_end",
+        toolCallId: "call_1",
+        toolName: "read",
+        result: { content: [{ type: "text", text: "ok" }] },
+        isError: false,
+      });
+      s.brain.emitter.emit("event", {
+        type: "message_end",
+        message: {
+          role: "assistant",
+          content: [],
+          stopReason: "error",
+          errorMessage: "429 rate limit exceeded",
+        },
+      });
+    });
+
+    const r = await getJson(port, "/api/prompt", "POST", {
+      text: "route me after tool",
+      sessionId: "route-tool-side-effect",
+      modelRouting: routePolicy,
+    });
+    await flushAsync();
+
+    expect(r.status).toBe(200);
+    expect(seenModels).toEqual(["openai/gpt-4"]);
+    expect(s.modelRouteState.activeCandidateKey).toBeUndefined();
+    expect(s.modelRouteState.cooldowns["openai/gpt-4"]).toBeGreaterThan(0);
+    expect(s.modelRouteState.attempts[0]).toMatchObject({
+      failureKind: "rate_limit",
+      fallbackBlockedReason: "tool_execution",
+    });
+    expect(s._extraEventBuffer.some((event) => event.type === "model_route_switch")).toBe(false);
+    expect(s._extraEventBuffer.find((event) => event.type === "model_route_exhausted")).toMatchObject({
+      failureKind: "rate_limit",
+      fallbackBlockedReason: "tool_execution",
+    });
+    expect(s._extraEventBuffer.some((event) => event.type === "tool_execution_end")).toBe(true);
+    expect(s._extraEventBuffer.some((event) =>
+      event.type === "message_end" && (event.message as any)?.stopReason === "error",
+    )).toBe(true);
+    expect(sm.persistModelRouteState).toHaveBeenCalledWith("route-tool-side-effect", s.modelRouteState);
+  });
+
   it("does not fallback on context overflow", async () => {
     const s = await sm.getOrCreate("route-context");
     const seenModels: string[] = [];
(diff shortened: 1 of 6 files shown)


LikiosSedo (commented):
## Review

Read the full diff and verified locally on the branch (`npx tsc --noEmit` clean; `model-routing` + `pi-agent-brain` + `http-server` suites = **111 tests pass**). Both mechanisms are well-built and well-tested. The `failureSource` distinction — a *preflight* `context_overflow` (`source: "setup"`) tries the next candidate, while a *runtime* `context_overflow` keeps the default no-fallback behavior — is a nice touch. Approvable; notes below are non-blocking.

### What I verified
- `session.compact()` and `session.settingsManager.getCompactionSettings()` are **new** calls on `AgentSession` not previously used by `PiAgentBrain` — `tsc` confirms the type contract holds against `pi-coding-agent@0.78.1`.
- Side-effect guard wiring is complete end-to-end: `hadToolExecution` → `fallbackBlockedReason` → no switch, cooldown still set, and brain events still flushed so the user sees the executed tool + error (not a silent drop). Covered by both the routing unit test and the http-server integration test.
- `context_overflow` has `DEFAULT_COOLDOWN = 0` and is in `DEFAULT_NO_FALLBACK_ON`, so a preflight-overflow candidate is skipped *without* a cooldown — correct, since waiting doesn't shrink history.

### Findings (non-blocking)

**1. [design / low] The replay guard blocks fallback after *any* tool, including read-only ones.** A read/grep/ls that ran before a transient `rate_limit` will now exhaust the route instead of retrying on a healthy fallback model, even though replaying an idempotent read is harmless. This is the safe default (can't know which tools have side effects, and `crash > duplicate side effect`), so it's the right v1 — just flagging the resilience cost. A future refinement could be a side-effect-free tool allowlist that still permits fallback.

**2. [efficiency / low] Preflight compaction is discarded on fallback.** `createPromptCheckpoint()` captures the leaf id at the *start* of `runAttempt`, before the preflight runs `compact()`. On a genuine fallback (no tool executed, e.g. plain `rate_limit`), `restorePromptCheckpoint` → `sessionManager.branch(oldLeaf)` branches back to *before* the compaction, so the summary work is dropped and the next candidate re-compacts from the full history. Correctness is fine (and per-candidate windows differ, so some re-compaction is justified), but the compact-succeeds-then-transient-failure case burns an extra compaction LLM call per candidate. Could capture/restore the checkpoint around the prompt only, or skip restore when the kept history is still valid for the next candidate. Worth confirming this matches your intent.

Solid work overall.

jacoblee-io:
## Review

This is a well-targeted hardening of the merged model-routing, and it closes the main open concern from the #308 review (replaying a prompt on a fallback model after side-effectful tools already ran). The logic is sound and the test coverage is strong (+155 model-routing, +88 pi-agent-brain, +57 http-server). Approvable.

**What it does (correct):**
- **Fallback-replay guard** — `runAttempt` tracks `hadToolExecution` (any `tool_execution_start/end` or a `toolResult` message_end). If a tool ran before the attempt failed, it sets `fallbackBlockedReason="tool_execution"` and blocks fallback, so a fallback model never replays a prompt whose tools already had side effects. Pre-tool failures (the common overload/billing/rate-limit case) still fall over normally.
- **Compact preflight** — `ensureContextForModelPrompt(model, text)` estimates `currentContext + prompt` vs `contextWindow - reserve`: fits → no-op; over-budget + compaction enabled → compact and re-check; over-budget + disabled → a clean `context_overflow/setup` failure. With `shouldTryNextCandidateForFailure`, a too-small fallback is skipped to the next candidate rather than hard-failing.

**Minor / notes (none blocking):**

1. **Fallback is blocked on *any* tool, including read-only ones** — `isToolExecutionEvent` doesn't distinguish side-effectful from read-only tools, so a primary failure right after e.g. a `kubectl get` loses fallback resilience. Safe/conservative default (classifying side-effect-freeness is hard), but worth a one-line comment so it reads as deliberate.

2. **Cooldown applied to the primary on a tool-blocked failure** — `if (attempt.fallbackBlockedReason && nextCandidate) { … state.cooldowns[key] … }`: if the post-tool failure kind is cooldown-worthy, a healthy primary gets cooled down and the next turn starts on the fallback. Defensible (mirrors the kind-based cooldown), just flagging the side effect.

3. **Preflight token estimate leans on internals** — `estimateCurrentContextTokens` reads `session.getContextUsage()` (last-turn usage) and falls back to `session.agent.state.messages`. It's a heuristic and reaches into pi-agent internal state (same fragility class as the `onResponse`/checkpoint casts), so a future pi-agent bump could silently drift it; a thin accessor + a test would harden it.

4. **Preflight runs on the primary too** for routing-enabled sessions, so they now get preemptive compaction when the primary would overflow (a behavior change, by design). Compaction events emitted during preflight are buffered into the attempt and replayed on success — likely the intended "compacted for fallback" UI, just noting.

Verdict: correct, well-tested, closes the #308 replay/side-effect gap. The four items are optional polish/comments, not blockers.

jacoblee-io (approved):
Approving. This closes the main open concern from the #308 review — replaying a prompt on a fallback model after side-effectful tools already ran — via the `hadToolExecution` fallback guard, and adds a sensible compact-preflight for smaller-context fallbacks. Logic is sound and test coverage is strong (model-routing / pi-agent-brain / http-server).

The four items from my comment are optional polish, not blockers:
- blocking fallback on any tool (incl. read-only) is a safe-but-conservative default — worth a one-line comment;
- primary cooldown on a tool-blocked failure can start the next turn on the fallback (defensible);
- the preflight token estimate leans on pi-agent internals (`getContextUsage` / `agent.state.messages`) — a thin accessor + test would harden it;
- preflight now runs on the primary for routing-enabled sessions (preemptive compaction, by design).

Please confirm CI is green before merge (reviewed statically). LGTM. 🚀

PR #313, opened 2026-06-09
Remember chat agent and session selection

LikiosSedo:
## Summary

- remember the last selected chat agent and each agent's last selected session in localStorage
- restore agent/session from URL params or stored selection when Chat mounts or receives a kept-alive route update
- point sidebar Chat and Agents-page Chat navigation at the remembered chat path

## Why

Returning to Chat from agent or skill configuration currently falls back to a default agent/session because only the current component state and the agent URL param are considered. This preserves the user's last Chat context across navigation and refreshes.

## Validation

- npm test -- src/lib/chatSelection.test.ts
- npm run build

## Test Deployment

- Built and pushed portal-only image: registry-cn-shanghai.siflow.cn/k8s/siclaw-portal:chat-selection-binding-20260609162343-1427357
- Image digest: sha256:29990267bf4a8ea35f05b1be2560658ff2b3f9ab90887313368b9b6fb274bfa7
- Deployed to sdliu-siclaw deploy/siclaw-portal
- Verified rollout and /api/health

### DIFF
--- a/portal-web/src/components/AgentChat.tsx
+++ b/portal-web/src/components/AgentChat.tsx
@@ -136,9 +136,11 @@ function SessionSidebar({
 
 interface AgentChatProps {
   agentId: string
+  selectedSessionId?: string | null
+  onSessionChange?: (sessionId: string | null) => void
 }
 
-export function AgentChat({ agentId }: AgentChatProps) {
+export function AgentChat({ agentId, selectedSessionId, onSessionChange }: AgentChatProps) {
   const toast = useToast()
   const confirmDialog = useConfirm()
   const [showSessions, setShowSessions] = useState(false)
@@ -146,6 +148,8 @@ export function AgentChat({ agentId }: AgentChatProps) {
   const [sessions, setSessions] = useState<ChatSession[]>([])
   const [activeSessionId, setActiveSessionId] = useState<string | null>(null)
   const [loading, setLoading] = useState(true)
+  const activeSessionIdRef = useRef<string | null>(null)
+  const selectedSessionIdRef = useRef<string | null | undefined>(selectedSessionId)
 
   // Panel state
   const [skillPanelMsg, setSkillPanelMsg] = useState<PilotMessage | null>(null)
@@ -155,6 +159,20 @@ export function AgentChat({ agentId }: AgentChatProps) {
 
   // Auto-title: track whether we already titled this session
   const titledSessionRef = useRef<string | null>(null)
+
+  const selectSession = useCallback((sessionId: string | null) => {
+    activeSessionIdRef.current = sessionId
+    setActiveSessionId(sessionId)
+    onSessionChange?.(sessionId)
+  }, [onSessionChange])
+
+  useEffect(() => {
+    activeSessionIdRef.current = activeSessionId
+  }, [activeSessionId])
+
+  useEffect(() => {
+    selectedSessionIdRef.current = selectedSessionId
+  }, [selectedSessionId])
 
   // Pilot-style chat hook
   const pilot = usePilotChat({ agentId, sessionId: activeSessionId })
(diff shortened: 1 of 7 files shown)


jacoblee-io (commented):
## Review

Solid change overall. `chatSelection.ts` is a clean, well-tested pure module (SSR-guarded, input-normalized, custom same-tab change event so the sidebar updates without a reload), and the ref-based reconciliation in `AgentChat` converges correctly — no infinite-loop risk. One blocking issue and one minor cleanup.

### 🔴 Blocking: chat pane flashes to a spinner on every toast

In `AgentChat.tsx`, the `fetchSessions` effect now depends on `toast`:

```ts
setLoading(true)            // ~L205
...
}, [agentId, selectSession, toast])   // ~L236
```

and the component is gated by `if (loading) return <Loader/>`.

`useToast()` is **not referentially stable** — `ToastProvider` rebuilds its value object literal on every render, and re-renders whenever a toast is added *or auto-dismissed* (the 4s timer):

```ts
// toast.tsx — fresh object every render
const value: ToastContextValue = { success: ..., error: ..., info: ... }
```

So: any toast fires → `useToast` ref changes → `AgentChat` re-renders → effect dep `toast` changed → effect re-runs → `setLoading(true)` → the whole chat pane is replaced by the loader → re-`GET /chat/sessions` → `setLoading(false)`. This triggers **twice per toast** (on show + on 4s auto-dismiss), including this feature's own "Session created/deleted" toasts. `selectSession` is stable within a mount (agent fixed via `key`), so `toast` is the sole offender.

**Suggested fix — memoize the provider value at the source (helps every consumer):**

```ts
// toast.tsx
const value = useMemo(() => ({
  success: (m: string) => add(m, "success"),
  error:   (m: string) => add(m, "error"),
  info:    (m: string) => add(m, "info"),
}), [add])
```

(`add` is already `useCallback([])`.) Alternatively, keep `toast` out of the effect deps via a ref — but fixing it at the source is cleaner.

### 🟡 Minor: dead branch in `handleSessionChange` (Chat.tsx)

```ts
if (sessionId) rememberChatSession(selectedAgentId, sessionId)
else           rememberChatSession(selectedAgentId, null)   // sessionId is already null here
```

Collapses to a single `rememberChatSession(selectedAgentId, sessionId)`.

### 🟢 Verified / no action needed

- `<Chat/>` is always mounted in `Layout`, so Chat's init `/agents` effect runs on every route. The `location.pathname.startsWith("/chat")` guards are therefore load-bearing and are all present/correct. Background `rememberChat*` writes on non-chat routes are guarded (prefer stored `lastAgentId`, won't clobber) and are what make the sidebar link meaningful — looks intended.
- `usePilotChat` already keys its message cache by `sessionId` and refetches on change, so driving it via the new `selectedSessionId` prop is safe.
- The reconciliation effect converges; stale/deleted stored session IDs correctly fall back to `sessions[0]` and rewrite storage + URL.

Just the toast-driven flicker to address before merge.

jacoblee-io (approved):
## Re-review — LGTM ✅

Both fixes from the previous round are verified correct:

- **Toast flicker fixed.** `ToastProvider`'s value is now `useMemo(..., [add])` and `add` is `useCallback([])`, so the context reference is stable — the `fetchSessions` effect no longer re-runs (and no longer flips the chat pane to a spinner) when toasts appear/dismiss.
- **Dead branch removed.** `handleSessionChange` collapsed to a single `rememberChatSession(selectedAgentId, sessionId)`.

I also re-checked the edge cases I was worried about, and they all hold up — no remaining blocking issues:

- **Deleting the active session** → `selectSession(nextSessions[0] || null)` round-trips through the parent; the reconciliation effect hits the "already equal" branch and no-ops. Converges, no loop.
- **Stale remembered agent/session (deleted on the backend)** → no transient "Failed to load sessions" toast, because `Chat`'s `if (loading) return <spinner>` gates the `<AgentChat>` mount until `/agents` has resolved and `nextAgentId` is validated (falls back to `list[0]` when invalid).
- **Stale session deep-link** → `AgentChat`'s reconciliation effect falls back to `sessions[0]` once the real list loads and rewrites storage + URL. Self-heals.
- **URL ↔ state** → an `AgentChat`-driven session change updates the URL, which re-runs the URL-sync effect, but the `setSelected*` calls are "equal → unchanged", so it's just one idempotent `rememberChatSession` + event dispatch. No render loop.

### 🟢 Two optional nits (non-blocking, fine to ignore)

1. That round-trip does one redundant `rememberChatSession` write + `CHAT_SELECTION_CHANGE_EVENT` dispatch (→ one extra `setChatPath` in `Layout`). Harmless.
2. `Layout`'s chatPath effect deps `[location.pathname, location.search]` re-register the listeners on every navigation; since the custom event + `storage` event already cover updates, the deps could just be `[]`.

Approving.

PR #315, opened 2026-06-10
fix: harden model fallback routing edge cases (#308 follow-up)

LikiosSedo:
## Summary

Follow-up hardening for the model fallback routing layer introduced in #308 (and patched in #312). A post-merge review of the feature surfaced a set of confirmed edge-case defects in the runner, the AgentBox plumbing, and the chat UI metadata path. This PR fixes the ones that are small, low-risk, and behavior-correct without product decisions.

Larger known gaps are intentionally **out of scope** (see below).

### Core runner (`src/core/model-routing.ts`)

- **Cooling candidates are deprioritized, not excluded.** Previously a candidate in cooldown was dropped from the run entirely: with `[A, B]` where B was cooling, an A failure exhausted the turn without ever trying B. Cooling candidates now sort to the end and are tried as a last resort.
- **Cooldowns now reflect terminal failures and successes.** A failure on the last candidate (or a tool-blocked / no-fallback failure) records its cooldown so the next turn's ordering sees it; a success clears that candidate's cooldown (it could previously stay "cooling" after serving a reply in the all-cooling path).
- **A user Stop between attempts halts the chain.** The abort flag was only consulted while `brain.prompt` was in flight; a Stop landing in the switch window (cooldown persist / checkpoint restore / setModel are all awaited) was swallowed and the cancelled prompt re-ran on the next candidate. The runner now polls a `shouldAbort` hook between attempts; both AgentBox prompt paths wire it to the session abort flag.
- **Message-end `aborted` results go through the classifier.** The `message_end` path hardcoded `user_abort`, bypassing the transport-vs-user abort distinction the classifier already implements — a transport-level abort surfacing as `stopReason: "aborted"` could not fall back. Genuine user stops still never fall back.
- **A manual model pin landing mid-run is preserved.** `PUT /api/sessions/:id/model` during an in-flight routed turn was silently clobbered by the runner's success handler (`activeCandidateSource` reset to `auto`). The success path now respects a user pin.
- **`content_policy` classification anchors on real provider phrasing.** Bare `policy` / `blocked` / `safety` matched unrelated errors ("blocked by upstream proxy", "retry policy exceeded") and mislabeled them as a no-fallback kind, intercepting kinds that would otherwise classify as fallback-worthy further down the chain.
- Drop dead `isModelRouteFailureKind`.

### AgentBox plumbing (`src/agentbox/session.ts`, `http-server.ts`)

- **Routed synthetic turns no longer leak raw brain events.** The synthetic prompt path computed `routeEnabled` but never set `_routeBrainEventsThroughExtra`, so an SSE client connected during a routed synthetic turn streamed every failed attempt's raw events live — the exact leak the runner's buffering exists to prevent. The flag is now set/reset exactly like the HTTP `/prompt` path (regression-tested).
- **Route-state persists are serialized per session.** Each write was atomic (tmp + rename) but writes were unordered, so a slow older write could rename over a newer one, leaving stale state on disk.
- **Unreadable state files warn instead of silently resetting.** Corrupt JSON / permission errors on `.model-route-state.json` now leave a trace; `ENOENT` (normal first run) stays quiet.

### Portal-web (`usePilotChat.ts`, `PilotArea.tsx`)

- **Model route indicator is now deterministic across reload.** The live path set route metadata on every routed reply while the gateway only persists it for fallback/recovery replies (`sse-consumer.ts`), so the model label vanished from normal routed replies after a refresh. The live predicate now matches the persistence predicate.
- **Remove the unreachable `ModelRouteNotice` renderer** and its now-unused helpers. `isVisibleChatMessage` has filtered `model_route_notice` rows out of the render list since `1a8f6f5`, so the component could never mount (~60 lines of dead UI).

## Known gaps left out of scope

These are real but need product/design decisions rather than spot fixes:

- Routed turns deliver the winning attempt as one burst (no live token streaming), which also skews persisted per-tool timing metadata.
- Persisted `model_route_notice` audit rows are never surfaced in any UI.
- Spawned sub-agents and the first post-restart synthetic turn bypass routing.
- `model_route_exhausted` is not consumed end-to-end.

## Validation

- `npm test` — 176 files, 3489 passed / 2 skipped
- `npm run build`, `npx tsc -p tsconfig.agentbox.json`
- `npm run build:web`, portal-web `vitest run` — 12 files, 113 passed
- New coverage: 7 runner unit tests (cooling last-resort, terminal cooldown, success clears cooldown, mid-run pin, inter-attempt abort, transport-vs-user abort, classifier narrowing) + synthetic-turn SSE suppression assertions in `notify-parent.test.ts`

### DIFF
--- a/portal-web/src/components/chat/PilotArea.tsx
+++ b/portal-web/src/components/chat/PilotArea.tsx
@@ -124,14 +124,6 @@ function ModelTimeLabel({ timing }: { timing: MessageTiming | undefined }) {
   )
 }
 
-function stringMeta(value: unknown): string | undefined {
-  return typeof value === "string" && value.length > 0 ? value : undefined
-}
-
-function numberMeta(value: unknown): number | undefined {
-  return typeof value === "number" && Number.isFinite(value) ? value : undefined
-}
-
 function modelRouteMetadata(message: PilotMessage): ModelRouteMetadata | null {
   const route = message.metadata?.model_route
   if (!route || typeof route !== "object" || Array.isArray(route)) return null
@@ -146,22 +138,10 @@ function isVisibleChatMessage(message: PilotMessage): boolean {
   return !message.hidden && !isModelRouteNoticeMessage(message)
 }
 
-function routeModelLabel(provider?: string, modelId?: string): string {
-  return provider && modelId ? `${provider}/${modelId}` : provider || modelId || "unknown"
-}
-
 function routeModelDisplayName(modelId?: string, provider?: string): string {
   if (!modelId) return provider || "unknown"
   const parts = modelId.split("/").filter(Boolean)
   return parts[parts.length - 1] || modelId
-}
-
-function formatCooldown(until: unknown): string | undefined {
-  const value = numberMeta(until)
-  if (value == null) return undefined
-  const date = new Date(value)
-  if (Number.isNaN(date.getTime())) return undefined
-  return date.toLocaleTimeString([], { hour: "2-digit", minute: "2-digit" })
 }
 
 function ModelRouteIndicator({ route }: { route: ModelRouteMetadata | null }) {
@@ -1502,10 +1482,6 @@ function MessageItem({
     return <DelegationStatusNotice content={message.content} />
   }
 
-  if (message.metadata?.kind === "model_route_notice") {
-    return <ModelRouteNotice message={message} />
-  }
-
   if (isTool) {
     if (message.toolName === "delegate_to_agents") {
       return <AgentWorkBatchCard message={message} />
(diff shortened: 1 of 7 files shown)


jacoblee-io (approved):
## Review

**Verdict: Approve.** Every defect claimed in the PR description is real (each one traced back in the pre-PR code), every fix carries a targeted regression test, and all relevant suites pass locally (67 model-routing/notify-parent tests, 34 session tests, 12 portal-web hook tests). The PR body is accurate, including its honest "known gaps" section.

### Verified point by point

- **Cooling candidates deprioritized, not excluded** (`model-routing.ts:493-499`): the old logic dropped cooling candidates from the turn entirely; the new ordering tries fresh candidates first and appends cooling ones. The all-cooling case degenerates to the original order, so behavior there is unchanged.
- **Cooldown recorded on terminal failure / cleared on success**: the old code only wrote cooldowns inside the switch path, so the last candidate's failure left no trace and a candidate that succeeded while cooling stayed "cooling". The new `willSwitch` logic matches the inline comments exactly (a zero-cooldown terminal failure such as user_abort keeps what it had).
- **Inter-attempt abort poll**: the `shouldAbort` check is correctly placed (`setModel` happens inside `runAttempt`, and the previous iteration's checkpoint restore has completed — nothing is left half-switched). Both wiring sites are safe against stale flags: the HTTP path resets `_aborted = false` at http-server.ts:385 and the synthetic path at session.ts:639, both at turn start.
- **`message_end` abort routed through the classifier**: the combined signal text is lowercased upstream (the new case-sensitive regex is safe); `transportAbortPattern` tests the *message* rather than the combined string containing the literal `stopReason: "aborted"`, so no false matches; a bare abort with no message still defaults to `user_abort` — the safe direction (never auto-fallback on a possible user stop).
- **`content_policy` regex narrowing**: the previously mislabeled examples ("blocked by upstream proxy" → `network`, "retry policy exceeded" → `unknown`) now flow to fallback-capable kinds; tests cover both directions.
- **Synthetic-turn SSE suppression**: `_routeBrainEventsThroughExtra` is set/reset exactly like the HTTP path (set before the runner, reset in `finally`); the new test asserts the flag is true during *both* attempts and reset after.
- **Persist serialization**: the payload is snapshotted synchronously before chaining, errors are caught inside the chain (one failed write can't wedge the queue), and the map entry is cleaned up only if it is still the tail. No leak, no deadlock.
- **Portal-web alignment**: the hook's new predicate (`is_fallback || recovered_from_candidate_key`) exactly mirrors the gateway's persistence predicate (`sse-consumer.ts:384-387`), and the field names match the snake_case `ModelRouteMetadata` shape. The deleted `ModelRouteNotice` was genuinely unreachable: `isVisibleChatMessage` filters `model_route_notice` rows before `MessageItem` ever renders.

### Two non-blocking nits

1. **Theoretical race in the two-pass cooling filter** (`model-routing.ts:497-499`): `freshCandidates` and `coolingCandidates` each call `now()` per candidate at slightly different times. A candidate whose cooldown expires between the two passes is cooling in the first filter and not cooling in the second — and silently vanishes from `ordered`. Vanishingly unlikely with real clocks, but the fix is free: capture `const t = now()` once and pass it to both filters (also cleaner for test-injected `now`).
2. **A user Stop reuses the `model_route_exhausted` event** (`model-routing.ts:522-528`) while the result says `exhausted: false`. Harmless today (the event isn't consumed end-to-end — the PR itself lists this as a known gap), but once a consumer lands, a user Stop would render as "all models exhausted". Consider a dedicated `model_route_aborted` event (or an `aborted: true` field) in a follow-up.

One observation, not a request: the persist chain serializes every write even when a burst of `onStateChange` calls makes intermediate payloads obsolete. Coalescing to "latest pending payload wins" would cut IO, but write volume here is tiny (a few per turn), so correctness-first is the right call.

LikiosSedo:
Addressed both review nits in 20fd702:

1. **Cooldown partition race** — fresh/cooling candidates are now partitioned against a single `now()` timestamp (also cleaner semantics for an injected `now`).
2. **User Stop event semantics** — the between-attempts Stop now emits a dedicated `model_route_aborted` event instead of reusing `model_route_exhausted`. Since no consumer of either event exists yet, the union change is purely additive; the regression test asserts no exhausted event is emitted for a manual stop.

The persist-chain coalescing observation is left as-is per your note (correctness over IO at this write volume).

Validation: routing suites green (101 tests), `npm run build` + `tsc -p tsconfig.agentbox.json` clean.

PR #317, opened 2026-06-10
feat(cluster): consolidate cluster_info into cluster_list (search + structured metadata)

jacoblee-io:
## Summary

`cluster_info` and `cluster_list` overlapped — same registry, and `cluster_info`'s output was a subset of `cluster_list`'s. This **merges them into one tool** and, in the same pass, **wires through the structured cluster metadata** admins maintain in sicore (RDMA type, scheduler, node model, …) which previously never reached the agent — only the free-text description did.

### Problem

Two issues, one surface:
1. **Redundant tools.** `cluster_info` returned `{name, infra_context(=description), meta}` — a strict subset of `cluster_list`, differing only by a `name` filter. Two fuzzy-bounded tools make the LLM choose needlessly.
2. **Lost metadata.** The structured metadata (`resource_metadata` × `metadata_keys`, the data behind virtual-grouping) is where admins actually maintain RDMA/scheduler/node facts, but the siclaw adapter channel dropped it — `ClusterMeta` had no field and no tool emitted it.

### Solution

**Tool surface**
- Remove `cluster_info`; `cluster_list` is the single source of cluster info.
- `cluster_list` gains a `name` param (case-insensitive substring search, e.g. `cluster_list changliu`) and emits `meta` (admin metadata flattened to `display_name → value`).
- Distinct hints for *search-miss* vs *unbound*.

**Metadata passthrough** (sicore `docs/design/siclaw-cluster-metadata-passthrough.md` §6)
- `credential-types.ts`: `ClusterMetaEntry` + optional `ClusterMeta.meta`.
- `credential-broker.ts`: `inferClusterMetaFromResponse` parses `metadata.meta` on the get path, filtering malformed entries; list path carried as-is.
- `cluster-meta.ts`: shared `flattenClusterMeta()` — the single output sink for both paths, defensively dropping malformed entries so the unvalidated list path can't leak stray keys.

`prompt.ts` + `tools.md` updated (`cluster_info` → `cluster_list`; also dropped the already-removed `knowledge_search` from the platform-exemption list).

**Compatibility.** The metadata side is additive — when sicore sends no `meta`, output is byte-identical to today. The tool removal is a behavior change (one fewer tool); end-to-end metadata requires sicore §5, but this side is safe to merge first.

## Test Plan

- [x] `cluster-list.test.ts` — name search (case-insensitive), search-miss vs unbound hints, flattened `meta`
- [x] `cluster-meta.test.ts` — flatten + key fallback + defensive malformed-entry filtering
- [x] `credential-broker.test.ts` — `metadata.meta` parsed on `credential.get`, malformed filtered; absent when none
- [x] `cluster_info` removed cleanly (no remaining refs); `npx tsc --noEmit` clean; full `src/` regression green (3520 passed)
- [ ] Manual end-to-end once sicore §5 lands: fill metadata on a cluster → `cluster_list` shows it; `registry` system key excluded; edit reflects without agentbox restart


LikiosSedo:
Thanks for wiring the Siclaw side of this through. The main path looks right to me, but I think this needs one contract cleanup before merge:

1. `flattenClusterMeta()` currently accepts any typed `{ key, value }` entry, so if upstream passes the existing system metadata key `registry`, `cluster_info` / `cluster_list` will expose it to the model. The PR body lists “system key `registry` excluded” as part of the intended E2E behavior, but the current Siclaw output sink does not enforce that. Since this helper is the final model-visible boundary, please either add a Siclaw-side denylist/test for `registry` (or future system keys), or make the upstream-only guarantee explicit with cross-repo test evidence.

2. The output uses `display_name` as the object key. Sicore only guarantees `(org_id, key)` uniqueness; `display_name` is editable display text and can collide. `Object.fromEntries()` will silently drop one value when two keys share the same display name. Please prefer the stable `key` as the map key, or preserve entries as an array / nested `{ key: { display_name, value } }` shape.

The build and touched Siclaw tests pass locally for me (`npm run build`; 45 touched tests), so this is about the cross-repo output contract rather than a mechanical breakage.

jacoblee-io:
Both are valid — thanks, especially (2), which I'd under-weighted in my own review (I'd treated the collision as a data-entry edge case rather than a structural one). Fixed in `flattenClusterMeta` (the single model-visible sink), force-pushed:

**(2) `display_name` collision** — now keyed by the stable `key` (sicore's `(org_id, key)` uniqueness guarantee) instead of the editable, non-unique `display_name`. Two entries sharing a display name no longer clobber each other. `display_name` stays in `ClusterMetaEntry` (faithful to the wire shape) but is no longer the map key. Output is now `{ rdma_type: "SR-IOV", scheduler: "volcano" }`; tool description and tests updated accordingly.

**(1) `registry` system key** — added a Siclaw-side `RESERVED_SYSTEM_KEYS` denylist enforced in the same sink, with a comment noting it mirrors sicore's `IsReservedSystemKey` and is defense-in-depth (sicore still filters upstream; this guarantees nothing reserved escapes the model-visible boundary even if a future key is missed upstream).

New tests cover both: no-collision on shared `display_name`, and `registry` dropped (incl. the all-reserved → no `meta` key case). Full `src/` suite green (3521 passed).

LikiosSedo (approved):
Looks good to me. The latest force-push addresses the two contract issues I raised: model-visible metadata is keyed by stable metadata key rather than display_name, and reserved system keys such as registry are filtered at the Siclaw output boundary. Build/tests are green.

PR #322, opened 2026-06-12
fix(model-routing): visible failures, reload-safe persistence, and live primary streaming

LikiosSedo:
Makes model-routing (fallback) turns behave like non-routed turns end to end: failures are visible, they survive a reload, and the happy path streams live instead of buffering. Three stacked pieces (all on the routing runner + its consumers).

## 1. Visible failures + stream after first tool (original)

- **Synthesize a terminal error event on setup-failure exhaustion** (`model-routing.ts`): when every candidate fails during attempt *setup* (context preflight, `model_not_found`, `setModel` throw) the runner emitted zero brain events and *returned* (only `prompt_error` rejects). The chat client rendered an empty turn — no answer, no error bubble, logged as "completed". Now it emits an assistant `message_end` with `stopReason:"error"`, the exact shape clients build error bubbles from. No frontend change needed.
- **Stream live once the first tool executes**: buffering past the first tool call bought no atomicity (fallback is blocked there, #312) and only killed streaming. Flush the buffered prefix and go live from that point.
- Log routed turns that finish unsuccessfully as such (not "Prompt completed"); count extra-channel events in the SSE replay log.

## 2. Failed turns survive a reload

The live error bubble is client-only (`role:"error"`, no DB row), so reloading history dropped it — a failed turn reappeared as just the user message. `sse-consumer.ts` now persists a failed turn as `role:"assistant"` + `metadata.kind:"error_response"` (the DB role enum is user|assistant|tool). **Paired sicore MR** renders that row back as an error bubble on reload and dedups it against the live-only bubble.

## 3. The primary candidate streams live (fallback UX == no-fallback UX)

Routing buffered every brain event until an attempt won, so a routed turn had a slow first token then one-shot rendering — even when the primary succeeded (the common case). But almost every fallback trigger (4xx/5xx/auth/model_not_found/overloaded) fires *before* the first token, where there's nothing to hide.

- **Stage 1 — optimistic live primary** (`model-routing.ts`): the primary candidate streams live from its first event; fallback candidates keep buffering (until their own first tool call). A live primary that fails emits a new `model_route_rollback` before the switch so consumers discard what it rendered; a failure before the first token emits nothing and sends no rollback. `optimisticPrimaryStream` (default on) lets synthetic/background turns opt out — they persist from collected events, so a live failed attempt would leak in.
- **Stage 2 — commit-gated persistence** (`sse-consumer.ts`): on a routed turn the assistant + error rows are deferred and written only at a commit point (tool_start / `model_route_success` / `model_route_exhausted`), and dropped on `model_route_rollback`. The live frontend already rendered them over SSE, so this only governs what survives a reload — without it a failed primary's reply/error would be left in the DB after a fallback took over. Tool rows are not deferred (already committed); non-routed turns persist inline.

**Stage 3 (separate follow-up):** frontend rollback rendering — clearing a failed primary's live bubbles on `model_route_rollback`. Until then, a *rare* live-primary failure briefly shows the failed output before the fallback's; a reload is already clean.

## Testing

- `model-routing.test.ts` (49), `http-server.test.ts` (42), `sse-consumer.test.ts` (36): primary streams live on success (no buffering, no rollback); live-fail → rollback → fallback ordering; commit-gating drops a rolled-back attempt's reply + error while persisting only the winner; exhaustion persists the error row.
- Full suite: 177 files, 3536 passed, 0 failed; `tsc --noEmit` clean.


src/gateway/sse-consumer.ts
```diff
@@ -296,6 +296,47 @@ export async function consumeAgentSse(opts: ConsumeAgentSseOptions): Promise<Sse
...
+  // next candidate. Drop everything it buffered and re-arm the per-attempt
+  // dedup flags so the next attempt can record (and surface live) its own
+  // error, and so the rolled-back attempt's error text doesn't leak into the
+  // returned run summary / notifications (mirrors commitRoutedTurn).
+  const discardRoutedAttempt = () => {
+    pendingAssistantOps.length = 0;
+    pendingErrorOps.length = 0;
+    errorPersisted = false;
+    streamErrorEmitted = false;
+    errorMessage = "";
```
jacoblee-io, line 338: `firstAssistantPersisted` is flipped when the assistant op is **enqueued** (~L775), but it is not reset here. On the path `primary streams text → message_end{stopReason:"error"} → model_route_rollback → fallback succeeds`, the primary's enqueued-then-discarded op leaves `firstAssistantPersisted = true`, so the surviving fallback reply is gated out of `ttft_ms` (the `!firstAssistantPersisted` check at message_end) — the turn then shows no TTFT on reload.

Repro (a scenario the current tests don't cover): primary emits text deltas + `message_end{error}`, then `model_route_rollback`, then fallback `message_end{stop}` + `model_route_success`. The persisted fallback row's `metadata.timing` comes out as `{turn_total_ms, thinking_ms, output_ms}` with **no `ttft_ms`**.

Mirror the other per-attempt resets just above:
```suggestion
    errorMessage = "";
    firstAssistantPersisted = false;
```
Low severity (telemetry-only), but it's the one asymmetry left after the patch-5 fixes.

jacoblee-io (approved):
## Review — Approve ✅

High-quality, well-staged change. I traced the routed-turn flows end to end and they hold together: the commit-gating between `model-routing.ts` and `sse-consumer.ts` is subtle but internally consistent. The live primary's `message_end` precedes `model_route_success` (deferred → flushed at the commit point), while a buffered fallback's `message_end` follows `model_route_success` (persists inline) — both land correctly, and `model_route_success` is emitted for the non-fallback primary too, so commit-gating always fires.

### Verified locally (checked out the branch, all green)
| Suite | Result |
|---|---|
| `model-routing.test.ts` | 49 ✅ |
| `sse-consumer.test.ts` | 39 ✅ |
| `http-server.test.ts` | 42 ✅ |
| `portal-web` `model-route-rollback.test.ts` | 5 ✅ |

### Findings

**1. [Low — confirmed by repro] `discardRoutedAttempt` does not reset `firstAssistantPersisted`.** See inline comment. Telemetry-only, on a rare path (primary must stream *text* before failing). Asymmetric with the patch-5 resets right above it. One-line fix + a test.

**2. [Note] Deployment coupling.** The gateway now emits `model_route_rollback` + commit-gated persistence; this PR ships the siclaw-native `portal-web` rollback rendering (patch 4), but the sicore-web rendering is the paired stage-3 MR. Until both ship, a *rare* live-primary failure briefly shows the failed output before the fallback (already acknowledged in the description). Data/reload is already correct — recommend landing the two together.

**3. [Note — not a bug] Recovered-primary reply loses inline `model_route` metadata under live streaming.** Because the live primary's `message_end` precedes `model_route_success`, `currentModelRouteMetadata` is null when the reply row is captured. This is intentional and consistent (the frontend ref-gating uses the same ordering, so live == reload, and the recovery notice still persists as its own row). Flagging only as a behavior change vs. the old fully-buffered path.

Approving — only #1 is actionable, and it is telemetry-only on a rare path, fine to fold in as a follow-up.

jacoblee-io (commented):
## Review summary

Carefully architected, well-tested PR — no blocking correctness issues found. I traced the base code against the diff focusing on event ordering in the rollback and commit-gating paths.

**The most important correctness property holds:** a live primary that executes a tool and *then* fails does **not** emit `model_route_rollback`. In that case `hadToolExecution` sets `fallbackBlockedReason="tool_execution"` and the runner takes the `model_route_exhausted` branch (`model-routing.ts:584`); the rollback emit (`model-routing.ts:434-441`) lives only in the fallback-taken branch, which requires `!fallbackBlockedReason`. So a tool whose side effects already ran is never told to be dropped by the frontend. 👍

Also verified: commit-gating covers success / exhausted / rollback / transport-drop, with the end-of-stream defensive flush as a backstop; `optimisticPrimaryStream:false` for synthetic/background turns is correct (interactive HTTP path at `http-server.ts:609` leaves it defaulted to true).

A few low-severity items and two things I'd like the author to confirm.

---

### 🟡 1. Rollback doesn't re-arm the timing flags → fallback reply loses `ttft_ms`
`discardRoutedAttempt` re-arms `errorPersisted` and `streamErrorEmitted`, but not `firstAssistantPersisted` (nor `firstTokenTime` / `assistantMsgFirstTextTime`). When a primary streams **partial text** before failing, the deferred-assistant branch sets `firstAssistantPersisted=true` (because `assistantContent` is non-empty), and the rollback doesn't reset it — so the fallback's persisted reply gets no `ttft_ms`, and the turn loses its TTFT telemetry. All rollback tests use empty-content failures (`content:[]`), so this path is **uncovered**. Either reset these flags in `discardRoutedAttempt`, or add a comment that omitting TTFT on a rolled-back-primary turn is intentional.

### 🟡 2. `dropFailedAttemptOutput` relies on an implicit invariant — please document it
It drops *all* visible rows after the last user message. That's only safe because rollback is emitted exclusively for the primary (a fallback that goes live has executed a tool → fallback blocked → exhaust, no rollback). If a future change (Stage 3+) lets fallbacks stream live, this would wrongly drop a prior fallback's good output. Worth a comment in the function asserting "rollback only ever fires for the primary, so everything after the user turn is the failed attempt's output."

### ❓ 3. PR description vs. diff mismatch on "Stage 3"
The description says **"Stage 3 (separate follow-up): frontend rollback rendering"**, but the diff already adds exactly that to `portal-web/src/hooks/usePilotChat.ts` (the `model_route_rollback` case + `dropFailedAttemptOutput` + a test). I assume the follow-up refers to the paired **sicore** frontend and portal-web is done here — please realign the description so reviewers don't misjudge scope.

### ❓ 4. Stage 2 reload rendering is missing in portal-web (only in sicore)
`toPilotMessage` (`usePilotChat.ts:384`) does **not** special-case `metadata.kind === "error_response"`. A persisted error row is `role:"assistant"` + that metadata, so on reload portal-web renders it as a **plain assistant bubble**, not an error bubble. The description says the *sicore* MR renders it back as an error bubble. Please confirm: (a) is portal-web's degraded rendering (error text shown as a normal reply) acceptable, and (b) is the sicore MR merging in lockstep — otherwise a reloaded failed turn in sicore shows a bare assistant row too.

### 📝 5. Docs
The new CLAUDE.md matrix row has `Must read: —` (no design doc). This changes the routing + persistence contract (optimistic streaming, commit-gated persistence) — per the project's own "documentation rule," consider capturing the rationale in `docs/design/`. The matrix row itself is admirably detailed.

### 🔹 6. Minor
An aborted routed turn whose winning `message_end` was already deferred gets flushed by the end-of-stream defensive flush as a *normal* row (no `incomplete:true`), unlike the non-routed abort path which marks partial text incomplete. Cosmetic.

---

_Note: I did not run the suite locally; the 3536-passed figure is from the PR description._

src/gateway/sse-consumer.ts
```diff
@@ -296,6 +296,52 @@ export async function consumeAgentSse(opts: ConsumeAgentSseOptions): Promise<Sse
...
+  // model_route_rollback: the live primary failed and will be retried on the
+  // next candidate. Drop everything it buffered and re-arm the per-attempt
+  // dedup flags so the next attempt can record (and surface live) its own
+  // error, and so the rolled-back attempt's error text doesn't leak into the
+  // returned run summary / notifications (mirrors commitRoutedTurn).
+  const discardRoutedAttempt = () => {
+    pendingAssistantOps.length = 0;
+    pendingErrorOps.length = 0;
+    errorPersisted = false;
+    streamErrorEmitted = false;
```
jacoblee-io, line 337: `discardRoutedAttempt` re-arms `errorPersisted` and `streamErrorEmitted` here, but not `firstAssistantPersisted` (nor `firstTokenTime` / `assistantMsgFirstTextTime`). When a primary streams **partial text** before failing, the deferred-assistant branch sets `firstAssistantPersisted=true` (because `assistantContent` is non-empty), and this reset doesn't clear it — so the fallback's persisted reply loses `ttft_ms` and the turn loses TTFT telemetry. The rollback tests all use empty-content failures (`content:[]`), so this path is uncovered. Either reset the timing flags here too, or add a note that dropping TTFT on a rolled-back-primary turn is intentional.

portal-web/src/hooks/usePilotChat.ts
```diff
@@ -339,6 +339,23 @@ function findLastMessageIndex<T>(items: T[], predicate: (item: T) => boolean): n
...
 
+/**
+ * Drop the live output of a failed model-routing primary attempt: the streaming
+ * assistant bubble and any error bubble rendered after the latest visible user
+ * message. The user turn, earlier history, and hidden ledger rows are kept, so
+ * the fallback candidate's reply streams in cleanly instead of stacking under
+ * the rolled-back attempt's output. Matches the gateway's commit-gated
+ * persistence (a rolled-back attempt is never written to the DB).
+ */
+export function dropFailedAttemptOutput(messages: PilotMessage[]): PilotMessage[] {
```
jacoblee-io, line 350: This drops **all** visible rows after the last user message, which is only safe because `model_route_rollback` is emitted exclusively for the primary candidate (a fallback that goes live has executed a tool → fallback is blocked for good → exhaust, no rollback). Worth stating that invariant in the doc comment: if a future change (Stage 3+) lets fallback candidates stream live, this would wrongly drop a prior fallback's good output.

PR #323, opened 2026-06-13
Guide chart tool toward valid structured data

LikiosSedo:
## Summary
- tighten render_chart guidance so models only call it with finalized structured numeric data
- steer uncertain/simple category charts toward Mermaid xychart-beta
- document that chart data must be an object, not a JSON string or placeholder expression

## Validation
- npx vitest run mcp/create-chart/src/handler.test.ts
- npm run build

### DIFF
--- a/mcp/create-chart/src/handler.ts
+++ b/mcp/create-chart/src/handler.ts
@@ -17,7 +17,7 @@ export const RENDER_CHART_INPUT_SCHEMA = {
     data: {
       type: "object",
       description:
-        "Chart data. Pie: {slices:[{label,value}]}. Bar: {categories:[string], series:[{name,values:[number]}]}. Line: {series:[{name, points:[{x:number|string, y:number}]}]}. For time series pass x as epoch seconds (number) and y as the metric value.",
+        "Chart data as a real JSON object, never as a JSON string. Pie: {slices:[{label,value}]}. Bar: {categories:[string], series:[{name,values:[number]}]}. Line: {series:[{name, points:[{x:number|string, y:number}]}]}. Every value must be a finite number; do not use placeholders, variables, or references to earlier messages.",
     },
     title: { type: "string" },
     width: { type: "integer", minimum: 200, maximum: 2400 },
@@ -29,7 +29,12 @@ export const RENDER_CHART_INPUT_SCHEMA = {
 } as const;
 
 export const RENDER_CHART_DESCRIPTION =
-  "Render a pie/bar/line chart from structured data. The tool returns a READY_TO_PASTE chart block as plain markdown plus metadata. Use proactively when the user asks to visualize, plot, or 画图/出图/画饼图/柱状图/曲线/折线/趋势图, AND chartable data is already in context. Typical sources: status_counts (pie), category_summary anomaly/no_anomaly counts (bar), time-series samples (line, x=epoch seconds or display string, y=value). In your final reply, paste the READY_TO_PASTE block exactly as returned. Do not rewrite, escape, quote, or wrap the chart JSON; the frontend renders ```chart fenced JSON blocks as SVG.";
+  [
+    "Render a pie/bar/line chart only when finalized structured numeric data is already in context and can be passed as valid tool arguments.",
+    "For qualitative diagrams, workflows, topology, decision trees, rough summaries, or small category/count visuals where tool arguments might be uncertain, use a ```mermaid fenced block instead; xychart-beta is suitable for simple bar charts.",
+    "Arguments must be one JSON object. data must be an object, never a JSON string. Use only literal finite numbers; never use placeholders, expressions, previous-message references, or bare tokens.",
+    "The tool returns a READY_TO_PASTE chart block as plain markdown plus metadata. In your final reply, paste the READY_TO_PASTE block exactly as returned. Do not rewrite, escape, quote, or wrap the chart JSON; the frontend renders ```chart fenced JSON blocks as SVG.",
+  ].join(" ");
 
 function chartBaseDir(): string {
   const root =
--- a/mcp/create-chart/src/handler.test.ts
+++ b/mcp/create-chart/src/handler.test.ts
@@ -28,6 +28,9 @@ describe("RENDER_CHART_INPUT_SCHEMA", () => {
     expect(RENDER_CHART_DESCRIPTION).toMatch(/READY_TO_PASTE/);
     expect(RENDER_CHART_DESCRIPTION).toMatch(/exactly/i);
     expect(RENDER_CHART_DESCRIPTION).toMatch(/Do not rewrite, escape, quote/);
+    expect(RENDER_CHART_DESCRIPTION).toMatch(/mermaid/i);
+    expect(RENDER_CHART_DESCRIPTION).toMatch(/data must be an object/i);
+    expect(RENDER_CHART_DESCRIPTION).toMatch(/never a JSON string/i);
   });
 });
 
--- a/portal-web/src/components/chat/Markdown.test.tsx
+++ b/portal-web/src/components/chat/Markdown.test.tsx
@@ -58,6 +58,20 @@ describe("Markdown Mermaid fences", () => {
       ok: true,
       kind: "timeline",
     })
+    expect(
+      validateMermaidSource(
+        [
+          "xychart-beta",
+          '  title "事故单AI可解决性分析"',
+          '  x-axis ["AI可解决", "AI不可解决", "有条件可解决"]',
+          '  y-axis "数量" 0 --> 2',
+          "  bar [1, 0, 2]",
+        ].join("\n"),
+      ),
+    ).toMatchObject({
+      ok: true,
+      kind: "xychart",
+    })
   })
 
   it("repairs leaked stream content prefixes inside Mermaid blocks", () => {
--- a/portal-web/src/components/chat/MermaidRenderer.tsx
+++ b/portal-web/src/components/chat/MermaidRenderer.tsx
@@ -13,7 +13,7 @@ const MAX_MERMAID_LINES = 160
 const MAX_MERMAID_EDGES = 80
 
 type MermaidValidation =
-  | { ok: true; source: string; kind: "flowchart" | "sequence" | "timeline" }
+  | { ok: true; source: string; kind: "flowchart" | "sequence" | "timeline" | "xychart" }
   | { ok: false; reason: string }
 
 type RenderState =
@@ -88,9 +88,12 @@ export function validateMermaidSource(raw: string): MermaidValidation {
   if (/^timeline\b/i.test(head)) {
     return { ok: true, source, kind: "timeline" }
   }
+  if (/^xychart-beta\b/i.test(head)) {
+    return { ok: true, source, kind: "xychart" }
+  }
   return {
     ok: false,
-    reason: "Only flowchart/graph, sequenceDiagram, and timeline diagrams are supported.",
+    reason: "Only flowchart/graph, sequenceDiagram, timeline, and xychart-beta diagrams are supported.",
   }
 }
 


mcp/create-chart/src/handler.ts
```diff
@@ -29,7 +29,12 @@ export const RENDER_CHART_INPUT_SCHEMA = {
 } as const;
 
 export const RENDER_CHART_DESCRIPTION =
-  "Render a pie/bar/line chart from structured data. The tool returns a READY_TO_PASTE chart block as plain markdown plus metadata. Use proactively when the user asks to visualize, plot, or 画图/出图/画饼图/柱状图/曲线/折线/趋势图, AND chartable data is already in context. Typical sources: status_counts (pie), category_summary anomaly/no_anomaly counts (bar), time-series samples (line, x=epoch seconds or display string, y=value). In your final reply, paste the READY_TO_PASTE block exactly as returned. Do not rewrite, escape, quote, or wrap the chart JSON; the frontend renders ```chart fenced JSON blocks as SVG.";
+  [
+    "Render a pie/bar/line chart only when finalized structured numeric data is already in context and can be passed as valid tool arguments.",
+    "For qualitative diagrams, workflows, topology, decision trees, rough summaries, or small category/count visuals where tool arguments might be uncertain, use a ```mermaid fenced block instead; xychart-beta is suitable for simple bar charts.",
```
jacoblee-io: **Blocking:** this steers the model toward Mermaid diagram types the frontend rejects.

`portal-web/src/components/chat/MermaidRenderer.tsx:82-94` only accepts `flowchart`/`graph`, `sequenceDiagram`, and `timeline`; anything else returns *"Only flowchart/graph, sequenceDiagram, and timeline diagrams are supported."* So a `xychart-beta` block (and mermaid `pie`) renders as an error rather than a chart — and "small category/count visuals" is precisely what `render_chart` does well (the old description used it for `status_counts`→pie and `category_summary counts`→bar).

Either drop the `xychart-beta` / "small category/count → mermaid" wording and limit Mermaid here to the qualitative diagrams the allowlist supports (workflows/topology/decision-trees → flowchart, sequence, timeline), or widen the `MermaidRenderer` allowlist to include `xychart-beta` first. As written, the guidance and the renderer disagree.

mcp/create-chart/src/handler.ts
```diff
@@ -17,7 +17,7 @@ export const RENDER_CHART_INPUT_SCHEMA = {
     data: {
       type: "object",
       description:
-        "Chart data. Pie: {slices:[{label,value}]}. Bar: {categories:[string], series:[{name,values:[number]}]}. Line: {series:[{name, points:[{x:number|string, y:number}]}]}. For time series pass x as epoch seconds (number) and y as the metric value.",
+        "Chart data as a real JSON object, never as a JSON string. Pie: {slices:[{label,value}]}. Bar: {categories:[string], series:[{name,values:[number]}]}. Line: {series:[{name, points:[{x:number|string, y:number}]}]}. Every value must be a finite number; do not use placeholders, variables, or references to earlier messages.",
```
jacoblee-io: Minor contradiction: this lists `Line: points:[{x:number|string, ...}]` and then says **"Every value must be a finite number."** The validator supports a string `x` and string bar `categories` (see `handler.test.ts:148-169` "keeps numeric and string x"), so a literal reading could make the model avoid string x-axis labels / categories. Suggest scoping it to numeric fields, e.g. *"every **numeric** field (y, values, slice value) must be a finite number; x-axis labels and categories may be strings."*

jacoblee-io (changes requested):
## Review — Request changes

The tightening of "only call render_chart with finalized numeric data" is good. But the new Mermaid steering points the model at diagram types this frontend cannot render, and the rewrite silently drops some product-specific guidance. Details inline; summary below.

### 🔴 Blocking — Mermaid steering conflicts with the frontend allowlist
The new description recommends `xychart-beta` for simple bar charts and steers "small category/count visuals" to a ```mermaid block. But `portal-web/src/components/chat/MermaidRenderer.tsx:82-94` only renders **flowchart/graph, sequenceDiagram, timeline** — everything else is rejected with *"Only flowchart/graph, sequenceDiagram, and timeline diagrams are supported."* So `xychart-beta` (and mermaid `pie`) blocks render as an error, not a chart — exactly the bar/pie/count cases that are `render_chart`'s core use. See inline on the description line.

Fix options: (a) drop the `xychart-beta` / "small category/count → mermaid" wording and reserve Mermaid for the *qualitative* diagrams the allowlist supports (flowchart/sequence/timeline), keeping numeric charts on `render_chart`; or (b) first widen `MermaidRenderer`'s allowlist to include `xychart-beta`, then this guidance is correct. The two must agree. (The sicore-web frontend referenced in #322 isn't visible here — please confirm its Mermaid allowlist too.)

### 🟡 Dropped product/domain guidance (not mentioned in the summary)
The rewrite removed, from the old description: the proactive-use trigger incl. the Chinese keywords `画图/出图/画饼图/柱状图/曲线/折线/趋势图`; the domain sources `status_counts (pie)`, `category_summary counts (bar)`; and the `time-series x = epoch seconds` convention. siclaw is primarily a Chinese-language SRE tool, so dropping the Chinese triggers + steering count/bar to (currently broken) Mermaid is likely to noticeably reduce `render_chart` usage. Please confirm this is intentional.

### 🟢 Minor — contradictory wording on `data`
See inline: the `data` field says `x: number|string` but then "Every value must be a finite number", while the validator (and `handler.test.ts:148-169`) explicitly supports string `x` and string bar categories.

### Tests
The 3 new assertions pass and the old ones stay compatible, but they're string matches — they can't catch the allowlist conflict above, which is the substantive issue.

LikiosSedo:
Addressed all three in c219ee8 (rebased onto latest main).

**🔴 guidance vs renderer — fixed by making them agree.** Rather than dropping the xychart steering, this PR now widens the renderer: `portal-web/src/components/chat/MermaidRenderer.tsx` accepts `xychart-beta` (mermaid v11 renders it; the render path is generic `mermaid.render(source)`). The sicore-web renderer got the same change separately, so both frontends now render `xychart-beta`. The description and the renderers agree.

**🟡 restored the dropped product guidance.** Brought back the proactive-use triggers incl. the Chinese keywords (`画图/出图/画饼图/柱状图/曲线/折线/趋势图`), the domain sources (`status_counts`→pie, `category_summary` counts→bar, time-series→line), and the `x=epoch seconds` convention — while keeping this PR's tightening ("only when finalized numeric data is in context", "never a JSON string"). Reconciled the count/bar steering: **finalized count/bar data stays on `render_chart`**; mermaid is reserved for qualitative diagrams (flowchart/sequence/timeline), with `xychart-beta` available for a quick bar block.

**🟢 data wording.** Scoped the finite-number rule to numeric fields: *"Numeric fields (y, values, slice value) must be finite numbers; x-axis labels and bar categories may be strings"* — matches the validator + `handler.test.ts` ("keeps numeric and string x").

Tests: `handler.test` 20/20 (added guards for the restored triggers + the numeric-vs-string scoping); `portal-web/Markdown.test` gains an `xychart-beta` validation case; `mcp/create-chart` tsc build clean. (pie is still render_chart-only — neither renderer does mermaid `pie`, and the description doesn't steer there.)

LikiosSedo:
Correction to my previous comment re 🟡.

Per the maintainer: dropping the Chinese trigger keywords in this PR is **intentional** — this is an English open-source project, and the rewrite deliberately tightens render_chart toward "only fire with finalized numeric data" rather than "fire eagerly on any visualize request". So I've **reverted** my earlier restoration of the Chinese keywords / proactive triggers / domain sources. The description stays the lean English version from this PR.

Net state now (4c921b8):
- **🔴 resolved by widening the renderer, not the description**: `portal-web/.../MermaidRenderer.tsx` now accepts `xychart-beta` (mermaid v11 renders it; generic render path), so the PR's `xychart-beta` steering renders correctly. sicore-web got the same change. Description and renderers agree.
- **🟡**: won't-fix by design — English-only triggers are intentional.
- **🟢** (data wording "every value must be a finite number" vs string x/categories): left as the PR's original for now; happy to scope it to numeric fields if you'd still like that tightening, but it's independent of the blocker.

So the PR is: the original lean-English chart-tool guidance + xychart-beta renderer support in both frontends.

jacoblee-io (commented):
Re-reviewed after `4c921b8`.

## ✅ Blocker resolved
🔴 The Mermaid allowlist conflict is fixed: `MermaidRenderer` now accepts `xychart-beta`, so the guidance and the renderer agree. Confirmed sound for portal-web — mermaid 11.15.0 renders `xychart-beta` (added in v10.3), and `kind:"xychart"` is only used as a download-filename label so nothing downstream breaks. create-chart handler tests (18) + Markdown tests (18) pass; `tsc` clean.

## Still open from the prior review

🟡 **Chinese triggers + domain sources still dropped.** The rewrite removed `画图/出图/画饼图/柱状图/曲线/折线/趋势图`, the `status_counts (pie)` / `category_summary counts (bar)` sources, and the `x = epoch seconds` convention. siclaw is a Chinese-first SRE tool, so dropping the Chinese triggers is a real behavioral regression for the primary user base — please restore them (or confirm dropping is intentional, as asked last round).

🟢 **`data` description self-contradicts.** It lists `points:[{x:number|string,…}]` and bar `categories:[string]`, then says "Every value must be a finite number." The validator + `handler.test.ts` support string x / string categories, so this wording can steer the model away from valid string labels. Suggest: "every **numeric** value must be finite; x / category may be a string label."

🟠 **sicore-web Mermaid allowlist (cross-frontend).** This PR widened only portal-web's renderer. The guidance now steers models to `xychart-beta`, but sicore-web (the separate frontend) has its own allowlist — if it doesn't also accept `xychart-beta`, the new guidance renders an error block there instead of a chart, exactly for the bar/count case. Please confirm sicore-web's allowlist (separate repo) before relying on this steering.

Once the `data` wording is corrected, the Chinese triggers restored (or the drop justified), and sicore-web confirmed, this is good to approve — the code change itself is clean.

jacoblee-io (approved):
Approving. ✅ The blocking item is resolved, and on reflection the rest are non-blocking — also correcting the framing of my earlier note.

## Resolved
🔴 **Mermaid allowlist conflict** → fixed in `4c921b8`: `MermaidRenderer` now accepts `xychart-beta`, so guidance and renderer agree. Verified for portal-web — mermaid 11.15.0 renders `xychart-beta` (added v10.3), and `kind:"xychart"` is only a download-filename label so nothing downstream breaks. create-chart handler (18) + Markdown (18) tests pass; `tsc` clean.

## Correction to my earlier comment
I framed the dropped-guidance point as a "Chinese-first" regression — that over-emphasized one aspect. The accurate, language-agnostic finding is that the rewrite dropped (1) the **proactive-trigger vocabulary** and (2) the **domain-source hints** (`status_counts → pie`, `category_summary counts → bar`, `x = epoch seconds`). The deleted trigger tokens happened to be the Chinese keywords (`画图/画饼图/柱状图/趋势图`), which is the concrete example — but the substance is "explicit triggers + domain hints removed," not "Chinese removed." And tightening proactive charting to "only finalized numeric data" is a reasonable, **intentional** author choice, so I'm treating it as a design decision rather than a blocker.

## Non-blocking follow-ups
🟢 `data` description self-contradicts: it allows `x:number|string` / `categories:[string]` yet says "Every value must be a finite number" (the validator + tests support string x/categories). Suggest: "every **numeric** value must be finite; x / category may be a string label." Worth a one-liner, not blocking.

🟠 sicore-web Mermaid allowlist (separate repo): the guidance now steers models to `xychart-beta`; please confirm sicore-web's renderer also accepts it, otherwise charts render as an error block in that frontend. A cross-frontend confirm, not a defect in this PR — fine to verify around merge.

Code change itself is clean and green — good to go.

PR #324, opened 2026-06-13
fix(chat): abort gateway SSE consumer on stop so tool rows finalize

jacoblee-io:
## Problem
After clicking Stop, refreshing the page makes a stopped conversation appear to "keep reasoning" (spinner / Thinking never clears) — for up to the frontend's 30-minute stale window before it's finally treated as timed_out.

## Root cause
`chat.abort` calls `client.abortSession()` to stop the agentbox prompt, but **never aborts the gateway's `consumeAgentSse` `AbortController`**. The consumer loop therefore ends only when the agentbox closes the stream naturally, skipping its abort-finalization — so an in-flight tool row (`tool_execution_start` persisted but no matching `tool_execution_end`) is left with `metadata.status` missing / `outcome=null`, which the frontend maps to `running`. On refresh, `hasRunningPersistedMessages` returns true and the UI re-enters the streaming/recovery state, looking like it's "still reasoning". The comment in `sse-consumer.ts` already assumes the signal is aborted on Stop, but that wiring was missing.

## Fix (`src/gateway/server.ts`)
- Add a per-session `activeStreamAborts: Map<string, AbortController>`.
- `chat.send` registers this turn's `abortCtrl` after `prompt()` succeeds (placed **after** the concurrent-send "already running" → steer early-return, so it can't clobber the in-flight prompt's controller) and clears it when the turn settles.
- `chat.abort` calls `activeStreamAborts.get(sessionId)?.abort()` **before** `client.abortSession()`, so `consumeAgentSse` breaks on `signal.aborted` and runs its finalization (in-flight tool rows → `stopped`, partial assistant text persisted).

## Testing
- New `src/gateway/server-chat-abort.test.ts`: drives the real `startRuntime` `chat.send`/`chat.abort` RPC handlers (data layer mocked) and asserts Stop aborts the in-flight consumer's signal. **The test fails when the abort wiring is removed**, confirming it guards the regression.
- `npm test` green (gateway suite 519 passed; full suite 3527 passed).
- Full backend-chain smoke locally (fake Anthropic streaming endpoint + real Portal→Runtime→AgentBox→consumer→DB): with the fix the tool row finalizes to `stopped`; with the fix line removed and the runtime restarted it stays `running` (bug reproduced).
- Built `siclaw-runtime:v0.1.9-fixstop` and deployed to siclaw-inner for verification (confirmed the fix is present in the image).

### DIFF
--- a/src/gateway/server-chat-abort.test.ts
+++ b/src/gateway/server-chat-abort.test.ts
@@ -1,0 +1,164 @@
+/**
+ * Regression test for the chat.abort → SSE-consumer wiring.
+ *
+ * Bug: clicking Stop aborted the agentbox prompt but NOT the gateway's
+ * consumeAgentSse signal, so the consumer ended "naturally" and skipped its
+ * abort-finalization — leaving in-flight tool rows persisted as "running".
+ * A page refresh then re-painted the turn as still reasoning.
+ *
+ * This test drives the real chat.send / chat.abort RPC handlers from
+ * startRuntime (with the data-layer + agentbox modules mocked) and asserts
+ * that chat.abort aborts the signal handed to the in-flight consumer.
+ */
+import { describe, it, expect, afterEach, vi } from "vitest";
+
+vi.mock("./chat-repo.js", () => ({
+  ensureChatSession: vi.fn(async () => {}),
+  appendMessage: vi.fn(async () => "msg-id"),
+  incrementMessageCount: vi.fn(async () => {}),
+}));
+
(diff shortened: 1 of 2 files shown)


LikiosSedo (approved):
LGTM. I reviewed the Runtime chat.abort wiring, the SSE consumer abort-finalization path, and the frontend reload mapping for stopped tool rows. Targeted gateway tests and typecheck pass locally; the remaining test-observability note is non-blocking.

PR #325, opened 2026-06-13
fix(chat): make Stop reliably halt all work across every scenario

jacoblee-io:
## Problem
After PR #324 (foreground tool rows no longer resume on refresh), Stop was still unreliable for **background work** and several edge windows. The reported case: send a message, the agent launches `host_exec(run_in_background)` (ping -c 100), click Stop immediately — the foreground turn stops but the background ping **runs to completion**.

## Root cause
Abort was a one-shot sweep + `brain.abort()`, with no persistent latch. A tool call mid-flight at Stop registers its background job **after** the single `stopSessionJobs()` sweep, so nothing ever stops it. Plus several pre/post-prompt and notification windows leaked. `_aborted` existed but was only read by the model-routing `shouldAbort` callback — never by the background-exec executor or sub-agent launcher.

## Changes (10)
- **#1 (primary) Re-sweep** `stopSessionJobs` in `/abort` AFTER `brain.abort()` resolves (the run loop is fully drained, so the job the in-flight tool launched during the drain is now registered → caught).
- **#2** Clear the steer queue (`brain.clearQueue()`) on Stop so a queued steer doesn't replay on the next prompt.
- **#3** Latch `createBackgroundExecExecutor` on parent `_aborted`: register a terminal "stopped" job and return a launched handle (never throw — throwing makes the tool fall back to a foreground run); no process spawned.
- **#4** Latch `startBackgroundSubagent` on parent `_aborted`: register stopped, skip `runSpawnedSubagent`.
- **#5** Pre-prompt latch in `/api/prompt`: if `_aborted` before the run starts, finish cleanly without prompting.
- **#6** Pre-spawn pending-abort (`markPendingAbort`/`consumePendingAbort`, 10-min TTL): `/abort` on a not-yet-created session returns `200 {pending:true}` instead of `404`; `/api/prompt` consumes it AFTER the unconditional `_aborted=false` reset (placing it before would be wiped).
- **#7** Discard buffered completion notifications + cancel the coalesce timer on Stop, and guard `flushPendingNotifications`/`runSyntheticPrompt` on `_aborted` — kills the "a job that finished just before Stop wakes the model via a synthetic turn" resurrection path (independent of #1–#6).
- **#8** `PiAgentBrain.abort()` also calls `session.abortCompaction()` — the run AbortController doesn't cover the separate compaction controllers, so a Stop during compaction otherwise leaks the LLM call and hangs `waitForIdle()`.
- **#9** `runSpawnedSubagent` bails on parent `_aborted` after `createSiclawSession` — the setup window where `job.abort` isn't wired and `stopJob` no-ops (job stays "running", so a job-status check would never fire; parent `_aborted` is the reliable signal).
- **#10** `PiAgentBrain` tracks an `aborted` flag so the empty-response retry loop can't fire a fresh, un-aborted re-prompt when Stop lands in the backoff sleep.

## Scenario coverage
pre-spawn (#6+#5), post-spawn pre-LLM (#5), LLM streaming / foreground tool (already worked via the run AbortSignal), foreground sub-agent (already worked), background sub-agent (#4+#9+#1), background job — the reported bug (#1+#3), compaction (#8), empty-response retry (#10), just-completed-job resurrection (#7), steer queue (#2).

Out of scope (documented): sub-tree job sweep is unnecessary (child sessions get no background executors, so no job registers under a child sessionId); a foreground `host_exec` that itself `nohup &`-backgrounds a remote process is unreachable by Stop.

## Testing
Targeted tests added/updated across `http-server.test.ts`, `session.test.ts`, `notify-parent.test.ts`, `pi-agent-brain.test.ts` (the prior pi-agent-brain abort test is flipped from the buggy `callCount===2` to the fixed `===1` — direct proof of no re-prompt after Stop). `npx tsc --noEmit` clean; full suite **3554 passed / 2 skipped**.

Built as `siclaw-runtime:v0.1.9-fixstop2` and deployed to `siclaw-inner` + `sicore-test` for verification.

### DIFF
--- a/src/agentbox/http-server.test.ts
+++ b/src/agentbox/http-server.test.ts
@@ -156,6 +156,9 @@ function makeFakeSessionManager() {
     list: () => Array.from(sessions.values()),
     get: (id: string) => sessions.get(id),
     stopSessionJobs: vi.fn(() => 0),
+    markPendingAbort: vi.fn(),
+    consumePendingAbort: vi.fn(() => false),
+    discardPendingNotifications: vi.fn(),
     getOrCreate: async (id?: string, _mode?: unknown, _systemPromptTemplate?: unknown, activeMode?: unknown) => {
       getOrCreateCalls.push({ id, activeMode });
       const key = id ?? "default";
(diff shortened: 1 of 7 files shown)


LikiosSedo:
Thanks for the comprehensive Stop hardening. The overall direction looks right to me, but I found one blocker and one follow-up risk before merging.

**Blocking:** the `#9` background sub-agent setup-window path still does not convert the registry job into a user-stopped job.

The race is:

1. `startBackgroundSubagent()` registers the job as `running`, before `job.abort` is wired.
2. User presses Stop while `runSpawnedSubagent()` is still in `createSiclawSession` / child setup.
3. `stopSessionJobs()` sees a running job with no `abort` hook, so `JobRegistry.stopJob()` returns `"starting up"` and does not set `status="stopped"` or `suppressNotifyTurn=true`.
4. The new parent `_aborted` check later calls `requestStop()` and correctly prevents the child prompt from starting, but it still does not patch the registry job status/suppression.
5. The background-subagent completion path then maps the job to `partial` and calls `notifyParent()` without a durable `suppressNotifyTurn` latch. Today this relies on `managed._aborted` still being true when the coalesce timer flushes; a quick next prompt resets `_aborted` and can let the cancelled sub-agent notification ride into the next turn.

Suggested fix: in the `#9` parent-aborted branch, once `opts.jobId` is wired, route through `jobs.stopJob(opts.jobId, { suppressNotifyTurn: true })` or otherwise set `status="stopped"` + `suppressNotifyTurn=true` before returning. Please add a regression test that asserts the final registry status is `stopped`, `suppressNotifyTurn === true`, and no parent synthetic prompt is triggered.

**Follow-up risk:** `markPendingAbort()` is a 10-minute sticky latch for any missing AgentBox session. That solves true pre-spawn Stop, but an abort against a released/stale AgentBox session, or a send that fails before it reaches `/api/prompt`, can cancel the next real prompt with the same `sessionId`. Consider binding pending aborts to a Runtime-known in-flight send/generation, or using a scoped generation token/shorter narrowly-owned window instead of a generic session-level latch.

nightmeng (approved):

PR #326, opened 2026-06-13
feat(vision): forward image attachments to vision-capable models

LikiosSedo:
## What
Thread image attachments from the chat path down to pi-agent so vision-capable models (e.g. the ChatGPT/Codex gpt-5.x subscription) read pasted/attached images as **native vision input**, instead of relying on OCR-to-text.

## Why
The chat transport was text-only at every hop (`chat.send` RPC → agentbox `/prompt` → `brain.prompt` → pi-agent), so image bytes were dropped before reaching the model. pi-ai + pi-coding-agent already support image input end-to-end (they serialize `{type:"image"}` to `input_image` data URLs and gate on `model.input` including `"image"`); the only missing piece was plumbing the bytes through siclaw's layers.

## Changes
- `BrainSession.prompt(text, images?)` carries `PromptImage[]` (raw base64).
- `PiAgentBrain` maps them to pi-ai `ImageContent` and passes via `session.prompt(text, { images })`; re-passed on empty-response retries.
- `runPromptWithModelRouting` / `runAttempt` thread images to each attempt.
- agentbox `/prompt` accepts + validates an `images` field (≤4, png/jpeg/webp, ≤8MB b64); image-only messages allowed with a minimal default instruction.
- gateway `chat.send` + `AgentBoxClient` `PromptOptions` carry images.

Sicore advertises `input:["text","image"]` for codex/vision models and forwards the raw image (separate sicore MR).

## Test
- New unit tests: image passthrough / image-only / malformed-drop (http-server), `{mimeType,data}`→`ImageContent` mapping (pi-agent-brain), routing threads images to each attempt.
- `npm test` green for model-routing / http-server / pi-agent-brain / brain-session (131 tests); `tsc --noEmit` clean.
- Verified end-to-end in a live cluster: gpt-5.5 read a pasted screenshot and extracted ticket IDs from it (OCR confirmed not involved).

### DIFF
--- a/src/agentbox/http-server.test.ts
+++ b/src/agentbox/http-server.test.ts
@@ -304,6 +304,14 @@ describe("http-server — prompt + session lifecycle", () => {
     expect(r.data.error).toMatch(/Missing.*text/);
   });
 
+  it("POST /api/prompt drops images whose data is not valid base64", async () => {
+    const r = await getJson(port, "/api/prompt", "POST", {
+      images: [{ mimeType: "image/png", data: "not base64!!!" }], // bad base64 → dropped
+    });
+    expect(r.status).toBe(400);
+    expect(r.data.error).toMatch(/Missing.*text/);
+  });
+
   it("resolves the active operating mode from DP markers and passes it to getOrCreate", async () => {
     const lastMode = () => sm.getOrCreateCalls[sm.getOrCreateCalls.length - 1].activeMode;
 
--- a/src/agentbox/http-server.test.ts
+++ b/src/agentbox/http-server.test.ts
@@ -269,6 +269,41 @@ describe("http-server — prompt + session lifecycle", () => {
     expect(r.data.error).toMatch(/Missing.*text/);
   });
 
+  it("POST /api/prompt forwards images to brain.prompt as vision input", async () => {
+    const session = await sm.getOrCreate("img-1");
+    const r = await getJson(port, "/api/prompt", "POST", {
+      text: "what is in this image?",
+      sessionId: "img-1",
+      images: [{ mimeType: "image/png", data: "aW1n" }],
+    });
+    expect(r.status).toBe(200);
+    expect(session.brain.prompt).toHaveBeenCalledWith(
+      "what is in this image?",
+      [{ mimeType: "image/png", data: "aW1n" }],
+    );
+  });
+
+  it("POST /api/prompt accepts an image-only message and defaults the text", async () => {
+    const session = await sm.getOrCreate("img-only");
+    const r = await getJson(port, "/api/prompt", "POST", {
+      sessionId: "img-only",
+      images: [{ mimeType: "image/png", data: "aW1n" }],
+    });
+    expect(r.status).toBe(200);
+    expect(session.brain.prompt).toHaveBeenCalledWith(
+      "Please analyze the attached image.",
+      [{ mimeType: "image/png", data: "aW1n" }],
+    );
+  });
+
+  it("POST /api/prompt drops malformed images and rejects when nothing usable remains", async () => {
+    const r = await getJson(port, "/api/prompt", "POST", {
+      images: [{ mimeType: "image/gif", data: "aW1n" }], // unsupported mime → dropped
+    });
+    expect(r.status).toBe(400);
+    expect(r.data.error).toMatch(/Missing.*text/);
+  });
+
   it("resolves the active operating mode from DP markers and passes it to getOrCreate", async () => {
     const lastMode = () => sm.getOrCreateCalls[sm.getOrCreateCalls.length - 1].activeMode;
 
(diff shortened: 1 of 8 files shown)


jacoblee-io (commented):
Reviewed — strong PR, well-tested, sound end-to-end. One thing to reconcile before merge plus a couple of notes.

## 🟡 Image limits exceed the agentbox body cap (reachable inconsistency)

`sanitizePromptImages` allows `MAX_PROMPT_IMAGES=4 × MAX_PROMPT_IMAGE_BASE64_CHARS=8MB = 32MB`, but `/api/prompt`'s `parseJsonBody` rejects any body over `MAX_BODY_SIZE = 10MB` (`req.destroy()`) **before** sanitize runs. So:

- the advertised "≤4 images, ≤8MB each" is unreachable — even 2 × 8MB (16MB) fails;
- the failure is a generic `Body exceeds 10MB byte limit`, not the clean image count/size rejection the validation implies;
- the portal chat-gateway separately allows ~32MB (`MAX_CHAT_REQUEST_BODY_BYTES`), and sicore forwards raw images, so a 2-image / large-image request from sicore would die at the agentbox layer with a confusing error.

Effective limit today is ~1 large image. Please reconcile: raise `MAX_BODY_SIZE` to fit `4×8MB + overhead`, or lower the per-image/count caps to fit within 10MB, or document the 10MB body cap as the real bound. (The 10MB cap does bound the memory-blowup risk, so it fails safe — just inconsistent with the stated limits.)

## 🟢 Otherwise solid
- Clean plumbing: `BrainSession` → `PiAgentBrain` → every routing attempt (incl. fallbacks) → agentbox `/prompt` → RPC → client all carry `images`.
- Good validation (mime allowlist png/jpeg/webp lowercased, malformed-drop, count cap), re-validated at agentbox even though sicore caps upstream.
- Graceful non-vision degradation (pi-ai placeholder); image-only handled with a default instruction; `detectLanguage(promptText)` avoids the now-possibly-undefined `body.text`; empty-response retry re-passes images.
- Backward compatible (optional `images?`). `tsc` clean; 133 tests pass incl. the 5 new image cases.

## 🟢 Scope note (not a defect)
siclaw's own Portal (`chat-gateway.ts`) is unchanged — it still OCRs pasted images and does not forward raw `images` in `chat.send`, so native vision is exercised only by sicore (separate MR). No regression, but worth knowing that pasting an image in siclaw's own Portal still goes through OCR until chat-gateway is wired to pass images.

## Minor
- `sanitizePromptImages` checks `data.length` but not that `data` is valid base64 — a non-base64 string passes and errors downstream at the provider. Very low severity.

LikiosSedo:
Thanks — addressed in 30b0234.

**🟡 Image caps vs body cap** — took the "lower the caps to fit 10MB" route: `MAX_PROMPT_IMAGE_BASE64_CHARS` dropped from 8MB → **2MB**, so `MAX_PROMPT_IMAGES(4) × 2MB = 8MB` now fits under the 10MB `MAX_BODY_SIZE` with headroom for the JSON envelope + text. All four advertised images are actually deliverable now. Added a comment on the constant tying it to `MAX_BODY_SIZE` so the two can't silently drift apart again. (Kept the 10MB body cap as the memory-blowup guard rather than raising it 3.6×.)

**Minor — base64 validation** — `sanitizePromptImages` now rejects data that isn't valid base64 (`length % 4` + charset regex) instead of letting it fail downstream at the provider. Added a regression test.

**🟢 Scope note (Portal still OCRs)** — confirmed and intended: this PR only wires the agentbox/runtime path; siclaw's own Portal `chat-gateway.ts` still OCRs pasted images and isn't forwarding raw `images` in `chat.send`. Native vision is exercised via sicore only; wiring the siclaw Portal to forward images is a separate follow-up, no regression here.

`tsc` clean; http-server suite green (52 tests incl. the new base64 case).

jacoblee-io (approved):
Re-reviewed after `30b0234` — approving. ✅

Both items from my comment are resolved:

1. **Image caps vs body cap** → fixed: `MAX_PROMPT_IMAGE_BASE64_CHARS` lowered 8MB → 2MB so `4 × 2MB = 8MB` fits under the 10MB `MAX_BODY_SIZE` with headroom for the JSON envelope + text. All four advertised images are now actually deliverable, and the new comment documents the invariant so the two limits can't silently drift again. Good fix.
2. **base64 validation** → `sanitizePromptImages` now rejects non-base64 (`data.length % 4` + `^[A-Za-z0-9+/]+={0,2}$`) instead of letting it fail downstream at the provider. (Linear regex, no ReDoS risk.)

Rest of the PR was already solid: clean end-to-end plumbing (BrainSession → PiAgentBrain → every routing attempt → agentbox `/prompt` → RPC → client), mime allowlist + malformed-drop, graceful non-vision degradation, image-only handling, retry re-passes images, backward compatible. Verified locally: `tsc --noEmit` clean; http-server + pi-agent-brain + model-routing suites green (134 tests).

Minor (non-blocking): the PR description still reads "≤8MB b64" for the agentbox `/prompt` validation — update it to ≤2MB/image (≤8MB total) to match the code.

PR #327, opened 2026-06-13
feat(model-params): apply per-model reasoning effort

LikiosSedo:
## What
Honour `modelConfig.params.reasoning_effort` delivered by the control plane (SiCore) so reasoning effort is configurable per model instead of the hardcoded `thinkingLevel: "high"` in `agent-factory.ts`. Pairs with the SiCore subscription-model change that lets operators set effort per model.

> An earlier prototype also carried a `service_tier`/"fast" knob; it was **dropped** — the target subscription plan 400s on `service_tier=fast` and honouring it required reaching into pi internals. **This PR is reasoning-effort only** (no `service_tier`, `agent-factory.ts` unchanged).

## How
- `brain-session`: `BrainModelParams` + optional `applyModelParams`.
- `pi-agent-brain`: `applyModelParams` maps `reasoningEffort` -> `session.setThinkingLevel` (validated against pi's `ThinkingLevel` union; invalid value -> no-op).
- `http-server`: after `setModel`, read `modelConfig.params.reasoning_effort` (snake_case) -> `applyModelParams({ reasoningEffort })` (camelCase).

## Wire contract
SiCore delivers `modelConfig.params = { reasoning_effort }` (snake_case, string); this reads exactly that key. `xhigh` is not clamped because SiCore advertises an identity `thinkingLevelMap` for codex.

## Behaviour note
Absent `reasoning_effort` is a **no-op** (it does NOT reset to the model default). For subscription models SiCore always sends an effort, so this isn't hit in practice; a cross-model reset-to-default is a known follow-up, not in this PR.

## Tests
`tsc --noEmit` clean; `pi-agent-brain` applyModelParams cases (valid / invalid / absent) + `http-server` / `guard-pipeline` / `agent-factory` suites green.

Stacked on `feat/codex-multimodal-vision` (lands after that base).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,6 @@
         "@anthropic-ai/sdk": "^0.71.2",
         "@clack/prompts": "^1.1.0",
         "@earendil-works/pi-agent-core": "^0.78.1",
-        "@earendil-works/pi-ai": "0.78.1",
         "@earendil-works/pi-coding-agent": "0.78.1",
         "@earendil-works/pi-tui": "^0.78.1",
         "@kubernetes/client-node": "^1.4.0",
@@ -5427,7 +5426,6 @@
       "resolved": "https://registry.npmjs.org/express/-/express-5.2.1.tgz",
       "integrity": "sha512-hIS4idWWai69NezIdRt2xFVofaF4j+6INOpJlVOLDO8zXGpUVEVzIYk12UUi2JzjEzWL3IOAxcTubgz9Po0yXw==",
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "accepts": "^2.0.0",
         "body-parser": "^2.2.1",
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,6 +12,7 @@
         "@anthropic-ai/sdk": "^0.71.2",
         "@clack/prompts": "^1.1.0",
         "@earendil-works/pi-agent-core": "^0.78.1",
+        "@earendil-works/pi-ai": "0.78.1",
         "@earendil-works/pi-coding-agent": "0.78.1",
         "@earendil-works/pi-tui": "^0.78.1",
         "@kubernetes/client-node": "^1.4.0",
@@ -5426,6 +5427,7 @@
       "resolved": "https://registry.npmjs.org/express/-/express-5.2.1.tgz",
       "integrity": "sha512-hIS4idWWai69NezIdRt2xFVofaF4j+6INOpJlVOLDO8zXGpUVEVzIYk12UUi2JzjEzWL3IOAxcTubgz9Po0yXw==",
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "accepts": "^2.0.0",
         "body-parser": "^2.2.1",
--- a/package.json
+++ b/package.json
@@ -42,7 +42,6 @@
     "@anthropic-ai/sdk": "^0.71.2",
     "@clack/prompts": "^1.1.0",
     "@earendil-works/pi-agent-core": "^0.78.1",
-    "@earendil-works/pi-ai": "0.78.1",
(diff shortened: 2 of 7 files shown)


jacoblee-io (commented):
Reviewed. The reasoning-effort path is well done, but I'd hold approval on two points — one is a real scope/diff mismatch.

## 🔴 The `service_tier` half isn't in the diff

The title ("…reasoning effort + service tier") and the **How** section describe service tier in detail — pi-agent-brain stashing `serviceTier`, `agent-factory` wrapping `agent.streamFn` to stamp `options.serviceTier`, `openai-codex-responses` sending `body.service_tier`, http-server applying it — and the **Wire contract** says it reads `{ reasoning_effort, service_tier }`.

But the 4-file diff implements only `reasoning_effort`:
- `BrainModelParams` has only `reasoningEffort` (no `serviceTier`)
- `http-server` reads only `p.reasoning_effort`
- `agent-factory.ts` is not changed
- no `serviceTier` / `service_tier` token appears anywhere in the PR or its base branch

So the PR doesn't deliver what it advertises. Either the service-tier commits were dropped/not pushed, or the description is stale — please reconcile (add the service-tier code, or re-scope the title/body to reasoning effort only) before merge.

## 🟡 `reasoning_effort` doesn't reset when absent (contradicts the description)

The body says *"absent fields reset so values never leak across turns."* The code doesn't:
- absent `reasoning_effort` → `applyModelParams({ reasoningEffort: undefined })`
- `applyModelParams` only calls `setThinkingLevel` when the value is present **and** valid, so `undefined` is a no-op
- `setModel` (which would reset the level to the model default) is skipped when the model is unchanged (`needsUpdate === false`)

→ if the control plane sends `reasoning_effort` once and later omits it for the same model, the prior effort persists into later turns. Either reset to the session/model default on absent/invalid effort (capture the default), or drop the "reset" claim. Worth a test for the absent-after-present case.

## 🟢 The reasoning-effort code itself is solid
- `setThinkingLevel(level: ThinkingLevel)` is a real `AgentSession` method (no crash), and `THINKING_LEVELS` exactly matches pi's `ThinkingLevel` union `off|minimal|low|medium|high|xhigh` — invalid values are safely no-op'd.
- `applyModelParams` is correctly ordered after `setModel`; the optional-method guard keeps non-pi brains safe.
- Defensive `params` parsing; additive/backward-compatible; the new `applyModelParams` unit tests (valid/invalid/empty) pass (pi-agent-brain.test.ts 36/36 on the branch).

(Noting the PR targets `feat/codex-multimodal-vision`, not `main` — stacked, so it lands after that base. Not a blocker.)

jacoblee-io (approved):
Re-reviewed after the latest pushes — approving. ✅

Both points from my earlier comment are resolved:

1. **service_tier scope mismatch** → resolved by descoping: title is now "apply per-model reasoning effort", the body explicitly documents that the fast/`service_tier` knob was dropped (target plan 400s on `service_tier=fast`, and honouring it needed pi internals) with `agent-factory.ts` left unchanged, and there's no `serviceTier`/`service_tier` token in the diff. Title/body/code are consistent now.
2. **reset-on-absent** → the body no longer claims "absent fields reset"; it now accurately states that absent `reasoning_effort` is a no-op (does NOT reset to default), with rationale (SiCore always sends an effort for subscription models) and a note that cross-model reset-to-default is a known follow-up. The contradiction is gone; it's a documented, justified limitation.

The reasoning-effort code itself remains clean: `setThinkingLevel` is a real `AgentSession` method, `THINKING_LEVELS` matches pi's `ThinkingLevel` union exactly (invalid → no-op), `applyModelParams` runs after `setModel`, optional-method guard keeps non-pi brains safe, and `params` parsing is defensive. Verified locally: `tsc --noEmit` clean; pi-agent-brain + http-server suites green (87 tests).

Note: this is stacked on `feat/codex-multimodal-vision` (#326), so it should land after that base merges.

PR #332, opened 2026-06-15
feat(metrics): Grafana-style time range picker on a unified from/to contract

chent1996:
## What

Ports the Grafana-style time range picker to the Metrics dashboard, replacing the 3-option `period` enum (today/7d/30d) with quick relative ranges + an absolute From/To, and unifying the `/api/v1` metrics endpoints on a single `from`/`to` timestamp contract.

## Why

The dashboard filtered time with a `period` enum while the Audit tab carried its own independent range selector, so the two could silently disagree, and neither could express an absolute window. This mirrors the picker already shipped on the downstream sicore dashboard.

## Changes

**Backend — `src/portal/siclaw-api.ts`**
- `resolveWindow` + a shared `parseTs` helper (unix-ms or ISO; both-missing → trailing 7d; invalid / `from >= to` → 400, fail-fast, no silent fallback)
- `summary` + `timing`: drop the `period` enum, add a `created_at <= to` upper bound to all five windowed queries
- `audit`: rename `startDate`/`endDate` → `from`/`to` (keeps `BETWEEN`)
- The `/api/internal` adapter (downstream sicore Portal contract) is **left untouched**

**Frontend — `portal-web`**
- `useMetrics`: `TimeRange` model + `resolveRange`/`rangeLabel`/`isValidRange`; relative expressions resolve at fetch time (on range change or manual Refresh), so a relative range re-resolves to the current `now` on each fetch while an absolute one stays fixed. (The `/live` card polls every 30s but is window-less; the summary/timing cards do not poll on a timer.)
- `TimeRangePicker` (new): hand-rolled popover (no shadcn/date-fns in this repo) with a dependency-free mini calendar + searchable quick-range list; `buildMonthGrid` is unit-tested
- Audit tab unified under the header picker (drops its own range select); the window is frozen per filter change so pagination cursors don't drift on a sliding range

## Testing

- Backend: `tsc --noEmit` clean; 53 portal tests pass (incl. the migrated invalid-window → 400 test)
- Frontend: `tsc --noEmit` clean; 120 portal-web tests pass (incl. 7 new `buildMonthGrid` tests covering leap year / month rollover / DST); `vite build` succeeds

Design notes: `docs/design/2026-06-15-metrics-time-range-picker.md`

### DIFF
--- a/docs/design/2026-06-15-metrics-time-range-picker.md
+++ b/docs/design/2026-06-15-metrics-time-range-picker.md
@@ -1,0 +1,205 @@
+# Metrics Time Range Picker (Grafana-style) — Design
+
+> Status: draft · 2026-06-15
+> Scope: `portal-web` Metrics dashboard + `src/portal/siclaw-api.ts` `/api/v1/siclaw/metrics/*`
+> Ports the Grafana-style time picker already shipped in the downstream sicore web dashboard.
+
+## Background & goal
+
+The Metrics dashboard (`portal-web/src/pages/Metrics.tsx`) filters time with a 3-option
+`<select>` — Today / Last 7 days / Last 30 days — driving `summary` and `timing`. The Audit
+tab carries its **own, independent** range `<select>` (Last 1h/6h/24h/7d/30d), so the header
+period and the audit window can silently disagree. Neither path can express an **absolute**
+window ("2026-06-10 09:00 → 12:00").
+
+Goal: replace the period enum with a **Grafana-style time picker** — quick relative ranges
+plus an absolute From/To — and unify the backend `live`/`summary`/`timing`/`audit` endpoints
+on a single `from`/`to` timestamp contract, matching what audit already half-did with
+`startDate`/`endDate`. Clean break, no compatibility shim.
+
+## What stays untouched (contract boundary)
+
+`src/portal/adapter.ts` `/api/internal/siclaw/metrics/*` is a **separate, parallel**
+implementation (independent SQL, `period` enum) consumed by the downstream sicore Portal over
+the internal-auth channel. It is a cross-system wire contract and is **out of scope** — exactly
+mirroring how the sicore port left its own `adapter/rpc.go` alone. Only `/api/v1/*` (siclaw's
+own portal-web) migrates.
+
+```
+┌─────────────────┐   /api/v1/siclaw/metrics/*    ┌──────────────────────┐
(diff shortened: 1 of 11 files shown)


LikiosSedo:
## Review — Grafana-style time range picker

Direction looks good: clean contract migration, all SQL parameterized, the sicore `/api/internal` adapter genuinely untouched, and the `buildMonthGrid` calendar math is correct. Verdict: **approve-with-comments**. Two things worth fixing:

### should-fix

**① `extractTime` regex corrupts a draft time that carries seconds / ISO** (`TimeRangePicker.tsx`)
`extractTime`'s `/(\d{1,2}:\d{2})\s*$/` matches `30:45` instead of `10:30` when the draft has seconds (`10:30:45`); re-picking a day then rewrites the field to `2026-06-15 30:45` (unparseable). A `Z`/ISO `T` draft drops the time-of-day entirely. `pickDay` calls `extractTime(cur)` on every calendar click, so this is reachable. Suggest splitting on `T`/space to isolate the time segment and taking the first 5 chars, or `/\b(\d{1,2}:\d{2})(?::\d{2})?/`.

**② audit endpoint dropped the `from >= to` validation that summary/timing now enforce** (`src/portal/siclaw-api.ts`, audit handler)
summary/timing go through `resolveWindow` and return 400 on `from >= to`, but audit is still `parseTs(query.from) ?? default` with no check, so a reversed window is silently swallowed into an empty list by `BETWEEN`. The same malformed input now behaves differently across endpoints (one 400s, one silently returns empty), which contradicts the fail-fast posture. Suggest routing audit through `resolveWindow` as well; if you want to keep the different default windows (summary: both-missing → 7d; audit: each bound independently 24h/now), please document the tradeoff explicitly.

### nit
- `resolveWindow` is asymmetric for a single open bound: missing `to` defaults to now, missing `from` → 400. The frontend always sends both, so impact is limited, but worth unifying or commenting.
- The PR body says relative ranges "slide on the 30s live refresh", but `useLive` only fetches the window-less `/live`, and summary/timing have no timer — the cards do not actually slide every 30s (now is only re-resolved on range change / manual Refresh). The design notes get this right; suggest syncing the PR body wording.
- The parsing core (`resolveExpr`/`resolveRange`/`extractTime`) has zero test coverage; backend also lacks `audit?from=2000&to=1000` (reversed window) and the to-missing-default / from-missing→400 cases. A single `extractTime` assertion would have caught ①.

LikiosSedo (approved):

PR #333, opened 2026-06-16
feat(metrics): federate agentbox metrics through the Gateway

chent1996:
## Background & Problem

agentbox business metrics (token / prompt / tool / skill / prompt_duration, etc.) are accumulated only inside the agentbox process, yet an agentbox Pod is a medium-lived "one pod per agent, shared by all of that agent's sessions, recycled after 5 min idle" workload. Having Prometheus scrape these recyclable targets directly is structurally unsound:

- After a Pod is idle-recycled the target disappears, the counter series breaks, and `rate()` produces holes at the gap
- The last <one-scrape-interval of increments before recycling is lost — never scraped
- In the extreme (a short session that immediately idles out) sampling is sparse

## Approach

Move collection from "Prometheus scrapes short-lived Pods" to **application-layer federation in the Gateway**: the Gateway pulls each Pod's prom-client snapshot → **delta-accumulates** it into one stable, monotonic series → re-exports from `gateway:3001/metrics` (the single stable scrape target).

Three core correctness decisions:

- **Delta accumulation**: counters/histograms add only the increase since the last pull, so a Pod being destroyed can never make an exported counter go down — this is what defends against short-lived targets
- **Keyed on `(boxId, incarnation)`**: boxId is derived from agentId alone and is reused when a Pod is rebuilt, so it cannot by itself distinguish two process incarnations; adding a per-process start nonce makes a rebuilt Pod a fresh series counted in full, and correctly detects counter resets
- **Gauge whitelist**: only `siclaw_sessions_active` is summed across Pods (meaningful); per-session gauges like `context_tokens_*` are skipped (summing them across Pods is meaningless)

> Why not Pushgateway: its replace-not-accumulate semantics + unbounded dead-series buildup + the need to wire up mTLS separately inside the mesh do not fit this system's constraints (existing Gateway pull loop + mTLS + zero new dependencies).

## Changes (6 modules)

1. **Data contract + agentbox snapshot export** — extend `MetricsSnapshot` with `incarnation` + a cumulative `prom` field; add a `getMetricsAsJSON` wrapper and a per-process incarnation nonce
2. **Federation aggregator** `PromFederationAggregator` — delta accumulation, `(boxId,incarnation)` key, reset detection, gauge whitelist, `siclaw_` export admission, grace-delayed eviction, idempotent ingest
3. **Gateway wiring + merged export** — the pull loop feeds federation and reconciles live instances; `/metrics` merges deduplicated (federation registry + a dedicated self-monitoring registry, zero name overlap, avoiding duplicate `# TYPE` parse failures)
4. **Federator self-monitoring** — `siclaw_federation_*` metrics (pull failures / flush counts / last-success timestamp / tracked instances / series count); a prerequisite for removing 9090 (failure visibility)
5. **agentbox SIGTERM final flush** — push a final frame before graceful exit; **boxId is taken from the mTLS certificate, not the request body** (anti-spoofing); best-effort, never blocks Pod recycling
6. **Remove the direct-scrape path** — delete the agentbox 9090 plaintext metrics server + PodMonitor (helm + k8s) + the Pod's 9090 port; the Gateway ServiceMonitor becomes the sole scrape entry point

## ⚠️ Breaking Changes

- Removes the agentbox 9090 `/metrics` plaintext port, the `PodMonitor`, and the Pod's 9090 containerPort
- The old "graceful-exit scrape window" of 9090 is replaced by the final flush (module 5)
- Business metric names are unchanged, so **existing Grafana dashboards need no modification**

## Testing

- 5 new test files with full hard-acceptance coverage: incarnation isolation, delta/reset, idempotency, grace race, gauge whitelist, prefix admission, cert anti-spoofing, dedup with no TYPE collision, histogram bucket precision
- Full suite **3555 tests passing**; both `tsc` and `tsc -p tsconfig.agentbox.json` pass
- `helm template` render: 0 PodMonitor / 1 ServiceMonitor, no 9090 residue

## Review History

Went through two review rounds. The design review corrected a wrong "Pod lifecycle premise" (it is one-pod-per-agent with idle recycling, not one-pod-per-session) and closed gaps in podId sourcing, the dedup contract, Gateway-restart loss of accumulated values, and federator self-monitoring; the code review fixed a dynamic-import escape risk in the flush route and justified why immediate incarnation eviction is safe for the graceful-replacement case.


LikiosSedo:
## Review — federate agentbox metrics through the Gateway

The architecture is sound, and the delta / incarnation / dedup design is correct and well-tested. Verified items: mTLS cert anti-spoofing (pull uses `pod.metadata.name`, flush uses the cert `serialNumber`, body `boxId` is ignored); delta/reset math; histogram accumulation (`_bucket{le}` / `_sum` / `_count` + `+Inf`); `/metrics` dedup with no duplicate `# TYPE`; the dynamic-import fix; `ingest` is fully synchronous (no torn-map races); helm renders 0 PodMonitor / 1 ServiceMonitor.

Verdict: **approve-with-changes**. There are 3 silent counter-distortion edges + 1 operational gap worth handling before merge (this PR makes `:3001` the sole scrape path, so a distortion goes unnoticed and a misconfig means total metric loss). Each repro below was run in a standalone harness.

### should-fix

**① `seriesKey` does not escape label values → distinct series collapse and get summed** (`prom-federation-aggregator.ts`, `seriesKey()`)
The key joins `k=v` with `,` and does not escape. When a label value contains `,` or `=` (MCP `tool_name`, user-defined `skill_name`, `user_id`, `model` all can), two distinct series merge into one and their values add up. Repro: `{a:"1,b=2"}` and `{a:"1",b:"2"}` collapse into one (1 family, expected 2). The `NUL=" "` comment only addresses the family separator, not the `,`/`=` inside `entries`. Suggest a sorted `JSON.stringify(entries)` key or a separator that cannot collide.

**② A late frame after eviction is treated as first-seen and re-counted in full** (same file, `retainInstances` / `ingest`)
Grace is only one extra round (~60s @ 30s interval). Once an instance is evicted (`lastSeen` deleted), a late pull / SIGTERM flush of the same incarnation is treated as first-seen and re-counted in full, inflating a monotonic counter. Repro: 50 → 100. The class doc's "idempotency does not depend on pull/flush arrival order" only holds while `lastSeen` survives — not after eviction. Suggest keeping a short TTL that rejects frames for a just-evicted `(boxId, incarnation)`, or lengthening grace.

**③ A reordered stale-incarnation frame inflates the counter** (same file, `evictStaleIncarnations`)
`pullAll` uses `Promise.all`, so an in-flight fetch can resolve out of order; during a Pod rebuild with two incarnations of the same boxId, a late old-incarnation frame is treated as first-seen and added. Repro: inc1=100 → inc2=10 (+10=110) → reordered inc1=120 (+120=230). Same root cause as ②; the TTL blocklist would cover this too.

**④ ServiceMonitor sends no token, but `/metrics` still goes through `checkMetricsAuth`** (`gateway-servicemonitor.yaml` + `server.ts`)
With no token configured it fails open, which is fine; but once `SICLAW_METRICS_TOKEN` is set, every Prometheus scrape 401s, and self-monitoring lives on the same endpoint so it 401s too — exactly the "metrics went flat, indistinguishable from idle" mode this PR warns about. Since this PR makes `:3001` the only scrape entry point, the misconfig means total metric loss. Suggest the ServiceMonitor support `bearerTokenSecret`, or document in values that the token must be unset in K8s.

### nit
- Gauge whitelist: `ingestGauge` only overwrites samples present in the current frame. The only whitelisted gauge `siclaw_sessions_active` is label-less (prom-client always emits 0), so no impact today; but a future labeled gauge would retain a stale value when a label combo stops being reported, inflating the cross-Pod sum. Suggest a note next to `FEDERATABLE_GAUGES`.
- `flushReceivedTotal.inc()` runs before body validation, so a malformed frame increments both `received` and `errors`; `received` semantics are slightly off (a test locks this, so ignore if intentional).
- `CHANGELOG.md` still advertises the ":9090 metrics server / metrics containerPort" as added features; this PR removes both — suggest a removed entry.

chent1996:
Thanks for the thorough review — all four findings and the nits are addressed in 42b521d. Per-item:

### should-fix

**① seriesKey escaping — fixed.** `seriesKey` now joins every field with `NUL` (which cannot appear in a metric name / label name / label value) instead of `=`/`,`. `{a:"1,b=2"}` and `{a:"1",b:"2"}` no longer collapse. Locked by a test that asserts the two stay distinct and never sum to 12.

**②③ over-count after eviction — fixed at the root, both with one mechanism.** The cause was deleting the counter baseline (`lastSeen`) on eviction, so a returning frame was treated as first-seen and re-counted in full. Now:
- counter baselines are **retained** (LRU-capped at 4096, pruned least-recently-ingested-first — bounded memory, never prunes a live instance);
- **gauges are still reconciled promptly** (grace eviction + stale-incarnation eviction) for cross-pod sum correctness.

So a late flush/pull (②) and a reordered stale-incarnation frame (③) both compute `max(0, cur-last)` against the retained baseline. New tests reproduce your exact scenarios: ② `50 → evict → 100` ⇒ 100 (not 150); ③ `inc1=100 → inc2=10 → reordered inc1=120` ⇒ 130 (not 230). The class docs now state the retention contract explicitly.

> Note on reachability: through the real plumbing these are hard to hit (pull has a 3s timeout and is intra-cycle; the SIGTERM flush completes before the pod is deleted; one boxId can't have two live endpoints). But you're right that the *stated* contract was stronger than the code, and on the now-sole scrape path it's worth making the contract actually hold — hence the fix rather than a doc caveat.

**④ ServiceMonitor auth — fixed.** Added an optional `metrics.serviceMonitor.bearerTokenSecret`; when set, the ServiceMonitor sends `Authorization: Bearer`. values.yaml documents that it must match `SICLAW_METRICS_TOKEN` (and that the token is unset/fail-open by default, relying on NetworkPolicy). Verified `helm template` renders it only when set.

### nits
- **gauge stale value** — added a note next to `FEDERATABLE_GAUGES`: the only entry is label-less so it can't happen today; before adding a labeled gauge, reconcile per-frame disappearance of label combos.
- **CHANGELOG** — corrected: the unreleased entries no longer advertise the `:9090` server / `metrics` container port / AgentBox PodMonitor; they now describe Gateway federation.
- **`flushReceivedTotal` before validation** — kept intentionally. `received` = all flush requests received, `errors` = the subset that failed; this `received ⊇ errors` relationship is the conventional one (cf. `http_requests_total` vs `http_errors_total`), and a test locks it.

Full suite green (3559 passing); both `tsc` and `tsc -p tsconfig.agentbox.json` pass.

LikiosSedo (approved):

PR #334, opened 2026-06-16
fix(lark-card): stop down-converting headings & tables for schema-2.0…

glow-Kev1n:
… markdown

The feishu reply card uses CardKit schema 2.0, whose markdown element renders ATX headings and GFM pipe tables natively. The old sanitizer down-converted them, which caused two rendering bugs:
- heading -> **...** emitted literal ** asterisks when the bold marker glued onto adjacent CJK/emoji text;
- table -> fenced code block rendered as a truncated, line-numbered, horizontally-scrolling monospace box instead of a real table.

Pass headings and tables through unchanged so the 2.0 markdown element renders them natively. Blockquote->｜ rewrite kept as-is. Tests updated to assert pass-through.

## Test Plan

- [ ] `npm test` passes

## Architecture Checklist

## General Checklist

### DIFF
--- a/src/gateway/channels/lark-card.test.ts
+++ b/src/gateway/channels/lark-card.test.ts
@@ -29,13 +29,16 @@ describe("sanitizeMarkdownForFeishu", () => {
     expect(sanitizeMarkdownForFeishu(md)).toBe(md);
   });
 
-  it("turns ATX headings into bold lines", () => {
-    expect(sanitizeMarkdownForFeishu("# Title")).toBe("**Title**");
-    expect(sanitizeMarkdownForFeishu("## Subtitle\nbody")).toBe("**Subtitle**\nbody");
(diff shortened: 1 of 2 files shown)


LikiosSedo:
Looks good overall. My read is that this is specifically letting the CardKit JSON 2.0 markdown element render ATX headings and GFM pipe tables natively, rather than adding a separate chart/visualization renderer. That matches the current `schema: "2.0"` card path and avoids the old heading-to-bold / table-to-codeblock degradation.

One small follow-up: the top-of-file design note in `src/gateway/channels/lark-card.ts` still says H1-H3 and GFM tables are unsupported and normalized by `sanitizeMarkdownForFeishu`. Since this PR intentionally reverses that for schema 2.0, please update that comment too so future maintainers do not reintroduce the down-conversion.

PR #340, opened 2026-06-22
Add A2A gateway for Siclaw agents

LikiosSedo:
## Summary

- add a Portal-hosted A2A HTTP+JSON gateway for Siclaw agents
- expose Agent Card discovery, message send/stream, task polling/listing, subscribe, and cancel routes
- add an `a2a_tasks` projection table plus tests for auth, streaming, task lifecycle, context mapping, and route parsing
- document the A2A design, standards alignment, deferred work, and future Sicore enablement options

## Validation

- `npm run build`
- `npm test -- src/portal/a2a-gateway.test.ts src/portal/migrate-sqlite.test.ts src/gateway/rest-router.test.ts src/portal/chat-gateway.test.ts src/portal/chat-gateway-events.test.ts`
- `npm test`
- `git diff --check`

### DIFF
--- a/src/portal/a2a-gateway.ts
+++ b/src/portal/a2a-gateway.ts
@@ -1,0 +1,1014 @@
+import crypto from "node:crypto";
+import type http from "node:http";
+import {
+  parseBody,
+  parseQuery,
+  RequestBodyTooLargeError,
+  type RestRouter,
+} from "../gateway/rest-router.js";
+import { getDb } from "../gateway/db.js";
+import type { RuntimeConnectionMap } from "./runtime-connection.js";
+import { authenticateApiKey, type ApiKeyAuthResult } from "./api-key-auth.js";
+import { resolveAgentModelBinding } from "./chat-gateway.js";
+
+const A2A_VERSION = "1.0";
+const A2A_JSON = "application/a2a+json; charset=utf-8";
+const MAX_A2A_BODY_BYTES = 1024 * 1024;
+const MAX_A2A_ID_BYTES = 255;
+const ASSISTANT_ARTIFACT_ID = "assistant-text";
+
+type A2aTaskState =
+  | "TASK_STATE_SUBMITTED"
+  | "TASK_STATE_WORKING"
+  | "TASK_STATE_COMPLETED"
+  | "TASK_STATE_FAILED"
+  | "TASK_STATE_CANCELED"
+  | "TASK_STATE_REJECTED";
+
+const A2A_TASK_STATES = new Set<A2aTaskState>([
+  "TASK_STATE_SUBMITTED",
+  "TASK_STATE_WORKING",
+  "TASK_STATE_COMPLETED",
+  "TASK_STATE_FAILED",
+  "TASK_STATE_CANCELED",
+  "TASK_STATE_REJECTED",
+]);
+
+interface A2aMessage {
+  messageId?: string;
+  contextId?: string;
+  taskId?: string;
+  role?: string;
+  parts?: Array<{ text?: string; raw?: string; url?: string; data?: unknown; mediaType?: string }>;
+  metadata?: Record<string, unknown>;
+}
+
+interface NormalizedA2aMessage extends A2aMessage {
+  messageId: string;
+  contextId?: string;
+}
+
+interface SendMessageRequest {
+  message?: A2aMessage;
+  configuration?: Record<string, unknown>;
+  metadata?: Record<string, unknown>;
+}
+
+interface A2aTaskRecord {
+  id: string;
+  agentId: string;
+  userId: string;
+  apiKeyId: string | null;
+  contextId: string;
+  sessionId: string;
+  state: A2aTaskState;
+  statusMessage: string | null;
+  artifactText: string;
+  error: string | null;
+  createdAt: string | null;
+  updatedAt: string | null;
+  lastEventAt: string | null;
+  completedAt: string | null;
+}
+
+type A2aStreamResponse =
+  | { task: Record<string, unknown> }
+  | { message: Record<string, unknown> }
+  | { statusUpdate: Record<string, unknown> }
+  | { artifactUpdate: Record<string, unknown> };
+
+interface ActiveTracker {
+  unsubscribe: () => void;
+  artifactText: string;
+}
+
+const activeTrackers = new Map<string, ActiveTracker>();
+const streamSubscribers = new Map<string, Set<(response: A2aStreamResponse) => void>>();
+
+class A2aHttpError extends Error {
+  constructor(readonly status: number, readonly code: string, message: string) {
+    super(message);
+  }
+}
+
+function isTerminalState(state: A2aTaskState): boolean {
+  return state === "TASK_STATE_COMPLETED"
+    || state === "TASK_STATE_FAILED"
+    || state === "TASK_STATE_CANCELED"
+    || state === "TASK_STATE_REJECTED";
+}
+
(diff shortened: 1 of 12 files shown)


src/portal/a2a-gateway.ts
```diff
@@ -0,0 +1,1014 @@
...
+    const text = name ? `Finished tool: ${name}` : "Siclaw tool finished";
+    await setTaskState(task.id, "TASK_STATE_WORKING", text);
+    emitTask(task.id, buildStatusUpdate(task, "TASK_STATE_WORKING", text, name ? { currentTool: name } : undefined));
+    return;
+  }
+
+  if (evt.type === "stream_error") {
+    const message = streamErrorMessage(evt);
+    await setTaskArtifact(task.id, tracker.artifactText);
+    await setTaskState(task.id, "TASK_STATE_FAILED", message, message);
```
chent1996, line 655: **Real correctness race: `stream_error` unconditionally overwrites terminal states.**

Compare line 661–666: `prompt_done` re-reads the task and bails out if it's already terminal. `stream_error` here does not. The race scenario:

1. User calls `/tasks/:taskId:cancel`. The handler awaits `chat.abort`.
2. Runtime fires `stream_error` as part of the abort. The tracker callback enqueues `handleTrackedEvent`, which awaits `setTaskArtifact`.
3. The cancel handler's `await chat.abort` resolves first → it writes `TASK_STATE_CANCELED` → emits `STATUS_CANCELED` → `stopTaskTracker`.
4. The in-flight tracker promise resumes → `setTaskState(FAILED)` **overwrites the user's CANCELED**, emits `STATUS_FAILED`.

Net: user clicked cancel, DB says FAILED with `status_message="Siclaw stream failed"`. Misleading. Fix is to mirror the `prompt_done` guard:

```typescript
if (evt.type === "stream_error") {
  const latest = await loadTaskRecordById(task.id);
  if (latest && isTerminalState(latest.state)) { stopTaskTracker(task.id); return; }
  const message = streamErrorMessage(evt);
  await setTaskArtifact(task.id, tracker.artifactText);
  ...
}
```

Or — cleaner — make `setTaskState` terminal-immutable by adding `AND state NOT IN (...terminal states...)` to the UPDATE WHERE clause. That fixes the symmetric case too (cancel racing a `prompt_done`).

src/portal/a2a-gateway.ts
```diff
@@ -0,0 +1,1014 @@
...
+
+  const tracker: ActiveTracker = {
+    artifactText: task.artifactText,
+    unsubscribe: () => {},
+  };
+
+  tracker.unsubscribe = connectionMap.subscribe(task.agentId, "chat.event", (data: unknown) => {
+    const envelope = data as { sessionId?: string; event?: Record<string, unknown> } | undefined;
+    if (!envelope?.event) return;
+    if (envelope.sessionId && envelope.sessionId !== task.sessionId) return;
```
chent1996, line 611: `envelope.sessionId &&` makes the filter permissive: when the runtime emits a `chat.event` without `sessionId` (or with an empty string), every active tracker for the same `agentId` processes it. That can leak another session's text deltas into this task's `artifactText`, and trigger spurious `tool_execution_*` status updates.

In practice Runtime always sets `sessionId`, so this is latent — but the A2A tracker is colder-state than the chat-gateway path (artifacts get **persisted** to `a2a_tasks.artifact_text` at terminal), so a corruption here would survive longer. Same pattern exists in `chat-gateway.ts:382` (`if (envelope.sessionId && envelope.sessionId !== sessionId) return;`) — worth tightening across both. Strictly: `if (envelope.sessionId !== task.sessionId) return;`. Non-blocking; pre-existing pattern, not introduced here.

src/portal/migrate.ts
```diff
@@ -312,6 +312,26 @@ const PORTAL_SCHEMA_SQLS: string[] = [
...
+    session_id CHAR(36) NOT NULL,
+    state VARCHAR(40) NOT NULL,
+    status_message TEXT,
+    artifact_text TEXT,
+    error TEXT,
+    created_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    updated_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    last_event_at TIMESTAMP NULL DEFAULT NULL,
+    completed_at TIMESTAMP NULL DEFAULT NULL,
+    CONSTRAINT fk_a2a_tasks_agent FOREIGN KEY (agent_id) REFERENCES agents(id) ON DELETE CASCADE
```
chent1996, line 332: **No indexes on `a2a_tasks` for non-PK lookups.**

The PK on `id` covers `loadTaskRecord` and `loadTaskRecordById`. But:
- `loadSessionIdForContext` filters `WHERE agent_id = ? AND api_key_id = ? AND context_id = ?` — every `message:send` with a `contextId` does a full table scan on this.
- `listTaskRecords` filters by `(agent_id, api_key_id [, context_id, state])` with `ORDER BY created_at DESC` — full scan + sort.

At MVP scale fine, but every accepted A2A request triggers the `context_id` lookup. Suggest:
```sql
INDEX idx_a2a_tasks_agent_key_ctx (agent_id, api_key_id, context_id),
INDEX idx_a2a_tasks_agent_key_created (agent_id, api_key_id, created_at)
```
CLAUDE.md §5 only constrains *legacy* index names; new indexes on a new table are fine. Non-blocking, but easier to add now than to migrate later.

chent1996 (approved):
**Solid A2A adapter. Approving with three notes.**

The design doc is excellent and the architectural placement is right (Portal as A2A boundary; Runtime untouched). I traced the auth, body limits, ID limits, page validation, route compilation change, migration DDL, in-memory tracker lifecycle, and SSE snapshot-before-deltas fix in the second commit. Things I verified:

- `agent_api_keys` hash lookup + `agentId` route binding → key rotation is a clean access-control boundary.
- `a2a_tasks` DDL respects the invariant §5 constraints (no `TIMESTAMP(3)`, no `ON UPDATE`, no `JSON`); explicit `SET updated_at = CURRENT_TIMESTAMP` is consistent with how the chat tables work; `ON DELETE CASCADE` to `agents` prevents orphan rows.
- `rest-router` `compilePath` rewrite is necessary for `:taskId:cancel` action suffixes and doesn't break any existing route — I enumerated all registered paths and none use multi-`:` segments outside the new A2A handlers.
- `replicas: 1` is explicitly documented in `values.yaml` with the rationale (in-process trackers + single Runtime phone-home), which closes what would otherwise be a multi-replica correctness landmine.
- `__resetA2aGatewayState` for graceful shutdown / test isolation — good.
- `activeTrackers` lifecycle: created in `submitA2aTask` and `streamTask`, idempotent on re-entry; stopped on every terminal branch. `streamSubscribers` cleaned up on `endStream`. No leaks I can see.
- The second commit's snapshot-before-deltas buffer (`buffered.push` until `snapshotSent`) is the right call — synchronous subscribe + buffered flush guarantees A2A clients see the Task frame first while not losing live events during the snapshot's Runtime round-trip.

Three notes inline. None blocking, but #1 is an actual (rare) correctness race that's worth fixing in this PR or a fast follow.

Note: I could not run the suite locally (no `node_modules` in this review checkout). Reviewed by reading; trusting the author's `npm test` validation.

Minor non-anchored observations:
- Agent Card is returned for any `:agentId` without checking whether it exists in `agents`. Subsequent calls 401/403 on auth mismatch, so no enumeration exposure, but it can mislead discovery clients. Cheap to add an existence check.
- `writeA2aSse` doesn't observe `res.write`'s return value (backpressure). Pre-existing pattern in `chat-gateway.ts`; not introduced here.

PR #341, opened 2026-06-22
feat(metrics): external-facing showcase dashboard with daily trend charts

chent1996:
## What

Repurposes the `Metrics → Dashboard` tab from internal ops telemetry into an **external-facing showcase** (scale + value), reading **only Portal DB**, desensitized to counts/distributions/trends — no individual drill-down.

## Changes

**Backend** (`src/portal/siclaw-api.ts` — `/api/v1/siclaw/metrics/summary`)
- New aggregates, all sharing the existing card filters so figures stay consistent:
  - `distinctUsers`, `toolCalls` (total), `skillsUsed` (distinct skills parsed from `tool_input` across the 4 script-exec tools, `ROW_LIMIT`-capped)
  - `inventory` snapshot: clusters / hosts / skills (overlay-excluded) / knowledge / agents / mcp
  - `dailySeries`: per-day prompts & tool calls (same `WHERE` as the scalar cards → each chart's Total equals its KPI card)
- `byUser` kept server-side (adapter REST + RPC mirrors also return it, locked by tests) — front-end simply stops rendering it; no protocol drift.

**Frontend** (`portal-web/`)
- Reorder: Managed scale (snapshot) → Usage (windowed) → Usage trend (daily charts).
- New `TrendChart` SVG component (smooth area + gradient + axis + Total/avg/Peak) — **no chart-lib dependency**.
- `KpiCards` split into two labelled groups; live / timing / By-User cards removed from the dashboard.
- User filter scoped to the Audit tab (external view never drills into individuals / exposes usernames).
- Trend shown for all periods (1–2 points at 24h); long-range x-axis uses ~7 evenly-spaced `MM-DD` labels with edge-anchored ends.

## Boundary / design

- Stays entirely within the Portal DB (no gateway/agentbox coupling); window queries are fixed-width so cost is bounded by recent activity, not total history.
- No new dependencies; chart is hand-rolled SVG.

## Testing

- `src/portal/siclaw-api.misc.test.ts` extended: skill-name parsing (JSON + regex fallback + dedup + skip-missing), daily-series gap-fill summing to period totals, inventory shape. **30/30 pass.**
- Type-clean on all touched files.

### DIFF
--- a/portal-web/src/components/metrics/KpiCards.tsx
+++ b/portal-web/src/components/metrics/KpiCards.tsx
@@ -1,12 +1,14 @@
-import { MessageSquare, Bot, Activity, Wrench } from "lucide-react"
+import { Users, MessageSquare, Activity, Wrench, Zap, Package, Server, FileCode, BookOpen, Bot, Plug } from "lucide-react"
 
 interface Props {
+  rangeLabel: string
+  distinctUsers: number
   totalSessions: number
   totalPrompts: number
-  activeSessions: number
-  wsConnections: number
-  toolCallsTotal: number
-  rangeLabel: string
+  toolCalls: number
+  skillsUsed: number
+  skillsUsedApprox?: boolean
+  inventory: { clusters: number; hosts: number; skills: number; knowledgeRepos: number; agents: number; mcpServers: number }
 }
 
 function fmt(n: number): string {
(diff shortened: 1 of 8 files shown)


LikiosSedo:
## Review — external-facing showcase dashboard

The dashboard logic is solid and well-tested (chart geometry, skill parsing, the `skillsUsedApprox` flag honestly surfaced, `toolCalls` an uncapped `COUNT(*)`, and TrendChart's empty/1-point/all-zero edges all safe — no NaN/divide-by-zero). But one item directly defeats the "desensitized external view" goal. Verdict: **request-changes**.

### blocking

**① `byUser` (raw userIds + per-user session/message counts) is still serialized to the browser** (`src/portal/siclaw-api.ts`, summary handler)
This endpoint is re-scoped as a desensitized, no-individual-drill-down external view, yet the handler still queries and returns `byUser` — the frontend simply stops rendering it, but the data still ships over the wire and is visible in the Network panel. "Frontend stops rendering" ≠ "removed from the response", and per-user data is exactly what the PR claims to desensitize. Please drop the `byUser` query and field from the handler (it has no frontend consumer left) and remove `SummaryData.byUser`. If it's still needed internally, expose it on a separate internal-only endpoint. (There is a `requireAdmin` gate, so this isn't anonymous exposure, but it's an unnecessary data surface and contradicts the stated goal.)

### should-fix

**② adapter REST + RPC mirrors were not updated → "no protocol drift" is not true** (`src/portal/adapter.ts`, REST + RPC `metrics.summary`)
The portal `/metrics/summary` gained 6 new fields (`distinctUsers` / `toolCalls` / `skillsUsed` / `inventory` / `dailySeries`), but both mirrors still return only `{ totalSessions, totalPrompts, byUser }`, so the three implementations are no longer equivalent. No in-repo consumer today, so impact is low, but the "no protocol drift" claim doesn't hold. When dropping `byUser`, remove it from all three.

**③ `dailySeries` "Total == KPI card" breaks when the DB timezone ≠ UTC, and the tests structurally cannot catch it** (`src/portal/siclaw-api.ts`, `bucketByDay` + gap-fill)
`DATE(m.created_at)` buckets by the DB's local day, but `bucketByDay` / gap-fill build keys via `toISOString()` (UTC). When the DB is non-UTC and the driver returns Date objects, a row near midnight is bucketed to the previous UTC day's key, which may fall outside the gap-fill window and be dropped — so the chart Total no longer equals the KPI scalar (a pure timestamp `COUNT(*)`, TZ-agnostic). The tests pass only because the mock keys are pre-computed with the same UTC `toISOString` as the handler, so the skew can never surface in test.

> ⚠️ Note: the production clusters are all UTC+8 (Malaysia / Shanghai / Beijing), so a non-UTC DB is the norm, not an edge case. This is likely to show up in production — a customer-facing dashboard whose totals don't add up. Before merge, please confirm whether mysql2 runs with `dateStrings:true`; if not, unify the timezone basis (`DATE(CONVERT_TZ(...))` or force `dateStrings`) and add a "driver returns Date object + non-UTC" regression test.

### nit
- `dailySeries` length is `seriesDays + 1` (7d → 8 points, 30d → 31), intentional (today + N prior days), but "7d yields 8 points" is mildly surprising — a one-line comment would help.
- The cap-exceeded branch (`skillsUsedApprox = true`) and the TZ-skew path currently have zero test coverage — the two highest-risk branches.

LikiosSedo (approved):

PR #342, opened 2026-06-22
Add native PDF media input support

LikiosSedo:
## Summary
- Add unified prompt media support for images and PDF files through AgentBox HTTP, gateway RPC, brain sessions, and model routing.
- Route PDF-capable requests only to candidates that advertise `pdf` input, while preserving image capability filtering.
- Convert PDF media into OpenAI file payload shapes for Responses and Chat Completions paths.

## Tests
- `npm run build`
- `npm test -- --run src/agentbox/http-server.test.ts src/core/model-routing.test.ts src/core/brains/pi-agent-brain.test.ts src/gateway/agentbox/client.test.ts src/core/openai-file-payload.test.ts`

### DIFF
--- a/src/agentbox/http-server.ts
+++ b/src/agentbox/http-server.ts
@@ -22,15 +22,21 @@ import { getSyncHandler, createClusterHandler, createHostHandler } from "./sync-
 import { GATEWAY_SYNC_DESCRIPTORS, type AgentBoxSyncHandler, type GatewaySyncType } from "../shared/gateway-sync.js";
 import { detectLanguage } from "../shared/detect-language.js";
 import {
+  candidateSupportsPromptMedia,
   clearModelRouteUserSelectionIfDifferent,
+  filterCandidatesForPromptMedia,
   markModelRouteUserSelection,
+  normalizeCandidates,
   normalizeModelRoutePolicy,
+  requiredInputsForPromptMedia,
   runPromptWithModelRouting,
   shouldUseModelRouteRunner,
+  unsupportedPromptMediaMessage,
+  type ModelRouteCandidate,
   type ModelRouteEvent,
   type ModelRoutePolicy,
 } from "../core/model-routing.js";
-import type { BrainSession, PromptImage } from "../core/brain-session.js";
+import type { BrainSession, PromptFile, PromptImage, PromptMedia } from "../core/brain-session.js";
 
 type RequestHandler = (
   req: http.IncomingMessage,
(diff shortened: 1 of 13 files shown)


src/agentbox/http-server.ts
```diff
@@ -128,6 +102,88 @@ const MAX_BODY_SIZE = 10 * 1024 * 1024; // 10MB
 const BODY_TIMEOUT_MS = 30_000; // 30s
 const DP_ACTIVATION_MARKER = "[Deep Investigation]\n";
 const DP_EXIT_MARKER = "[DP_EXIT]";
+const MAX_PROMPT_IMAGES = 4;
+const MAX_PROMPT_IMAGE_BASE64_CHARS = 2 * 1024 * 1024;
+const SUPPORTED_PROMPT_IMAGE_MIMES = new Set(["image/png", "image/jpeg", "image/webp"]);
+const MAX_PROMPT_FILES = 4;
+const MAX_PROMPT_FILE_BASE64_CHARS = 8 * 1024 * 1024;
```
chent1996, line 109: **Per-file cap × file count exceeds the request body cap.** `MAX_PROMPT_FILE_BASE64_CHARS = 8 MiB × MAX_PROMPT_FILES = 4` ⇒ 32 MiB worst case of file data, but `parseJsonBody` enforces `MAX_BODY_SIZE = 10 MiB` (line 101). Two max-size PDFs always reject at the body parser before `sanitizePromptFiles` ever runs, with a vague `Error("Body exceeds 10485760 byte limit")`.

Single 8 MiB PDF + envelope (~8.4 MiB total) still fits, so one-PDF workflows are fine — but the contract is broken for the documented 4-file limit. Either:
- raise `MAX_BODY_SIZE` to ≥ `MAX_PROMPT_FILES × MAX_PROMPT_FILE_BASE64_CHARS + envelope` (~33 MiB), and audit AgentBox's memory budget for buffering that, or
- lower `MAX_PROMPT_FILE_BASE64_CHARS` so `4 ×` fits comfortably under 10 MiB (≈ 2 MiB each), or
- explicitly say only one PDF per request (`MAX_PROMPT_FILES = 1` when files contain PDFs).

While here: `parseJsonBody` rejects with a plain `Error`, which the route handler turns into a 500 — would be better to bubble a 413 like `chat-gateway.ts` does via `RequestBodyTooLargeError`.

src/core/openai-file-payload.ts
```diff
@@ -0,0 +1,126 @@
+import type { PromptFile } from "./brain-session.js";
+
+const PDF_DATA_URL_PREFIX = "data:application/pdf;base64,";
+const DEFAULT_PDF_FILENAME = "attachment.pdf";
+const MAX_REMEMBERED_FILES = 256;
+const rememberedPdfFilenames = new Map<string, string>();
```
chent1996, line 6: **`rememberedPdfFilenames` is a process-global Map keyed by full data URLs — up to ~2 GiB of memory for an ~8-char filename lookup.**

Each key is `data:application/pdf;base64,<entire PDF bytes>` — for an 8 MiB PDF, that's an 8 MiB+ key string. `MAX_REMEMBERED_FILES = 256` means the steady-state cap is `256 × 8 MiB ≈ 2 GiB` just to remember filenames. In K8s mode this is per-pod (bounded but excessive); in local mode (one process for all users per CLAUDE.md §1.1) it accumulates across users, and identical PDFs uploaded by two users overwrite each other's filename mapping.

The fragility goes beyond memory: this is an **out-of-band lookup** tied to call ordering — `rememberPromptFiles` (in pi-agent-brain) must run before `convertOpenAIPdfPayload` (in agent-factory's `onPayload`). If pi-coding-agent ever caches and replays a payload after eviction (e.g., a queued retry), filenames silently fall back to `"attachment.pdf"`.

Two cleaner approaches:
1. Key by `crypto.createHash("sha256").update(file.data).digest("hex").slice(0,16)` — ~32-byte keys, no semantic change.
2. **Better**: pack the filename into the prompt content from the start (e.g., emit `{type:"file", data, mimeType, filename}` and have the payload hook read filename **from the content array** rather than from a module global). That eliminates the global entirely.

Option 2 is the architectural fix; option 1 is the quick mitigation.

src/core/model-routing.ts
```diff
@@ -506,14 +541,24 @@ export async function runPromptWithModelRouting(
...
+  media?: PromptMedia,
 ): Promise<ModelRouteRunResult> {
   if (!isModelRoutePolicyEnabled(policy)) {
-    await brain.prompt(text, images);
+    await brain.prompt(text, media);
     return { success: true, exhausted: false, attempted: [] };
   }
 
-  const candidates = normalizeCandidates(policy.candidates);
+  const candidates = filterCandidatesForPromptMedia(normalizeCandidates(policy.candidates), media);
```
chent1996, line 551: Small but real fragility: this filter silently picks a different primary when the user's configured primary is text-only and a fallback is PDF-capable. `primaryCandidate = candidates[0]` after filtering can be a different model than the user-configured one for non-PDF turns in the same session. Behaviorally correct (matches the PR description), but invisible in logs.

Suggest emitting a single `console.warn` or `guardLog`-style note when the filtered primary differs from the input list's primary, so operators can see "prompt contained PDF; rerouted from gpt-4o-text to gpt-4o-pdf". Non-blocking.

chent1996 (approved):
**Clean plumbing for native PDF support; approving with three notes.**

I traced the end-to-end path: HTTP body → `sanitizePromptFiles` → `PromptMedia { images, files }` → `brain.prompt(text, media)` → `rememberPromptFiles` + `session.prompt({ images: [...{type:"file",...}] })` → pi-coding-agent emits an OpenAI payload → `onPayload` hook → `convertOpenAIPdfPayload` rewrites PDF `input_image` parts into `input_file` (Responses) / `file` (Chat Completions). Capability filtering in `runPromptWithModelRouting` correctly skips candidates that don't advertise `"pdf"` in their `model.input` list.

Things I verified are right:
- `BrainSession.prompt` signature change from `images?` to `media?` is migrated at every implementation and call site (`PiAgentBrain`, both `model-routing` paths). No external break.
- Sanitizer enforces mimetype, base64 charset (`BASE64_RE`), `data.length % 4 === 0`, `MAX_PROMPT_FILES`, basename-only filename, no CR/LF. Solid.
- `convertOpenAIPdfPayload` is idempotent (second pass is a no-op when nothing matches), so wrapping it around `previousOnPayload` won't double-convert.
- `filterCandidatesForPromptMedia` returns the *same order* as the original candidate list, preserving the routing priority; an explicit `unsupportedPromptMediaMessage` surfaces a `format_error` rather than silently trying a text-only model with a PDF.

Three inline notes below. The body-size and Map design points are non-blocking but worth fixing in this PR.

Body-only observations (no anchor needed):
- Today `chat-gateway.ts` doesn't yet forward `files` to the gateway RPC, so the new path is reachable only via direct `chat.send` callers (a2a-gateway, programmatic clients). The plumbing lands ahead of the Portal wiring — whoever lights up the Portal side will be the first to hit the body-size mismatch in #1.
- `filterCandidatesForPromptMedia` may re-elect a different primary when the user's configured primary is text-only and a fallback is PDF-capable. This is intentional per the PR description, but currently silent; consider a one-line `console.warn` when the primary changes due to media capability so operators can see auto-rerouting in logs.

Note: I could not run the suite locally (no `node_modules` in this review checkout). Reviewed by reading; trusting the author's reported test invocations.

PR #344, opened 2026-06-22
refactor(model-routing): route every prompt through one runner entry

LikiosSedo:
## Summary

Collapse the dual model-routing prompt entry into a single one. Every prompt now flows through `runPromptWithModelRouting`; the `routeEnabled ? runner : brain.prompt` ternary (duplicated in `http-server.ts` and `session.ts`) is gone. This makes downstream event collection uniform — every turn carries its model identity — which is the prerequisite for clean per-model Langfuse/OTel attribution.

### Problem

The prompt path forked on `shouldUseModelRouteRunner`: the routing runner for real fallback routing, a bare `brain.prompt` otherwise. The gate (`_routeBrainEventsThroughExtra`) leaked into SSE, persistence (`sse-consumer.ts`), and the tracing recorder — each had to branch on which path produced the events. Worse, a non-routed turn emitted no `model_route_*` events, so anything reconstructing per-model usage/cost had no model identity for the common case.

The bare path only existed to preserve live streaming back when the runner was pure-buffering. `optimisticPrimaryStream` (already on `main`) removed that constraint — a single-candidate run streams live and is behaviourally identical to a bare prompt.

### Solution

New helper `resolveEffectivePolicy(configured, state, currentModel)`:
- real multi-candidate routing applies → the configured policy, unchanged;
- otherwise (routing off / single candidate / user-pinned) → a single-candidate policy built from the current model;
- no current model → `undefined` (the runner's own guard does a bare `brain.prompt`).

Both callers invoke the runner unconditionally. `_routeBrainEventsThroughExtra` now means "a policy resolved", false only for the no-current-model edge.

Preserved contracts: user-pin = no automatic fallback; optimistic primary + `model_route_rollback`; commit-gated persistence; setup-failure `message_end(stopReason:"error")`; exhausted runs RESOLVE (only `prompt_error` rejects). The single candidate omits `modelConfig`, so `runAttempt`'s `modelNeedsUpdate` guard skips a redundant `setModel` and keeps applied runtime params.

Rationale and full contract list: `docs/design/2026-06-22-unified-model-routing-entry.md`.

### Behaviour changes (call-outs)

- **Every turn is now a routing turn.** It emits `model_route_start` + `model_route_success` (single candidate → `isFallback:false`, no switch), so commit-gating in `sse-consumer.ts` applies to every turn. Consumers already gate "switched model" UI on `is_fallback || recovered`.
- **Preflight on every turn.** A previously-bare prompt now runs `ensureContextForModelPrompt`: an over-budget turn compacts proactively (compaction on) or fails cleanly in preflight (compaction off) instead of mid-stream.
- **Deploy ordering.** agentbox now emits `model_route_*` on every turn — the paired gateway `sse-consumer.ts` must be a version that commit-gates routed turns.

## Test Plan

- [x] Full suite: `vitest run` — 3663 passed / 2 skipped
- [x] New unit tests: `resolveEffectivePolicy` (real-routing passthrough / single-candidate / user-pin / no-model)
- [x] New integration test: a routing-disabled plain turn still emits `model_route_success(isFallback:false)` with no switch
- [x] Updated `http-server.test.ts` assertions for single-entry behaviour (user-pin + lone-candidate now run through the runner)
- [ ] Live verification of the every-turn preflight path against a real brain (over-budget turn, compaction on/off) — not covered by the mocked suite


docs/design/2026-06-22-unified-model-routing-entry.md
```diff
@@ -0,0 +1,58 @@
...
+- **Synthetic turns** still pass `optimisticPrimaryStream:false` (they persist from collected events and have no live viewer).
+- All prior routing contracts are preserved: optimistic primary + `model_route_rollback`, commit-gated persistence, setup-failure `message_end(stopReason:"error")` synthesis, exhausted runs RESOLVE (only `prompt_error` rejects).
+
+## Consequences
+
+- **Behaviour change — preflight on every turn.** A turn that was previously a bare prompt now runs the runner's preflight (`ensureContextForModelPrompt`). For an over-budget turn this means proactive compaction (compaction on) or a clean preflight failure (compaction off) instead of a mid-stream failure. This is a deliberate, arguably cleaner outcome; it is invisible to the mocked unit suite (no real brain) — verify against a live brain when changing context budgeting.
+- **Deploy ordering.** agentbox now emits `model_route_*` on every turn; the paired gateway `sse-consumer.ts` MUST be a version that commit-gates routed turns. Do not run a new agentbox against a pre-routing-commit-gating gateway.
```
chent1996, line 49: **Deploy-ordering is a real risk worth a release-note callout, not just a doc line.**

agentbox now emits `model_route_*` on EVERY turn. A rolling deploy that lands the new agentbox image before the gateway is upgraded means:

- New agentbox → old gateway `sse-consumer.ts`: the old consumer doesn't recognize `model_route_start` as the signal to defer assistant rows, so it persists them inline. Then `model_route_success` arrives and the new commit-gating logic (if the gateway is partway upgraded) would also try to flush — depending on timing, you get either duplicated assistant rows or one row + a dangling pending op.
- The old `routeEnabled ? runner : brain.prompt` branch is gone, so there's no longer a path for the old gateway to handle "no routing events emitted, persist inline."

CLAUDE.md captures this, but the failure mode is silent (no obvious error — just duplicated/missing chat history rows on reload). Two improvements:

1. **Helm chart / release notes**: hard-link the agentbox image bump to a minimum-gateway-version requirement, the same way mTLS bumps are usually paired. A `helm install` should refuse to proceed if the gateway version is too old.
2. **Runtime probe**: agentbox could read its expected gateway protocol version at boot and refuse to register if the version is below the commit-gating cutoff.

Non-blocking, but the next person rolling out a hot agentbox image at 3am will appreciate the guard rail.

src/agentbox/http-server.ts
```diff
@@ -600,7 +598,10 @@ export function createHttpServer(
       managed._aborted = true;
       console.log(`[agentbox-http] Consumed pre-spawn pending abort for session ${managed.id}`);
     }
-    managed._routeBrainEventsThroughExtra = routeEnabled;
+    // Default to the extra channel; refined to false only for the no-current-model
+    // edge before the run kicks off (see effectivePolicy below). No brain prompt
+    // events fire during model setup, so this default is never observed wrongly.
+    managed._routeBrainEventsThroughExtra = true;
```
chent1996, line 604: **Setting `_routeBrainEventsThroughExtra = true` before model setup is safe today, but it's a fragile invariant.**

The comment says "No brain prompt events fire during model setup, so this default is never observed wrongly." That holds because `brain.setModel(...)` and `brain.registerProvider(...)` don't fire prompt events — and `ensureContextForModelPrompt` happens later inside `runAttempt` (after the flag is refined).

But the invariant lives in a comment, not in the type system. A future change that adds (say) a model-warmup probe call inside the setup block — even something as innocent as a `brain.health()` that internally subscribes to its own emission — would silently route those events through the wrong channel, and the bug would only surface as missing telemetry that nobody notices.

Two cheap defenses:
1. **Set the flag to `false` initially**, then set it to its final value AFTER `effectivePolicy` is computed. With current setup never emitting events, this default change is invisible; if a future setup step does emit, events go through the live `_eventBuffer` (the legacy path) rather than the routing extra channel — degrading to "looks like an unrouted turn" instead of dropped/misrouted events.
2. **Assert no brain events arrived during setup**: a counter on the live subscription, asserted ===0 right before the flag is refined.

Minor / non-blocking — flagging because the comment is the only thing protecting the invariant.

src/agentbox/http-server.ts
```diff
@@ -865,25 +866,35 @@ export function createHttpServer(
...
+      managed.modelRouteState,
+      managed.brain.getModel?.(),
+    );
+    // Only the no-current-model edge falls back to a bare brain.prompt (runner
+    // guard) whose events flow through the live _eventBuffer subscription.
+    managed._routeBrainEventsThroughExtra = effectivePolicy !== undefined;
+    const promptPromise = runPromptWithModelRouting(
+      managed.brain,
+      promptText,
+      effectivePolicy,
```
chent1996, line 885: **Preflight on every turn is the right cleanup, but worth verifying against real workloads before this hits production.**

The PR description and design doc both call out that a previously-bare prompt now runs `ensureContextForModelPrompt` here. Two production-side concerns:

1. **Turns that previously squeaked by** — an over-budget turn that the LLM tolerated (e.g., the provider clipped the front of context silently) will now fail preflight when compaction is off, or trigger compaction when it's on. Both are "more correct" outcomes, but they change observable behavior for sessions that previously appeared healthy. Worth a heads-up to operators monitoring `context_overflow` and `compaction_triggered` counters.

2. **Latency on hot path** — `ensureContextForModelPrompt` is now in the critical path of every turn, including snappy text-only turns. For most brains this is a cheap `estimateMessagesTokens` call, but if any implementation does a tokenizer load or remote call, that latency now lands on every turn rather than only routed ones.

The author flags this as "invisible to the mocked unit suite — verify against a live brain when changing context budgeting." Strongly agree; the verification should happen BEFORE merge if possible. Specifically:
- An over-budget turn with compaction OFF should now fail at preflight (not mid-stream) with a clean error.
- An over-budget turn with compaction ON should now compact proactively.
- A snappy small-context turn should add ≤ few-ms latency.

Non-blocking on the code; blocking on "have we actually run this against a real model yet".

chent1996 (approved):
**Well-reasoned consolidation — approving with three notes.**

I traced the unification end-to-end:

- `resolveEffectivePolicy(configured, state, currentModel)` is the right shape — three branches map cleanly to "real multi-candidate routing / single-candidate from current model / no-model bare-prompt edge", and the synthetic single-candidate intentionally omits `modelConfig` so `runAttempt`'s `modelNeedsUpdate` guard finds the current model already pinned and skips a redundant `setModel` (which would otherwise drop applied `reasoning_effort`).
- Both call sites mirror each other: `_routeBrainEventsThroughExtra = effectivePolicy !== undefined` is the **only** carry-over of the old `routeEnabled` gate, narrowed to "did a policy resolve." Synthetic turns still pass `optimisticPrimaryStream:false` so a live failed primary can't leak into events the synthetic path persists from.
- Tests cover the four `resolveEffectivePolicy` cases (real routing pass-through / single-candidate / user-pin / no-model) and the http integration assertion that a disabled-routing turn now emits `model_route_success(isFallback:false)`. That matches what consumers were already assuming via the "is_fallback || recovered" gate, so SSE/persistence stay consistent.
- The CLAUDE.md change-impact matrix is updated with the new contract and the deploy-ordering caveat. Good — exactly the kind of cross-cutting change that needed the matrix updated.

Three things worth flagging inline; none are blockers.

Body-only note: the design doc's "Partial win, honestly" section is unusually direct about what was NOT removed. That's the right tone — the `_eventBuffer` / `_routeBrainEventsThroughExtra` gate is still there, just narrower. A follow-up that removes the no-current-model edge (e.g., by failing setup before reaching the runner when no model is configured) would let the gate go away entirely.

Note: I could not run the suite locally (no `node_modules` in this review checkout). Reviewed by reading; trusting the author's reported `vitest run` (3628 passed / 2 skipped).

PR #345, opened 2026-06-22
fix(a2a): terminal-state immutability, strict session filter, api-key index prefix

LikiosSedo:
Follow-ups from code review of the A2A gateway (#340), now that it has merged.

## Changes

1. **Terminal-state immutability (correctness race).** `setTaskState` now updates only when the current state is non-terminal (`WHERE state NOT IN (<terminal>)`), so the first terminal write wins. Previously a `stream_error` could overwrite a state that was already `CANCELED`/`COMPLETED` — e.g. a user cancels, the runtime emits `stream_error` while aborting, and the task ends up `FAILED` ("Siclaw stream failed") despite the cancel. This also closes the symmetric races (cancel vs `prompt_done`, and a late `tool_execution_*` resurrecting a settled task to `WORKING`). `stream_error` additionally re-checks terminal state up front, mirroring the existing `prompt_done` guard.

2. **Strict session match in the A2A tracker.** The `chat.event` filter required `envelope.sessionId && envelope.sessionId !== task.sessionId`; the leading truthiness check meant a `sessionId`-less event fell through to *every* task on the agent. Because the A2A path persists `artifact_text`, that cross-session bleed would survive longer than on the chat path. Tightened to a strict `!==` match. (The same loose pattern exists in `chat-gateway.ts` and is intentionally left out of scope here.)

3. **Index prefixes match the access pattern.** Every A2A query is scoped by `(agent_id, api_key_id)`, but the `a2a_tasks` composite indexes led with `agent_id` / `context_id` only — so `loadSessionIdForContext` (run on every `message:send` with a `contextId`) and `listTaskRecords` scanned more than necessary. The two composites now lead with `(agent_id, api_key_id, …)`. Index names are unchanged.

## Validation

- `npx tsc --noEmit`
- `npm test` — 3690 passed, 2 skipped
- Added a deterministic terminal-immutability regression test (gated `chat.abort` so a `stream_error` commits `FAILED` first, then a cancel must not overwrite it); verified it fails without the guard and passes with it.


jacoblee-io:
## Review

Good follow-up set — the terminal-immutability guard and the strict session filter are both correct and well-tested. One issue with the index change (#3) that I think blocks the stated goal, plus a couple of minor notes.

### 🔴 #3 — the index redefinition is a no-op on every existing deployment

`ensureIndex` only checks for the index **by name** (`migrate-compat.ts:62-81`): if the name exists it returns early, regardless of the column list. `idx_a2a_tasks_agent` and `idx_a2a_tasks_context` were already created by #340 (now merged), so on any DB that has run migration since then — all current MySQL deployments and every persisted local `.siclaw/data/portal.db` — the new `(agent_id, api_key_id, …)` prefixes will **never be applied**. The optimization only lands on freshly-created databases. No correctness impact (queries still run on the old indexes), but the change silently doesn't do what the PR says it does where it matters.

There's a precedent for exactly this case a few lines down in the same file (the skills block): `dropIndexIfExists` the old index, then create under a **new name**:

```ts
await dropIndexIfExists(db, "skills", "uq_skills_org_name");
await ensureIndex(db, "skills", "idx_skills_org_name", "org_id, name");
```

I'd mirror that — drop `idx_a2a_tasks_agent` / `idx_a2a_tasks_context` and recreate them under new names (e.g. `_agent_key` / `_context_key`). Renaming keeps it one-time and idempotent; a `dropIndexIfExists` + same-name `ensureIndex` would instead drop+recreate on every boot. Worth noting no schema test pins index *columns* (only names), which is why this slipped through `npm test`.

### ✅ #1 — terminal-state immutability: correct

`WHERE … AND state NOT IN (<terminal>)` on both branches is the right shape, and putting it on the non-terminal branch too is what closes the "late `tool_execution_*` resurrects a settled task to WORKING" race. The inlined `TERMINAL_STATES_SQL` is safe (constant enum values, no user input) and portable across both drivers. The gated-`chat.abort` regression test is deterministic and actually exercises the race.

Minor: the `stream_error` recheck is described as "belt-and-suspenders," but it does more than that — `setTaskArtifact` runs *before* `setTaskState` and is **not** terminal-guarded, so without the early return a late `stream_error` would still overwrite the artifact text of an already-terminal task even though the state wouldn't flip. So the recheck is load-bearing for artifact integrity, not purely redundant.

### ✅ #2 — strict session match: verified safe

Confirmed every `chat.event` publisher sets a concrete `sessionId` (`server.ts:256/262/270/275/292/296`, `internal-api.ts:574`), so dropping the leading truthiness check won't strand any legitimate runtime event. Good tightening, and good call to scope the symmetric `chat-gateway.ts` case out.

Overall: #1 and #2 are ready; #3 needs the drop-and-rename (or it won't take effect on existing DBs).

jacoblee-io (approved):
Follow-up `fee7f6c` resolves both points from my earlier review — verified:

**#3 (blocker) — fixed.** The drop-and-rename (`idx_a2a_tasks_agent` / `_context` → `_agent_key` / `_context_key`) mirrors the skills `uq → idx` precedent and is one-time idempotent: existing post-#340 DBs drop the old names and recreate with the `(agent_id, api_key_id, …)` prefixes; fresh DBs no-op the drop; re-runs no-op both. Schema test updated and no stale references to the old names remain.

**#1 follow-up — `setTaskArtifact` now terminal-guarded.** Confirmed no happy-path regression: both `prompt_done` and `stream_error` write the artifact while the row is still `WORKING` (before the terminal transition), and both early-return on an already-terminal task, so the guard is a pure backstop. Comment corrected.

**#1 / #2** were already correct. LGTM.

PR #346, opened 2026-06-22
fix(agentbox): reject invalid native media payloads

LikiosSedo:
## Summary
AgentBox now validates native image/PDF prompt media instead of silently dropping malformed or oversized blocks. Valid images up to the media item limit are forwarded to the brain, while invalid media receives a clear 400 response.

## Problem
The prompt media sanitizer skipped unsupported, invalid, or oversized image blocks and continued with the remaining text. That made media-only or media-dependent prompts look like plain text prompts when the image was not forwarded.

## Solution
- Align AgentBox media/body limits with the four-attachment prompt contract.
- Replace silent media skipping with explicit validation errors.
- Return clear request errors for malformed JSON, body timeouts, and oversized bodies.
- Cover large valid images and invalid media for prompt/steer requests.

## Compatibility Notes
- The media cap is intentional: image items now use the same 8 MiB base64 ceiling as PDF items, and the JSON body limit allows up to four media items plus request overhead. This is an AgentBox ingress ceiling; provider-specific media limits may still be lower and will surface as provider errors instead of being hidden by a silent drop.
- Invalid media is now a hard request error for `/api/prompt` and `/api/sessions/:id/steer`. Non-frontend clients that previously sent unsupported or malformed media alongside text should now fix the payload instead of relying on lenient drop-and-continue behavior.

## Test Plan
- [x] `npm test -- --run src/agentbox/http-server.test.ts`
- [x] `npm run build`

### DIFF
--- a/src/agentbox/http-server.test.ts
+++ b/src/agentbox/http-server.test.ts
@@ -325,20 +325,45 @@ describe("http-server — prompt + session lifecycle", () => {
     );
   });
 
-  it("POST /api/prompt drops malformed images and rejects when nothing usable remains", async () => {
+  it("POST /api/prompt forwards large valid images to brain.prompt", async () => {
+    const session = await sm.getOrCreate("img-large");
+    const data = "A".repeat(3 * 1024 * 1024);
     const r = await getJson(port, "/api/prompt", "POST", {
-      images: [{ mimeType: "image/gif", data: "aW1n" }], // unsupported mime → dropped
+      sessionId: "img-large",
+      modelProvider: "openai",
+      modelId: "gpt-4",
+      modelConfig: modelConfigWithInput(["text", "image"]),
+      images: [{ mimeType: "image/jpeg", data }],
     });
-    expect(r.status).toBe(400);
-    expect(r.data.error).toMatch(/Missing.*text/);
+    expect(r.status).toBe(200);
+    expect(session.brain.prompt).toHaveBeenCalledWith(
+      "Please analyze the attached image.",
+      { images: [{ mimeType: "image/jpeg", data }] },
+    );
   });
 
-  it("POST /api/prompt drops images whose data is not valid base64", async () => {
+  it("POST /api/prompt rejects malformed images instead of dropping them", async () => {
     const r = await getJson(port, "/api/prompt", "POST", {
-      images: [{ mimeType: "image/png", data: "not base64!!!" }], // bad base64 → dropped
+      images: [{ mimeType: "image/gif", data: "aW1n" }],
     });
     expect(r.status).toBe(400);
-    expect(r.data.error).toMatch(/Missing.*text/);
+    expect(r.data.error).toMatch(/images\[0\]\.mimeType/);
+  });
+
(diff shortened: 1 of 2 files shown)


jacoblee-io:
## Review

Solid, well-scoped change. Replacing silent media-dropping with explicit `400`s is implemented cleanly, both ingress routes are covered, and the error-status plumbing is a genuine improvement. A few things worth confirming before merge, mostly around the bundled limit increase.

### What's good
- **Clean behavior fix.** Silent drop → field-precise `400` (`images[0].mimeType …`, `images[0].data … base64`) is exactly what the problem statement asks for.
- **Both callers updated.** `validatePromptMedia` replaces both `sanitizePrompt*` calls (`/api/prompt` and steer) — and those were the only two callers, so nothing is left on the old silent path.
- **Status-code plumbing is a real improvement.** `HttpRequestError` + the dispatcher branch maps malformed JSON → `400`, timeout → `408`, oversized body → `413` instead of the old blanket `500`. This now benefits *every* route that calls `parseJsonBody`.
- **Backward-compat risk is low for the official frontend.** `portal-web/.../InputArea.tsx` gates uploads to the same supported set (`image/png|jpeg|webp` + `application/pdf`) before sending, so the hard-reject won't fire on normal Portal flows. The gateway's `parseBody` has no default `maxBytes`, so the agentbox raise isn't silently capped upstream.
- `image/jpg → image/jpeg` normalization is a nice touch, and the new tests cover large-valid-image forward, the three invalid-media reject cases, and a steer case asserting `brain.steer` is *not* called.

### Worth confirming

**1. [Medium] The limit increase is bundled and under-advertised.** Beyond the validation refactor, this PR also raises:
- per-image base64 limit **2 MB → 8 MB** (4×)
- `MAX_BODY_SIZE` **10 MB → ~32.5 MiB** (3.25×)

`parseJsonBody` buffers the whole body as a JS string, and in **local mode all users share one process**, so peak memory per concurrent request roughly triples. Separately, an 8 MB base64 image (~6 MB decoded) can exceed downstream provider per-image limits (e.g. Anthropic's ~5 MB), turning a clean client-side `400` into a more confusing downstream API error. Please confirm the 8 MB image ceiling is intentional and within the limits of the providers we forward to — and ideally call it out in the description.

**2. [Low] Behavior is now a hard contract for non-frontend clients.** Any API consumer/script that previously relied on lenient silent-drop (e.g. a stray `gif`, or a slightly-oversized image alongside valid text) now gets a `400` and *nothing* processes, where before the text still went through. Low risk given frontend gating, but it's a public-surface change to `/api/prompt` and steer — worth a line in the PR notes.

**3. [Nit] Error wording.** `Body exceeds 32.5 MiB byte limit` reads redundantly ("MiB byte limit"); consider `Body exceeds 32.5 MiB limit`. `formatBytes` is shared between the per-item "base64 character limit" and the body "byte limit" messages, so the unit phrasing is slightly inconsistent — fine, but noting it.

None of these are blockers. If the 8 MB / 32.5 MiB ceilings are deliberate, I'd just want them acknowledged in the description; the rest is polish.

jacoblee-io (approved):

PR #347, opened 2026-06-22
Bring Feishu personal bot channel support to main

LikiosSedo:
## Summary
- Bring the PR #339 Feishu/Lark channel work from `feat/feishu-visual-image-pass-through` to `main`.
- Includes Feishu personal bot session support and OAuth guidance for personal binding.
- Preserves the channel-test work so Sicore personal authorization can be consumed by Siclaw Runtime on mainline builds.

## Context
PR #339 was merged into `channel-test`, but `main` does not contain the personal Feishu bot runtime commits (`c3b8a07`, `4033691`). As a result, mainline runtime builds still route personal bot messages through the old chat binding path and can silently ignore authorized personal messages.

## Validation
- Verified `origin/main` does not contain `c3b8a07` or `4033691`.
- Verified `origin/feat/feishu-visual-image-pass-through` contains the PR #339 commits.


mcp/create-chart/src/sicore-export.ts
```diff
@@ -0,0 +1,130 @@
...
+    executablePath,
+    headless: true,
+    env: {
+      ...process.env,
+      HOME: homeDir,
+      XDG_CONFIG_HOME: configDir,
+      XDG_CACHE_HOME: cacheDir,
+    },
+    args: [
+      "--no-sandbox",
```
chent1996, line 107: **Chromium `--no-sandbox` disables the process-level sandbox.** Combined with `--disable-setuid-sandbox` below it, this is the standard workaround for running Chromium inside a Docker container without `CAP_SYS_ADMIN` or a custom seccomp profile — but it means a Chromium 0-day exploit gets process-level access inside AgentBox.

The defense-in-depth that makes this acceptable today:
- AgentBox runs as the `sandbox` OS user (per security.md ADR-010), not root
- The AgentBox container drops all capabilities and uses `RuntimeDefault` seccomp
- AgentBox has read-only rootfs in K8s mode

So worst case is "Chromium exploit → sandbox-user privileges inside one AgentBox pod" — still contained, but the **content driving Chromium is now LLM-generated markdown** (chart JSON, mermaid source, visual-card JSON). Until now the AgentBox sandbox protected against tool-output going wrong; it now also has to protect against LLM-content driving a browser engine.

Recommend documenting this trust-boundary expansion in `docs/design/security.md` and ensuring the AgentBox Pod Security Standard stays `restricted` (no privilege escalation, no host networking). Non-blocking, but the security doc should reflect that the Chromium attack surface is now part of the AgentBox threat model.

mcp/create-chart/src/sicore-export.ts
```diff
@@ -0,0 +1,130 @@
...
+  image: Buffer;
+}
+
+interface SicoreExportOptions {
+  baseUrl?: string;
+  theme?: "light" | "dark";
+  timeoutMs?: number;
+}
+
+const DEFAULT_EXPORT_URL = "http://sicore-web:3000/siclaw-visual-export";
```
chent1996, line 17: `SICORE_VISUAL_EXPORT_URL` defaults to a hardcoded internal hostname (`http://sicore-web:3000/...`). Three implications worth being explicit about:

1. **Hard runtime dependency**: when Sicore Web isn't deployed (OSS-only Siclaw installs, dev environments, the K8s `siclaw` Helm chart without Sicore), every `render_chart` / `render_mermaid` / `render_visual_card` call throws. The MCP doesn't fall back to inline data URLs or a local render path. The agent will then either retry (cost) or return the source-only block (channel users see source instead of image).
2. **No TLS / no auth**: plain HTTP, no shared secret. Same posture as the OCR backend from PR #288 — fine for cluster-internal services but worth being aware of.
3. **Default hostname assumes the Sicore release name is exactly `sicore-web`**: same shape as the original `siclaw-runtime` bug fixed in PR #273. If anyone deploys Sicore under a different fullname, they'll have to override the env.

Suggest: (a) log + degrade gracefully when the export call fails (preserve source block, skip image), and (b) document the env override in `docs/design/` next to the Helm values. Non-blocking.

src/gateway/channels/background-delivery.ts
```diff
@@ -0,0 +1,68 @@
...
+export type BackgroundChannelDelivery = (message: ChannelDeliveryPayload) => Promise<boolean>;
+
+interface DeliveryEntry {
+  deliver: BackgroundChannelDelivery;
+  expiresAt: number;
+  timer: NodeJS.Timeout;
+}
+
+const DEFAULT_TTL_MS = 30 * 60 * 1000;
+const deliveries = new Map<string, DeliveryEntry>();
```
chent1996, line 29: Same single-replica constraint as `a2a-gateway.ts` from PR #340 — module-global `Map<sessionId, DeliveryEntry>` on the gateway process. With multiple Portal replicas, an agent's `channel_update` could end up on a replica that doesn't hold this session's delivery handle, and the milestone update silently disappears.

The `replicas: 1` rationale is already documented in `helm/siclaw/values.yaml` for the Portal+Runtime path, but this file doesn't reference it — recommend a one-line top-of-file note like:

```ts
// Process-local map: this requires the Portal+Runtime pair to run single-replica
// (see helm/siclaw/values.yaml `portal.replicas`). Horizontal scale would need a
// shared pub/sub bus, same constraint as src/portal/a2a-gateway.ts.
```

Also: the 30-minute TTL is fixed and **does not refresh on use**. A channel session that runs longer than 30 minutes (long investigations are exactly the use case `channel_update` targets) loses its delivery handle silently. Either bump the default, or refresh `expiresAt` on each successful `deliver()` call. Non-blocking but worth tightening before this hits a real long-running incident.

src/gateway/internal-api.ts
```diff
@@ -556,7 +566,18 @@ export async function handleDelegationEvents(
...
       case "delegation.append_message": {
-        response = { ok: true, id: await appendDelegationMessage(frontendClient, event.message) };
+        const deliveredToChannel = await deliverBackgroundChannelMessage(event.message);
+        const channelRegistered = hasBackgroundChannelDelivery(event.message.sessionId);
+        try {
+          response = { ok: true, id: await appendDelegationMessage(frontendClient, event.message) };
+        } catch (err) {
+          if (!deliveredToChannel && !channelRegistered) throw err;
+          console.warn(
+            `[internal-api] Portal append failed for channel background session=${event.message.sessionId} delivered=${deliveredToChannel}:`,
```
chent1996, line 576: **Swallowing Portal append errors for channel-delivered messages creates a chat-history divergence.** If `appendDelegationMessage` throws but `deliverBackgroundChannelMessage` succeeded, we return `{ ok: true }` with only a `console.warn` — the message lands in Feishu but never enters Portal's chat_messages table.

This is **deliberate** (the comment chain explains: channel users care about the reply, not the historical record), and the failure mode is preferable to "user sees the reply but later it disappears from the Portal UI re-render." But the channel-delivered + Portal-failed case is an audit gap. Two improvements:

1. **Metrics hook**: emit a counter for `portal_append_failed_but_channel_delivered` so operators can see when this divergence is happening (it shouldn't be common).
2. **Compensating write attempt**: if the failure is transient (DB hiccup), a single retry-with-jitter in the catch block would close most of the gap.

Pre-existing fail-soft pattern in the codebase — non-blocking, but if a `chat_messages` audit later contradicts what a Feishu group remembers, this is the first place to look.

chent1996 (approved):
**Substantial back-merge of the Feishu/Lark feature branch — approving with four notes.**

This is a 19-commit back-merge that lands a coherent set of channel features on `main`: personal bot p2p sessions, the `channel_update` agent tool, a Sicore-rendered visual image pipeline, CardKit 2.0 sanitizer relaxation, and a new `mcp/create-chart` MCP. I traced the seams that matter:

- **Session isolation by `(binding_id, session_key)`** — group messages use `chat:<chat_id>`, p2p uses `open_id:<sender>`. The new `channel_binding_sessions` table + `channel_bindings.session_id` column are migrated via `safeAlterTable`, the migration DDL respects invariants §5 (no `TIMESTAMP(3)` / `ON UPDATE` / `JSON`; new index has its own name `idx_channel_binding_sessions_session`), and the per-binding queue key `${bindingId}:${sessionKey}` keeps p2p users serialized independently. Sound.
- **`channel_update` tool is gated correctly** — `available: refs => Boolean(refs.channelMessageExecutor && refs.sessionIdRef)` and `modes: ["channel"]` means CLI/web/task agents never see it. The tool only expresses intent; gateway owns coalesce/cap/suppress policy.
- **Feishu in-flight ACK race** is handled with the `setImmediate(...)` detach trick (lark.ts:135–142) — Feishu re-delivers after ~10s if the event handler doesn't return, so the handler resolves in <1ms and kicks the work onto a detached task. Good comment explaining why.
- **CardKit 2.0 sanitizer relaxation** is the right fix for the literal `**` + truncated-table rendering bugs. The doc comment at `lark-card.ts:71–82` explicitly warns the next maintainer not to reintroduce the down-conversion.
- **AgentBox Dockerfile** now pulls in `chromium` + `fonts-noto-cjk` — adds ~150 MiB but unavoidable for the visual export path. Runtime gets `fonts-noto-cjk` only (PNG composition happens in gateway).
- **`src/core/prompt.ts` change is additive** — `CHANNEL_SECTION` only appended when `mode === "channel"`, so CLI/web/task prompts are untouched. (CLAUDE.md flags this file as HUMAN APPROVAL — author presumably has sign-off; I'm flagging the rule for posterity.)

Four inline notes below; none are blockers. Top concern is #1 (Chromium sandbox) — flagging because the trust boundary changed even though the deployment posture still mitigates it.

Body-only notes (no anchor):
- `lark.ts:138` swallows handler errors with `console.error` only — a partial handler failure in production won't surface via Sentry-style alerting unless something else hooks into stderr. Pre-existing pattern; not worse than before.
- `channel_update` tool description tells the agent "Gateway may coalesce, update the existing card, suppress, or cap messages" — good guidance, but make sure operators can SEE when the gateway suppresses (a `console.log` at the suppression point would help diagnose "agent says it sent an update but user never saw it").
- The bundle of 19 commits would benefit from a squash-or-explicit-merge commit message that lists the user-visible behaviour changes, since reviewers landing on `main` later will only see the merge commit.

Note: I could not run the suite locally (no `node_modules` in this review checkout). Reviewed by reading; trusting the author's reported `npm test` runs.

PR #350, opened 2026-06-23
feat(agentbox): configurable per-agent idle self-destruct window

jacoblee-io:
## Summary

Make an AgentBox pod's idle self-destruct window **per-agent configurable** from the Portal (agent → Basic settings), defaulting to 300s, with `0` meaning **resident** (never auto-destroys). Also routes idle teardown through the same graceful shutdown as SIGTERM instead of a raw `process.exit(0)`.

### Problem

The idle window was hard-coded to 5 minutes in `http-server.ts` — not tunable per agent, and no way to keep an agent resident. Separately, the idle path called `process.exit(0)` directly, **skipping graceful shutdown**: debug pods were never evicted (orphaned in the cluster) and the final metrics flush to the Gateway was lost.

### Solution

- **DB**: new `agents.idle_timeout_sec INT NOT NULL DEFAULT 300` (CREATE + idempotent `safeAlterTable`, MySQL + SQLite compatible).
- **API/UI**: `idle_timeout_sec` threaded through agent CRUD; new "Idle Timeout (seconds)" field in the Basic tab (`0` = resident).
- **Spawn transport**: gateway `resolveAgentSpawnEnv` reads the value (best-effort) and injects `SICLAW_AGENTBOX_IDLE_TIMEOUT` at spawn; the AgentBox reads it into `config.server.idleTimeoutSec`. Applies on the agent's next pod spawn (K8sSpawner ignores env for an already-running pod).
- **Graceful idle teardown**: idle timer now calls `onIdleShutdown` → the same flush + debug-pod eviction + session close path as SIGTERM (idempotent). Idle timer is `unref`'d.

Chose env injection (vs the runtime config snapshot) because the window is needed at AgentBox **startup**, before any snapshot fetch.

## Test Plan

- [x] `npx tsc --noEmit` (backend + portal-web) clean
- [x] New `http-server.test.ts` idle cases: fires at window, resident at `0`/negative, `disableIdleShutdown`, re-check guard, default fallback
- [x] `npm test` for migrate (sqlite + schema-invariants), agent-api, gateway server, manager/spawner — all green
- [x] Deployed to `siclaw-inner` (`v0.1.9-idle1`): pods Running, `[portal-migrate] added agents.idle_timeout_sec` confirmed, Basic-tab field renders


src/gateway/server.ts
```diff
@@ -140,6 +140,36 @@ export async function startRuntime(opts: StartRuntimeOptions): Promise<RuntimeSe
...
+        return { SICLAW_AGENTBOX_IDLE_TIMEOUT: String(sec) };
+      }
+    } catch (err) {
+      console.warn(`[gateway] Failed to resolve idle timeout for agent ${agentId}:`, err);
+    }
+    return undefined;
+  };
+
+  // Spawn (or reuse) the AgentBox for an agent, threading its per-agent spawn env.
+  const getOrCreateAgentBox = async (agentId: string) => {
```
LikiosSedo: **[Perf — chat hot path]** `getOrCreateAgentBox` resolves the spawn env via a `config.getAgent` RPC on **every** `chat.send` / `chat.abort` / `chat.steer` / `chat.clearQueue`. But `getOrCreateK8s` returns the running pod without ever consuming `env` when a pod already exists — and abort/steer/clearQueue operate on a live session, so the pod is guaranteed to be running there. That makes the RPC a wasted round-trip to Portal on the common path (warm pod), and unconditionally wasted on three of the four call sites.

Suggest resolving env only when a cold spawn is actually about to happen — e.g. pass a lazy thunk that `getOrCreateK8s` invokes only on its spawn branch, or skip env resolution entirely for abort/steer/clearQueue (a default-window respawn is fine if the pod somehow vanished).

src/portal/agent-api.ts
```diff
@@ -153,7 +154,7 @@ export function registerAgentRoutes(
     // Build dynamic SET clause
     const fields = [
       "name", "description", "status", "model_provider",
-      "model_id", "system_prompt", "is_production", "icon", "color",
+      "model_id", "system_prompt", "is_production", "idle_timeout_sec", "icon", "color",
```
LikiosSedo, line 167: **[Robustness — low]** `idle_timeout_sec` is threaded through the dynamic SET, but the handler pushes `body[field]` verbatim with no server-side coercion/clamp. Admin-only endpoint so risk is low, but a float / string / oversized value lands directly in the `INT` column (SQLite stores it verbatim; downstream `parseInt` masks it, but DB cleanliness/portability suffers). Consider clamping server-side — `Math.max(0, Math.floor(Number(v)))` — mirroring the UI input handler.

src/agentbox-main.ts
```diff
@@ -78,8 +78,40 @@ async function main() {
   if (process.env.USER_ID) sessionManager.userId = process.env.USER_ID;
   if (process.env.SICLAW_AGENT_ID) sessionManager.agentId = process.env.SICLAW_AGENT_ID;
 
-  // Start HTTP server
-  const server = createHttpServer(sessionManager);
+  // Graceful shutdown — defined before the server so the idle self-destruct
+  // timer can route through it (see onIdleShutdown below). Idempotent: the idle
+  // timer firing and a SIGTERM racing must not run the teardown twice.
+  let shuttingDown = false;
+  const shutdown = async () => {
```
LikiosSedo, line 92: **[Nit — latent footgun]** This `shutdown` closure references `server` (~L112) and `federationFlushEnabled` (~L140), both declared after it. Safe today because `onIdleShutdown` only ever fires via the `setTimeout` in `checkIdle()`, never synchronously inside `createHttpServer` — but that's an implicit contract. A one-line comment noting `onIdleShutdown` must never be invoked synchronously would prevent a future `createHttpServer` refactor from turning this into a TDZ `ReferenceError`.

LikiosSedo (commented):
Reviewed the full diff plus surrounding context. Solid PR — and the graceful idle-teardown change is a genuine bug fix on its own, not just a config knob: routing the idle path through `shutdown()` instead of `process.exit(0)` stops orphaned debug pods and recovers the final metrics flush. Migration (CREATE + idempotent `safeAlterTable`, MySQL/SQLite), the `0 = resident` semantics (consistent across UI / config / http-server), and the test coverage all look good.

A few findings inline. One is worth addressing before merge (the per-request `config.getAgent` RPC on the chat hot path); the rest are low / nit.

**Merge coordination:** this PR and #352 both rewrite the same lines of `docs/design/invariants.md` §1.3 (the self-destruct paragraph + the **Source** line) and conflict with each other — verified via `git merge-tree` (CONFLICT content). Each merges cleanly into `main` on its own, but the second to land will hit a conflict there. Resolution: keep this PR's expanded idle-window paragraph + #352's CA-fingerprint bullet + a combined **Source** line.

jacoblee-io:
Thanks for the review @LikiosSedo — all three addressed in `cd4ee92`:

- **Perf (chat hot path)** — the `config.getAgent` RPC is now resolved via a **lazy thunk** that `AgentBoxManager.getOrCreate` invokes **only on the cold-spawn branch**. Warm-pod reuse (the common `chat.send` case, and always abort/steer/clearQueue) calls the resolver zero times — no RPC. All four call sites keep using `getOrCreateAgentBox`, so a genuinely-vanished pod still respawns with the correct window (one RPC, only then). Added manager tests: resolver not called on warm reuse, called once on cold spawn, no-env when it yields undefined.
- **Robustness** — `idle_timeout_sec` is now clamped server-side in both POST and PUT via `clampIdleTimeoutSec` (`Math.max(0, Math.floor(Number(v)))`, non-numeric/missing → 300), mirroring the UI handler.
- **Nit (TDZ)** — added a comment at the `shutdown` closure spelling out that `onIdleShutdown` must only ever fire asynchronously (the forward-ref to `server`/`federationFlushEnabled` would TDZ otherwise).

**Merge coordination** — confirmed: this PR and #352 both touch `docs/design/invariants.md` §1.3. They merge cleanly into `main` independently; the second to land resolves by keeping this PR's idle-window paragraph + #352's CA-fingerprint bullet + a combined **Source** line (already validated on the shared inner test branch).

Also note: full-link E2E on siclaw-inner caught that the runtime is DB-free (the original `getDb()` in `resolveAgentSpawnEnv` threw), fixed in `5a216a2` to use the `config.getAgent` RPC — re-verified live (UI 150 → pod env `SICLAW_AGENTBOX_IDLE_TIMEOUT=150`).

LikiosSedo:
### Follow-up: the per-agent idle window only takes effect when the Web chat path is the cold-spawner

The lazy-resolver fix in `cd4ee92` is correct for the four `chat.*` RPC handlers. But `getOrCreateAgentBox` (the wrapper that threads the resolver) is wired **only** into `src/gateway/server.ts`. Every other spawn entry point calls `agentBoxManager.getOrCreate(agentId)` with no resolver, so it spawns the pod **without** `SICLAW_AGENTBOX_IDLE_TIMEOUT`:

- `src/gateway/channels/dingtalk.ts:294, 363`
- `src/gateway/channels/lark.ts:487, 570`
- `src/gateway/task-coordinator.ts:274`

There's no cluster-wide fallback either: the spawner's forwarded-env allowlist is only `["SICLAW_SUBAGENT_CONCURRENCY"]`, so such a pod falls back to `config.server.idleTimeoutSec` → default **300s**.

Because it's one pod per agent, **whichever entry point cold-spawns first wins the window** for the life of that pod. So an agent configured with `idle_timeout_sec = 0` (resident) — or any non-default value — still gets a 300s pod if a Lark/DingTalk message or a scheduled task happens to spawn it first; the configured value is silently ignored on that path. This bites hardest for exactly the "resident" use case (always-on channel bots), which is one of the PR's stated goals ("no way to keep an agent resident").

This is **pre-existing** in the PR, not introduced by `cd4ee92` — flagging it as a coverage gap in the feature, not a regression. Whether channel/cron spawns should honor the per-agent window is a scope call (v1 may be intentionally Web-only). If in scope, the clean fix is to share the resolver across all cold-spawn entry points — e.g. have `AgentBoxManager` own the per-agent env resolution from an injected `getAgent` lookup — rather than wiring it only at `server.ts`.

---

### Merge coordination update (re #352)

`cd4ee92` made this PR also touch `src/gateway/agentbox/manager.ts`, so #350 and #352 now overlap in **3 files** instead of 1. Test-merging the two current heads:

- `src/gateway/agentbox/manager.ts` — auto-merges cleanly **and correctly**: the combined `getOrCreateK8s` keeps both the `spawnEnvResolver` (this PR) and the `isCertFresh` reuse gate (#352). I extracted the merged blob and checked it line-by-line — no action needed.
- `src/gateway/agentbox/manager.test.ts` — trivial conflict: both PRs append a `describe(...)` block at EOF; keep both.
- `docs/design/invariants.md` — §1.3 content conflict (as before): merge into one paragraph keeping this PR's idle-window text + #352's CA-fingerprint bullet + a combined **Source** line.

Whichever lands second just needs those two manual resolutions.

jacoblee-io:
Good catch on the coverage gap @LikiosSedo — fixed in `d0691d0`, taking your recommended approach.

Moved per-agent env resolution into `AgentBoxManager` itself: `setSpawnEnvResolver(fn)` is registered once in `startRuntime`, and `getOrCreate` consults it on **every** cold spawn. Since bootstrap-runtime hands the *same* manager instance to the chat RPCs, the Lark/DingTalk channels, and the task-coordinator, all of them now honour the per-agent window — no more "first spawner wins the default". Dropped the per-call thunk + `getOrCreateAgentBox` wrapper; the four chat sites call plain `getOrCreate(agentId)` again, and channels/cron need zero changes. Still lazy: the resolver is only invoked on an actual spawn, never on warm-pod reuse.

This matters most for exactly the resident (`idle_timeout_sec = 0`) always-on-bot case you flagged. Added a manager test asserting a channel/cron-style `getOrCreate(agentId)` (no extra args) still gets the env.

Re **merge coordination** — moot now: #352 landed, and I've rebased this PR onto `main`. Resolved the two overlaps exactly as you predicted — `manager.ts` composes `spawnEnvResolver` + `isCertFresh` cleanly, `manager.test.ts` keeps both `describe` blocks, `invariants.md` §1.3 keeps the idle paragraph + CA-fingerprint bullet + combined **Source** line. `tsc` clean, 415 tests green.

LikiosSedo (approved):
Approving — all review findings across both rounds are resolved, and I re-verified the latest follow-up (`d0691d0`).

**Round 1 + 2 findings** (`cd4ee92`):
- Perf hot path → lazy resolver, no `config.getAgent` RPC on warm-pod reuse ✓
- Server-side clamp on `idle_timeout_sec` (POST + PUT) ✓
- TDZ comment on the `shutdown` closure ✓

**Coverage gap** (`d0691d0`) — the important one: per-agent env resolution is now owned by `AgentBoxManager` via a single injected `setSpawnEnvResolver`, registered once in `startRuntime`. Verified:
- One manager instance from `bootstrap-runtime.ts:78` is shared by the chat RPCs, the Lark/DingTalk `ChannelManager`, and the `TaskCoordinator` — so **every** cold-spawn entry point now honours the per-agent window, not just the Web path. The resolver is registered inside the awaited `startRuntime`, before channels/cron can fire.
- Still lazy: consulted only on an actual spawn, never on warm-pod reuse.
- Composes correctly with #352's `isCertFresh` (now on `main`): a CA-stale respawn also gets the per-agent window.
- All `startRuntime` test fakes updated (no missed mocks); added tests for the channel/cron-style `getOrCreate(agentId)` path, no-resolver, and undefined-resolver.

Rebased onto `main` (incl. #352), no conflicts; CI green (Type Check / Test / Portal Web Test). LGTM.

PR #351, opened 2026-06-23
feat(tools): per-agent tool capabilities via capability groups

chent1996:
## Summary

Per-agent tool capabilities: restrict each agent to a multi-select of **capability groups** that resolve to an `allowedTools` whitelist. Empty/unset selection = unrestricted (all tools), so the feature is **strictly opt-in** and every existing agent is unaffected.

### Problem

Tool availability had two axes: a hard-coded `ToolEntry.platform` exemption (some tools could never be disabled) and `allowedTools`. There was no way to give one agent a reduced tool set (e.g. a read-only Q&A agent that cannot run commands), and the file I/O tools bypassed `allowedTools` entirely.

### Solution

- **Unify availability under a single `allowedTools` axis** — drop `platform`; file tools + MCP flow through `appendAllowedTools` (no exemptions; `null` = all tools).
- **Capability-group model** stored in `agents.tool_capabilities` (TEXT; null/empty = unrestricted). Group keys → tool names are resolved at the **gateway boundary**; delivered over the gateway-sync framework as a new `tools` sync type (per-box handler, `invalidate`-on-reload — the same session-rebuild contract as MCP). K8s does a per-pod fetch; LocalSpawner injects the whitelist at spawn time.
- **MCP exempt** (governed by `agent_mcp_servers`); spawned sub-agents **inherit the parent whitelist** (privilege-escalation guard).
- Portal REST **decodes JSON-in-TEXT** agent columns at the boundary; frontend adds a capability-group multi-select (create dialog + a **Tools** tab) with a backend/frontend **drift-guard test**.
- Incidental fixes folded in: LocalSpawner `http://`→`https://` (live reload was broken), agentbox image stops compiling gateway code (compile fix), and a diagnostic log of the final restricted tool set.

## Test Plan

- [ ] `npm test` — **not run in full this session**; frontend capability tests (23) + `npm run build` (tsc && vite) pass
- [x] Manual (`siclaw local`): restricting an agent to `read_files` resolves to exactly `read, grep, find, ls`; live reload re-resolves on save; unrestricted/existing agents unaffected

## Architecture Checklist

- [x] **Deployment mode**: verified in local (LocalSpawner spawn-time inject) and designed for K8s (per-pod fetch in `agentbox-main`); the tools handler is per-box and writes only in-memory `allowedToolsState` (no shared-FS write — safe under LocalSpawner multi-spawn)
- [x] **Security model**: no new shell paths; restriction is a name whitelist; sub-agents inherit the parent whitelist
- [x] **DB parity**: `tool_capabilities TEXT` added via idempotent `safeAlterTable`, MySQL + SQLite compatible; null/empty = unrestricted (backward compatible)
- [x] **Tool protocol**: tools still register through `tool-registry.ts`; the sole availability axis is now the existing `allowedTools` whitelist (no exemptions)

## General Checklist

- [x] Single logical change (per-agent tool capabilities + the unification it requires)
- [x] Conventional Commits
- [x] No unrelated changes (the agentbox compile fix is required to build the images)

> Note: `tool-refract` is currently ~4 commits behind `main`; a rebase touches ~10 shared core files and needs manual conflict resolution (tracked separately).

### DIFF
--- a/src/agentbox-main.ts
+++ b/src/agentbox-main.ts
@@ -12,7 +12,8 @@ import { createHttpServer } from "./agentbox/http-server.js";
 import { AgentBoxSessionManager } from "./agentbox/session.js";
 import { loadConfig, reloadConfig, getConfigPath } from "./core/config.js";
 import { GatewayClient } from "./agentbox/gateway-client.js";
-import { syncAllResources } from "./agentbox/resource-sync.js";
+import { syncAllResources, syncResource } from "./agentbox/resource-sync.js";
+import { createToolsHandler } from "./agentbox/sync-handlers.js";
 import { debugPodCache } from "./tools/infra/debug-pod.js";
 
 // Side-effect: register metrics subscriber. Also imported in http-server.ts,
@@ -119,6 +120,37 @@ async function main() {
   const server = createHttpServer(sessionManager, {
     onIdleShutdown: () => { void shutdown(); },
   });
+
+  // Per-pod initial tool-capabilities fetch. The tools sync type is
+  // initialSync:false (its handler is per-box, not in the module-level registry
+  // that syncAllResources walks, and needs the sessionManager that doesn't exist
+  // at the syncAllResources call site above). This is the K8s analog of
+  // LocalSpawner's spawn-time injection: resolve the agent's whitelist into
+  // sessionManager.allowedToolsState so a restricted agent is restricted from
+  // its FIRST turn — not only after the next admin-triggered reload push.
+  //
+  // Awaited BEFORE server.listen (like syncAllResources above) so the box never
+  // accepts a prompt before its whitelist lands — first-turn restriction is the
+  // whole point of a security-relevant tool gate. On failure (after retries) the
+  // agent starts unrestricted (allowedToolsState stays null) until the next
+  // reload push — the safe-open default.
+  if (sessionManager.gatewayClient) {
+    const boxClient = sessionManager.gatewayClient.toClientLike();
+    try {
+      // Reuse syncResource's exponential-backoff retry (descriptor.retry =
+      // 3 attempts, 1s base) to shrink the fail-open window from a transient
+      // gateway blip. Pass the per-box handler since `tools` is not in the
+      // module-level registry that syncResource would otherwise look up.
+      const count = await syncResource("tools", boxClient, createToolsHandler(sessionManager, boxClient));
+      console.log(`[agentbox] Initial tool-capabilities synced: ${count === 0 ? "unrestricted" : `${count} tools`}`);
+    } catch (err) {
+      // Fail-open after all retries: a fresh box has no prior whitelist to fall
+      // back to, and a failed fetch usually means broader gateway unreachability
+      // (MCP/skills can't sync either). Loud warn so the gap is observable.
+      const msg = err instanceof Error ? err.message : String(err);
+      console.warn(`[agentbox] Initial tool-capabilities sync failed after retries (starting unrestricted): ${msg}`);
+    }
+  }
 
   const protocol = server instanceof https.Server ? "https" : "http";
   server.listen(PORT, () => {
--- a/src/core/tool-capabilities.ts
+++ b/src/core/tool-capabilities.ts
@@ -1,0 +1,100 @@
+/**
+ * Per-Agent Tool Capabilities — registry + resolution
+ *
+ * Capability groups are the user-visible, multi-selectable configuration unit.
+ * They decouple the vocabulary an admin reasons about ("read files", "run
+ * commands") from the internal tool names, so tools can be renamed or
+ * regrouped without invalidating an agent's stored selection.
+ *
+ * This is a pure module (no heavy deps) by design — mirroring `tool-append.ts`
+ * — so it stays unit-testable. `agent-factory.ts` pulls in ssh2 transitively
+ * and cannot be imported under vitest; keep the resolution logic here.
+ *
+ * Semantics mirror `appendAllowedTools` / `ToolRegistry.resolve()`:
+ *   null / empty selection → null (whitelist OFF; every tool passes).
+ * "Selecting nothing defaults to selecting everything" — this is the
+ * backward-compatibility hinge: an agent that never set `tool_capabilities`
+ * resolves to null and keeps today's full tool set.
+ */
+
+/**
+ * Capability group key → the internal tool names it grants.
+ *
+ * Copied verbatim from the design (per-agent-tool-capabilities-DESIGN.md
+ * "接口与数据结构"). The group keys are the stable contract stored in
+ * `agents.tool_capabilities`; the tool-name arrays may evolve as tools are
+ * added/renamed without changing stored selections.
+ */
+export const CAPABILITY_GROUPS: Record<string, string[]> = {
+  read_files:      ["read", "grep", "find", "ls"],
+  write_sandbox:   ["write", "edit", "skill_preview"],   // 含技能创作
+  inspect_infra:   ["cluster_list", "cluster_probe", "host_list", "resolve_pod_netns"],
+  run_commands:    ["bash", "node_exec", "pod_exec", "host_exec"],
+  run_scripts:     ["node_script", "pod_script", "local_script", "host_script"],
+  search_memory:   ["memory_search", "memory_get"],
+  plan_tasks:      ["task_create", "task_update", "task_list", "task_get"],     // 拆分①
+  spawn_subagents: ["spawn_subagent", "task_output", "job_stop"],               // 拆分①（权限放大）
+  scheduling:      ["manage_schedule"],
+  session_output:  ["task_report", "save_feedback", "channel_update"],   // 含 IM 渠道可见更新
+};
+
+/**
+ * Resolve a set of capability group keys to a concrete `allowedTools` list.
+ *
+ * - `null` / `undefined` / `[]` → `null` (whitelist off — all tools allowed).
+ *   "Selecting nothing defaults to selecting everything." This is the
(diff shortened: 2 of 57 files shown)


src/core/tool-capabilities.ts
```diff
@@ -0,0 +1,100 @@
...
+  write_sandbox:   ["write", "edit", "skill_preview"],   // 含技能创作
+  inspect_infra:   ["cluster_list", "cluster_probe", "host_list", "resolve_pod_netns"],
+  run_commands:    ["bash", "node_exec", "pod_exec", "host_exec"],
+  run_scripts:     ["node_script", "pod_script", "local_script", "host_script"],
+  search_memory:   ["memory_search", "memory_get"],
+  plan_tasks:      ["task_create", "task_update", "task_list", "task_get"],     // 拆分①
+  spawn_subagents: ["spawn_subagent", "task_output", "job_stop"],               // 拆分①（权限放大）
+  scheduling:      ["manage_schedule"],
+  session_output:  ["task_report", "save_feedback"],
+};
```
jacoblee-io, line 39: **Add a registry↔groups coverage guard.** Every name in `allToolEntries` should appear in some `CAPABILITY_GROUPS` entry, otherwise a restricted agent can never reach that tool (and "Select All" can't either). A test asserting this would have caught `channel_update` automatically and will catch the next un-grouped tool. On the rebase onto main, `channel_update` needs adding here (probably `session_output`).

src/core/tool-registry.ts
```diff
@@ -363,20 +360,15 @@ export class ToolRegistry {
...
       }
-      return {
-        def,
-        platform: e.platform ?? false,
-      };
+      return def;
     });
 
-    // 3. allowedTools whitelist (platform tools exempt)
+    // 3. allowedTools whitelist (sole availability axis; no exemptions)
```
jacoblee-io, line 381: Removing `platform` from `ToolEntry` will break the build after the rebase onto main: `src/tools/workflow/channel-update.ts` (not on this branch yet — from the Feishu merge f8ffe01) still declares `platform: true`, which becomes a TS2353 excess-property error. `tsc --noEmit` is green today only because that file isn't present. Drop `platform: true` from channel-update.ts during the rebase.

jacoblee-io (commented):
**Overall: strong, well-architected PR.** The `null = unrestricted` backward-compat hinge is held rigorously at every layer (resolve / encode / sync / snapshot), and test coverage is genuinely good — boundary coercion, the frontend↔backend drift guard, fail-safe paths, and the sync-descriptor contract are all exercised. The boundary-resolution design (group-keys→tool-names at the gateway, delivered over the existing sync framework with `invalidate`-on-reload) mirrors the established MCP/cluster/host pattern correctly.

One must-fix tied to the deferred rebase, plus a few lower-severity notes. Inline comments below.

### 🔴 Must fix — the deferred rebase breaks the build *and* silently drops a tool

`tool-refract` is behind `main` and lacks `src/tools/workflow/channel-update.ts` (added in the Feishu merge `f8ffe01`). That file still carries `platform: true`. After this PR removes `platform` from `ToolEntry`, the rebase produces:

1. **Compile break** — `channel-update.ts`'s `registration: ToolEntry = { …, platform: true }` becomes a TS2353 excess-property error. `npx tsc --noEmit` is green today only because the file isn't on the branch yet.
2. **Coverage gap** — `channel_update` is in **no** `CAPABILITY_GROUPS` entry. Once `platform` is gone it falls under the whitelist like everything else, so a restricted channel-mode agent (Feishu bot) silently loses the ability to post updates, and even "Select All" can't re-grant it.

On rebase: drop `platform: true` from `channel-update.ts` **and** add `channel_update` to a group (likely `session_output`).

### 🟡 Fail-open is a deliberate policy on a security gate
Both initial-fetch paths (K8s in `agentbox-main.ts`, Local in `local-spawner.ts`) start **unrestricted** on error. Reasoning is sound and documented, and reload-push failures correctly preserve the prior whitelist — but a transient gateway blip at pod start means a read-only agent boots with full tool access until the next reload. Worth a conscious sign-off.

### 🟢 Minor
- **Global `config.allowedTools` interaction**: `agent-factory.ts` falls back to the global when per-agent state is null. It defaults to `null` (so the "existing agents unaffected" claim holds), but any deployment that set a non-null global whitelist relied on the `platform` exemption keeping task/memory/schedule/spawn tools always-on — those will now be filtered unless explicitly listed. One-line note in `tools.md` would help.
- Subagent inheritance verified correct (child runs `web` mode, reports via captured output, no hidden `task_report` dependency).
- Dockerfile/tsconfig agentbox slimming verified safe — zero imports of the removed gateway modules anywhere in the agentbox build graph.

src/agentbox-main.ts
```diff
@@ -78,9 +79,41 @@ async function main() {
...
+      // gateway blip. Pass the per-box handler since `tools` is not in the
+      // module-level registry that syncResource would otherwise look up.
+      const count = await syncResource("tools", boxClient, createToolsHandler(sessionManager, boxClient));
+      console.log(`[agentbox] Initial tool-capabilities synced: ${count === 0 ? "unrestricted" : `${count} tools`}`);
+    } catch (err) {
+      // Fail-open after all retries: a fresh box has no prior whitelist to fall
+      // back to, and a failed fetch usually means broader gateway unreachability
+      // (MCP/skills can't sync either). Loud warn so the gap is observable.
+      const msg = err instanceof Error ? err.message : String(err);
+      console.warn(`[agentbox] Initial tool-capabilities sync failed after retries (starting unrestricted): ${msg}`);
```
jacoblee-io, line 151: Fail-open on a security/restriction gate: a transient gateway error at pod start boots the agent **unrestricted** until the next reload push. The reasoning (fresh box, no prior state, likely broader unreachability) is reasonable, and the reload-push path correctly preserves the prior whitelist on failure — just flagging this as a conscious policy decision to sign off on, not a silent default.

LikiosSedo:
## Review: LGTM — high-quality, well-tested feature (high confidence)

Clean design (the Gateway resolves group keys → tool names at its boundary; the AgentBox only ever consumes an already-resolved `allowedTools` list — a symmetric, well-drawn boundary), with rationale on every tradeoff. I found no blocking bugs. I verified the load-bearing invariants against a checkout rather than trusting the description.

### What I verified (all pass ✅)

| Check | Result |
|---|---|
| `platform` exemption fully removed | ✅ No leftover refs in registrations/registry (only unrelated `process.platform` / the skills `platform` dir) |
| **Coverage invariant** — all 26 registered tools' real `.name` belong to a capability group | ✅ Including the prime suspect `restricted-bash.js` → `.name="bash"` (matches `run_commands`). `tool-capabilities-coverage.test.ts` passes; no silently-dropped tool |
| Frontend ↔ backend group definitions identical | ✅ 10 groups, per-tool match → the drift guard passes |
| File I/O tools `read/edit/write/grep/find/ls` | ✅ Framework factories; `appendAllowedTools` filters by `name` at the right chokepoint; the `read_files` path has a manual smoke |
| `/api/reload-tools` dispatch | ✅ Route loop uses `perServerHandlers[type] ?? getSyncHandler(...)` so `tools` hits the per-box handler (correct mTLS cert → correct agentId); short-circuits to 200 without a gateway client |
| Dockerfile/tsconfig slimming is **runtime-safe** | ✅ The agentbox runtime graph imports zero `src/gateway/*` modules (only `shared/gateway-sync`), so this is safe at require-time, not just under tsc |
| Sub-agent privilege inheritance | ✅ The `session.ts` spawn path passes `allowedTools: this.allowedToolsState` to the child session |
| DB migration | ✅ Idempotent `safeAlterTable` + CREATE TABLE, MySQL+SQLite, `null = unrestricted` backward-compatible |
| Boundary decode | ✅ `decodeAgentRow` on POST/GET/PUT/clone; the empty-selection → null encoding hinge is tested |

### ⚠️ Two design tradeoffs admins must understand (not bugs, but load-bearing)

**1. MCP is fully exempt from the whitelist.** Capability restriction governs **built-in tools only**. A "read-only" agent restricted to `read_files` can still invoke any tool from a bound MCP server (including ones that run shell). This is least-privilege over built-ins, **not a full sandbox**. The code and UI say so ("MCP server tools are controlled separately"), but it's the easiest thing for an admin to misjudge.

**2. Initial-resolution failure is fail-open.** A failed K8s initial fetch / Local DB read starts the agent **unrestricted** (loud warn only). So the gate silently degrades to "no restriction" under gateway failure — it's a guardrail / least-privilege convenience, **not a hard security boundary that holds under failure**. Already flagged as reviewed policy with fail-closed hardening tracked as follow-up. Given the UI banner says "Everything else (except MCP) is **blocked**", please make both points explicit in admin-facing docs so the feature isn't over-trusted.

### Non-blocking nits

- **Unknown / forward group keys**: `encodeToolCapabilitiesForDb` intentionally does not validate that keys are known (forward-compat). But a selection of **only unknown keys** resolves to `[]` (a real restrictive whitelist that blocks **all** built-in tools), not `null`. The frontend can't trigger this (it only sends known keys), but the admin API accepts arbitrary string arrays — and the "forward-compatible storage" rationale half-conflicts with this (a purely-forward-key selection bricks built-in tools until that version ships). Worth a comment or a warn on the non-UI path.
- **Test-plan gap**: the backend `npm test` was not run in full this session (only frontend 23 + build). I hand-verified the three riskiest invariants (coverage / name-match / drift) and they hold, so risk is low — but since this is security-relevant, please run at least `tool-capabilities*`, `agent-api`, `sync-handlers`, `internal-api`, `gateway-sync`, and `local-spawner` before merge.
- Minor: the two `agent.reload` notify payloads differ in shape (`is_production` omits `agentId`; `tools` includes it). Harmless, pure consistency.

jacoblee-io (approved):

PR #353, opened 2026-06-24
feat(channels): Feishu group auto-bind + authorized gating + live progress card

LikiosSedo:
Runtime side of the Feishu channel work (companion to sicore MR k8s/sicore!546).

## 1. Group routing + open-group auto-bind (`818be80`)
- `channel-manager`/`lark`: thread `sender_open_id` through `channel.resolveBinding` so the Portal can auto-bind groups and pick the session key.
- standalone `adapter`: `channel.list` injects `group_channel_id` for open personal-bot channels; `resolveChannelBinding` adds an open-group fallback (one shared session per chat). Authorized stays Sicore-only.

## 2. Authorized-group access hint (`1c11aac`)
- `channel.resolveBinding` can return a `walled` result (sender unbound / no read access). `channel-manager` exports `ChannelAccessDenied` + `isChannelAccessDenied`; `lark` replies a short locale-aware hint instead of silently ignoring.

## 3. Surface Feishu send errors (`16df084`)
- The Feishu SDK doesn't throw on non-zero API codes. `openTypingCard`'s card-post and `replyToLark` now check the code and log it — a missing `im:message` scope used to look like a silent no-op.

## 4. Claude-tag progress card (`e276bb8`, `e1aa0f4`)
- Long channel investigations render an evolving checklist: ✅ done steps, ⏳ current, then the answer — instead of a "thinking…" spinner that jumps straight to the final reply.
- Milestones are **auto-derived from the agent's own narration** (first line of each intermediate assistant turn) — no reliance on the model calling a tool, no prompt nudge. Explicit `channel_update` milestones feed the same list.
- Bounded + safe: card shows the last 10 (`… (+k)` overflow), coalesced updates (never concurrent → respects Feishu's card update rate), consecutive-dedup.

## Tests
Full `vitest run`: 3805 passed. New cases: open-group / authorized resolution, walled hint, milestone derivation + checklist render + overflow cap.

## Verified live
open-group auto-bind, authorized DM, and the progress card all exercised on the sendong-sicore-test deployment (real Feishu group).

## Review fixes (session contract)
- **[P1]** open group now uses a **per-sender session** (`open_id:<sender>`, runs as the fixed owner) instead of a binding-level shared one; the group path threads the **server-authoritative `binding.sessionKey`** into the queue key + `/new` (open_id / sicore_user / "" binding-level). Fixes: concurrent writes to one shared session, and `/new` resetting the wrong/empty key (open + authorized). (sicore: per-sender open branch.)
- **[P2]** tests updated to the per-sender contract + added authorized `/new` reset (`sicore_user:<id>`).

### DIFF
--- a/src/gateway/channels/lark.ts
+++ b/src/gateway/channels/lark.ts
@@ -663,10 +663,16 @@ function formatPersonalPairReply(
 
 async function replyToLark(larkClient: any, messageId: string, text: string): Promise<void> {
   try {
-    await larkClient.im.message.reply({
+    // Feishu's SDK does NOT throw on a non-zero API code (e.g. missing
+    // im:message send scope) — it returns {code,msg} in the body. Surface it,
+    // otherwise a permission failure looks like a silent no-op.
+    const resp = await larkClient.im.message.reply({
       path: { message_id: messageId },
       data: { content: JSON.stringify({ text }), msg_type: "text" },
     });
+    if (resp && typeof resp.code === "number" && resp.code !== 0) {
+      console.error(`[lark] reply API returned non-zero code for messageId=${messageId}: code=${resp.code} msg=${resp.msg}`);
+    }
   } catch (err) {
     console.error(`[lark] Failed to reply to messageId=${messageId}:`, err);
   }
--- a/src/gateway/channels/lark.ts
+++ b/src/gateway/channels/lark.ts
@@ -27,6 +27,7 @@ import {
   openTypingCard,
   updateCardContent,
   finalizeCard,
+  buildMilestoneCardMarkdown,
   PLACEHOLDER_BY_LOCALE,
   EMPTY_RESULT_NOTICE_BY_LOCALE,
   localeForDomain,
@@ -65,7 +66,10 @@ const GROUP_ACCESS_DENIED_NOTICE_BY_LOCALE = {
   "zh-CN": "❌ 你没有这个助手的访问权限，请联系管理员授权。",
   "en-US": "❌ You don't have access to this assistant. Ask an admin to grant access.",
 } as const;
-const MAX_AGENT_SELECTED_UPDATES = 2;
+// Max milestones kept in the accumulating Claude-tag checklist. channel_update
+// milestones are meant to be sparse; if an agent over-emits we drop the oldest
+// so the card stays within Feishu's element size limits.
+const MILESTONE_CAP = 20;
 const MAX_LARK_BINDING_QUEUE = 20;
 
 interface QueuedLarkTask {
@@ -279,6 +283,11 @@ export async function handleLarkMessage(
   const senderOpenId = getLarkSenderOpenId(data);
   const sessionKey = buildLarkSessionKey(senderOpenId, chatId);
 
+  // Raw receipt log: fires for EVERY delivered event before any drop, so a
+  // group message that arrives but is filtered (non-text, empty after @-strip)
+  // is still visible. Lets us tell "never delivered" from "silently dropped".
+  console.log(`[lark] recv event chat=${chatId} chat_type=${chatType} msg_type=${msgType} sender=${senderOpenId ?? "?"} channelCfg=${channelId}`);
+
   if (msgType !== "text") return;
 
   let text: string;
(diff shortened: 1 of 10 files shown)


src/portal/adapter.ts
```diff
@@ -294,6 +301,34 @@ async function resolvePersonalChannelBinding(
...
+  if (personalBot.access_mode !== "open") return null;
+  if (personalBot.group_auto_bind === false) return null;
+  const sessionKey = `chat:${routeKey}`;
+  const session = await resolveChannelBindingParticipantSession(db, channel.id, sessionKey);
+  return {
+    agentId: personalBot.agent_id,
+    bindingId: channel.id,
+    sessionId: session.sessionId,
+    sessionKey: session.sessionKey,
+    createdBy: personalBot.owner_user_id ?? channel.created_by,
```
jacoblee-io, line 327: 🟠 **Open-group auto-bind + run-as-owner, enabled by default — worth a conscious sign-off before merge.**

`resolveOpenGroupBinding` makes an open personal bot auto-serve **any group it's added to**, and every turn runs as the **fixed owner** (`owner_user_id ?? channel.created_by`) with **no per-sender identity or authorization**. Combined with `channel.list` injecting `group_channel_id` by default (`group_auto_bind !== false`):

1. **Trust boundary widens from 1:1 → N**: anyone sharing a group with the bot can `@`-mention it and drive the owner's agent — including whatever clusters/hosts/tools that agent can reach. In an SRE context that can mean "in the group + @-mentioned the bot" ⇒ production `kubectl`/`ssh`.
2. **Silent enablement on deploy**: existing DM-only open bots will *start* answering in groups they're already in (since `group_auto_bind` defaults on).

Existing mitigations: the @bot gate limits it to deliberate invocations, `group_auto_bind: false` opts out, and `sicore_authorized` (Sicore-only) provides real per-user gating. This is clearly by design for "open" mode — but given the default-on + run-as-owner combination, I'd suggest either (a) defaulting `group_auto_bind` to **off** for open bots, or (b) a release/security note so operators consciously accept it.

src/portal/adapter.ts
```diff
@@ -2611,6 +2646,18 @@ export function buildAdapterRpcHandlers(): Map<string, (params: any, agentId: st
...
     for (const row of rows as any[]) {
       if (row.config !== undefined) row.config = safeParseJson(row.config, null);
+      // A per-agent open bot also serves the groups it joins. Advertise
+      // group_channel_id so the Runtime stops ignoring group messages (lark.ts
+      // gates the group path on it). DM-only bots set group_auto_bind:false.
+      const cfg = row.config;
+      if (
+        cfg && typeof cfg === "object" &&
+        cfg.personal_bot && typeof cfg.personal_bot === "object" &&
+        cfg.personal_bot.group_auto_bind !== false &&
```
jacoblee-io, line 2656: This is the default-on injection point: unless `group_auto_bind === false` is explicit, every open personal bot gets `group_channel_id` injected, which makes the Runtime start handling group messages — the source of the run-as-owner exposure above being on by default. If you adopt "default off", flipping the check here (require an explicit `=== true` to inject) is the one-line change.

src/gateway/channels/lark.ts
```diff
@@ -717,7 +886,17 @@ export async function collectChannelResponse(
         const blocks = Array.isArray(ev.message.content) ? ev.message.content : [];
         if (options.includeImages) collectImageAttachments(blocks, images, seenImageKeys);
         const turnText = contentBlocksToMarkdown(blocks);
-        if (turnText) lastAssistantText = turnText;
+        if (turnText) {
+          // A NEW assistant turn means the PREVIOUS one was an intermediate
+          // step (the agent narrated, then called a tool) — surface its first
+          // line as a progress milestone. The final turn is never followed by
+          // another, so it stays the answer, not a milestone.
+          if (lastAssistantText && options.onMilestone) {
```
jacoblee-io, line 894: 🟡 The first line of each *intermediate* assistant turn is surfaced to the channel as a milestone. It's the model's own narration (same trust as the final answer; `SAFETY_SECTION` applies), so not a new leak class — but the channel now exposes previously-hidden intermediate reasoning. Combined with the open-group exposure, all group members can see the investigation trail. Just confirm that's intended.

jacoblee-io (commented):
Read the full diff, traced the session-key/queue/`/new` flow end-to-end, and ran the 6 affected suites at the PR head — **321/321 pass**. Code quality and test coverage are strong, and the P1 session-contract fix is correct: explicitly-bound groups keep per-sender isolation (`buildLarkSessionKey` always returns a truthy key → `resolveChannelBindingParticipantSession` → non-null per-sender `sessionKey`), and the `?? ""` / `?? "__binding__"` fallbacks only apply to Sicore-returned legacy bindings — standalone never collapses existing per-sender group sessions into a shared one. `/new` resets the server-authoritative key because it runs inside `processQueuedLarkMessage` (where `sessionKey = effectiveSessionKey`); PAIR is handled before the @-mention gate, so it isn't blocked.

The **one thing to consciously sign off before merge is the open-group auto-bind exposure** (inline below). Everything else is a positive or a minor note.

### 🟢 Security positives in this PR
- The `@所有人` (@all) gating (`isBotMentioned`) is a genuine defensive fix — without it the bot replied to group-wide broadcasts it was never aimed at. Strict `open_id` match, with a sane `@_all`-exclusion degraded path.
- Surfacing non-zero API codes (`openTypingCard`/`replyToLark`) turns a silent missing-`im:message`-scope failure into a logged error.
- The `walled`/`ChannelAccessDenied` path correctly distinguishes "turn the sender away with a hint" from "ignore (no binding)"; `resolveQueuedBinding` re-checks access at dequeue (revoked between enqueue and run → treated as gone).

### 🟡 Minor
- **Milestone narration is now surfaced to the channel**: `onMilestone` emits the first line of every *intermediate* assistant turn as a visible checklist item. It's the model's own narration (same trust as the final answer; `SAFETY_SECTION` still applies), so not a new leak class — but the channel is now much more verbose and exposes previously-hidden intermediate reasoning. Combined with the open-group exposure below, that means all group members can see the investigation steps. Worth confirming it's intended.
- **Orphaned CardKit card on post-failure**: `openTypingCard` returns `null` on a non-zero code (good — falls back to text), but the card already created in CardKit is left dangling. Resource-only, not user-visible.

### What I verified
- 321/321 tests pass across `lark`, `lark-card`, `dingtalk`, `channel-manager`, `adapter-rpc`, `adapter.routes` at head `48f7a62`.
- Session-key invariant, `/new` resetting the server key, PAIR before the gate, and dequeue-time access re-check — all confirmed.
- adapter.ts bound path (row exists) is untouched by this PR — only the `!row` open-group fallback branch was added.

LikiosSedo:
Thanks for the thorough review — and for running the affected suites. Resolutions on the three points:

**🟠 Open-group auto-bind default-ON + run-as-owner (`adapter.ts:327` / `:2656`)** — Intentional; we're keeping the default ON. The core value of this feature is "pull one bot into N groups with zero PAIR", and flipping the default off defeats that. This is open-mode semantics by design: `open` already means every turn runs as the fixed owner for anyone who can reach the bot — groups widen the audience under the *same* model, they don't add a new identity/privilege path (tools still run with the agent's own resources, per R1). The exposure stays bounded by the existing controls:

- **@bot mention gating** — the bot only acts when actively @-invoked; it ignores `@all` and ambient group chatter.
- **`group_auto_bind: false`** — per-bot opt-out to keep a bot DM-only.
- **`sicore_authorized`** — for any group that isn't fully trusted, this gives real per-sender `im_bindings` + RBAC gating, and is the recommended mode when run-as-owner isn't acceptable.

Operational guidance (already in the design doc): place **open** bots only in groups where run-as-owner is acceptable; otherwise use `sicore_authorized`. We'll call out in the release notes that deploying this can make an existing open bot start answering in groups it already sits in, so operators accept it knowingly.

**🟡 Milestones surfaced to the channel (`lark.ts:894`)** — Intended. The progress card is the point of the Claude-tag UX, and it's the model's own narration (same trust level as the final answer; `SAFETY_SECTION` still applies). In open groups it follows the same "everyone in the group sees it" model as the auto-bind decision above, so no separate change.

**🟡 Orphaned CardKit card on post failure** — Acknowledged. It's resource-only and invisible to users, and CardKit cards are short-lived, so we're leaving it as-is for now; happy to add an explicit cleanup if it ever shows up in practice.

jacoblee-io (approved):

PR #354, opened 2026-06-24
refactor(metrics): remove dead dashboards and live-counter pipeline

chent1996:
## Summary

Removes two dead metrics-dashboard features plus their entire backend pipeline. The Portal "live" and "timing" dashboard widgets had no remaining frontend consumer, and removing them orphaned the whole in-memory live-counter subsystem ("path 1"). The Prometheus federation path ("path 2", which backs external Grafana) is kept fully intact.

See **ADR-014** (`docs/design/decisions.md`) for the decision record.

## What's removed

- **Timing dashboard**: `TimingStatsCard`, `useTimingStats`, `GET /metrics/timing`, `summariseLatency`
- **Live dashboard**: `RankedTable`, `useLive`, `GET /metrics/live`, the gateway `metrics.live` RPC
- **Live-counter subsystem (path 1)**: the whole `LocalCollector`, the aggregator's per-(user,agent) tool/skill counter maps + `mergeSnapshot`. `MetricsAggregator` is now a federation-only pull loop.
- Unified the agentbox pull-snapshot and the SIGTERM flush onto one `MetricsFlushPayload { incarnation, prom }`.

## What's kept

- **Prometheus federation (path 2)** — every federated `siclaw_*` series is unchanged (`sessions_active` / `tool_calls_total` / `skill_calls_total` come from prom-client, not LocalCollector). `GET /metrics`, the SIGTERM flush, and the federation aggregator are untouched.
- Per-message `ttft_ms` (still consumed by the chat timing badge).
- The `summary` / `audit` dashboards.

## Deployment note (rolling upgrade)

The agentbox `/api/internal/metrics-snapshot` payload slimmed from `{activeSessions, toolCallDeltas, skillCallDeltas, incarnation, prom}` to `{incarnation, prom}`. An **old gateway pulling a new agentbox** throws a caught, transient `TypeError` in its 30s pull loop (skips that federation cycle, self-heals, no data loss). **Roll the gateway Deployment before agentboxes** to avoid it; a new gateway tolerates the old payload. (ADR-014, Consequences.)

## Verification

- Backend + portal-web `tsc --noEmit`: clean
- Full backend suite: **3772 passed** / 2 skipped (188 files); portal-web: **144 passed**
- AgentBox boot smoke (module-graph resolution): `MODULE_EVAL_OK` — catches a dangling `local-collector` side-effect import that `tsc`/`vitest` are blind to (fixed in `agentbox-main.ts`)
- Net: **18 files, +88 / -1236**, 4 files deleted


LikiosSedo:
## Review: LGTM — clean, well-scoped dead-code removal (high confidence)

I read the full diff and verified the load-bearing claims against a checkout of the branch rather than just trusting the description. The claims hold up.

### What I verified (with evidence)

**1. No dangling references ✅**
Grepped all removed symbols across `src` + `portal-web` (`LocalCollector`/`localCollector`, `MetricsSnapshot`/`ToolCallStats`/`SkillCallStats`, `useLive`/`useTimingStats`/`TimingStats`/`RankedTable`/`TimingStatsCard`, `metrics.live`/`metrics/live`/`metrics/timing`, `summariseLatency`/`mergeSnapshot`) — **zero hits**. `Metrics.tsx`/`KpiCards.tsx` are untouched and tsc is clean, which confirms these widgets were genuinely orphaned (no remaining consumer).

**2. "External Grafana unaffected" is actually true ✅ (the key claim)**
This is the premise that makes deleting path ① safe. Confirmed in `src/shared/metrics.ts`: `toolCallsTotal.inc()` / `skillCallsTotal.inc()` / `sessionsActive.inc()` (metrics.ts:215 / 219 / 207) are an **independent prom-client subscriber** on the same diagnostic event bus (`tool_call` / `skill_call` / `session_created`), entirely separate from the removed `LocalCollector`. Removing `LocalCollector` cannot affect the federated series.

**3. Constructor signature change is consistent ✅**
`MetricsAggregator(mode, localRef, …)` → `(podLister, …)` is applied at every call site (`server.ts:514` + 7 sites across the two test files). In `server.ts`, `metricsAggregator` has exactly 3 references (decl `| undefined`, k8s construction, `?.destroy()`); local mode correctly leaves it `undefined`.

**4. Rolling-upgrade note holds ✅**
- New gateway ↔ old agentbox: `pullAll` reads only `snapshot.incarnation` / `snapshot.prom`, ignoring extra fields → tolerated.
- Old gateway ↔ new `{incarnation, prom}`: old `mergeSnapshot` throws `TypeError` on the absent `toolCallDeltas`, caught by `setInterval(() => this.pullAll().catch(...))` in `startPullLoop` (unchanged by this PR). So it really is a caught, skipped cycle. "Roll the gateway first" is the right mitigation.

### Non-blocking nits

- **ADR-014 wording**: "self-heals on the next pull" is slightly off — an old gateway pulling a permanently-new agentbox keeps throwing every 30s until the *gateway itself* is rolled; it doesn't recover on the next pull while the version skew persists. Suggest: "self-heals once the gateway is rolled / the version skew clears." (It's a transient, metrics-only gap, not data loss — that part is fine.)
- **Boundary re-check in `pullAll`** (`incarnation && prom` even though the type requires them): correct and intentional — the payload crosses the mTLS wire from another pod. Skipping a malformed pod instead of throwing is the right resilience choice. 👍
- **Editing the historical design doc** (`2026-06-15-metrics-time-range-picker.md`) to strip timing/live mentions is acceptable since ADR-014 is the authoritative new record; minor, since a design doc is somewhat a point-in-time artifact.

### Not done
Did not run the full `tsc`/vitest locally (shallow clone, no `node_modules`). The main thing tsc would catch — dangling references — is already disproven by grep, and the PR reports 3772 passing, so residual risk is low.

jacoblee-io (approved):
LGTM — clean, well-scoped deletion. Verified locally on the branch:

- Backend + portal-web `tsc --noEmit`: clean
- Zero dangling references to any removed symbol (`LocalCollector`/`mergeSnapshot`/`RankedTable`/`useLive`/`useTimingStats`/`TimingStatsCard`/`summariseLatency`/`/metrics/live`/`/metrics/timing`/`metrics.live`/`toolCallDeltas`/`skillCallDeltas`) across prod **and** test code; no remaining import of the deleted `local-collector` module
- Affected suites green (metrics-aggregator, federation-self-metrics, federation-snapshot, prom-federation, http-server — 87 tests)

Correctness:
- Path ② (Prometheus federation) genuinely intact — `metrics.ts` comment-only, `MetricsFlushPayload` kept (now carried byte-identically by both pull + SIGTERM flush), `/metrics` + federation aggregator untouched.
- Federation-only aggregator is robust: `startPullLoop` wraps the cycle in `.catch`, per-pod fetch in try/catch→null, `incarnation && prom` re-validated at the mTLS boundary. Constructor starting the loop unconditionally is fine since it's only built in K8s mode (`undefined` in local, guarded `?.destroy()`).
- Local mode: aggregator + `metrics.live` RPC removed with no remaining consumer; local `/metrics` still served from prom-client directly. No regression.
- The dangling `import "./shared/local-collector.js"` side-effect import in agentbox-main.ts is removed.
- ADR-014 is thorough and accurate.

One operational note (not blocking, already in the PR): the rolling-upgrade window is *old gateway pulling new agentbox* → caught transient TypeError, self-heals next cycle. Since agentbox has no Deployment (spawned from `SICLAW_AGENTBOX_IMAGE`), in practice that means **roll the runtime/gateway image first, then bump `SICLAW_AGENTBOX_IMAGE`**. New-gateway-pulls-old-agentbox is safe.

PR #357, opened 2026-06-24
fix(agentbox): enforce 300s minimum on idle self-destruct window

jacoblee-io:
## Summary

Add a **300s floor** to the per-agent idle self-destruct window, while keeping `0` (resident) as the explicit escape hatch. Follow-up to the configurable-idle feature (#350).

### Problem

The window accepted any non-negative integer with no lower bound. A small positive value (e.g. `1`–`30`s) churns pods: any pause longer than the window between turns drops the (one-per-agent) pod, so the next message pays a full cold start — pod spawn + mTLS cert + skills/config sync + brain init + JSONL session restore. For interactive chat (people routinely pause >30s) that's a laggy resume on nearly every message.

### Solution

Shared `normalizeIdleTimeoutSec` in `config.ts`:
- unset (`null`/`undefined`/`""`) → `300` (default)
- explicit `0` or negative → `0` (**resident**, never floored — the deliberate opt-out)
- `1`..`299` → `300` (floor)
- `>= 300` → itself (floored to int); invalid → `300`

Applied at **both** ends so no source can smuggle a sub-floor window:
- **write** — `agent-api` POST/PUT (stored + UI-displayed value is already compliant)
- **consumption** — `loadConfig()` final value, covering env var (`SICLAW_AGENTBOX_IDLE_TIMEOUT`), `settings.json`, and even a **legacy DB row** written before this floor (it's re-normalized when injected as env and read by the agentbox).

UI help text updated to state the 300 minimum + 0=resident.

## Test Plan

- [x] New `src/core/config.test.ts`: 0/neg→0, 1..299→300, ≥300 pass-through, unset/invalid→300
- [x] `tsc --noEmit` backend + portal-web: clean
- [x] agent-api / http-server / adapter-rpc suites green (235 tests)


chent1996 (approved):
LGTM

PR #358, opened 2026-06-24
fix(agent): apply idle-timeout change when leaving Resident + preset UI

jacoblee-io:
## Summary

Two related changes to the per-agent **Idle Timeout** feature:

1. **Fix: changing idle from Resident (0) → a finite window never took effect on a running instance.** The idle window is delivered as `SICLAW_AGENTBOX_IDLE_TIMEOUT` only at **cold spawn** and read once at AgentBox startup. A resident pod (idle=0) never self-destructs, so it never cold-spawns to pick up the new value — the `0 → 300` change was silently a no-op until the pod was manually recreated. (Other transitions — finite→finite, →0 — self-heal via normal idle recycling.)

2. **UI: preset picker** for the Idle Timeout field (matches the requested design).

## Changes

- **`src/portal/agent-api.ts`**: on a PUT that changes `idle_timeout_sec` from `0` → a positive value, reuse the runtime's existing `agent.terminate` RPC to stop the running box so the next message cold-spawns with the new window. Reads the old value before the UPDATE to detect the transition; scoped to the `0→positive` case so live boxes aren't disrupted for self-healing changes. No frontend↔backend coordination and no runtime code change needed — the runtime already exposes `agent.terminate` + lazy spawn-env resolution.
- **`portal-web/src/components/AgentSettings.tsx`** (Basic tab): replace the bare seconds input with quick presets (∞ Resident / 5 min / 15 min / 30 min / 1 h) + a seconds input + clearer help text noting that **switching off Resident now applies immediately**, while other changes still take effect on the next restart.

## Test Plan

- [x] `vitest run src/portal/agent-api.test.ts` — 40 pass, incl. 3 new cases (0→300 terminates; 300→600 does not; 0→0 does not)
- [x] `portal-web` `npm run build` — clean; `vitest run` — 167 pass

## Notes

- The create-agent dialog still has no idle field (new agents default to 300 server-side) — can add the same preset picker there in a follow-up if wanted.
- A sub-300 positive value typed in the box is still floored to 300 by the backend (unchanged); the help text explains it.

### DIFF
--- a/portal-web/src/components/AgentSettings.tsx
+++ b/portal-web/src/components/AgentSettings.tsx
@@ -335,6 +335,53 @@ export function AgentSettings({ agent, onUpdate, initialTab }: AgentSettingsProp
 
 // ── Tab Components ──────────────────────────────────────
 
+// Quick-pick windows for the idle self-destruct timer. Resident = 0 (the pod
+// never auto-destroys). Values are seconds; the backend floors any positive
+// value below 300 up to 300 (a shorter window churns instances).
+const IDLE_TIMEOUT_PRESETS: { label: string; value: number }[] = [
+  { label: "∞ Resident", value: 0 },
+  { label: "5 min", value: 300 },
+  { label: "15 min", value: 900 },
+  { label: "30 min", value: 1800 },
+  { label: "1 h", value: 3600 },
+]
+
+function IdleTimeoutField({ value, onChange }: { value: number; onChange: (v: number) => void }) {
+  return (
+    <div className="space-y-2">
+      <label className="text-[13px] font-medium text-foreground">Idle Timeout</label>
+      <div className="flex flex-wrap gap-2">
+        {IDLE_TIMEOUT_PRESETS.map(p => {
+          const active = value === p.value
+          return (
+            <button
+              key={p.value}
+              type="button"
+              onClick={() => onChange(p.value)}
+              className={`h-9 px-3 text-[13px] rounded-md border transition-colors ${active ? "border-primary bg-primary/10 text-foreground" : "border-border text-muted-foreground hover:text-foreground hover:bg-secondary/40"}`}
+            >
+              {p.label}
+            </button>
+          )
+        })}
+      </div>
+      <div className="relative w-44">
+        <input
+          type="number"
+          min={0}
+          value={value}
+          onChange={e => onChange(e.target.value === "" ? 0 : Math.max(0, Math.floor(Number(e.target.value))))}
+          className="w-full h-10 pl-3 pr-12 text-[15px] rounded-md border border-border bg-background focus:outline-none focus:ring-1 focus:ring-ring"
+        />
+        <span className="absolute right-3 top-1/2 -translate-y-1/2 text-[12px] text-muted-foreground pointer-events-none">sec</span>
+      </div>
+      <p className="text-[12px] text-muted-foreground/80 leading-relaxed">
+        The running instance self-destructs after this long with no active sessions (default 300s, minimum 300 — a smaller positive value is raised to 300, since a shorter window churns instances); choose Resident (0s) to keep it alive. Switching off Resident applies immediately; other changes take effect on the agent's next restart.
+      </p>
+    </div>
+  )
+}
+
 function BasicTab({ name, setName, description, setDescription, systemPrompt, setSystemPrompt, isProduction, setIsProduction, idleTimeoutSec, setIdleTimeoutSec }: {
   name: string; setName: (v: string) => void; description: string; setDescription: (v: string) => void
   systemPrompt: string; setSystemPrompt: (v: string) => void; isProduction: boolean; setIsProduction: (v: boolean) => void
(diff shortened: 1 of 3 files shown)


LikiosSedo:
## Review: LGTM — correct, well-scoped fix (high confidence)

Verified against a checkout of the branch. The root-cause analysis is right, the transition matrix is complete, and the implementation reuses existing primitives cleanly.

### What I verified

| Check | Result |
|---|---|
| **Transition matrix is complete** | `0 → positive` really is the *only* non-self-healing case (every other `old` is finite, so the box recycles on its own window); `0 → 0` is a no-op. The `old === 0 && new > 0` guard hits exactly it. ✅ |
| **Store and decision use the same `normalizeIdleTimeoutSec`** | Same function for the SET clause (L206) and the terminate decision (L277) — they can't drift; and `normalizeIdleTimeoutSec(0) === 0` preserves Resident, positives `< 300` floor to 300. ✅ |
| **404 fires before the notify** | `rows.length === 0 → 404 return` precedes every notify, so a missing agent never triggers terminate. ✅ |
| **Notify is fire-and-forget** | The PUT returns 200 immediately; terminate runs async and doesn't block the request. ✅ |
| **`agent.terminate` handles "no running box"** | It filters matching boxes and is a graceful no-op (stops 0/0, `ok: true`) when none are running. ✅ |
| **Tests** | 3 new cases cover the matrix (0→300 terminates / 300→600 doesn't / 0→0 doesn't). ✅ |

### Suggestions (all non-blocking)

**1. (low–med) `agent.terminate` is a hard stop — it interrupts an in-flight turn on a Resident agent.**
`agentBoxManager.stop()` resolves to `spawner.stop()` (K8s: pod delete / SIGTERM + grace; Local: stop + drop). So `0 → finite` *immediately* stops the running box — if that resident agent is actively serving someone, their current turn is cut off (the next message cold-spawns and restores from the persisted session, so it's not data loss, but it's a visible interruption). This is the necessary cost of making the setting apply, and it's correctly scoped to the only non-self-healing case. But the help text says "Switching off Resident applies immediately" without noting the running instance is stopped — consider adding a few words (e.g. "restarts the running instance"), and/or a "terminate when next idle" graceful variant as a follow-up if interruption matters.

**2. (micro-nit) `oldIdleTimeoutSec === 0` is a strict compare on a raw DB value.** INT columns come back as JS numbers from mysql2/sqlite, so it's fine today; but a driver that returned `"0"` as a string would silently miss the transition (no terminate, fix doesn't fire). `Number(oldIdleTimeoutSec) === 0` would be more robust.

**3. (minor, pre-existing UX)** Clearing the number input silently means `0` = Resident (`value === "" ? 0`, unchanged from before); a sub-300 value typed in the box shows as typed but is floored to 300 server-side and only reflects after reload. Help text already explains the floor.

### Nice
- Accurate root cause; exhaustive transition matrix; only acts on the one case that can't self-heal, leaving self-healing changes to not disturb a live box.
- Zero runtime change (reuses `agent.terminate` + lazy spawn-env resolution).
- Same normalize fn for store + decision (no consistency gap); 404 before notify; terminate is a no-op when nothing is running.

jacoblee-io:
### Follow-up (not in this PR): make idle-timeout a live-reload sync type

This PR fixes the worst symptom — changing idle `0` → finite never took effect on a resident pod — by notifying the existing `agent.terminate` RPC on the `0 → positive` transition. That works, but it **stops the shared per-agent box**, which can interrupt an in-progress reply if a turn is running at save time.

The cleaner long-term fix is to convert idle-timeout into a live-reload **Gateway-sync type** (same framework as `tools`), so a change hot-reloads the running box in place instead of killing it:

- Add a descriptor (`reloadPath: /api/reload-idle`, `gatewayPath: /api/internal/idle-timeout`, `initialSync:false`, per-box handler mirroring `createToolsHandler`).
- Portal: on `idle_timeout_sec` change, push `agent.reload {resources:["idle"]}` instead of `agent.terminate`.
- AgentBox: make `resolvedIdleMs` mutable and have the handler re-arm `checkIdle()` / clear the resident state in place (no `process.exit`, no pod kill).

**Benefits**
- No interrupted turns — applies with zero disruption.
- Handles all transitions uniformly (`0→N`, `N→M`, `N→0`), removing the cold-spawn-only special case and the `agent.terminate` heuristic in `agent-api.ts`.
- Closes the gap where the change silently doesn't apply if the gateway WS is briefly down at save time (next reload retries).

**Notes**
- Cold-spawn env stays as the initial value (like `tools`' out-of-band initial fetch); the sync type only governs live updates.
- Requires an `http-server.ts` idle-timer refactor (`resolvedIdleMs` const → mutable + a re-arm function) + a per-box handler.

Tracking here as a follow-up rather than expanding this PR's scope.

LikiosSedo:
**Review summary** — verified the diff against the #350 → #357 → #358 chain.

## Background this fixes

The idle window is injected as `SICLAW_AGENTBOX_IDLE_TIMEOUT` only at **cold spawn** and read once at AgentBox startup, so exactly one transition can't self-heal:

| Transition | Self-heals? | Why |
|---|---|---|
| finite → finite | ✅ | box self-destructs on its current window; next spawn reads the new value |
| finite → 0 (resident) | ✅ | same — dies once on the old window, respawns resident |
| **0 (resident) → finite** | ❌ | a resident pod never self-destructs → never cold-spawns → never picks up the new value |

So `0 → 300` was a silent no-op until manual pod recreation. This PR forces a cold spawn only for that one stuck transition.

## Looks correct

- Transition guard `oldIdle === 0 && (newIdle ?? 0) > 0` precisely targets the only non-self-healing case and leaves live boxes untouched for the rest.
- Number-coercion defenses are right and covered by tests: `raw == null ? null : Number(raw)` stops a missing/null row from reading as resident (`Number(null) === 0` would otherwise cause a spurious terminate), and coerces a stringified `"0"` so the `=== 0` check still fires.
- The terminate decision uses the **normalized** new value, so `0 → 100` correctly terminates and persists the floored `300` (asserted in tests).
- UPDATE persists before the terminate, so the next cold spawn reads the new value via `resolveAgentSpawnEnv`. The extra `SELECT` only runs when `idle_timeout_sec` is in the body.

## Non-blocking notes

1. `0 → finite` hard-terminates a resident box, which can interrupt an in-progress reply. It's the only option here (a resident pod has no natural idle to wait for) and the help text now warns of it — worth a quick confirm that `agent.terminate` drains the active session gracefully rather than hard-killing.
2. A sub-300 value typed in the seconds box shows as-typed until reload, while the backend floors it to 300 (same minor display lag as #357).
3. `as any` on the query result matches the file's existing style but is a smell.

LGTM.

LikiosSedo (approved):
LGTM — approving. Correct, well-scoped fix; details in my review comment above. The 3 items there are non-blocking suggestions, not change requests.

PR #359, opened 2026-06-24
feat(lark): show only the current step on the Feishu card, replace with the conclusion

LikiosSedo:
## What
The group progress card accumulated every milestone into a growing checklist, and the final card kept that whole trail above the conclusion. This switches to a live single-step indicator:

- **During work**: the card shows only the latest step (⏳), replaced in place — no accumulating checklist.
- **On finalize**: the step is replaced entirely by the conclusion — no step trail remains on the card.

`milestones[]` is now just an internal dedup buffer; renders use the latest entry only.

## Why
Follow-up to #353. The accumulating checklist made the card verbose and exposed the full intermediate investigation trail in groups; the ask is for the card to show only what's happening now and end as just the answer. This also reduces how much of the intermediate trail is exposed in open groups (addresses the milestone-exposure note from the #353 review).

## Tests
- Updated the `lark.test.ts` integration test to assert latest-step-only rendering and conclusion-only finalize.
- `npx vitest run src/gateway/channels/` — 158 passed.

### DIFF
--- a/src/gateway/channels/lark.test.ts
+++ b/src/gateway/channels/lark.test.ts
@@ -1049,7 +1049,7 @@ describe("handleLarkMessage — streaming card flow", () => {
     clearBackgroundChannelDelivery(sessionId);
   });
 
-  it("accumulates agent milestones into a Claude-tag checklist on the card", async () => {
+  it("shows only the latest step on the card and replaces it with the conclusion on finalize", async () => {
     resolveBindingMock.mockResolvedValue(makeBinding());
     promptMock.mockResolvedValue({ sessionId: "s-explicit-channel" });
     let releaseStream: () => void = () => {};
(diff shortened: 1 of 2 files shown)


jacoblee-io (approved):
LGTM — clean, focused, and well-covered. Approving. The notes below are non-blocking; feel free to take or leave them.

**What this does** (verified by walking each call path): the group progress card switches from an accumulating checklist to a live single-step indicator. During work it renders only `milestones.slice(-1)` (⏳ latest step, replaced in place); on `final`/finalize it passes `milestones: []` + `finalText`, so the card ends as just the conclusion. `milestones[]` is now only a dedup buffer. The pure `buildMilestoneCardMarkdown` is left unchanged — the behavior change is entirely at the call sites, which is the right call.

Correctness checks I ran through:
- streaming `slice(-1)` → `all.length === 1`, `isFinal=false` → renders `⏳ <latest>` ✅
- finalize `{ milestones: [], finalText }` → just the conclusion ✅
- no-milestone-yet → `slice(-1)` is `[]` → function returns `""` → `if (md.trim())` skips the update, placeholder card stays (unchanged from before) ✅
- `addMilestone` dedup (`milestones[len-1] === t`) still runs against the full buffer, unaffected by the render-time slice ✅
- coalesced `do/while` flush re-renders the latest step each pass ✅

Non-blocking nits:

1. **Stale docstring on `buildMilestoneCardMarkdown` (`lark-card.ts:62-76`).** It still describes "accumulated milestones as a checklist" with the ✅-done / ⏳-doing UX. The function is still generic and correct, but its only production caller (`lark.ts`) now always passes ≤1 milestone, so that documented multi-step checklist UX no longer happens in production. A one-line note ("current callers pass a single step; the multi-milestone checklist path is currently unused") would save a future reader some confusion.

2. **`lark-card.test.ts` is unchanged** and still asserts the multi-milestone checklist behavior of the pure function. That's fine (the function didn't change), but it means the unit tests assert the old UX while the integration test (`lark.test.ts`) asserts the new single-step UX. Pairs with nit 1 — a comment, or eventually pruning the now-unused checklist branch, would reconcile them.

3. `maxVisible` / `MILESTONE_CAP=20` are now effectively dead for the render path (inputs are always ≤1). `MILESTONE_CAP` is kept as a memory bound and the comment explains that — no change needed; just flagging.

One product confirmation (not a code issue): the final card intentionally drops the step trail, so observers can't see which steps the agent took after it finishes. The PR description frames this as the intended ask (and it addresses the #353 milestone-exposure note), so this is by design — just calling out that it's deliberate information loss, not a regression.

PR #360, opened 2026-06-24
feat(metrics): maintainer audit — per-entry axis, Sessions/Tools tabs, timing, read-only session snapshot

jacoblee-io:
Builds out the Portal **Metrics** page into a maintainer audit (**Dashboard / Sessions / Tools**) with a **per-entry-form axis** — Web / API / A2A / Channel / Scheduled, plus a combined Overview. Feedback is intentionally left out for now: siclaw has no feedback data source (`save_feedback` is a no-op stub).

## Commits (logical units)

1. **`feat(audit)` — full + categorizable session audit across all entry forms**
   Persist `chat_sessions.origin` for every entry form so the audit can categorize them: web=`NULL`, `api`, `a2a`, `channel` (Feishu/DingTalk unified), `task` (scheduled), `delegation` (sub-agent trace). Previously Feishu only persisted the user message and DingTalk persisted nothing — IM sessions showed up hollow or invisible.

2. **`feat(metrics)` P1 — per-entry audit axis + timing & sessions endpoints**
   - `src/portal/metrics-entry.ts`: shared entry predicates with **delegation inheritance** (a sub-agent's tool calls count under its parent's entry).
   - `metrics-timing.ts`: `summariseLatency` (count/avg/min/max/nearest-rank p90) + `extractTimingMs`.
   - Entry-aware `/metrics/summary` + `/metrics/audit`; new `/metrics/timing` and `/audit/sessions`.

3. **`feat(metrics)` P2/P3 — Sessions tab, shared Entry selector, timing card**
   - Page-level **Entry selector** (Overview / Web / API / A2A / Channel / Scheduled) shared across Dashboard / Sessions / Tools.
   - **Sessions** tab: per-session message/tool/error counts grouped by agent.
   - Dashboard **TimingStatsCard** (TTFT / thinking / per-tool p90).
   - Tools tab (was "Audit") entry-driven; rows show agent name.

4. **`feat(metrics)` — admin read-only session snapshot (modal, chat-identical render)**
   The Sessions tab lists every user's session, but the owner-scoped chat messages endpoint 404s on another user's session. Add an admin-only, read-only `GET /audit/sessions/:id/messages` (not user-scoped) + a centered modal that renders through the **same `PilotArea`** the live chat uses (additive `readOnly` prop hides composer/edit/steer; live chat untouched).

## Notes
- `src/portal/adapter.ts` (the internal mTLS mirror) intentionally **not** extended — portal-web (`siclaw-api.ts`, JWT) is the only consumer of these metrics endpoints.
- All metrics endpoints are `requireAdmin`.
- A >1000-message session shows the oldest 1000 (folding needs contiguous chronological order), flagged by a `truncated` banner. Rare.

## Verification
- `npx tsc --noEmit` clean (backend + portal-web)
- Backend: `siclaw-api.misc.test.ts` (45), `metrics-entry.test.ts`, `metrics-timing.test.ts` green
- portal-web: `npm run build` + vitest (172) green
- E2E on siclaw-inner: entry selector isolates per-entry counts; Sessions snapshot opens any user's transcript read-only in a modal.

### DIFF
--- a/src/gateway/channels/lark.ts
+++ b/src/gateway/channels/lark.ts
@@ -23,6 +23,7 @@ import {
 import type { FrontendWsClient } from "../frontend-ws-client.js";
 import { sessionRegistry } from "../session-registry.js";
 import { appendMessage, ensureChatSession } from "../chat-repo.js";
+import { buildRedactionConfigForModelConfig, redactText } from "../output-redactor.js";
 import {
   openTypingCard,
   updateCardContent,
@@ -638,6 +639,10 @@ async function processQueuedLarkMessage(ctx: QueuedLarkMessageContext): Promise<
     const collected = await collectChannelResponse(client, promptResult.sessionId, "lark", {
       includeImages: true,
       onMilestone: addMilestone,
+      // Audit: persist assistant + tool rows so the channel transcript matches
+      // web/api/a2a (origin="channel" set on the session above). Tool output on
+      // this stream is already sanitized at the agentbox boundary.
+      persist: { agentId },
     });
     resultText = collected.text;
     replyImages = collected.images;
@@ -855,15 +860,29 @@ export async function collectResponse(
   client: AgentBoxClient,
   sessionId: string,
   logPrefix = "lark",
+  options: { persist?: ChannelPersistContext } = {},
 ): Promise<string> {
-  return (await collectChannelResponse(client, sessionId, logPrefix)).text;
+  return (await collectChannelResponse(client, sessionId, logPrefix, { persist: options.persist })).text;
+}
+
+/**
+ * Opt-in audit persistence for the channel path. When set, collectChannelResponse
+ * writes the same user/assistant/tool transcript that web/api/a2a get via the
+ * runtime's sse-consumer, so IM-channel sessions are fully auditable (not just
+ * the inbound user message). `modelConfig` drives the same apiKey/baseUrl
+ * redaction the sse-consumer applies. The caller has already persisted the user
+ * message + ensured the session row (origin="channel").
+ */
+export interface ChannelPersistContext {
+  agentId: string;
+  modelConfig?: { apiKey?: string; baseUrl?: string };
 }
 
 export async function collectChannelResponse(
   client: AgentBoxClient,
   sessionId: string,
   logPrefix = "lark",
-  options: { includeImages?: boolean; onMilestone?: (text: string) => void } = {},
+  options: { includeImages?: boolean; onMilestone?: (text: string) => void; persist?: ChannelPersistContext } = {},
 ): Promise<CollectedChannelResponse> {
   const parts: string[] = [];
   const images: RenderedReplyImage[] = [];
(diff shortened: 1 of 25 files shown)


src/portal/siclaw-api.ts
```diff
@@ -3072,6 +3095,187 @@ export function registerSiclawRoutes(router: RestRouter, config: SiclawConfig, c
...
+        AND m.created_at >= ? AND m.created_at <= ? AND ${msg.predicate}`;
+    if (userFilter) { tSql += " AND s.user_id = ?"; tParams.push(userFilter); }
+    tSql += " ORDER BY m.created_at DESC LIMIT ?";
+    tParams.push(TIMING_ROW_LIMIT + 1);
+    const [tRows] = await db.query(tSql, tParams) as [Array<{ toolName: string; durationMs: number }>, unknown];
+    const byTool = new Map<string, number[]>();
+    for (const r of tRows.slice(0, TIMING_ROW_LIMIT)) {
+      const arr = byTool.get(r.toolName) ?? [];
+      arr.push(Number(r.durationMs));
+      byTool.set(r.toolName, arr);
```
chent1996, line 3148: Non-blocking — `metrics/timing` only reads `metadata.timing.{ttft_ms,thinking_ms}` from assistant rows, but those metadata fields are written exclusively by `src/gateway/sse-consumer.ts`, which is on the web/api/a2a path. The new channel persist in `collectChannelResponse` writes assistant rows without that metadata, so with `entry=channel` the TTFT/thinking cards will always be `count=0` while `tools[].count` populates fine from `duration_ms`. Worth either greying those cards in the UI when entry=channel, or noting the limitation in the dashboard copy so it doesn't look like a bug.

src/gateway/channels/lark.ts
```diff
@@ -869,14 +888,63 @@ export async function collectChannelResponse(
...
+        const name = (ev.toolName as string) || (ev.name as string) || "tool";
+        pushQ(toolInputs, name, ev.args ? JSON.stringify(ev.args) : "");
+        pushQ(toolStarts, name, Date.now());
+      }
+
+      if (ev.type === "tool_execution_end" || ev.type === "tool_end") {
+        if (options.includeImages) collectImageAttachments(ev.result?.content, images, seenImageKeys);
+        if (persist) {
+          const name = (ev.toolName as string) || (ev.name as string) || "tool";
+          const resultText = Array.isArray(ev.result?.content)
```
chent1996, line 933: Same FIFO-by-toolName limitation as `sse-consumer.ts` (which you explicitly model after) — multiple concurrent same-name tool calls can mispair start/end and produce a wrong `durationMs`. Pre-existing behaviour upstream so fine to inherit; just flagging since `metrics/timing` will now surface it for channel sessions too if anyone investigates anomalies.

chent1996 (commented):
Nice work bringing the audit axis closer to sicore parity. The per-entry predicate with delegation inheritance (`entryMessagePredicate`) is the right shape, and routing the admin snapshot through `PilotArea`'s real renderer (rather than re-implementing the chat view) is a clean reuse — `buildPilotMessages` extraction is exactly the seam that needed to exist.

A couple of small non-blocking notes inline. The PR branch is one commit behind main (PR #359, `d428bcd`, lark single-step card UX) — please rebase before merge so the in-progress card behaviour stays as #359 intended; the two diffs touch disjoint regions of lark.ts so it should be mechanical.

src/portal/siclaw-api.ts
jacoblee-io: Good catch — fixed in d58b578. TimingStatsCard now shows a one-line hint when `entry=channel` ("TTFT & thinking aren't recorded for channel (IM) sessions — only tool latency below"), since those come from assistant `metadata.timing` which only the sse-consumer path writes. Tool latency still populates from `duration_ms`.

src/gateway/channels/lark.ts
jacoblee-io: Agreed — inherited from sse-consumer by design. Updated the comment to admit it honestly instead of claiming pairing is always correct: concurrent same-name calls finishing out of order can mispair and skew that row's durationMs (audit metric only, never the reply). d58b578.

jacoblee-io:
Thanks for the review! Addressed both notes and rebased.

- **Rebased** onto current `main` (now includes #359 `d428bcd`). The lark.ts regions were disjoint as you predicted — clean rebase, `tsc` + lark/dingtalk/metrics suites green (608 tests).
- **channel timing cards** → hint added so a 0 count doesn't read as a bug (d58b578).
- **lark FIFO** → comment corrected to admit the inherited same-name-concurrency caveat (d58b578).

Branch head is now `d58b578`.

chent1996 (approved):
LGTM

PR #361, opened 2026-06-25
feat(agentbox): per-agent PVC persistence (runtime mechanism only)

glow-Kev1n:
Lower the session/memory persistence decision from a runtime-global env flag to an optional per-agent value, so one runtime can host both ephemeral agents (emptyDir; context cleared on pod restart/idle) and durable agents (PVC-backed; context survives restarts), without siclaw's own portal needing to store or expose the toggle.

Mechanism only — siclaw core does NOT persist or surface a per-agent flag. A downstream portal that wants per-agent persistence fills `binding.persistence` in its own `config.getModelBinding` handler; siclaw resolves it server-side by `agentId` on cold spawn. When the portal does not set it (siclaw's native portal, legacy callers, tests), the value is undefined and the spawner falls back to its global config — behaviour identical to upstream.

- types: AgentBoxConfig.persistence? (per-agent override)
- agent-model-binding: ResolvedModelBinding.persistence? (carried, never set by core)
- manager: setPersistenceResolver(agentId → binding.persistence), consulted only on a cold spawn (mirrors setSpawnEnvResolver); every getOrCreate entry point honours it, so the entry point is irrelevant
- server: register the resolver in startRuntime; chat.send no longer carries a per-request persistence param (single source of truth = the agentId-keyed resolver). task-coordinator still passes binding.persistence explicitly (explicit value wins)
- k8s-spawner: resolve boxConfig.persistence ?? global; mount shared PVC only when a claimName is configured, else degrade to emptyDir + warn (never spawn an unmountable pod)
- bootstrap-runtime: claimName has no hardcoded default (differs per deployment) and is decoupled from the global enabled flag, so a per-agent opt-in can mount the PVC even when the global default is off
- k8s-spawner.test: +5 persistence-matrix cases

Tests: tsc --noEmit clean; full suite 3935 passed / 2 skipped.

## Test Plan

- [ ] `npm test` passes

## Architecture Checklist

## General Checklist

### DIFF
--- a/src/gateway/agent-model-binding.ts
+++ b/src/gateway/agent-model-binding.ts
@@ -31,6 +31,12 @@ export interface ResolvedModelBinding {
   modelRouting?: ModelRoutePolicy;
   /** Agent's custom system prompt template (agents.system_prompt). Null/absent = built-in default. */
   systemPrompt?: string | null;
+  /**
+   * Per-agent session/memory persistence toggle. siclaw core leaves this
+   * undefined (no native per-agent store); a product portal that wants per-agent
+   * persistence resolves it from its own data and carries it over chat.send.
+   */
+  persistence?: boolean;
 }
 
 export async function resolveAgentModelBinding(
--- a/src/gateway/agentbox/manager.ts
+++ b/src/gateway/agentbox/manager.ts
@@ -105,6 +105,15 @@ export class AgentBoxManager {
    * on warm-pod reuse, so the chat hot path and channel/cron paths pay nothing
    * when the pod already exists.
    */
+  /**
+   * Per-agent config (e.g. `config.persistence`) is only consumed on a COLD
+   * spawn. A pod is keyed by agentId; when one is already running it is reused
+   * as-is, so persistence is effectively anchored at the agent's first spawn.
+   * Changing it on a later call does NOT recycle the pod — K8s cannot hot-change
+   * a running pod's volume mounts — and takes effect only on the next cold spawn
+   * (after the pod restarts or is idle-released). Callers are expected to send a
+   * stable value per agent.
+   */
   async getOrCreate(agentId: string, config?: Partial<AgentBoxConfig>): Promise<AgentBoxHandle> {
     if (!agentId) throw new Error("AgentBoxManager.getOrCreate requires an agentId");
     if (this.isK8s) {
@@ -118,6 +127,9 @@ export class AgentBoxManager {
 
     const info = await this.spawner.get(name);
     if (info && info.status === "running" && info.endpoint && this.isCertFresh(info)) {
+      // Warm reuse: return the running pod without spawning, so a changed
+      // config.persistence never reaches the pod spec (anchored at cold spawn —
+      // see getOrCreate docstring).
       return { boxId: name, endpoint: info.endpoint, agentId };
     }
     if (info && info.status === "running" && !this.isCertFresh(info)) {
@@ -143,6 +155,8 @@ export class AgentBoxManager {
       existing.lastActiveAt = new Date();
       const info = await this.spawner.get(existing.handle.boxId);
       if (info && info.status === "running") {
+        // Warm reuse: cached running box returned without spawning, so a changed
+        // config.persistence is ignored until the next cold spawn (see getOrCreate).
         return existing.handle;
       }
       this.boxes.delete(agentId);
--- a/src/gateway/agentbox/manager.ts
+++ b/src/gateway/agentbox/manager.ts
@@ -40,6 +40,7 @@ export class AgentBoxManager {
   private healthCheckTimer?: ReturnType<typeof setInterval>;
   private readonly isK8s: boolean;
   private spawnEnvResolver?: (agentId: string) => Promise<Record<string, string> | undefined>;
+  private persistenceResolver?: (agentId: string) => Promise<boolean | undefined>;
 
   constructor(spawner: BoxSpawner, config?: AgentBoxManagerConfig) {
     this.spawner = spawner;
@@ -65,6 +66,22 @@ export class AgentBoxManager {
    */
   setSpawnEnvResolver(fn: (agentId: string) => Promise<Record<string, string> | undefined>): void {
     this.spawnEnvResolver = fn;
+  }
+
+  /**
+   * Inject a resolver for the per-agent PVC persistence mode. Same contract as
+   * setSpawnEnvResolver: consulted on EVERY cold spawn (from any entry point —
+   * chat RPCs, channel webhooks, cron tasks, abort/steer) and NEVER on warm
+   * reuse. This is what makes persistence a true agent-level property: the
+   * value is resolved by agentId, independent of which entry point first
+   * cold-spawns the (one-per-agent) pod. Without it, only entry points that
+   * happened to pass `config.persistence` would honour it, so a pod cold-spawned
+   * by e.g. a Lark message would silently fall to the global default and ignore
+   * the agent's configured mode. Returns undefined to fall back to the global
+   * config (the spawner gates the actual mount on a claimName regardless).
+   */
+  setPersistenceResolver(fn: (agentId: string) => Promise<boolean | undefined>): void {
+    this.persistenceResolver = fn;
   }
 
   startHealthCheck(): void {
@@ -100,19 +117,19 @@ export class AgentBoxManager {
   }
 
   /**
-   * Get a running AgentBox for the agent, or spawn one. On a cold spawn the
-   * injected `spawnEnvResolver` (if any) is consulted for per-agent env — never
-   * on warm-pod reuse, so the chat hot path and channel/cron paths pay nothing
-   * when the pod already exists.
-   */
-  /**
-   * Per-agent config (e.g. `config.persistence`) is only consumed on a COLD
-   * spawn. A pod is keyed by agentId; when one is already running it is reused
-   * as-is, so persistence is effectively anchored at the agent's first spawn.
-   * Changing it on a later call does NOT recycle the pod — K8s cannot hot-change
-   * a running pod's volume mounts — and takes effect only on the next cold spawn
-   * (after the pod restarts or is idle-released). Callers are expected to send a
-   * stable value per agent.
+   * Get a running AgentBox for the agent, or spawn one.
+   *
+   * Per-agent config — the injected `spawnEnvResolver` (env, e.g. idle timeout)
+   * and `persistenceResolver` (PVC mode) — is resolved ONLY on a cold spawn,
+   * never on warm-pod reuse, so the chat hot path and channel/cron paths pay no
+   * RPC when the pod already exists.
+   *
+   * Because a pod is keyed by agentId, the persistence/env mode is resolved on
+   * each cold spawn (including a cert-stale recreate) from the agent-level
+   * resolver — NOT from whichever entry point happens to call first. The volume
+   * mount is fixed at pod creation (K8s cannot hot-change a running pod's
+   * mounts), so a configuration change applies on the agent's next cold spawn
+   * (after restart/idle-release), not immediately on a warm pod.
    */
   async getOrCreate(agentId: string, config?: Partial<AgentBoxConfig>): Promise<AgentBoxHandle> {
     if (!agentId) throw new Error("AgentBoxManager.getOrCreate requires an agentId");
@@ -127,9 +144,9 @@ export class AgentBoxManager {
 
     const info = await this.spawner.get(name);
     if (info && info.status === "running" && info.endpoint && this.isCertFresh(info)) {
-      // Warm reuse: return the running pod without spawning, so a changed
-      // config.persistence never reaches the pod spec (anchored at cold spawn —
-      // see getOrCreate docstring).
+      // Warm reuse: return the running pod without spawning. Per-agent config
+      // (env/persistence) is NOT re-resolved here — the pod's volume mount is
+      // already fixed, so a changed mode applies on the next cold spawn.
       return { boxId: name, endpoint: info.endpoint, agentId };
     }
     if (info && info.status === "running" && !this.isCertFresh(info)) {
@@ -142,6 +159,7 @@ export class AgentBoxManager {
     const handle = await this.spawner.spawn({
       ...config,
       agentId,
+      persistence: await this.resolvePersistence(agentId, config?.persistence),
       env: Object.keys(resolvedEnv).length > 0 ? resolvedEnv : undefined,
     });
 
@@ -155,8 +173,8 @@ export class AgentBoxManager {
       existing.lastActiveAt = new Date();
       const info = await this.spawner.get(existing.handle.boxId);
       if (info && info.status === "running") {
-        // Warm reuse: cached running box returned without spawning, so a changed
-        // config.persistence is ignored until the next cold spawn (see getOrCreate).
+        // Warm reuse: cached running box returned without spawning. Per-agent
+        // config (env/persistence) is NOT re-resolved — applies on next cold spawn.
         return existing.handle;
       }
       this.boxes.delete(agentId);
@@ -168,6 +186,7 @@ export class AgentBoxManager {
     const handle = await this.spawner.spawn({
       ...config,
       agentId,
+      persistence: await this.resolvePersistence(agentId, config?.persistence),
       env: Object.keys(resolvedEnv).length > 0 ? resolvedEnv : undefined,
     });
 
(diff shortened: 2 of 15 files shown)


LikiosSedo:
Thanks for the clean split in the runtime code here. I think the K8s spawner logic is directionally right, but I see one deployment-contract issue that should be fixed before merge.

The PR intends to support `global persistence default = off` while allowing an external/product portal to opt a specific agent into PVC persistence via `chat.send.persistence=true`. In the Helm deployment path, though, the shared PVC capability is still gated by `.Values.agentbox.persistence.enabled`:

- `gateway-deployment.yaml` only injects `SICLAW_PERSISTENCE_CLAIM_NAME` when `agentbox.persistence.enabled=true`
- the chart-managed `siclaw-data-pvc.yaml` is only rendered under the same condition
- default values keep `agentbox.persistence.enabled=false`

That means a standard chart deployment with the new intended policy (`enabled=false`, per-agent opt-in later) will not give the runtime any `claimName`. The new spawner then correctly falls back to `emptyDir`, so `persistence: true` from `chat.send` does not actually make that agent durable.

I think the chart needs to separate two concepts:

1. PVC infrastructure availability / claim name, e.g. create or reference a shared RWX PVC and inject `SICLAW_PERSISTENCE_CLAIM_NAME`.
2. The global default policy, i.e. whether callers that omit `persistence` should use the PVC.

With that split, deployments can set up the PVC while keeping the global default off, and per-agent `persistence: true` can actually mount it.

One smaller follow-up to document or handle: this value only affects cold spawns. If an agent already has a running pod, `AgentBoxManager.getOrCreateK8s()` reuses it by `agentId` before `boxConfig.persistence` reaches the pod spec. If toggling per-agent persistence is expected to take effect immediately, the resolved persistence mode probably needs to be stamped on the pod and mismatches should recycle it. If “next pod restart/idle release” is the intended behavior, please make that explicit and add a test around the warm-pod reuse behavior.

Local validation on the PR head passed for me: `git diff --check`, `npx tsc --noEmit`, focused agentbox/task tests, and full `npm test` (`3928 passed / 2 skipped`).

glow-Kev1n:
Thanks for the careful review — both points addressed in 129fbdf.

**1. Chart: PVC availability decoupled from the global default**

You were exactly right — the runtime code separated `claimName` (infra) from `enabled` (policy), but the chart still gated claimName injection, the `user-data` mount, and PVC creation all behind `agentbox.persistence.enabled`, so "global default off + per-agent opt-in" couldn't be expressed by a standard deploy.

I added a helper `siclaw.persistence.pvcAvailable` = `enabled OR (claimName set)`, and split the two concepts:

- **PVC availability (infra)** → injects `SICLAW_PERSISTENCE_CLAIM_NAME` + mounts `user-data` whenever `pvcAvailable`.
- **Global default (policy)** → injects `SICLAW_PERSISTENCE_ENABLED=true` only when `enabled`.
- **Chart-managed PVC** → still created only when `enabled`; a non-empty `claimName` with `enabled=false` references a pre-existing RWX PVC the deployer brings themselves (no hardcoded default — the name differs per deployment).

So you can now set up a PVC while keeping the global default off, and per-agent `persistence:true` actually mounts it.

Backward compatibility verified with `helm template` diffs (ignoring the per-render ephemeral CA):
- default (`enabled=false`, no claimName) → renders identically to before (no PVC, no mount, no persistence env).
- `enabled=true` → renders identically to before.
- new: `enabled=false` + `claimName=<existing-pvc>` → no chart PVC, `SICLAW_PERSISTENCE_CLAIM_NAME` injected, `SICLAW_PERSISTENCE_ENABLED` absent, `user-data` mounted.

**2. Warm-pod reuse — intended behavior (option b), now documented + tested**

Confirmed: per-agent persistence is anchored at cold spawn, not applied immediately. This is by design — K8s can't hot-change a running pod's volume mounts, and `getOrCreate` reuses a running pod by `agentId` before `boxConfig.persistence` reaches the pod spec. Changing the value takes effect on thstart/idle-release). Callers are expected to send a stable value per agent.

I documented this on `getOrCreate` and both warm-reuse branches, and added two tests in `m asserting that flipping `persistence` on a running pod performs zero spawns.

Tests: `tsc --noEmit` clean; full suite 3930 passed / 2 skipped.

src/lib/bootstrap-runtime.ts
```diff
@@ -162,15 +162,21 @@ function createSpawner(
   if (kind === "k8s") {
     const image = opts.k8sImage ?? process.env.SICLAW_AGENTBOX_IMAGE ?? "siclaw-agentbox:latest";
     const namespace = opts.k8sNamespace ?? process.env.SICLAW_K8S_NAMESPACE ?? "default";
-    const claimName =
-      opts.k8sPersistenceClaimName ?? process.env.SICLAW_PERSISTENCE_CLAIM_NAME ?? "siclaw-data";
+    // claimName identifies the shared PVC the deployer actually created — its name
+    // differs per deployment, so there is NO hardcoded default. It is "available"
+    // only when explicitly configured (helm opt or SICLAW_PERSISTENCE_CLAIM_NAME).
+    const claimName = opts.k8sPersistenceClaimName ?? process.env.SICLAW_PERSISTENCE_CLAIM_NAME;
```
chent1996, line 168: **Silent breaking change worth a warning log.** Previously `claimName` defaulted to `"siclaw-data"`, so a deployer who only set `SICLAW_PERSISTENCE_ENABLED=true` via raw env (no Helm, no `SICLAW_PERSISTENCE_CLAIM_NAME`) got persistence wired to a PVC literally named `siclaw-data` — would either bind to an actual PVC of that name or fail loudly at pod-create time. After this PR, the same env-only config silently degrades: `claimName=undefined` → `persistence=undefined` → all agents get `emptyDir` with no signal at startup.

The new Helm chart docs cover this for chart users, but a raw-env upgrade path is now a silent regression. Suggest a one-time warning at this point:

```ts
if (process.env.SICLAW_PERSISTENCE_ENABLED === "true" && !claimName) {
  console.warn("[runtime] SICLAW_PERSISTENCE_ENABLED=true but no SICLAW_PERSISTENCE_CLAIM_NAME configured — persistence will silently fall back to emptyDir (no shared PVC mounted)");
}
```

src/gateway/agentbox/manager.ts
```diff
@@ -105,6 +105,15 @@ export class AgentBoxManager {
    * on warm-pod reuse, so the chat hot path and channel/cron paths pay nothing
    * when the pod already exists.
    */
+  /**
```
chent1996: Nit — this is a second `/** */` block immediately after the existing one (lines 102-107). Most IDE tooltips only show the last block before a declaration, so the original "`spawnEnvResolver` not consulted on warm reuse" note will get visually shadowed. Worth merging the two docstrings into one block.

src/gateway/agentbox/manager.ts
```diff
@@ -118,6 +127,9 @@ export class AgentBoxManager {
 
     const info = await this.spawner.get(name);
     if (info && info.status === "running" && info.endpoint && this.isCertFresh(info)) {
+      // Warm reuse: return the running pod without spawning, so a changed
```
chent1996: Non-blocking — the "callers are expected to send a stable value per agent" contract is documented but not enforced. In a multi-portal deployment (or even a single portal that mis-resolves the value), if two callers reach the same agentId with different `persistence` values, the first cold spawn wins silently and subsequent mismatches are invisible until pod recycle. Consider stashing the cold-spawn value on `existing` (local) / probing the pod spec (k8s) and emitting a one-line warn when a warm-reuse call's `config.persistence` disagrees. Strictly an observability improvement — fine to defer.

chent1996 (commented):
Clean refactor — the per-agent-override-with-fallback-to-global shape is exactly right for a runtime-mechanism PR, and decoupling the Helm chart's PVC availability from the global default policy is a real architectural improvement (no more conflating infra with default). The k8s-spawner persistence matrix test set is the right level of coverage. A couple of non-blocking notes inline; one is a silent breaking change I want to call out before merge.

chent1996 (approved):
LGTM

src/gateway/server.ts
```diff
@@ -211,6 +211,11 @@ export async function startRuntime(opts: StartRuntimeOptions): Promise<RuntimeSe
     const modelRouting = params.modelRouting as PromptOptions["modelRouting"];
     const images = params.images as PromptOptions["images"];
     const files = params.files as PromptOptions["files"];
+    // Per-agent PVC persistence toggle, carried over the chat.send protocol.
+    // siclaw's own portal does not set this (no native per-agent UI); a product
+    // portal (e.g. gpu-cloud) resolves it from its own store and passes it through.
+    // undefined when omitted → spawner falls back to its global config.
+    const persistence = typeof params.persistence === "boolean" ? params.persistence : undefined;
```
jacoblee-io: **Source-of-truth inconsistency (pre-merge).** Here `persistence` is taken from the **client-supplied** `params.persistence`, but `task-coordinator.ts:274` resolves it **server-side by `agentId`** via `binding.persistence`. Two different sources for the same conceptual value.

Since persistence is logically an *agent* property (it shouldn't depend on which entry point triggered the agent), and the pod is one-per-agent anchored at cold spawn, a client-supplied per-request value makes the outcome entry-point-dependent: IM-channel handlers (`lark.ts`/`dingtalk.ts`) and `chat.abort`/`steer`/`clearQueue` pass nothing, so whichever path cold-spawns first wins and a later `persistence:true` is silently ignored on warm reuse.

Suggest sourcing `persistence` from `binding.persistence` (agent-level) on this path too, so all entry points agree.

src/gateway/task-coordinator.ts
```diff
@@ -271,7 +271,7 @@ export class TaskCoordinator {
       // One pod per agent — shared across users who call the agent.
       // Caller/task-owner attribution flows to Upstream via the session registry.
       sessionRegistry.remember(sessionId, userId, agentId);
-      const handle = await this.manager.getOrCreate(agentId);
+      const handle = await this.manager.getOrCreate(agentId, { persistence: binding.persistence });
```
jacoblee-io, line 274: This is the *correct* pattern — `persistence` resolved by `agentId` via the binding, so it's a true agent property. The issue is that it's only applied here and in `chat.send` (and `chat.send` uses a different, client-supplied source). The two live IM-channel entry points (`lark.ts` ~L630, `dingtalk.ts` ~L296) — which also cold-spawn this same per-agent pod — don't pass `binding.persistence` at all. Recommend threading `binding.persistence` through every `getOrCreate` call site so the agent's persistence is honored regardless of entry point.

src/gateway/agent-model-binding.ts
```diff
@@ -31,6 +31,12 @@ export interface ResolvedModelBinding {
   modelRouting?: ModelRoutePolicy;
   /** Agent's custom system prompt template (agents.system_prompt). Null/absent = built-in default. */
   systemPrompt?: string | null;
+  /**
+   * Per-agent session/memory persistence toggle. siclaw core leaves this
+   * undefined (no native per-agent store); a product portal that wants per-agent
+   * persistence resolves it from its own data and carries it over chat.send.
+   */
+  persistence?: boolean;
```
jacoblee-io, line 39: If this `binding.persistence` becomes the single agent-level source of truth (recommended — see server.ts/task-coordinator.ts comments), then all `getOrCreate` callers (web chat, channels, task-coordinator) should read it, and the per-request `chat.send` param can be dropped or demoted to an explicit override. That keeps the "core never stores the flag" constraint (downstream populates this field in its `resolveAgentModelBinding` extension) while making the behavior entry-point-independent.

jacoblee-io (commented):
Nice, well-tested mechanism — the K8sSpawner volume selection, the Helm `pvcAvailable` decoupling, and the warm-pod anchoring are all sound, and as a pure mechanism it is behavior-preserving for upstream siclaw (core never sets the flag → every path falls to the global config → identical to today).

But there is one **pre-merge logic issue** that undermines the headline capability, plus two minor notes. Details inline; summary here:

### Blocking: `persistence` has two different sources of truth across the two wired paths

The value should be an **agent property** (the agent decides whether it persists — independent of which entry point triggers it). The PR doesn't implement it that way consistently:

| Path | Where `persistence` comes from |
|---|---|
| `chat.send` (`server.ts`) | `params.persistence` — a **client-supplied per-request** value |
| `task-coordinator.ts` | `binding.persistence` — **server-resolved by `agentId`** via `resolveAgentModelBinding` |

Because a pod is one-per-`agentId` and the persistence mode is anchored at cold spawn + warm-reused thereafter, the volume mode is decided by *whichever entry point first cold-spawns the pod*. The IM-channel live paths (`lark.ts` ~L630, `dingtalk.ts` ~L296) and `chat.abort`/`chat.steer`/`chat.clearQueue` call `getOrCreate(agentId)` with **no** persistence at all.

Consequence on the PR's headline deploy (global default OFF + per-agent opt-in): if a durable agent receives a Lark/DingTalk message before any `persistence:true` `chat.send`, its pod cold-spawns as `emptyDir`, and the later `persistence:true` is **silently ignored** (warm reuse, no respawn). The agent appears non-persistent despite the portal opting in — no error.

### Suggested fix

Make `persistence` a single **agent-level source of truth**: resolve it server-side by `agentId` (the `binding.persistence` path, as `task-coordinator` already does) and have **every** `getOrCreate` call read it. Then the entry point is irrelevant — every path resolves the same value for the same agent. The `chat.send` client param should either be dropped or treated only as an explicit override, not as the primary source. This also keeps the "core does not store the flag" constraint intact (downstream fills `binding.persistence` in its `resolveAgentModelBinding` extension).

### Minor (non-blocking)
- `bootstrap-runtime.ts`: dropping the hardcoded `"siclaw-data"` claimName default is a behavior change. Fine via Helm (chart always injects the claim name when `pvcAvailable`), but a hand-rolled deploy that set only `SICLAW_PERSISTENCE_ENABLED=true` and relied on the old default now silently degrades to `emptyDir`. Worth a release note.
- `manager.ts` docstring says persistence is anchored at *first* spawn; a cert-stale recycle is also a cold respawn that re-reads the triggering caller's value, so it's really "anchored at each cold spawn, using that caller's value." Minor wording.

glow-Kev1n:
Thanks — all three addressed in `f8c5da8`.

**Blocking — single agent-level source of truth.** Done as suggested: `persistence` is now resolved server-side by `agentId`, and **every** `getOrCreate` honours it, so the entry point is irrelevant.

- Added `AgentBoxManager.setPersistenceResolver(agentId => …)`, mirroring the existing `setSpawnEnvResolver` (idle_timeout) pattern: consulted **only on a cold spawn**, never on warm reuse, so the hot path pays no extra RPC.
- Registered in `startRuntime` to resolve `binding.persistence` via `resolveAgentModelBinding` (the `config.getModelBinding` path the cron path already used).
- **Dropped** the per-request `params.persistence` from `chat.send` — the agentId-keyed resolver is the sole source. `task-coordinator` keeps passing `binding.persistence` explicitly (an explicit value wins in `resolvePersistence`, avoiding a second RPC).
- The channel / `chat.abort` / `chat.steer` / `chat.clearQueue` call sites are **unchanged** — they get the agent's mode for free on cold spawn. So a durable agent first reached via Lark/DingTalk now cold-spawns with the PVC, not emptyDir.

**Minor — claimName release note.** Added a one-time startup warning in `bootstrap-runtime.ts` when `SICLAW_PERSISTENCE_ENABLED=true` but no claim name is configured (the dropped `"siclaw-data"` default), plus a release note in the PR description. Helm users are unaffected (the chart injects the claim name whenever a PVC is available).

**Minor — docstring wording.** Reworded to "resolved on each cold spawn (including a cert-stale recreate) from the agent-level resolver, entry-point independent," and merged the two stacked docstrings.

On RPC placement: I kept `persistence` on `binding` / `config.getModelBinding` (as you suggested, matching `task-coordinator`) rather than moving it next to `idle_timeout_sec` on `config.getAgent`. That costs a second cold-spawn RPC but avoids reworking the cron path and the downstream portal's existing handler — happy to consolidate if you'd prefer.

Tests: `tsc --noEmit` clean; full suite **3935 passed / 2 skipped**. New `manager.test.ts` cases cover resolver-on-cold-spawn (K8s + Local), explicit-config-wins-over-resolver, and resolver-not-consulted-on-warm-reuse.

jacoblee-io (approved):
## Approved ✅ — blocking concern resolved

The new commit `f8c5da8` ("resolve per-agent persistence by agentId across all entry points") fixes the pre-merge issue exactly the right way: **persistence is now a single agent-level source of truth, resolved in the common chokepoint — no per-entry-point/per-channel logic.**

Verified the implementation:

- **Single source of truth, agent-level**: `setPersistenceResolver` registered in `startRuntime` resolves `binding.persistence` by `agentId` via `resolveAgentModelBinding`. Mirrors the existing `setSpawnEnvResolver` (idle_timeout) pattern — consistent with the codebase.
- **Resolved in the common funnel**: `getOrCreate` consults `resolvePersistence(agentId, config?.persistence)` on **both** cold-spawn paths (K8s + Local), only on cold spawn, never on warm reuse. All 9 `getOrCreate` call sites (chat, Lark, DingTalk, cron, abort/steer/clearQueue) now land the same mode for the same agent with **zero per-call-site changes** — so an agent that disables PVC takes effect regardless of entry point (web/api/a2a/channels).
- **Dual source eliminated**: `chat.send` no longer reads the client-supplied `params.persistence`; it calls plain `getOrCreate(agentId)`. Explicit config still wins (task-coordinator keeps passing `binding.persistence`, saving a redundant RPC).
- **Upstream behavior unchanged**: native `getModelBinding` returns no persistence → resolver yields `undefined` → global fallback. Core still stores nothing.

The two earlier minor notes are also addressed: `bootstrap-runtime` now warns at startup when `SICLAW_PERSISTENCE_ENABLED=true` but no claim name is configured, and the `getOrCreate` docstring was reworded to "resolved on each cold spawn (incl. cert-stale recreate), entry-point independent."

Test coverage added in `manager.test.ts` (resolver-on-cold-spawn K8s+Local, explicit config wins, resolver not consulted on warm reuse). LGTM.

PR #362, opened 2026-06-25
feat(dingtalk): reply with agent-produced images via robot OpenAPI

silence8013:
## Summary

DingTalk replies were text/markdown only, so agent-produced image artifacts (charts, screenshots, etc.) were silently dropped. This adds image delivery for the DingTalk channel, mirroring the existing Lark pattern (`replyVisualImages`): images are sent as **separate robot messages**, not inlined into markdown.

Only images that the agent produces as **structured image content blocks** (collected via `collectChannelResponse({ includeImages: true })`) are delivered. Arbitrary agent-authored markdown image URLs are still stripped by `sanitizeMarkdownForDingTalk`, so the prompt-injection / data-exfiltration surface is unchanged.

## 上线效果 / Screenshots

image

## What changed

- **`dingtalk-api.ts` (new)** — DingTalk OpenAPI helpers:
  - `getAccessToken` — clientId-scoped token cache, refreshes 1 min before expiry.
  - `uploadImageMedia` — uploads a PNG/JPEG/WebP buffer to DingTalk media (`/media/upload`) → `media_id`. Enforces 20MB / supported-type limits.
- **`dingtalk-image.ts` (new)** — robot-message image delivery:
  - `sendImageMessage` — `sampleImageMsg` (`photoURL=media_id`) via `oToMessages/batchSend` (1:1, `userIds`) or `groupMessages/send` (group, `openConversationId`).
  - `deliverImages` — uploads + sends each collected image; best-effort, returns delivered count.
- **`dingtalk.ts`** — capture `senderStaffId` from the inbound frame; thread channel `config` into the handler; switch to `collectChannelResponse({ includeImages: true, persist })` to collect text + images while preserving the existing audit persistence; deliver images as separate messages **after** the text reply.
- Unit tests for both new modules.

## How it works

```
tool returns image block → collectChannelResponse collects RenderedReplyImage[]
   → uploadImageMedia (media_id) → sendImageMessage (robot OpenAPI) → DingTalk
```

Text reply goes out first (over `sessionWebhook`), images follow as separate robot messages.

## Security

- All OpenAPI calls run in the trusted Runtime against pinned DingTalk hosts (`api.dingtalk.com`, `oapi.dingtalk.com`); token/secret are redacted in logs.
- Image bytes come only from the agent's structured image blocks, never from agent-authored URLs.
- Image delivery is best-effort and post-text: a failed upload/send (e.g. missing permission) never breaks the primary reply — it degrades to text-only.

## Testing

- `npx tsc --noEmit` — clean.
- `vitest` — 68 passing (10 `dingtalk-api`, 9 `dingtalk-image`, 49 existing `dingtalk` handler suite incl. audit persistence).
- Real-device build pushed and exercised: image replies were received in DingTalk.

## Requires (DingTalk console)

- **Media file upload** + **robot message send** permissions (without them token resolves but send is rejected → degrades to text-only).

## Known limitations / follow-ups

- `senderStaffId` (1:1) and group `conversationId`-as-`openConversationId` are validated against the live frame shape; covered by the degradation path if a field is absent.
- AI streaming card (image blocks inside a live card) remains a separate Phase 2 (reserved `DingTalkStreamingCard` seam).
- Non-multimodal model routing (`api: openai-completions`) cannot feed tool-result images back into the model context, so the agent may not "see" the image it just sent — orthogonal to delivery, tracked separately.

### DIFF
--- a/src/gateway/channels/dingtalk-api.ts
+++ b/src/gateway/channels/dingtalk-api.ts
@@ -1,0 +1,173 @@
+/**
+ * DingTalk OpenAPI helpers shared by the channel handler.
+ *
+ * Phase 1 scope: an access-token cache plus image media upload, which together
+ * let the handler turn an agent-produced image (a binary artifact, NOT an
+ * agent-authored URL) into a DingTalk `media_id`. That `media_id` is then
+ * embedded into the `sessionWebhook` markdown reply (`![alt](media_id)`) the
+ * same way the upstream `@soimy/dingtalk` plugin's `sendBySession` does.
+ *
+ * Security note: every call here runs in the trusted Runtime process (inside
+ * the cluster), never in the agent sandbox. We only ever talk to the pinned
+ * DingTalk OpenAPI hosts (`api.dingtalk.com` for the token, `oapi.dingtalk.com`
+ * for media upload). The bytes we upload come from the agent's structured image
+ * blocks collected by `collectChannelResponse`, so this path does not reopen
+ * the prompt-injection / data-exfiltration hole that markdown image-stripping
+ * closes — arbitrary agent-authored image URLs are still stripped.
+ */
+
+import type { DingTalkChannelConfig } from "./dingtalk.js";
+
+/** DingTalk OpenAPI endpoints (pinned — do not accept these from frame data). */
+const TOKEN_URL = "https://api.dingtalk.com/v1.0/oauth2/accessToken";
+const MEDIA_UPLOAD_URL = "https://oapi.dingtalk.com/media/upload";
+
+/** DingTalk hard limit for image media uploads. */
+const IMAGE_UPLOAD_MAX_BYTES = 20 * 1024 * 1024;
+
+/** Network timeout for OpenAPI calls (token + upload). */
+const OPENAPI_TIMEOUT_MS = 10_000;
+
+/** Supported outbound image MIME types and the filename extension we tag them with. */
+const IMAGE_MIME_EXTENSION: Record<string, string> = {
+  "image/png": "png",
+  "image/jpeg": "jpg",
+  "image/webp": "webp",
+};
+
+interface TokenCacheEntry {
+  accessToken: string;
+  /** Epoch ms at which the cached token should be considered stale. */
+  expiresAt: number;
+}
+
+/**
+ * Access-token cache keyed by clientId so multiple DingTalk channels (each a
+ * distinct app) keep independent tokens. In-memory only: a Runtime restart
+ * simply re-fetches on first use.
+ */
+const tokenCache = new Map<string, TokenCacheEntry>();
+
+/** Test-only helper to clear the token cache between cases. */
+export function resetDingTalkTokenCacheForTest(): void {
+  tokenCache.clear();
+}
+
+/**
+ * Redact secrets from any string before it reaches a log collector. The token
+ * endpoint takes the AppSecret in the request body and the upload endpoint puts
+ * the access token in the query string, so a naive error dump could leak both.
+ */
+export function redactSecrets(input: string): string {
+  return input
+    .replace(/(appSecret|appsecret|access_token|accessToken)["=:\s]+[^"&\s,}]+/gi, "$1=***");
+}
+
+export async function fetchWithTimeout(
+  url: string,
+  init: RequestInit,
+  timeoutMs = OPENAPI_TIMEOUT_MS,
+): Promise<Response> {
+  const controller = new AbortController();
+  const timer = setTimeout(() => controller.abort(), timeoutMs);
+  try {
+    return await fetch(url, { ...init, signal: controller.signal });
+  } finally {
+    clearTimeout(timer);
+  }
+}
+
+/**
+ * Fetch (or return a cached) DingTalk access token for the given channel app.
+ * Refreshes one minute before expiry to avoid near-boundary failures. Returns
+ * `null` on any failure so callers degrade gracefully (text/markdown reply
+ * without the image) rather than throwing into the message handler.
+ */
+export async function getAccessToken(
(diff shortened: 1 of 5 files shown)


src/gateway/channels/dingtalk.ts
```diff
@@ -359,6 +372,27 @@ export async function handleDingTalkMessage(
     ? buildTextMessage(finalBody)
     : (resultText ? buildMarkdownMessage(finalBody) : buildTextMessage(finalBody));
```
jacoblee-io, line 373: **UX (medium): chart/mermaid source fences are no longer stripped — you now get both the raw source block *and* the rendered image.**

The Lark path strips source fences when images are present:
```ts
// lark.ts
stripVisualBlocks(finalBody, { stripSourceBlocks: replyImages.length > 0 })
```
`sanitizeMarkdownForDingTalk` only neutralizes `` syntax; it preserves code fences. So a ```chart / ```mermaid / ```visual-card block will show as raw source in the text bubble while its rendered image arrives as a separate message. Suggest running `finalBody` through `stripVisualBlocks(finalBody, { stripSourceBlocks: replyImages.length > 0 })` before `buildMarkdownMessage`, mirroring Lark (this also gives you the visual-only fallback notice for the image-only case).

src/gateway/channels/dingtalk.ts
```diff
@@ -359,6 +372,27 @@ export async function handleDingTalkMessage(
...
   await replyToDingTalk(sessionWebhook, body);
+
+  // Deliver any agent-produced images as separate robot messages (mirrors
+  // Lark's replyVisualImages). Best-effort and post-text: image delivery needs
+  // an access token + robot-send permission (config), and a failed image must
+  // never break the primary reply. Skipped when the agent errored, produced no
+  // images, or the channel config was not threaded through (older callers).
+  if (!agentError && config && replyImages.length > 0) {
```
jacoblee-io, line 381: **Security note (non-blocking, inherent): image artifacts bypass output sanitization.**

The sanitization pipeline (docs/design/sanitization.md) redacts secrets in *text* output. Binary images are uploaded as-is, so a chart/screenshot that visually contains sensitive data (e.g. a terminal screenshot with a token) would be posted to the chat un-redacted. This is not introduced by this PR — Lark already delivers images the same way, and tools here are read-only/whitelisted, so residual risk is bounded — but it's worth accepting *deliberately* (a line in sanitization.md) rather than assuming the sanitizer covers it.

src/gateway/channels/dingtalk-api.ts
```diff
@@ -0,0 +1,173 @@
...
+const IMAGE_UPLOAD_MAX_BYTES = 20 * 1024 * 1024;
+
+/** Network timeout for OpenAPI calls (token + upload). */
+const OPENAPI_TIMEOUT_MS = 10_000;
+
+/** Supported outbound image MIME types and the filename extension we tag them with. */
+const IMAGE_MIME_EXTENSION: Record<string, string> = {
+  "image/png": "png",
+  "image/jpeg": "jpg",
+  "image/webp": "webp",
```
jacoblee-io, line 35: **Minor: `image/webp` is likely rejected by the media server.** DingTalk's `/media/upload?type=image` historically supports jpg/png/gif/bmp, not webp. Behavior is safe (server returns `errcode != 0` → `null` → image skipped, no crash), but the accepted-type map overstates what actually delivers. Since the chart pipeline emits PNG this is rare; consider dropping webp or adding a comment that it may not upload.

src/gateway/channels/dingtalk-image.ts
```diff
@@ -0,0 +1,118 @@
...
+ * image is independent, and any failure is logged without aborting the rest or
+ * the primary text reply. Returns the count successfully delivered.
+ */
+export async function deliverImages(
+  config: DingTalkChannelConfig,
+  target: DingTalkImageTarget,
+  images: RenderedReplyImage[],
+): Promise<number> {
+  let delivered = 0;
+  for (const { image, mimeType } of images) {
```
jacoblee-io, line 112: **Minor (DoS surface): no cap on image count.** Each image is bounded (20MB + type check) and delivery is best-effort, but the loop has no upper bound on the number of images — a turn that produces many image blocks means many sequential upload+send round-trips (each with a 10s timeout) and could hit DingTalk rate limits. Consider a `MAX_REPLY_IMAGES` cap (Lark has the same gap, so optional).

jacoblee-io (approved):
## Approved ✅

Clean, well-scoped, well-documented. Faithfully mirrors the Lark pattern (`collectChannelResponse({includeImages:true})` → separate image messages), preserves the existing trust boundary, and degrades gracefully on every failure path. Good unit coverage on the two new modules.

**Security review — no blocking issues.** I verified the two highest-risk paths:
- **No SSRF / no agent-controlled fetch.** Image bytes come only from structured base64 image blocks; `visual-image.ts` accepts `data:` URLs only and rejects remote URLs, so there is no "agent writes a URL → Runtime fetches it" path. Agent-authored markdown image URLs are still stripped by `sanitizeMarkdownForDingTalk`. Injection/exfil surface is unchanged. ✔
- **No addressing amplification.** `senderStaffId` (1:1) and `openConversationId` (group) come from the signed inbound frame, not message content, and only ever address the originating sender/group. `resolveBinding` gates authorization *before* delivery. ✔

Few non-blocking notes inline. The one worth a conscious decision: binary images bypass the text output-sanitization pipeline (a chart/screenshot could visually contain secrets and would be posted un-redacted). This is inherent and shared with Lark — fine to accept, just accept it deliberately.

Real-device verification + the two cross-API assumptions (v1.0 token used against the legacy `oapi` upload endpoint; group `conversationId` == `openConversationId`) are confirmed by the author; both degrade to text-only if wrong.

PR #363, opened 2026-06-25
fix(chat): restore context-usage meter on session open/refresh

jacoblee-io:
## What

The chat composer's context-usage meter was populated **only** from a live `agent_end` event, so opening or refreshing a session (or reopening a cold one after idle self-destruct) showed nothing until the next turn. `chat_messages` never persisted token usage and the frontend never restored it from history.

This persists the `agent_end` context-usage snapshot onto the turn's **last assistant message metadata** (reusing the existing `chat.*` RPC + `metadata` field — **no schema change, no new RPC, no sicore Go change**), and restores it on the frontend when a session loads.

## Why

Backend + provider `usage` were verified healthy on siclaw-inner (the meter shows correctly *after* a live turn, `Context 24/100k`); the gap was purely persist/restore. This is **not** a usage or compaction issue — earlier "missing usage / estimate-fallback" theories were disproven and reverted.

## How

- **`src/gateway/sse-consumer.ts`** — capture `contextUsage` at `agent_end`. Ordering matters: every turn is routed, and `agent_end` fires **before** the commit (`model_route_success`) that runs the deferred assistant persist — so the row isn't written yet at `agent_end`. We stash the snapshot (`capturedContextUsage`) and the deferred `persistAssistant` folds it into the row's metadata; the immediate/non-routed path patches via `updateMessage` (re-sending content so the handler's `content ?? ""` doesn't wipe it). Best-effort — a persistence failure never breaks the SSE stream.
- **`portal-web/src/hooks/usePilotChat.ts`** — `latestContextUsageFromItems()` restores the meter from the most recent persisted snapshot on load. Guarded to (a) only set when a snapshot is found (never blank a cached/live value) and (b) require `percent > 0` to match the composer's render condition.

## Tests

- `sse-consumer.test.ts`: snapshot folded into the deferred row (routed order) + patched (immediate order) + no-op without usage. 83 pass.
- `tsc` clean (backend + portal); portal hook tests pass.

## Verified on siclaw-inner (Playwright, real session)

After a turn completes → **full page reload, no new message** → meter renders immediately: `↑24.3k ↓138 24% (Context 24,316/100,000)`. Restored from DB.

## Out of scope

- The SDK usage-gated mid-turn compaction trigger (separate; backend usage is healthy).
- sicore dashboards / Go apiserver.
- Any `chat_sessions` schema change (deliberate — keeps the change deployment-agnostic).

### DIFF
--- a/portal-web/src/hooks/usePilotChat.ts
+++ b/portal-web/src/hooks/usePilotChat.ts
@@ -412,6 +412,25 @@ export function buildPilotMessages(items: ChatMessage[]): PilotMessage[] {
   return annotateSubagentCompletions(annotateExecJobCompletions(annotateDelegationSynthesis(items.map(toPilotMessage))))
 }
 
+/**
+ * Recover the most recent persisted context-usage snapshot from loaded history.
+ * The Runtime stores it on the latest assistant row's `metadata.context_usage`
+ * (see sse-consumer.ts agent_end handling), so the context meter can render on
+ * session open/refresh instead of staying blank until the next live turn.
+ * Items are chronological (oldest first), so scan from the end.
+ */
+function latestContextUsageFromItems(items: ChatMessage[]): ContextUsage | null {
+  for (let i = items.length - 1; i >= 0; i--) {
+    const md = normalizeMetadata(items[i]?.metadata)
+    const cu = md?.context_usage as ContextUsage | undefined
+    // Require percent > 0 to match the meter's render guard (InputArea), so a
+    // zero-percent snapshot (e.g. usage-less turn) doesn't shadow an earlier
+    // real one and leave the meter blank.
+    if (cu && typeof cu === "object" && typeof cu.percent === "number" && cu.percent > 0) return cu
+  }
+  return null
+}
+
 export function toPilotMessage(m: ChatMessage): PilotMessage {
   const toolArgs = m.tool_input ? tryParseJson(m.tool_input) : undefined
   const metadata = normalizeMetadata(m.metadata)
@@ -829,6 +848,13 @@ export function usePilotChat({ agentId, sessionId }: UsePilotChatOptions): UsePi
         const { items, pilotMsgs } = await fetchSessionPage1(agentId, sessionId!)
         if (cancelled) return
         setMessages(pilotMsgs)
+        // Restore the context meter from persisted history so it shows on open/
+        // refresh — without it the meter stays blank until the next live
+        // agent_end. Only set when a snapshot is found: never blank an existing
+        // value (a cache-hit restore at line ~833, or a live agent_end that
+        // landed while this fetch was in flight).
+        const restoredUsage = latestContextUsageFromItems(items)
+        if (restoredUsage) setContextUsage(restoredUsage)
         setHasMore(items.length >= PAGE_SIZE)
         const hasRunning = hasRunningPersistedMessages(pilotMsgs)
 
--- a/src/gateway/sse-consumer.test.ts
+++ b/src/gateway/sse-consumer.test.ts
@@ -100,6 +100,32 @@ describe("consumeAgentSse — assistant message flow", () => {
     expect(assistantRow.sessionId).toBe("sid");
   });
 
+  it("merges the agent_end context-usage snapshot onto the last assistant row's metadata", async () => {
+    // Lets the frontend restore the context meter on session reopen/refresh.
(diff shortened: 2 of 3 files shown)


src/gateway/sse-consumer.ts
```diff
@@ -402,6 +417,35 @@ export async function consumeAgentSse(opts: ConsumeAgentSseOptions): Promise<Sse
...
+    //   • immediate persist (non-routed): the row already exists, so patch it now.
+    // Best-effort: a persistence failure must never break the SSE stream.
+    if (eventType === "agent_end" && persist) {
+      const contextUsage = (evt as Record<string, unknown>).contextUsage;
+      if (contextUsage && typeof contextUsage === "object") {
+        capturedContextUsage = contextUsage as Record<string, unknown>;
+        if (lastAssistantDbMessageId) {
+          try {
+            await updateMessage({
+              messageId: lastAssistantDbMessageId,
```
chent1996, line 437: Non-blocking pre-existing wart you're working around: re-sending `content` here is purely to defeat `chat.updateMessage`'s `params.content ?? ""` (adapter.ts:2390), which would otherwise wipe the row. The adapter already uses `COALESCE(?, delegation_id)` for the optional `delegation_id` column right next to it — making the `content` column match that pattern (`COALESCE(?, content)`) would let callers patch metadata-only without having to remember to re-thread content. Not for this PR, just flagging since this comment makes the workaround visible.

src/gateway/sse-consumer.test.ts
```diff
@@ -348,6 +374,27 @@ describe("consumeAgentSse — routed turn commit gating", () => {
...
+      { type: "message_end", message: { role: "assistant", content: [{ type: "text", text: "hi" }], stopReason: "stop" } },
+      { type: "agent_end", contextUsage: cu },
+      { type: "model_route_success", attempt: 1, candidateKey: "openai/gpt-4", provider: "openai", modelId: "gpt-4", isFallback: false, primaryCandidateKey: "openai/gpt-4" },
+    ];
+    await consumeAgentSse({ client: mkClient(events), sessionId: "sid", userId: "u", persistMessages: true });
+    const row = appendCalls.find((r) => r.role === "assistant" && r.content === "hi");
+    expect(row).toBeDefined();
+    expect(row.metadata.context_usage).toEqual(cu);
+    // No post-hoc patch needed on the routed path.
+    expect(updateCalls.find((u) => u.metadata?.context_usage)).toBeUndefined();
```
chent1996, line 395: Coverage gap, non-blocking: the routed-deferred test covers a single-assistant turn. A multi-assistant turn (assistant → tool → assistant → agent_end) takes the `updateMessage` patch path on the LAST assistant via `lastAssistantDbMessageId` (overwritten on each `message_end`). The contract is right by inspection, but worth a test so a future refactor doesn't silently regress to stamping usage on the FIRST assistant of the turn.

portal-web/src/hooks/usePilotChat.ts
```diff
@@ -412,6 +412,25 @@ export function buildPilotMessages(items: ChatMessage[]): PilotMessage[] {
...
+ * The Runtime stores it on the latest assistant row's `metadata.context_usage`
+ * (see sse-consumer.ts agent_end handling), so the context meter can render on
+ * session open/refresh instead of staying blank until the next live turn.
+ * Items are chronological (oldest first), so scan from the end.
+ */
+function latestContextUsageFromItems(items: ChatMessage[]): ContextUsage | null {
+  for (let i = items.length - 1; i >= 0; i--) {
+    const md = normalizeMetadata(items[i]?.metadata)
+    const cu = md?.context_usage as ContextUsage | undefined
+    // Require percent > 0 to match the meter's render guard (InputArea), so a
```
chent1996, line 426: Minor — the `percent > 0` guard is the right call to match the meter's render condition, but worth noting it also silently swallows the (admittedly edge-case) case where the very first turn legitimately closed near-empty (e.g. an aborted turn that recorded `percent=0`). For practical purposes fine; just a known asymmetry between persist (writes everything) and restore (filters).

chent1996 (approved):
Clean, minimal fix — exactly the right altitude. Reusing `chat_messages.metadata` (zero schema/RPC change) is the deployment-agnostic move; the two-order handling (deferred-fold on routed turns vs post-hoc patch on the no-route edge) reads correctly against the PR #344 commit-gating contract. Tests pin both orders + the no-usage no-op.

A couple of small non-blocking notes inline.

PR #364, opened 2026-06-26
Pass agent model binding to Lark channel prompts

LikiosSedo:
## Summary

- Resolve each Lark channel message's agent model binding before prompting AgentBox.
- Forward `modelProvider`, `modelId`, `modelConfig`, `modelRouting`, and the agent system prompt into the channel prompt payload.
- Reuse the resolved model config for channel audit redaction.
- Add a Lark channel regression test that verifies the model binding reaches `AgentBoxClient.prompt`.

## Root Cause

Web chat already forwards the Sicore-resolved model binding to Runtime, so AgentBox switches to the agent's configured model. The Lark channel path only sent `text`, `agentId`, `mode`, and `sessionId`, so AgentBox fell back to its default model. In the test environment that default model was Kimi with an invalid API key, producing `401 authorization api key not valid` and an empty Feishu reply card.

## Validation

- `npm test -- src/gateway/channels/lark.test.ts`
- `npm run build`

### DIFF
--- a/src/gateway/channels/lark.test.ts
+++ b/src/gateway/channels/lark.test.ts
@@ -48,6 +48,12 @@ vi.mock("../channel-manager.js", () => ({
     v !== null && typeof v === "object" && (v as { walled?: unknown }).walled === true,
 }));
 
+const resolveAgentModelBindingMock = vi.fn();
+
+vi.mock("../agent-model-binding.js", () => ({
+  resolveAgentModelBinding: (...args: unknown[]) => resolveAgentModelBindingMock(...args),
+}));
+
 const ensureChatSessionMock = vi.fn();
 const appendMessageMock = vi.fn();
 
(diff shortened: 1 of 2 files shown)


jacoblee-io (approved):
LGTM — approving. The Lark fix is correct and mirrors the established pattern; one follow-up below (non-blocking).

**Verified correct:**
- Field names match `PromptOptions`: `systemPromptTemplate` (not `systemPrompt`), `modelConfig`, `modelRouting: ModelRoutePolicy`, `modelProvider`, `modelId`.
- Mirrors the existing `task-coordinator.ts:268-286` pattern exactly (`resolveAgentModelBinding(agentId, frontendClient)` + `modelRouting` + `systemPromptTemplate: binding.systemPrompt`).
- Graceful fallback: `resolveAgentModelBinding` swallows RPC errors → `null` → pre-PR default behavior; `frontendClient` is reliably present on this path (binding resolution already uses `frontendClient!`).
- `persist.modelConfig` is genuinely consumed — `collectChannelResponse` builds the redactor via `buildRedactionConfigForModelConfig(persist.modelConfig)` (`lark.ts:~901`), so channel audit redaction now uses the agent's real model config. Real improvement, not a no-op.
- Regression test covers it; `beforeEach` defaults the mock to `null` so other cases keep the prior behavior.

**Follow-up (should not block this PR, but needs a fast follow):** DingTalk has the identical gap. `dingtalk.ts:303` resolves only `resolveAgentSystemPrompt` (system prompt), and its `promptOpts` (`:325`) omits `modelProvider/modelId/modelConfig/modelRouting` — so a DingTalk-bound agent with a custom model still falls back to AgentBox's default model and will hit the same `401 / empty reply` symptom this PR fixes for Lark. Cleanest fix is the same shape: swap `resolveAgentSystemPrompt` → `resolveAgentModelBinding` and forward all fields.

**Minor (optional):** both channels now inline "resolve binding + build promptOpts"; a shared `buildChannelPromptOptions(agentId, frontendClient, …)` helper would prevent exactly this kind of lark/dingtalk drift.

PR #366, opened 2026-06-27
feat(lark): accept inbound images; resolve text image URLs at the gateway boundary

chent1996:
## What

The Feishu/Lark channel could only **send** images out — images a user posted were dropped at the non-text gate and never reached the agent. This adds an inbound pipeline that fills `PromptOptions.images`, with **text-image-URL resolution unified at the `AgentBoxClient.prompt` boundary** so every front-end that reaches AgentBox via this client (Feishu / DingTalk / Portal Web chat / a2a / cron) shares one implementation.

## Changes

- **`extractInbound()`** parses `text` / `image` / `post` messages into `{ text, imageRefs }`; `post` rich-text is walked for embedded `img` nodes (and accepts the locale-nested receive shape, surfacing title + hyperlink href). The handler gate accepts `text/image/post` and only drops a turn with **neither** text nor image.
- **Native Lark images** (`channels/inbound-image.ts`): downloaded via the receive-side `im.messageResource.get` (`getReadableStream`, aggregated in memory — never the upload-side `im.images` endpoint). **Vision-gated** like the URL path — skipped for non-vision models (the `[image]` placeholder still records the image).
- **Text image URLs** are resolved generically in the new **`agentbox/image-url-ingest.ts`**, applied once at `AgentBoxClient.prompt` (NOT in the channel), covering Feishu / DingTalk / Web chat / a2a / cron:
  - **Vision-gated** via `modelOptionsSupportImageInput`: resolve only when the prompt's model/route declares image input. A non-vision model leaves the URL as **plain text** (the model replies it can't open it) — no fail-closed turn, no OCR fallback.
  - **Fail-closed host-allowlist SSRF guard** (`SICLAW_IMAGE_URL_ALLOWLIST`): https-only, per-redirect re-validation, streamed size cap, `Content-Type` + magic-byte checks. URLs are fetched **in parallel under one overall timeout** (`SICLAW_IMAGE_URL_TOTAL_TIMEOUT_MS`, default 8s) so a slow allowlisted host can't stall the turn. The outbound fetch + allowlist stay in the **Gateway process** (the AgentBox pod is network-isolated and must not fetch arbitrary user URLs).
  - Per-image failures are skipped with a warning; total capped at the AgentBox media budget.
- An **image-only turn** gets a placeholder caption so the session title, persisted user row, and prompt stay non-empty (audit transcript intact).

## Notes

- **Env renamed (no compat shim):** `SICLAW_LARK_IMAGE_URL_ALLOWLIST` → `SICLAW_IMAGE_URL_ALLOWLIST` (plus `_ALLOW_HTTP` / `_FETCH_TIMEOUT_MS`). **Deployment values that set the old names must be updated**, otherwise text-URL resolution stays fail-closed.
- Group `@`-mention gating is **unchanged**: pure drag-in images apply to p2p personal bots; groups use `post(@bot+img)` or text image URLs.

## Tests

- `src/gateway/agentbox/image-url-ingest.test.ts` (new) — URL extraction, SSRF allowlist, redirect re-validation, size caps, budget merge, parallel + overall-timeout bound
- `src/core/model-routing.test.ts` — `modelOptionsSupportImageInput` (single model / routed / disabled)
- `src/gateway/agentbox/client.test.ts` — vision-gated resolution at `prompt()`
- `src/gateway/channels/{inbound-image,lark}.test.ts` — native download (vision-gated), post receive shapes (nested/flat/href), channel hands URL text to the unified layer

### DIFF
--- a/src/gateway/agentbox/client.ts
+++ b/src/gateway/agentbox/client.ts
@@ -8,7 +8,7 @@
 import https from "node:https";
 import { GATEWAY_SYNC_DESCRIPTORS, type GatewaySyncType } from "../../shared/gateway-sync.js";
 import { modelOptionsSupportImageInput, type ModelRoutePolicy } from "../../core/model-routing.js";
-import { enrichImagesFromText } from "./image-url-ingest.js";
+import { enrichImagesFromText, redactImageUrlsInText } from "./image-url-ingest.js";
 
 export interface AgentBoxTlsOptions {
   cert: string;
@@ -146,11 +146,21 @@ export class AgentBoxClient {
    * process and never reach here, so this never scans system text. The fetch +
    * SSRF allowlist stay in the Gateway process (the AgentBox pod is
    * network-isolated and must not fetch arbitrary user URLs).
+   *
+   * The text forwarded to AgentBox (model context + session history) has any
+   * signed-URL credentials stripped — the fetch above already used the full URL,
+   * so resolution is unaffected, but `Signature`/`AccessKeyId` don't get
+   * persisted/sent in plaintext.
    */
   async #withResolvedImageUrls(opts: PromptOptions): Promise<PromptOptions> {
-    if (!modelOptionsSupportImageInput(opts)) return opts;
-    const images = await enrichImagesFromText(opts.text, opts.images ?? []);
-    return images.length ? { ...opts, images } : opts;
+    const images = modelOptionsSupportImageInput(opts)
+      ? await enrichImagesFromText(opts.text, opts.images ?? [])
+      : opts.images;
+    const text = redactImageUrlsInText(opts.text);
+    if (text === opts.text && images === opts.images) return opts;
+    const next: PromptOptions = { ...opts, text };
+    if (images && images.length) next.images = images;
+    return next;
   }
 
   /**
--- a/src/gateway/agentbox/client.ts
+++ b/src/gateway/agentbox/client.ts
@@ -7,7 +7,8 @@
 
 import https from "node:https";
 import { GATEWAY_SYNC_DESCRIPTORS, type GatewaySyncType } from "../../shared/gateway-sync.js";
-import type { ModelRoutePolicy } from "../../core/model-routing.js";
+import { modelOptionsSupportImageInput, type ModelRoutePolicy } from "../../core/model-routing.js";
+import { enrichImagesFromText } from "./image-url-ingest.js";
 
 export interface AgentBoxTlsOptions {
   cert: string;
@@ -133,14 +134,34 @@ export class AgentBoxClient {
   }
 
   /**
+   * Resolve image URLs in the prompt text into vision `images`, but ONLY when the
+   * prompt's model/route can take image input (see DESIGN decision 1). Non-vision:
+   * the URL is left as plain text — the model replies it can't open it; no
+   * fail-closed turn.
+   *
+   * This is the SINGLE backend place text-URL images are resolved, covering all
+   * Gateway front-ends (Feishu / Web chat / a2a / cron). Only user→agent prompts
+   * flow through THIS client; system/synthetic prompts run inside the AgentBox
+   * process and never reach here, so this never scans system text. The fetch +
+   * SSRF allowlist stay in the Gateway process (the AgentBox pod is
+   * network-isolated and must not fetch arbitrary user URLs).
+   */
+  async #withResolvedImageUrls(opts: PromptOptions): Promise<PromptOptions> {
+    if (!modelOptionsSupportImageInput(opts)) return opts;
+    const images = await enrichImagesFromText(opts.text, opts.images ?? []);
+    return images.length ? { ...opts, images } : opts;
+  }
+
+  /**
    * Send a prompt
    */
   async prompt(opts: PromptOptions): Promise<PromptResponse> {
-    console.log(`[agentbox-client] prompt sessionId=${opts.sessionId ?? "new"}`);
+    const sendOpts = await this.#withResolvedImageUrls(opts);
+    console.log(`[agentbox-client] prompt sessionId=${sendOpts.sessionId ?? "new"}`);
     const resp = await this.fetch("/api/prompt", {
       method: "POST",
       headers: { "Content-Type": "application/json" },
-      body: JSON.stringify(opts),
+      body: JSON.stringify(sendOpts),
     });
     const result: PromptResponse = await resp.json();
     console.log(`[agentbox-client] prompt → ok=${result.ok} sessionId=${result.sessionId}`);
--- a/src/gateway/agentbox/client.ts
+++ b/src/gateway/agentbox/client.ts
@@ -139,8 +139,9 @@ export class AgentBoxClient {
    * the URL is left as plain text — the model replies it can't open it; no
    * fail-closed turn.
    *
-   * This is the SINGLE backend place text-URL images are resolved, covering all
-   * Gateway front-ends (Feishu / Web chat / a2a / cron). Only user→agent prompts
+   * This is the SINGLE backend place text-URL images are resolved, covering ALL
+   * front-ends that reach AgentBox via this client — Feishu, DingTalk, Web chat,
+   * a2a, and cron (every `client.prompt` caller). Only user→agent prompts
    * flow through THIS client; system/synthetic prompts run inside the AgentBox
    * process and never reach here, so this never scans system text. The fetch +
    * SSRF allowlist stay in the Gateway process (the AgentBox pod is
--- a/src/gateway/agentbox/image-url-ingest.ts
+++ b/src/gateway/agentbox/image-url-ingest.ts
@@ -1,0 +1,247 @@
+import type { Readable } from "node:stream";
+
+/**
+ * Prompt-text image URL ingestion — the SINGLE backend place that turns image
+ * URLs found in a prompt's text into the `{ mimeType, data }` shape that
+ * `PromptOptions.images` (AgentBox vision input) expects.
+ *
+ * Why it lives here, in the AgentBox client layer (not in each channel, not in
+ * the AgentBox pod):
+ *   - Every Gateway→AgentBox prompt funnels through `AgentBoxClient.prompt()`
+ *     (Feishu directly; Portal Web chat / a2a / cron via the gateway `chat.send`
+ *     RPC), so enriching here covers all front-ends with one implementation.
+ *   - The outbound fetch + SSRF allowlist belong in the Gateway process, which
+ *     already has an egress path. The AgentBox pod is network-isolated (only
+ *     dials the Gateway over mTLS) — fetching arbitrary user URLs from inside it
+ *     would breach that isolation and hand SSRF reach into the cluster.
+ *
+ * Native Lark images (the receive-side resource download) are channel-specific
+ * and stay in `channels/inbound-image.ts`; they share the size/sniff helpers
+ * below but cannot be generalised across front-ends.
+ *
+ * Whether to run this at all is a VISION decision made by the caller
+ * (`AgentBoxClient`): only when the prompt's model/route can take image input do
+ * we resolve URLs; otherwise the URL is left as plain text for the model to see.
+ */
+
+/** Same shape as AgentBox `PromptOptions.images[]` (base64, no `data:` prefix). */
+export interface InboundImage {
+  mimeType: string;
+  data: string;
+}
+
+// AgentBox contract (src/agentbox/http-server.ts): MAX_PROMPT_MEDIA_ITEMS = 4
+// (images + files share the budget), single item base64 ≤ 8MB, mime ∈
+// {png,jpeg,webp}. Mirror the caps here to fail fast before hitting AgentBox.
+export const MAX_INBOUND_IMAGES = 4;
+export const MAX_IMAGE_BASE64_CHARS = 8 * 1024 * 1024;
+// 6MiB raw → ≤8MiB base64, so the AgentBox base64 cap can never be exceeded.
+export const MAX_IMAGE_BYTES = 6 * 1024 * 1024;
+const MAX_REDIRECTS = 3;
+const DEFAULT_FETCH_TIMEOUT_MS = 5000;
+
+// Extension-anchored: an OSS signed URL keeps its `.jpg` before the `?query`.
+// Extensionless URLs are intentionally out of scope for v1 (no HEAD probe).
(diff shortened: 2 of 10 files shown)


jacoblee-io (commented):
Code review: the overall design (single text-URL resolution point at the `AgentBoxClient.prompt` boundary, fetch kept in the Gateway process, fail-closed allowlist, vision gating) is sound and test coverage is good. Findings below, tagged by priority.

**Address before merge:** #1 (no overall timeout on the prompt path — a slow allowlisted host can stall every turn), #3 (signed-URL credentials persisted in plaintext).
**Confirm + add tests:** #2 (DingTalk silently affected), #4 (post receive-format assumption).
The rest are minor.

src/gateway/agentbox/client.ts
```diff
@@ -132,15 +133,35 @@ export class AgentBoxClient {
...
+   * fail-closed turn.
+   *
+   * This is the SINGLE backend place text-URL images are resolved, covering all
+   * Gateway front-ends (Feishu / Web chat / a2a / cron). Only user→agent prompts
+   * flow through THIS client; system/synthetic prompts run inside the AgentBox
+   * process and never reach here, so this never scans system text. The fetch +
+   * SSRF allowlist stay in the Gateway process (the AgentBox pod is
+   * network-isolated and must not fetch arbitrary user URLs).
+   */
+  async #withResolvedImageUrls(opts: PromptOptions): Promise<PromptOptions> {
```
jacoblee-io, line 155: **[Blocking / availability — address before merge]** Image resolution is `await`-ed here **before** `prompt()` sends to AgentBox, and `enrichImagesFromText` fetches the URLs **sequentially** — each URL can take `(MAX_REDIRECTS+1)` hops × `AbortSignal.timeout(5000)` each. Worst case ≈ 4 URLs × 4 × 5s ≈ **80s of blocking**, with no **global** deadline.

A single slow-but-allowlisted OSS host can stall or hang every turn — the channel user waits with no feedback, and the a2a/cron RPC blocks.

Suggestion: add an overall budget (e.g. one ≤8s `AbortSignal` for the whole resolution step) and fetch the images in parallel with `Promise.all`.

src/gateway/agentbox/client.ts
```diff
@@ -132,15 +133,35 @@ export class AgentBoxClient {
     return resp.json();
   }
 
+  /**
+   * Resolve image URLs in the prompt text into vision `images`, but ONLY when the
+   * prompt's model/route can take image input (see DESIGN decision 1). Non-vision:
+   * the URL is left as plain text — the model replies it can't open it; no
+   * fail-closed turn.
+   *
+   * This is the SINGLE backend place text-URL images are resolved, covering all
```
jacoblee-io: **[Behavior spread — confirm + add tests]** This comment lists Feishu / Web chat / a2a / cron, but `client.prompt()` actually has 4 callers and `dingtalk.ts` is one of them — it also populates `modelProvider/modelId/modelConfig/modelRouting`. So **DingTalk** silently gains text-URL image resolution too, yet there is no DingTalk test and no mention in the PR description/comment.

Either this is intended (then document it + add a DingTalk case) or it is an unvetted side effect.

src/gateway/channels/lark.ts
```diff
@@ -530,14 +598,14 @@ async function processQueuedLarkMessage(ctx: QueuedLarkMessageContext): Promise<
...
+  console.log(`[lark] Message channel=${channelId} chat=${chatId} sender=${senderOpenId ?? "unknown"} → agent=${agentId} session=${sessionId}: "${effectiveText.slice(0, 80)}" images=${imageRefs.length}`);
 
   try {
-    await ensureChatSession(sessionId, agentId, binding.createdBy, text, text, "channel");
+    await ensureChatSession(sessionId, agentId, binding.createdBy, effectiveText, effectiveText, "channel");
     await appendMessage({
       sessionId,
       role: "user",
-      content: text,
+      content: effectiveText,
```
jacoblee-io: **[Security / privacy — address before merge]** `safeUrl()` in `image-url-ingest.ts` deliberately strips the query before logging, and its comment notes a signed image URL may carry `Signature`/`AccessKeyId` credentials — but the `effectiveText` persisted here is still the **raw text containing the full signed URL**. It lands in the user row, session history, and the model context in plaintext.

The logging path is guarded; the persistence path is not. Suggestion: apply the same redaction to image URLs in the message text before persisting (or at least assess the risk of signed URLs in the DB).

src/gateway/channels/lark.ts
```diff
@@ -285,6 +286,67 @@ function isBotMentioned(message: any, botOpenId?: string): boolean {
...
+    return { text: "", imageRefs: imageKey ? [{ imageKey }] : [] };
+  }
+
+  if (msgType === "post") {
+    const imageRefs: LarkImageRef[] = [];
+    const textParts: string[] = [];
+    // im.message.receive_v1 delivers post content already locale-resolved as
+    // `{ title, content: Node[][] }` (the `{zh_cn:{…}}` nesting is a SEND-API
+    // shape, not a receive event), so parse the flat form directly.
+    const paragraphs: any[][] = Array.isArray(raw?.content) ? raw.content : [];
```
jacoblee-io: **[Robustness — confirm + add tests]** This assumes `raw.content` is a flat `Node[][]`. The comment asserts receive_v1 is always flat and that `{zh_cn:{…}}` is a send-side shape only — but Feishu post has historically delivered the locale-nested form on **receive** in some API/SDK versions. If a tenant/version sends the nested shape, `Array.isArray(raw?.content)` is false → no text and no images → the **entire post message is silently dropped** (the gate returns on "neither text nor image"), with no error.

Suggestion: verify against a real captured post receive payload. Also `title` text is dropped, and `tag:"a"` (hyperlink) / `tag:"at"` nodes are ignored — an image shared via a rich-text hyperlink would be lost.

src/gateway/channels/lark.ts
```diff
@@ -516,6 +579,11 @@ async function processQueuedLarkMessage(ctx: QueuedLarkMessageContext): Promise<
     return;
   }
 
+  // After the command branch: an image-only message has empty text, so give it a
+  // placeholder. Replace `text` uniformly here (not just in promptOpts) so the
+  // session title, persisted user row, and logs all show "user sent image(s)".
+  const effectiveText = text.length === 0 && imageRefs.length > 0 ? IMAGE_ONLY_PLACEHOLDER : text;
```
jacoblee-io, line 609: **[Edge case]** `effectiveText` is assigned **after** the command branch, which keys off the raw `text`. An image whose caption starts with `/` (e.g. `/foo look at this`) is routed as a command and the image is dropped. Minor, but currently undocumented and untested.

src/gateway/channels/lark.ts
```diff
@@ -634,8 +702,13 @@ async function processQueuedLarkMessage(ctx: QueuedLarkMessageContext): Promise<
   const modelBinding = frontendClient
     ? await resolveAgentModelBinding(agentId, frontendClient)
     : null;
+  // Download native Lark images at execution time — close to the prompt, and
+  // skipped entirely when the queue rejects. Text image URLs are NOT handled
+  // here: they are resolved generically (and vision-gated) at the
+  // `AgentBoxClient.prompt()` boundary, shared with Portal Web chat / a2a / cron.
+  const images = await collectInboundImages({ imageRefs, larkClient, messageId });
```
jacoblee-io: **[Consistency]** Native drag-in images (`collectInboundImages`) are **not** vision-gated — they are attached to `promptOpts.images` unconditionally — whereas the text-URL path is vision-gated (a non-vision model leaves the URL as plain text and degrades gracefully). Result: sending a native image to a non-vision model reaches AgentBox and **fails closed** at media filtering, so the same user action degrades gracefully on one path but errors on the other.

Suggestion: run native images through `modelOptionsSupportImageInput` too, dropping them (keeping the `[image]` placeholder) for non-vision models.

src/gateway/agentbox/image-url-ingest.ts
```diff
@@ -0,0 +1,247 @@
...
+  const out: InboundImage[] = [...existing];
+  const urls = extractImageUrls(text);
+  if (urls.length === 0) return out;
+
+  for (const url of urls) {
+    if (out.length >= MAX_INBOUND_IMAGES) break;
+    try {
+      const img = await fetchUrlImage(url);
+      if (img.data.length > MAX_IMAGE_BASE64_CHARS) {
+        console.warn(`[image-url-ingest] skip oversize image (url=${safeUrl(url)}): ${img.data.length} base64 chars`);
```
jacoblee-io: **[Minor / dead code]** This `MAX_IMAGE_BASE64_CHARS` (8MiB) check can never fire: `fetchUrlImage` already enforces `MAX_IMAGE_BYTES`=6MiB on the raw bytes, which guarantees base64 ≤ 8MiB. `inbound-image.ts` has the same redundant copy. Consider removing it or consolidating the size checks in one place.

src/gateway/agentbox/image-url-ingest.ts
```diff
@@ -0,0 +1,247 @@
...
+  for (let hop = 0; hop <= MAX_REDIRECTS; hop++) {
+    assertAllowedImageUrl(current); // re-validate every hop (redirect can't escape the allowlist)
+    const res = await fetch(current, {
+      redirect: "manual",
+      signal: AbortSignal.timeout(timeoutMs),
+    });
+    if (res.status >= 300 && res.status < 400) {
+      const loc = res.headers.get("location");
+      if (!loc) throw new Error(`redirect without location: ${safeUrl(current)}`);
+      current = new URL(loc, current).toString();
```
jacoblee-io, line 227: **[Minor]** On a 3xx redirect this `continue`s without consuming/cancelling the previous hop's `res.body`. Under undici an undrained response body can hold the connection until GC. Low impact, but you could `res.body?.cancel()` before `continue`.

src/gateway/agentbox/image-url-ingest.ts
```diff
@@ -0,0 +1,270 @@
...
+  if (out.length >= MAX_INBOUND_IMAGES) return out;
+  const urls = extractImageUrls(text);
+  if (urls.length === 0) return out;
+
+  // One deadline for the whole step (a slow allowlisted host must not stall the
+  // turn), and fetch the URLs in parallel rather than sequentially. The raw-byte
+  // cap in fetchUrlImage (MAX_IMAGE_BYTES) already bounds each image's size.
+  const overall = AbortSignal.timeout(totalTimeoutMs());
+  const settled = await Promise.all(
+    urls.map(async (url) => {
```
jacoblee-io, line 272: **[Blocking — DoS/SSRF amplification]** The switch to parallel fetching dropped the per-image cap that the old sequential loop enforced via `break`. `urls.map(...)` now fires a fetch for **every** extracted URL, and `MAX_INBOUND_IMAGES` (4) is only applied afterwards when collecting results. `extractImageUrls` dedups but does **not** bound the count.

So a single channel message with N distinct allowlisted image URLs (any group member can send one) triggers **N concurrent outbound fetches** from the Gateway — e.g. 50 URLs → 50 connections × up to 6 MiB each ≈ 300 MiB transient. The overall timeout bounds wall-clock but not concurrency or memory.

Fix: bound the fan-out to the remaining slots before mapping, e.g.
```js
const slots = MAX_INBOUND_IMAGES - out.length;
const settled = await Promise.all(urls.slice(0, slots).map(async (url) => { ... }));
```

src/gateway/channels/lark.ts
```diff
@@ -634,8 +725,24 @@ async function processQueuedLarkMessage(ctx: QueuedLarkMessageContext): Promise<
...
+  // Text image URLs are NOT handled here — they are resolved generically (and
+  // vision-gated) at the `AgentBoxClient.prompt()` boundary, shared with Portal
+  // Web chat / a2a / cron.
+  const visionCapable = modelOptionsSupportImageInput({
+    modelProvider: modelBinding?.modelProvider,
+    modelId: modelBinding?.modelId,
+    modelConfig: modelBinding?.modelConfig,
+    modelRouting: modelBinding?.modelRouting,
+  });
+  const images = visionCapable
```
jacoblee-io, line 746: **[Minor — UX]** With a non-vision model, native images are now dropped entirely (correct, mirrors the text-URL path), but the user gets no signal: the turn is just `[image]` placeholder text and the model sees nothing about an image. On the text-URL path the model at least receives the URL and can reply "I can't open it."

Consider a one-line hint to the user ("this model can't read images") instead of silently ignoring the attachment. Not blocking.

jacoblee-io (commented):
Re-review after the "address review feedback" commit — thanks, this is in good shape. The earlier findings are well handled: total timeout + parallel fetch (#1), locale-nested/`tag:"a"` post parsing (#4), vision-gated native images (#6), dead base64 checks removed (#7), redirect body drained (#8).

One **blocking** item remains (the parallelization introduced it) plus one minor and a deployment note:

- **Blocking:** parallel fan-out is no longer bounded by the image cap — see inline on `image-url-ingest.ts`.
- **Minor:** non-vision models silently swallow native images — see inline on `lark.ts`.
- **Deployment:** the env var was renamed to `SICLAW_IMAGE_URL_ALLOWLIST` with no compat shim. Confirm prod sets the new name before/at rollout, otherwise text-URL resolution silently stays fail-closed.

With the fan-out bound added, this is good to merge from my side.

jacoblee-io (approved):
Approving. All earlier review findings are resolved and verified:

- **Parallel fan-out** is now bounded to the remaining media budget before firing (`slice(0, budget)`), with tests proving 50 URLs → 4 fetches and existing-2 → 2 fetches. No more unbounded concurrent egress.
- **Non-vision native images** now append a note so the model can tell the user it can't read images, mirroring the text-URL path.
- **Signed-URL credentials** are stripped (`redactImageUrlsInText`) from both the persisted row and the text forwarded to AgentBox, while the fetch still uses the full URL — verified by tests.

Diagnosis double-checked: `IMAGE_URL_RE` is only used via `.match`/`.replace` (no stateful `.test`/`.exec`), so no shared-regex lastIndex hazard.

Non-blocking follow-ups (fine for a later PR): the redaction is wired into Lark's gateway-DB persistence + the shared client layer but not DingTalk's own user-row write (`dingtalk.ts`), and the redaction strips the entire query (not just credential params) on every path — acceptable given vision attaches the image and non-vision can't use it anyway.

Deployment reminder: the env var is renamed to `SICLAW_IMAGE_URL_ALLOWLIST` with no compat shim — set the new name in prod or text-URL resolution stays fail-closed.

PR #367, opened 2026-06-29
feat(agentbox): configure embedding endpoint and node scheduling for K8s pods

silence8013:
## Summary

Two related K8s/AgentBox pod configuration gaps for production deployments: there was no way to (1) supply the memory indexer's embedding endpoint, or (2) constrain which nodes AgentBox pods schedule onto. Both are now configurable via the runtime env and surfaced through the Helm chart.

### Problem

In K8s mode the runtime spawns one AgentBox pod per agent, but the pod spec was effectively fixed:

- **Embedding endpoint** — Portal's `config.getSettings` serves only providers, the agentbox only auto-writes `{ providers }` to `settings.json`, and its config volume is an `emptyDir` with no ConfigMap mount. With memory enabled but no embedding config, `getEmbeddingConfig()` returns null and memory silently degrades to FTS-only (no vector recall).
- **Node scheduling** — the pod spec carried no `nodeSelector`, so operators couldn't pin agent pods to a dedicated/labeled node pool (e.g. an SSD or GPU pool); pods landed on whatever node the scheduler picked.

### Solution

- **Embedding** — add infrastructure env overrides (`SICLAW_EMBEDDING_BASE_URL` / `_MODEL` / `_DIMENSIONS` / `_API_KEY`) in `loadConfig()`, forward them from the runtime into every spawned AgentBox via the k8s-spawner allowlist, and expose a `runtime.embedding` block in the chart. `SICLAW_EMBEDDING_API_KEY` stays in the sanitize-env blocklist so it never leaks into child shell processes. An empty embedding key now only inherits the default LLM provider's key when the embedding endpoint is the **same origin** as that provider; a cross-origin self-hosted endpoint stays unauthenticated so the LLM credential is never sent to a different trust domain.
- **nodeSelector** — add an optional `nodeSelector` to `K8sSpawnerConfig`, written onto each pod spec when non-empty. Forwarded via a JSON `SICLAW_AGENTBOX_NODE_SELECTOR` env var (malformed input is warned and ignored, never crashes startup), and exposed as an `agentbox.nodeSelector` Helm block rendered with `toJson`. An empty map omits both the env var and the pod field, so default behaviour is unchanged. This targets **labeled** node pools; pinning to a *tainted* pool (which additionally needs `tolerations`) is out of scope for this PR.

## Test Plan

- [x] `npx tsc --noEmit`
- [x] `npm test` (197 files, 3965 passing)
- [x] `npm run build`
- [x] `helm template` render with and without each option set
- [x] Embedding verified live in-cluster (9/9 chunks embedded with Qwen3-Embedding-0.6B)


helm/siclaw/templates/gateway-deployment.yaml
```diff
@@ -52,6 +52,26 @@ spec:
...
+            {{- if .model }}
+            - name: SICLAW_EMBEDDING_MODEL
+              value: {{ .model | quote }}
+            {{- end }}
+            {{- if .dimensions }}
+            - name: SICLAW_EMBEDDING_DIMENSIONS
+              value: {{ .dimensions | quote }}
+            {{- end }}
+            {{- if .apiKey }}
+            - name: SICLAW_EMBEDDING_API_KEY
```
chent1996, line 70: Non-blocking hygiene — `SICLAW_EMBEDDING_API_KEY` is rendered as a plain `value:` env entry. For a literal key this lands in the Deployment manifest in cleartext (visible to anyone with `kubectl get deployment` rights). The standard K8s pattern would be `valueFrom: secretKeyRef`. 

Suggestion for a follow-up: support both shapes via the chart, e.g. `runtime.embedding.apiKey` (literal, current) and `runtime.embedding.apiKeySecretRef: { name, key }` (production). Not for this PR since the stated common case is unauthenticated self-hosted TEI — just flagging.

helm/siclaw/templates/gateway-deployment.yaml
```diff
@@ -74,6 +94,12 @@ spec:
             - name: SICLAW_DEBUG_IMAGE
               value: {{ .Values.agentbox.debugImage | quote }}
             {{- end }}
+            {{- if .Values.agentbox.nodeSelector }}
+            # Node scheduling constraint forwarded by the runtime spawner onto
+            # every spawned AgentBox pod (parsed as a JSON label map).
```
chent1996, line 99: Minor — `{{- if .Values.agentbox.nodeSelector }}` is **truthy for an empty map `{}`** in Sprig (it's a non-nil reference, not the same as Go's `nil`). So the values.yaml default `nodeSelector: {}` will still emit `SICLAW_AGENTBOX_NODE_SELECTOR="{}"` into the Deployment manifest. The runtime side handles this gracefully (`parseNodeSelector` returns `undefined` when `Object.keys(result).length === 0`, and the k8s-spawner test explicitly covers `nodeSelector: {}` ⇒ no pod field), so the actual behaviour is correct — but a deployer reading their rendered manifest will see a confusing env var they didn't set.

Guard could be `{{- if and .Values.agentbox.nodeSelector (gt (len .Values.agentbox.nodeSelector) 0) }}` to suppress on the empty-default path.

src/core/config.embedding-env.test.ts
```diff
@@ -0,0 +1,87 @@
...
+  it("applies defaults (bge-m3 / 1024) when only baseUrl is set", () => {
+    process.env.SICLAW_EMBEDDING_BASE_URL = "http://tei.svc/v1";
+    reloadConfig();
+    expect(getEmbeddingConfig()).toEqual({
+      baseUrl: "http://tei.svc/v1",
+      apiKey: "",
+      model: "BAAI/bge-m3",
+      dimensions: 1024,
+    });
+  });
```
chent1996, line 71: Coverage gap, non-blocking: the suite uses an empty SICLAW_CONFIG_DIR (no settings.json) so it only exercises pure-env construction. The implementation explicitly supports **field-level merge over an existing file config** (`envBaseUrl ?? existing?.baseUrl ?? ""` etc.), but no test pins that: a future refactor could silently revert to wholesale env-replace-or-fallthrough and the tests would still pass. One case where env partially overrides a populated `cached.embedding` would lock the contract.

src/lib/bootstrap-runtime.ts
```diff
@@ -184,12 +184,56 @@ function createSpawner(
...
+ * string-valued object, warning once so misconfiguration is visible without
+ * crashing runtime startup.
+ */
+function parseNodeSelector(raw: string | undefined): Record<string, string> | undefined {
+  if (!raw || !raw.trim()) return undefined;
+  let parsed: unknown;
+  try {
+    parsed = JSON.parse(raw);
+  } catch {
+    console.warn(
```
chent1996, line 217: Out-of-scope thought, not for this PR: `nodeSelector` is applied to spawned AgentBox pods but the runtime gateway Deployment itself still schedules anywhere. For deployments that taint a dedicated pool for the gateway as well (CPU/RAM/security separation from agents), operators would want symmetric control via `runtime.nodeSelector`. Probably worth a separate small PR if/when an operator asks.

chent1996 (dismissed):
Clean scoping — two independent operational gaps (embedding endpoint for memory, nodeSelector for agent pods) bundled in one focused PR, each with its own commit and test set. The defensive JSON parsing for nodeSelector (`parseNodeSelector` with warn-and-ignore for malformed input), the env-forwarding allowlist extension over arbitrary forwarding, and the file/env merge with field-level fallback are all the right patterns. PR description is unusually clear about problem and verification.

Security check: `SICLAW_EMBEDDING_API_KEY` is correctly blocked from child shells by `sanitize-env.ts`'s SICLAW_* catch-all rule (anything not in `SICLAW_SAFE` is stripped) — verified via the existing test `blocks SICLAW_EMBEDDING_API_KEY` at sanitize-env.test.ts:11. 

Four non-blocking notes inline — all hygiene/edge-case, none change the shipping decision.

src/core/config.ts
```diff
@@ -316,6 +316,29 @@ export function loadConfig(): SiclawConfig {
...
+  // Each field overrides the file value individually; missing fields keep the
+  // file value (or fall back to getEmbeddingConfig()'s defaults). The block is
+  // only constructed when at least one var is set, so non-memory deployments are
+  // unaffected. `getEmbeddingConfig()` still returns null when baseUrl is empty.
+  {
+    const envBaseUrl = process.env.SICLAW_EMBEDDING_BASE_URL;
+    const envModel = process.env.SICLAW_EMBEDDING_MODEL;
+    const envApiKey = process.env.SICLAW_EMBEDDING_API_KEY;
+    const envDimensions = process.env.SICLAW_EMBEDDING_DIMENSIONS;
+    if (envBaseUrl || envModel || envApiKey || envDimensions) {
```
jacoblee-io, line 330: **[Security — pre-existing fallback, newly reachable via this path]** This new env path is the first way (in K8s) to point memory at a self-hosted embedding endpoint with an empty key — which is exactly the use case `values.yaml` documents (`apiKey: "" # empty for unauthenticated self-hosted endpoints`).

But `getEmbeddingConfig()` (config.ts, unchanged here) has a long-standing fallback: when `embedding.apiKey` is empty it substitutes the **default LLM provider's** API key:
```js
let apiKey = config.embedding?.apiKey ?? "";
if (!apiKey) { const d = getDefaultLlm(); if (d) apiKey = d.apiKey; }
```
So with memory enabled + an embedding `baseUrl` set + no embedding key, the agent's main LLM key (Anthropic/OpenAI/etc.) is sent as `Authorization: Bearer …` to that self-hosted TEI/vLLM endpoint. That contradicts the "empty = unauthenticated" doc and leaks a high-value credential to a different trust domain.

The fallback is reasonable when embedding and LLM share one provider, but not for the independent-endpoint case this PR enables. Suggestion (any one): for env-configured embedding, don't inherit the LLM key; or send no auth header when the embedding key is empty; or at minimum fix the `values.yaml` comment to state that the default LLM key is sent when left empty.

src/gateway/agentbox/k8s-spawner.ts
```diff
@@ -261,6 +277,9 @@ export class K8sSpawner implements BoxSpawner {
         subdomain: "agentbox-hs",
         automountServiceAccountToken: false,
         restartPolicy: "Never",
+        ...(this.config.nodeSelector && Object.keys(this.config.nodeSelector).length > 0
```
jacoblee-io, line 280: **[Functional gap vs. the PR description]** `nodeSelector` alone pins pods to *labeled* nodes, but the PR description lists pinning to a "dedicated/**tainted** node pool" as a goal. A dedicated pool is usually both labeled **and** tainted, and a `nodeSelector` does not tolerate taints — pods would land on the selected nodes only if those nodes aren't tainted, otherwise they stay `Pending`.

Either add an optional `tolerations` passthrough alongside `nodeSelector`, or drop "tainted" from the PR/description so operators don't pin to a tainted pool and get unschedulable agent pods. Not blocking for the labeled-pool case.

jacoblee-io (commented):
Clean PR, good test coverage and defensive parsing. Two findings not already covered by the existing comments (the cleartext-key and empty-map points are well handled there):

- A pre-existing embedding-key fallback that this PR's new env path now triggers (inline on `config.ts`) — worth a security look.
- `nodeSelector` covers dedicated *labeled* pools, but the PR description also mentions *tainted* pools, which need tolerations (inline on `k8s-spawner.ts`).

Neither is blocking on its own; #1 is the one I'd most like addressed before this is used with memory enabled.

src/core/config.ts
silence8013: Good catch — fixed in 336e91b. `getEmbeddingConfig()` now only inherits the default LLM provider's key when the embedding endpoint is the **same origin** (scheme + host + port) as that provider, i.e. one provider serving both chat and embeddings. A cross-origin self-hosted endpoint with an empty key now sends no auth, so the LLM credential is never shipped to a different trust domain — "empty = unauthenticated" holds for the case this PR enables. Malformed URLs fail closed (no inheritance). Updated the `values.yaml` comment to document the same-origin behavior, and added tests covering cross-origin (no inherit) / same-origin (inherit) / explicit key (always wins).

src/gateway/agentbox/k8s-spawner.ts
silence8013: Thanks — scoping this PR to the labeled-pool case. I've dropped "tainted" from the PR description so operators won't try to pin to a tainted pool and get `Pending` agent pods. A `tolerations` passthrough alongside `nodeSelector` is a reasonable follow-up if/when someone needs a dedicated tainted pool, but it's out of scope here.

jacoblee-io (approved):

PR #368, opened 2026-06-29
fix(core): pass raw JSON Schema as MCP tool parameters

### LINKED ISSUE #355: MCP tools with integer/object params fail validation (`must be integer`) — TypeBox 0.34 vs typebox 1.x mismatch disables arg coercion
## Description

Any MCP tool that declares an `integer`/`object`/`boolean` parameter fails at the **argument validation stage** (before the call ever reaches the MCP server) whenever the model emits that argument as a string (e.g. `"10"` instead of `10`). The call is rejected with errors like:

```
Validation failed for tool "mcp__grafana__get_panel_image":
  - panelId: must be integer
```

Root cause: `src/core/mcp-client.ts` builds MCP tool parameter schemas with **`@sinclair/typebox` `^0.34`**, but the pi runtime validates/coerces tool arguments with a *different* package, **`typebox` `1.1.38`** (bundled under `@earendil-works/pi-ai`). The two are incompatible at the metadata level, which silently disables the string→number coercion that should normalize these arguments.

This is **not** a serialization bug in the MCP bridge itself — `McpClientManager.createToolDefinition().execute` passes `args` through to `client.callTool` untouched. The string value originates from the model output, and the coercion step that is supposed to fix it is dead.

## Steps to Reproduce

1. Configure a Grafana MCP server in `.siclaw/config/settings.json` (`mcpServers`).
2. Ask the agent to render a panel, e.g. call `get_panel_image` with `dashboardUid` + `panelId: 10` (+ `timeRange`, `width`, `height`).
3. The argument arrives as a string (`"10"`) and is rejected during validation; the call never reaches Grafana.

Any MCP tool with an integer/object/boolean parameter reproduces this; pure-string tools do not.

## Expected Behavior

String-encoded numeric/boolean/object arguments are coerced to the schema type (`"10"` → `10`) before validation, so the MCP call succeeds — matching the behavior pi-ai already implements for raw JSON Schema tools (`coerceWithJsonSchema`).

## Actual Behavior

Validation throws `panelId: must be integer` and the tool call is dropped before reaching the MCP server.

### Why coercion silently fails

In `@earendil-works/pi-ai` `validateToolArguments` (`dist/utils/validation.js`):

```js
const args = structuredClone(toolCall.arguments);
Value.Convert(tool.parameters, args);                 // (1) intended string→int coercion
...
if (!hasTypeBoxMetadata(tool.parameters) && isJsonSchemaObject(tool.parameters)) {
    const coerced = coerceWithJsonSchema(args, tool.parameters);  // (2) robust JSON-Schema fallback
    ...
}
if (validator.Check(args)) return args;               // (3) otherwise throws "must be integer"
```

The two TypeBox packages identify a schema's kind differently:

- `@sinclair/typebox` 0.34 marks kind with the symbol `Symbol.for('TypeBox.Kind')` (`build/cjs/type/symbols/symbols.js`).
- `typebox` 1.x reads a plain string property `~kind` (`build/type/types/schema.mjs` → `IsKind` checks `value["~kind"] === kind`).

Because our schema is built by 0.34:

1. **`Value.Convert` (typebox 1.x) is a no-op** — `FromType` dispatches via `IsInteger`/`IsObject`, which check `~kind`. A 0.34 schema has no `~kind`, so every branch is false and `FromType` returns the value unchanged (it never even recurses into object properties). `"10"` stays a string.
2. **The robust fallback is skipped** — `hasTypeBoxMetadata` checks exactly the 0.34 symbol `Symbol.for('TypeBox.Kind')`, which *is* present, so it returns `true` and the `if (!hasTypeBoxMetadata ...)` guard skips `coerceWithJsonSchema` (the path that correctly coerces `"10"`→`10`, `"true"`→`true`, arrays/nested objects/`oneOf`).
3. **The compiled validator still enforces `type: "integer"`**, so `Check` fails and the error is thrown.

Secondary issue in the same file: `jsonSchemaToTypebox` handles `object` by iterating `properties` only and **drops `additionalProperties`**, so schemas like `get_panel_image`'s `variables` lose fidelity regardless of the coercion bug.

## Proposed Fix

Stop converting the MCP `inputSchema` through `@sinclair/typebox` in `src/core/mcp-client.ts` and pass the raw JSON Schema as the tool's `parameters`:

```ts
// createToolDefinition()
const inputSchema = mcpTool.inputSchema ?? { type: "object", properties: {} };
const parameters = inputSchema; // raw JSON Schema — no 0.34 Kind symbol
```

Effect:
- `hasTypeBoxMetadata` becomes `false` → pi-ai runs `coerceWithJsonSchema`, which correctly coerces `"10"`→`10` (and handles `additionalProperties`, `oneOf`, arrays, nested objects).
- Advertising to the LLM is unaffected — providers read `parameters.properties` / `parameters.required` directly (`pi-ai` `providers/anthropic.js`, `openai-completions.js`: *"TypeBox already generates JSON Schema"*).
- Removes the dependency coupling on the mismatched `@sinclair/typebox` 0.34 for MCP tools.

Open question for maintainers: is keeping `@sinclair/typebox` 0.34 intentional, or should siclaw align on the same `typebox` 1.x the pi runtime uses? The raw-JSON-Schema passthrough fixes this regardless, but it would be good to confirm the preferred direction (and whether `jsonSchemaToTypebox` should be removed or retained as a fallback).

## Deployment Mode

TUI (local CLI) — but the defect is in shared core (`src/core/mcp-client.ts`), so it affects all modes.

## Node.js Version

24.10.0 (also reproduces on the documented >= 22.19.0 baseline)

## Additional Context

- siclaw `0.1.4`
- `@earendil-works/pi-coding-agent` / `pi-ai` / `pi-agent-core` `0.78.1`
- `@sinclair/typebox` `0.34.48` (siclaw direct dep)
- `typebox` `1.1.38` (nested under the pi packages)

Relevant locations:
- `src/core/mcp-client.ts` — `jsonSchemaToTypebox()`, `createToolDefinition()`
- `@earendil-works/pi-ai/dist/utils/validation.js` — `validateToolArguments()`, `hasTypeBoxMetadata()`, `coerceWithJsonSchema()`

silence8013:
## Summary

MCP tools that declare `integer`/`object`/`boolean` parameters fail argument validation (`must be integer`) whenever the model emits the value as a string (`"10"`), so the call never reaches the MCP server. This passes the MCP `inputSchema` through as raw JSON Schema instead of converting it to a `@sinclair/typebox` schema, which re-enables the pi runtime's argument coercion.

Closes #355

### Problem

`src/core/mcp-client.ts` built tool parameter schemas with `@sinclair/typebox` 0.34, but the pi runtime (`@earendil-works/pi-ai`) validates and coerces arguments with `typebox` 1.x. The two tag schema "kind" differently (`Symbol.for('TypeBox.Kind')` vs the `~kind` string prop), so:

- `Value.Convert` is a no-op — `"10"` stays a string.
- pi-ai's robust JSON-Schema coercion fallback is skipped, because `hasTypeBoxMetadata` detects the 0.34 symbol.
- The compiled validator still enforces `type: "integer"`, so the call is dropped with `panelId: must be integer` (e.g. Grafana's `get_panel_image`).

A secondary issue: the old `jsonSchemaToTypebox` dropped `additionalProperties` (e.g. the `variables` field), losing schema fidelity.

### Solution

Add `normalizeMcpInputSchema`, which returns a plain JSON Schema (no TypeBox kind metadata) while guaranteeing the `type: "object"` + `properties` shape and preserving `required` / `additionalProperties` / other fields. `createToolDefinition` now uses it for tool `parameters`.

With the kind metadata absent, pi-ai's `coerceWithJsonSchema` runs (`"10"` → `10`), and providers still advertise tools straight from `.properties` / `.required`. `jsonSchemaToTypebox` is retained (no longer used for params) for compatibility.

Rejected alternative: aligning siclaw on `typebox` 1.x — a larger dependency change, deferred to maintainer guidance per the open question in #355.

## Test Plan

- [x] `npm test` (`npm run build`) passes
- [x] `vitest run src/core/mcp-client.test.ts` — 22 passed, including new `normalizeMcpInputSchema` cases asserting no TypeBox kind symbol and preserved `additionalProperties`

## Architecture Checklist

- [x] **Tool protocol**: MCP tools continue to register as `ToolDefinition`s; `parameters` now carries raw JSON Schema (accepted by the pi runtime), which fixes the coercion path. Change is confined to shared core (`src/core/mcp-client.ts`), so it applies to all deployment modes.

## General Checklist

### DIFF
--- a/src/core/mcp-client.test.ts
+++ b/src/core/mcp-client.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect } from "vitest";
-import { jsonSchemaToTypebox, buildMcpToolName, isMcpTool, MCP_TOOL_PREFIX, mcpContentToAgentContent, McpClientManager } from "./mcp-client.js";
+import { jsonSchemaToTypebox, normalizeMcpInputSchema, buildMcpToolName, isMcpTool, MCP_TOOL_PREFIX, mcpContentToAgentContent, McpClientManager } from "./mcp-client.js";
 
 describe("jsonSchemaToTypebox", () => {
   it("converts string type", () => {
@@ -166,5 +166,44 @@ describe("createToolDefinition server description", () => {
   it("applies the server context to the fallback description too", () => {
     const def = makeDef("Monitoring tenant ID: t-123");
     expect(def.description).toBe('[Server "grafana" context: Monitoring tenant ID: t-123]\nMCP tool query from grafana');
+  });
+});
+
+describe("normalizeMcpInputSchema", () => {
+  const TYPEBOX_KIND = Symbol.for("TypeBox.Kind");
+
+  it("returns a plain JSON Schema without TypeBox kind metadata", () => {
+    // This is the core invariant behind scitix/siclaw#355: MCP tool parameters
+    // must NOT carry the @sinclair/typebox Kind symbol, otherwise the pi runtime
+    // (typebox 1.x) skips its JSON-Schema arg coercion and string-encoded integers
+    // like "10" fail validation with "must be integer".
+    const result = normalizeMcpInputSchema({
+      type: "object",
+      properties: { panelId: { type: "integer" } },
+      required: ["panelId"],
+    });
+    expect(Object.getOwnPropertySymbols(result)).not.toContain(TYPEBOX_KIND);
+    // By contrast, the old TypeBox conversion DID carry the symbol.
+    const viaTypebox = jsonSchemaToTypebox({ type: "object", properties: {} });
+    expect(Object.getOwnPropertySymbols(viaTypebox)).toContain(TYPEBOX_KIND);
+  });
+
+  it("preserves properties, required and additionalProperties", () => {
+    const result = normalizeMcpInputSchema({
+      type: "object",
+      properties: { panelId: { type: "integer" }, dashboardUid: { type: "string" } },
+      required: ["panelId", "dashboardUid"],
+      additionalProperties: { type: "string" },
+    }) as any;
+    expect(result.type).toBe("object");
+    expect(result.properties.panelId).toEqual({ type: "integer" });
+    expect(result.required).toEqual(["panelId", "dashboardUid"]);
+    expect(result.additionalProperties).toEqual({ type: "string" });
+  });
+
+  it("guarantees object shape for missing/empty/invalid schemas", () => {
+    expect(normalizeMcpInputSchema(undefined) as any).toMatchObject({ type: "object", properties: {} });
+    expect(normalizeMcpInputSchema({}) as any).toMatchObject({ type: "object", properties: {} });
+    expect(normalizeMcpInputSchema([1, 2]) as any).toMatchObject({ type: "object", properties: {} });
   });
 });
--- a/src/core/mcp-client.ts
+++ b/src/core/mcp-client.ts
@@ -1,7 +1,7 @@
 /**
  * MCP Client Manager — connects to external MCP servers and exposes their tools
- * as pi-agent ToolDefinitions (TypeBox schema) for the pi-agent brain,
- * and as raw config for the Claude SDK brain (native MCP support).
+ * as pi-agent ToolDefinitions (raw MCP JSON Schema as parameters) for the
+ * pi-agent brain, and as raw config for the Claude SDK brain (native MCP support).
  *
  * Supports three transport types: stdio, sse, streamable-http.
  * Config loaded from .siclaw/config/settings.json mcpServers field.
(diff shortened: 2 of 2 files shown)


LikiosSedo (approved):
Reviewed the current PR head against origin/main. The raw JSON Schema path matches the active pi-ai 0.78.1 validation behavior, preserves schema fidelity for required/additionalProperties, and the merge-result focused checks passed. No blocking issues found.

src/core/mcp-client.ts
```diff
@@ -285,7 +314,18 @@ export class McpClientManager {
...
+    // arguments with a *different* TypeBox package (`typebox` 1.x, bundled under
+    // @earendil-works/pi-ai) whose schema-kind detection (`~kind` string prop) is
+    // incompatible with @sinclair/typebox 0.34's metadata (`Symbol.for('TypeBox.Kind')`).
+    // Converting here silently disables string→number/boolean coercion (e.g. "10" → 10):
+    // `Value.Convert` no-ops and pi-ai's JSON-Schema coercion fallback is skipped because
+    // `hasTypeBoxMetadata` detects the 0.34 symbol — so integer/object params such as
+    // get_panel_image's `panelId` fail validation with "must be integer". Raw JSON Schema
+    // leaves the kind metadata absent, letting pi-ai's coercion path run, and providers
+    // advertise tools straight from `.properties`/`.required`. See scitix/siclaw#355.
+    const parameters = normalizeMcpInputSchema(inputSchema);
```
jacoblee-io, line 328: **[Scope / root cause — not blocking]** This correctly fixes MCP tools, but the underlying coercion-skip isn't MCP-specific. Native siclaw tools build `parameters` with `@sinclair/typebox` 0.34 (`Type.Object`/`Type.Number`, e.g. `host_exec`'s `timeout_seconds`), which carry the same `Symbol.for("TypeBox.Kind")`. By the identical logic in pi-ai's `validation.js`:
```js
Value.Convert(tool.parameters, args);                 // no-op on a 0.34-tagged schema
if (!hasTypeBoxMetadata(tool.parameters) && isJsonSchemaObject(...)) {
  coerced = coerceWithJsonSchema(args, tool.parameters); // skipped — 0.34 symbol present
}
```
so a native numeric/boolean param sent as a string (`"30"`) would fail the same way (`must be number`). This PR is a targeted patch on one symptom; the real fix (align siclaw on typebox 1.x, deferred per #355) would cover native tools too. Worth tracking that as the root-cause follow-up so the next person doesn't rediscover this on a native tool.

src/core/mcp-client.test.ts
```diff
@@ -168,3 +168,42 @@ describe("createToolDefinition server description", () => {
     expect(def.description).toBe('[Server "grafana" context: Monitoring tenant ID: t-123]\nMCP tool query from grafana');
   });
 });
+
+describe("normalizeMcpInputSchema", () => {
```
jacoblee-io, line 172: **[Test depth — minor]** These cases assert the *precondition* (no TypeBox Kind symbol, fields preserved) but not the actual behavior the PR fixes — that a string-encoded `"10"` coerces to `10` through pi-ai's validation path. That end-to-end was verified manually (Grafana `get_panel_image`). Consider an integration test that runs a stringy-int arg through the pi-ai coercion so the fix is locked against a future pi-ai/typebox upgrade silently regressing it.

jacoblee-io (commented):
Verified the root cause in pi-ai's `validation.js` (`hasTypeBoxMetadata` gate + `Value.Convert` no-op on a 0.34-tagged schema) and the version split (siclaw `@sinclair/typebox` 0.34.x vs pi-ai's bundled typebox 1.x). The fix is correct and the `additionalProperties`/`required` preservation is a real bonus (it also re-enables nested coercion). Also confirmed nothing in siclaw consumes `tool.parameters` as a real TypeBox schema, so the `as unknown as TSchema` cast is safe today.

Two points not raised by the existing approval — neither blocking this fix:

- The same coercion-skip affects **native tools**, not just MCP (inline on `mcp-client.ts`).
- New tests assert the precondition but not the end-to-end coercion (inline on the test file).

PR #370, opened 2026-06-29
fix(tools): drop `setsid -w` for portability across older util-linux/busybox

### LINKED ISSUE #356: host_exec/host_script fail on older SSH targets: `setsid -w` requires util-linux >= 2.24 (CentOS 7 ships 2.23.2)
## Description

Every command dispatched over SSH by `host_exec` (and `host_script`) is wrapped in a killable session via `setsid -w` in `wrapBackgroundSession()` (`src/tools/infra/bg-session.ts`):

```ts
export function wrapBackgroundSession(innerCmd: string, pgidFile: string): string {
  const launch = `echo $$ > ${pgidFile}; ${innerCmd}; rc=$?; rm -f ${pgidFile}; exit $rc`;
  return `setsid -w sh -c ${shellEscape(launch)}`;   // <-- requires util-linux >= 2.24
}
```

The `-w` / `--wait` option was only added to `setsid` in **util-linux 2.24** (2013). On older SSH targets the option does not exist, so the wrapper fails immediately and the user's command never runs.

This affects both the foreground and background paths of `host_exec`/`host_script`, since both go through `wrapBackgroundSession()`.

A code comment currently asserts this dependency is "present on any normal SSH target incl. BusyBox" — that assumption is inaccurate: BusyBox `setsid` does not support `-w` either, and neither does util-linux < 2.24.

## Steps to Reproduce

1. Bind an SSH host running CentOS 7 (util-linux 2.23.2) as a target.
2. Ask the agent to read/inspect a file on that node (any `host_exec` command).
3. The agent invokes `host_exec`, which dispatches a `setsid -w sh -c ...` wrapped command.

## Expected Behavior

A general-purpose SSH command-execution tool should run on any target with a POSIX shell. Reading a file on a CentOS 7 node should succeed regardless of the target's `setsid`/util-linux version.

## Actual Behavior

The remote command exits immediately without running, with:

```
Exit code: 1
STDERR: setsid: invalid option -- 'w'
Usage:
 setsid [options] <program> [arguments ...]
Options:
 -c, --ctty     set the controlling terminal to the current one
 -h, --help     display this help and exit
 -V, --version  output version information and exit
For more details see setsid(1).
```

## Target environment

```
# cat /etc/redhat-release
CentOS Linux release 7.8.2003 (Core)

# rpm -qf /bin/setsid
util-linux-2.23.2-65.el7_9.1.x86_64
```

`setsid -w` works on a newer host, confirming this is purely a version dependency.

## Root cause

`-w` makes `setsid` wait for the child and return its real exit status (needed by the foreground path, which relies on synchronous `sshExec` to capture the exit code). The new *session* (rather than just a process group) is used so that GNU `timeout`'s own process group is also reaped by `pkill -s <sid>` on `job_stop`. The behavior is correct, but it hard-depends on a util-linux version that CentOS 7 / RHEL 7 (and BusyBox) do not satisfy.

## Possible fixes (for discussion)

1. **Capability probe + fallback** — probe `setsid -w true` once per target and fall back to a `-w`-free path when unsupported (extra SSH round-trip; fallback still needs to solve wait + exit-code).
2. **Pure-POSIX session wrapper (preferred)** — replace `setsid -w` with a shell construct that records the session/process group and waits without depending on a specific util-linux version, preserving the `job_stop` reap semantics. Centralized in `bg-session.ts`, shared by `host_exec`/`host_script`/`node_exec`/`pod_script`, so it needs a full regression pass (including the e2e remote-reap test).
3. **Graceful degradation** — when `setsid`/`timeout` are missing or incompatible, run the bare command and clearly report reduced `job_stop` guarantees.

Happy to open a PR — would like maintainer guidance on the preferred direction (esp. #2 vs #1) before implementing.

## Deployment Mode

TUI (local CLI)

## Additional Context

Relevant code: `src/tools/infra/bg-session.ts`, `src/tools/cmd-exec/host-exec.ts`, `src/tools/script-exec/host-script.ts`.

silence8013:
## Summary

`host_exec`/`host_script` wrap every SSH command in a killable session via `setsid -w`, but the `-w`/`--wait` flag only exists in util-linux >= 2.24. On CentOS 7 / RHEL 7 (util-linux 2.23.2) and BusyBox, `setsid` aborts with `invalid option -- 'w'` and the command never runs. This PR removes the `-w` dependency while preserving the wait + real-exit-status and `pkill -s <sid>` reap semantics. Implements option #2 (pure-POSIX session wrapper) from the issue.

### Problem

Every command dispatched over SSH goes through `wrapBackgroundSession()` in `src/tools/infra/bg-session.ts`, which emitted `setsid -w sh -c '...'`. The `-w` flag (added in util-linux 2.24, 2013) is absent on older targets and BusyBox, so the wrapper fails immediately:

```
setsid: invalid option -- 'w'
```

Both the foreground and background paths of `host_exec`/`host_script` are affected, so basic operations (e.g. reading a file on a CentOS 7 node) fail. A stale code comment also claimed this dependency was "present on any normal SSH target incl. BusyBox", which is inaccurate.

### Solution

Drop `-w` and instead run `setsid sh -c '<launch>'; exit $?`. The trailing `exit $?` is a second command, so the launching shell cannot exec-optimize the lone `setsid` and must run it as a forked, non-process-group-leader child. In that case `setsid` exec's the program in place (no fork) on **every** setsid version, so the launching shell's `wait` still observes the real exit status, and `exit $?` propagates it. Session creation is unchanged, so `pkill -s <sid>` still reaps timeout's whole subtree on `job_stop`.

The fix is centralized in `bg-session.ts` and shared by `host_exec` / `host_script` / `node_exec`, with the rationale documented inline.

## Test Plan

- [x] `npx vitest run` for the affected suites — 32 tests pass (`bg-session`, `host-exec`, `host-exec-background`, `node-exec-foreground`, `host-script`)
- [x] Tests updated to assert `setsid sh -c` is emitted and `setsid -w` is absent
- [x] Manual: run a `host_exec` command against a CentOS 7 (util-linux 2.23.2) SSH target and confirm it executes and that `job_stop` still reaps the process tree

## Architecture Checklist

- [x] **Security model**: No new shell execution path; the change is confined to the existing `bg-session.ts` session wrapper shared by the cmd/script tools. No bypass of `command-sets.ts`.

## General Checklist

Closes #356

### DIFF
--- a/src/tools/infra/bg-session.test.ts
+++ b/src/tools/infra/bg-session.test.ts
@@ -24,7 +24,9 @@ describe("bg-session", () => {
 
   it("wraps a command as a setsid session that records its session id and cleans up", () => {
     const wrapped = wrapBackgroundSession("timeout 600 sh -c 'do work'", "/tmp/x.pgid");
-    expect(wrapped.startsWith("setsid -w sh -c ")).toBe(true);
+    expect(wrapped.startsWith("setsid sh -c ")).toBe(true);
+    expect(wrapped).not.toContain("setsid -w");            // no util-linux >= 2.24 dependency
+    expect(wrapped.endsWith("; exit $?")).toBe(true);      // forces a forked (non-leader) setsid child
     expect(wrapped).toContain("echo $$ > /tmp/x.pgid");   // record session id
     expect(wrapped).toContain("timeout 600 sh -c");        // inner command preserved
     expect(wrapped).toContain("rm -f /tmp/x.pgid");        // cleanup on normal exit
--- a/src/tools/infra/bg-session.test.ts
+++ b/src/tools/infra/bg-session.test.ts
@@ -14,6 +14,12 @@ const {
   killRemoteSessionViaSsh,
 } = await import("./bg-session.js");
 
+// NOTE: the assertions below check SHELL-STRING COMPOSITION only — they intentionally do not
+// exercise runtime semantics (real exit-status propagation, or the `pkill -s <sid>` reap chain),
+// which are the two guarantees this wrapper exists to provide. Those depend on `setsid(2)` /
+// process-group behavior that can't be observed without a real Linux shell, so they are verified
+// manually on a CentOS 7 / BusyBox target (see the PR's manual test plan) rather than in CI. A
+// green run of this file alone does not prove the runtime contract holds.
 describe("bg-session", () => {
   it("builds a unique, sanitized pidfile path", () => {
     const a = backgroundPgidFile("functions.host_exec:0");
--- a/src/tools/infra/bg-session.ts
+++ b/src/tools/infra/bg-session.ts
@@ -13,6 +13,14 @@ import { sshExec, type SshTarget } from "./ssh-client.js";
  * process group, so a single `kill -<pgid>` of the launcher's group misses timeout's subtree.
  * setsid starts a new session whose id every descendant inherits (across timeout's sub-group),
  * so `pkill -s <sid>` reaps them all.
+ *
+ * Why NOT `setsid -w`: the `-w`/`--wait` flag only exists in util-linux >= 2.24 (2013). CentOS 7 /
+ * RHEL 7 ship util-linux 2.23.2, and busybox `setsid` has no `-w` either, so `setsid -w` aborts
+ * with "setsid: invalid option -- 'w'" and the command never runs. We get the same "wait + real
+ * exit status" behavior portably by running setsid as a NON process-group-leader child (see
+ * {@link wrapBackgroundSession}): when its caller isn't a process-group leader, setsid exec's the
+ * program in place (no fork), so the launching shell's `wait` observes the real exit status on any
+ * setsid version. The session is still created, so `pkill -s <sid>` reaping is unchanged.
  */
 
 /** A unique pidfile path holding the setsid session id (`$$` of the leader) for one job. */
@@ -25,10 +33,22 @@ export function backgroundPgidFile(toolCallId: string): string {
  * Wrap `innerCmd` so it runs as its own session leader (setsid), records the session id to
  * `pgidFile`, and removes the pidfile on normal exit. Returns ONE remote-shell command string.
  * `echo $$` consumes no stdin, so a piped script body flows through to `innerCmd` unchanged.
+ *
+ * The wrapper is `setsid sh -c '<launch>'; exit $?` (NOT `setsid -w`): the trailing `exit $?` is a
+ * second command, so the launching shell can't exec-optimize the single `setsid` invocation and
+ * must run it as a forked CHILD. That child is never a process-group leader, so setsid exec's
+ * `sh -c '<launch>'` in place (no fork) on EVERY setsid version — incl. CentOS 7 / RHEL 7's
+ * util-linux 2.23.2 and busybox, neither of which support `setsid -w`. Because there's no fork,
+ * the launching shell `wait`s for the real command and `exit $?` propagates its true exit status.
+ * It runs synchronously (no `&`), so a script body piped on stdin still reaches `innerCmd` (an `&`
+ * async list would have its stdin redirected to /dev/null when job control is off).
+ *
+ * The inner `sh -c` records its own pid via `echo $$`; setsid has already made that pid the new
+ * session id, so `pkill -s <sid>` still reaps the whole tree (incl. timeout's own process group).
  */
 export function wrapBackgroundSession(innerCmd: string, pgidFile: string): string {
   const launch = `echo $$ > ${pgidFile}; ${innerCmd}; rc=$?; rm -f ${pgidFile}; exit $rc`;
-  return `setsid -w sh -c ${shellEscape(launch)}`;
+  return `setsid sh -c ${shellEscape(launch)}; exit $?`;
 }
 
 /**
--- a/src/tools/infra/bg-session.ts
+++ b/src/tools/infra/bg-session.ts
@@ -45,6 +45,16 @@ export function backgroundPgidFile(toolCallId: string): string {
  *
  * The inner `sh -c` records its own pid via `echo $$`; setsid has already made that pid the new
  * session id, so `pkill -s <sid>` still reaps the whole tree (incl. timeout's own process group).
+ *
+ * LOAD-BEARING CONTRACT — two assumptions this correctness depends on:
+ *   1. The launching shell must NOT exec-optimize a multi-command sequence. POSIX shells
+ *      (bash/dash/ash/busybox sh) run `<cmd>; exit $?` as a forked child, which is exactly what
+ *      keeps setsid a non-leader so it exec's in place. A shell that tail-call-optimized the
+ *      sequence would let setsid inherit leader status, fork internally, and — with no `-w` —
+ *      the parent would exit 0 immediately, silently dropping the real exit code.
+ *   2. Do NOT "simplify" this to `exec setsid …`. `exec` makes setsid inherit the launching
+ *      shell's process-group-leader status, which triggers the same internal fork + lost exit
+ *      code regression. The forked-child path (no `exec`) is required, not incidental.
  */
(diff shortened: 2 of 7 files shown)


src/tools/infra/bg-session.ts
```diff
@@ -25,10 +33,22 @@ export function backgroundPgidFile(toolCallId: string): string {
...
+ * It runs synchronously (no `&`), so a script body piped on stdin still reaches `innerCmd` (an `&`
+ * async list would have its stdin redirected to /dev/null when job control is off).
+ *
+ * The inner `sh -c` records its own pid via `echo $$`; setsid has already made that pid the new
+ * session id, so `pkill -s <sid>` still reaps the whole tree (incl. timeout's own process group).
  */
 export function wrapBackgroundSession(innerCmd: string, pgidFile: string): string {
   const launch = `echo $$ > ${pgidFile}; ${innerCmd}; rc=$?; rm -f ${pgidFile}; exit $rc`;
-  return `setsid -w sh -c ${shellEscape(launch)}`;
+  return `setsid sh -c ${shellEscape(launch)}; exit $?`;
```
chent1996, line 61: Correctness of this line hinges on two assumptions that the docstring above states but doesn't quite spell out as **load-bearing**:

1. **The launching POSIX shell does not exec-optimize a multi-command sequence.** bash/dash/ash/busybox sh all behave this way, but if anyone ever pipes this through a shell that DOES tail-call-optimize `<cmd>; exit $?`, `setsid` would inherit the shell's leader status, fall into the `getpgrp() == getpid()` branch, and fork — and without `-w` the parent would exit immediately with status 0, silently losing the real exit code.

2. **`exec setsid ...` is NOT a valid simplification here** — it would make `setsid` inherit the shell's leader status (same fork-and-lose-exit-code regression). Worth a one-liner saying so, since the natural "why not just `exec`?" follow-up question will come up.

A short addition to the docstring covering both — "requires a POSIX shell that does not exec-optimize multi-command sequences; do NOT replace with `exec setsid …`" — would lock the contract.

src/tools/infra/bg-session.test.ts
```diff
@@ -24,7 +24,9 @@ describe("bg-session", () => {
 
   it("wraps a command as a setsid session that records its session id and cleans up", () => {
     const wrapped = wrapBackgroundSession("timeout 600 sh -c 'do work'", "/tmp/x.pgid");
-    expect(wrapped.startsWith("setsid -w sh -c ")).toBe(true);
+    expect(wrapped.startsWith("setsid sh -c ")).toBe(true);
```
chent1996, line 33: Coverage gap, non-blocking: these assertions check string composition only — they cannot detect a regression in the actual exit-status propagation or in the `pkill -s <sid>` reap chain (the two things this wrapper exists to guarantee). Even a single integration-style test that spawns a real `bash -c` of the wrapped command against a fake `innerCmd`, asserts `$?` propagation, and probes `/proc/<pid>` for cleanup, would catch the failure modes the unit tests can't.

Not in scope for this PR if it's a larger refactor, but please at least call out in the docstring or PR description that the unit tests intentionally do not cover the runtime semantics, so a future contributor doesn't assume green CI means the contract holds.

chent1996 (changes requested):
The mechanism is sound — the trailing `; exit $?` reliably forces the launching shell to fork rather than exec-optimize, putting `setsid` in a non-pgroup-leader child where the `getpgrp() == getpid()` check fails and `setsid(1)` calls `setsid(2)` + `execvp` in place. With no internal fork, `-w` isn't needed: the launching shell's `wait` observes the real exit status of the exec'd `sh -c`, and `exit $?` propagates it. `pkill -s <sid>` reaping is structurally unchanged. This relies on `setsid(2)` Linux semantics, not on any util-linux version feature, so it should be portable as advertised.

That said, I'm holding back approval until two things are addressed:

**1. The CentOS 7 manual verification (the PR's whole reason for existing) is still unchecked.** Unit tests only assert string composition, which is the cheap part of the contract. The expensive part — "this actually runs on util-linux 2.23.2" and "`job_stop` still reaps `timeout`'s subtree end-to-end" — has no automated coverage in this codebase, and the manual checkbox under Test Plan is still empty. Please run a real command against a CentOS 7 (or BusyBox) SSH target, abort it mid-flight via `job_stop`, confirm both (a) the command actually executes (no `invalid option -- 'w'`) and (b) `ps` on the target shows no residual `timeout`/child after abort. Paste the output into the PR.

**2. Two small documentation gaps inline.** Details below — non-blocking individually, but worth fixing before merge so the next person reading this doesn't repeat the analysis from scratch.

Happy to flip to approve once the manual test result is posted.

silence8013:
@chent1996 
image

image

image


image

src/tools/infra/bg-session.ts
silence8013: Good call — locked the contract into the docstring in f101bb6. The `wrapBackgroundSession` doc now has a "LOAD-BEARING CONTRACT" section spelling out both assumptions explicitly:

1. the launching shell must NOT exec-optimize `<cmd>; exit $?` (POSIX shells fork the child, which is what keeps setsid a non-leader so it exec's in place); and
2. an explicit "do NOT simplify to `exec setsid …`" with the reason — `exec` makes setsid inherit leader status, triggering the same internal fork + lost-exit-code regression.

So the "why not just `exec`?" follow-up is now answered inline.

src/tools/infra/bg-session.test.ts
silence8013: Agreed on the coverage gap. I kept it in scope as the docstring/PR-description callout you suggested rather than a larger refactor: f101bb6 adds a note at the top of `bg-session.test.ts` stating these assertions cover shell-string composition only, and that the runtime exit-status propagation and `pkill -s <sid>` reap chain depend on real `setsid(2)`/process-group behavior that can't be observed without a Linux shell — so they're verified manually on a CentOS 7 / BusyBox target (per the PR's manual test plan), not in CI. That way a future contributor won't read green CI as proof the runtime contract holds.

(I prototyped a Linux-gated integration test that runs the real wrapped command, but dropped it to stay consistent with the repo's existing test approach — happy to revisit in a follow-up if you'd prefer real runtime coverage.)

chent1996 (approved):
All my prior concerns are addressed. Flipping to approve.

**Manual verification** (the blocking item): the 4 screenshots are exactly what was needed. They confirm in one chain:
1. Target is CentOS Linux 7, kernel 3.10.0-1127.el7 — the environment this PR exists to fix.
2. `lsblk --version` / `lscpu --version` both report `util-linux 2.23.2` — the version with no `-w` support.
3. `host_exec` runs to completion on that target (no more `invalid option -- 'w'`); `run_in_background: true` also works.
4. `ps aux` on the target shows the actual remote process tree matches the new wrapper exactly:
   ```
   bash -c setsid sh -c '…; timeout 620 sh -c '\''sleep 600'\''; …'; exit $?
   sh -c …; timeout 620 sh -c 'sleep 600'; …
   timeout 620 sh -c 'sleep 600'
   sh -c sleep 600
   sleep 600
   ```
   — parent bash forks setsid as a non-leader child (correct), setsid exec's `sh -c` in place (correct), `timeout` subtree is established under the new session id. After `job_stop`, `ps` shows only `grep` itself remains — the entire `setsid` / `timeout` / `sleep` subtree was reaped via `pkill -s <sid>`. The pgidfile was cleaned up too.

**Docstring fixes** (f101bb6): the `wrapBackgroundSession` doc now carries a clearly labelled "LOAD-BEARING CONTRACT" section spelling out both assumptions (no exec-optimize; do not simplify to `exec setsid …`). `bg-session.test.ts` got the requested note at the top explicitly stating the unit tests cover shell-string composition only and the runtime semantics are verified manually on CentOS 7 / BusyBox. Both lock the contracts I was worried would silently rot.

Good fix — clean root-cause solution that exploits `setsid(2)`'s built-in Linux semantics instead of probing versions or adding fallbacks, and now properly documented and empirically verified.

PR #372, opened 2026-06-30
feat(channels,metrics): attribute & filter channel audit by sender

jacoblee-io:
## Summary

Channel (Lark/DingTalk) sessions are now audited by the **actual sender** —
the siclaw-native `sender_external_id` (Lark `open_id` / DingTalk `staffId`)
scoped to its `channel_id` — never the binding owner. The Metrics admin UI can
filter and group channel Sessions/Tools by **sender + channel**, so an operator
can isolate one person's activity in a shared group bot.

## Why

A channel agent can be used by anyone in a group (open/PAIR-bound), and today
the audit collapses every sender into the binding owner — there was no way to
tell the senders apart. The sender's `open_id` is the stable "same person" key
siclaw already has (it keys per-sender sessions on `open_id:<sender>`).

**Dependency direction (the rule that shaped this):** SiCore depends on siclaw,
never the reverse. So siclaw has **no SiCore-user concept** — an earlier draft of
this branch added `sicore_user_id` / `ResolvedChannelBinding.senderUserId`; those
are all removed here (a reverse dependency). Mapping a Lark account to a SiCore
identity is solved separately on the SiCore side, out of scope for siclaw.

## What changed

- **Schema** (`migrate.ts`): `chat_sessions` carries `sender_external_id` +
  `channel_id`. `chat_messages` carries neither (redundant with its session;
  a channel session is per-sender). No `sicore_user_id` anywhere.
- **Channels** (`lark.ts`, `dingtalk.ts`, `chat-repo.ts`, `adapter.ts`): stamp
  the sender id + channel on the session; never fall back to the owner; NULL when
  the event omits the sender.
- **Metrics** (`metrics-entry.ts`, `siclaw-api.ts`): `actorUserColumn` =
  `CASE channel THEN sender_external_id ELSE user_id`. The channel-senders
  endpoint returns each sender with an occurrence count + last-seen, ordered by
  recency, so the opaque ids are recognizable.
- **Metrics UI** (`Metrics.tsx`, hooks, tables): the "who" axis is origin-aware —
  the portal-user ("All Users") filter shows/applies **only off-channel**; the
  channel entry hides it and uses the **sender (open_id)** filter + a channel
  picker. `web`/`api`/`a2a` attribution is unchanged.

Design doc: `docs/design/2026-06-29-channel-sender-audit-filter.md`.

## Not in scope

- No SiCore identity / account linking (SiCore's own layer).
- No sender display-name resolution (deferred; only worth it if the event already
  carries the name — no extra API/dependency).
- API-key "show-once" hardening is tracked separately.

## Testing

- `npx tsc --noEmit` clean (backend + portal-web).
- Full backend `npm test` green (7648 passed; the only failures are in a stale
  `.claude/worktrees/` copy, unrelated). portal-web vitest green.

## Deploy

Verified on siclaw-inner: `siclaw-portal:v0.2.3-chanuser4` +
`siclaw-runtime:v0.2.3-chanuser3` (runtime code byte-identical to this head).

### DIFF
--- a/src/portal/siclaw-api.ts
+++ b/src/portal/siclaw-api.ts
@@ -3044,9 +3044,14 @@ export function registerSiclawRoutes(router: RestRouter, config: SiclawConfig, c
     const conds: string[] = ["m.role = 'tool'", "m.created_at BETWEEN ? AND ?", msg.predicate];
     const params: unknown[] = [from, to];
     // Origin-aware actor: channel rows filter by the actual sender, never owner.
+    // channel_id / sender_external_id live on the PARENT channel session; a
+    // delegation child inherits them via parent_s (COALESCE) so its tool rows —
+    // which the entry predicate already counts under the parent's channel entry —
+    // don't vanish the moment a channel/sender filter is applied. (user_id IS set
+    // on the child, so the actor filter needs no parent fallback.)
     if (query.userId) { conds.push(`${actorUserColumn("s")} = ?`); params.push(query.userId); }
-    if (channelId) { conds.push("s.channel_id = ?"); params.push(channelId); }
-    if (query.senderExternalId) { conds.push("s.sender_external_id = ?"); params.push(query.senderExternalId); }
+    if (channelId) { conds.push("COALESCE(s.channel_id, parent_s.channel_id) = ?"); params.push(channelId); }
+    if (query.senderExternalId) { conds.push("COALESCE(s.sender_external_id, parent_s.sender_external_id) = ?"); params.push(query.senderExternalId); }
     if (query.toolName) { conds.push("m.tool_name = ?"); params.push(query.toolName); }
     if (query.outcome) { conds.push("m.outcome = ?"); params.push(query.outcome); }
     if (query.cursorTs && query.cursorId) {
@@ -3061,7 +3066,7 @@ export function registerSiclawRoutes(router: RestRouter, config: SiclawConfig, c
       `SELECT m.id, m.session_id AS sessionId, m.tool_name AS toolName,
               SUBSTR(m.tool_input, 1, 500) AS toolInput,
               m.outcome, m.duration_ms AS durationMs, m.created_at AS timestamp,
-              s.user_id AS userId, s.sender_external_id AS senderId, s.agent_id AS agentId, s.origin AS origin,
+              s.user_id AS userId, COALESCE(s.sender_external_id, parent_s.sender_external_id) AS senderId, s.agent_id AS agentId, s.origin AS origin,
               a.name AS agentName
        FROM chat_messages m
        LEFT JOIN chat_sessions s ON m.session_id = s.id
--- a/src/portal/siclaw-api.ts
+++ b/src/portal/siclaw-api.ts
@@ -54,7 +54,7 @@ import {
   normalizeChatSessionTitle,
   truncateChatSessionTitle,
 } from "./chat-session-fields.js";
-import { normalizeEntry, entrySessionPredicate, entryMessagePredicate } from "./metrics-entry.js";
+import { normalizeEntry, entrySessionPredicate, entryMessagePredicate, actorUserColumn } from "./metrics-entry.js";
 import { summariseLatency, extractTimingMs } from "./metrics-timing.js";
 
 /** Trace viewer message limit — matches siclaw_main.cron-limits.MAX_TRACE_MESSAGES */
@@ -2853,12 +2853,23 @@ export function registerSiclawRoutes(router: RestRouter, config: SiclawConfig, c
     // Message/tool-level: count a delegation child's rows under its PARENT's
     // entry so sub-agent tool calls stay auditable. msg.join adds the parent LEFT JOIN.
     const msg = entryMessagePredicate(entry);
+    // Channel sub-axis: narrow to one channel (only meaningful for entry=channel).
+    const channelId = entry === "channel" ? (query.channelId || null) : null;
+    // User filter is origin-aware: channel sessions group by the actual sender
+    // (sender_external_id), never the binding owner. The channelId filter is
+    // session-level. Appends to `params` in order.
+    const userChanCond = (alias: string, params: unknown[]): string => {
+      let s = "";
+      if (userFilter) { s += ` AND ${actorUserColumn(alias)} = ?`; params.push(userFilter); }
+      if (channelId) { s += ` AND ${alias ? alias + "." : ""}channel_id = ?`; params.push(channelId); }
+      return s;
+    };
 
     const db = getDb();
 
     const sessionParams: unknown[] = [from, to];
     let totalSessionsSql = `SELECT COUNT(*) AS c FROM chat_sessions WHERE created_at >= ? AND created_at <= ? AND ${tablePred}`;
-    if (userFilter) { totalSessionsSql += " AND user_id = ?"; sessionParams.push(userFilter); }
+    totalSessionsSql += userChanCond("", sessionParams);
     const [sRows] = await db.query(totalSessionsSql, sessionParams) as [Array<{ c: number }>, unknown];
     const totalSessions = Number(sRows[0]?.c ?? 0);
 
@@ -2869,15 +2880,15 @@ export function registerSiclawRoutes(router: RestRouter, config: SiclawConfig, c
       WHERE m.role = 'user' AND m.created_at >= ? AND m.created_at <= ?
         AND ${msg.predicate}
         AND (m.metadata IS NULL OR m.metadata NOT LIKE '%"kind":"delegation_event"%')`;
-    if (userFilter) { totalPromptsSql += " AND s.user_id = ?"; pParams.push(userFilter); }
+    totalPromptsSql += userChanCond("s", pParams);
     const [pRows] = await db.query(totalPromptsSql, pParams) as [Array<{ c: number }>, unknown];
     const totalPrompts = Number(pRows[0]?.c ?? 0);
 
     // ── distinctUsers (window) ──
     const duParams: unknown[] = [from, to];
-    let distinctUsersSql = `SELECT COUNT(DISTINCT user_id) AS c FROM chat_sessions
+    let distinctUsersSql = `SELECT COUNT(DISTINCT ${actorUserColumn("")}) AS c FROM chat_sessions
       WHERE created_at >= ? AND created_at <= ? AND ${tablePred}`;
-    if (userFilter) { distinctUsersSql += " AND user_id = ?"; duParams.push(userFilter); }
+    distinctUsersSql += userChanCond("", duParams);
     const [duRows] = await db.query(distinctUsersSql, duParams) as [Array<{ c: number }>, unknown];
     const distinctUsers = Number(duRows[0]?.c ?? 0);
 
@@ -2888,7 +2899,7 @@ export function registerSiclawRoutes(router: RestRouter, config: SiclawConfig, c
       ${msg.join}
       WHERE m.role = 'tool' AND m.created_at >= ? AND m.created_at <= ?
         AND ${msg.predicate}`;
-    if (userFilter) { toolCallsSql += " AND s.user_id = ?"; tcParams.push(userFilter); }
+    toolCallsSql += userChanCond("s", tcParams);
     const [tcRows] = await db.query(toolCallsSql, tcParams) as [Array<{ c: number }>, unknown];
     const toolCalls = Number(tcRows[0]?.c ?? 0);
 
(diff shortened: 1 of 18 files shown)


LikiosSedo (commented):
Reviewed end-to-end — design is solid, test coverage is good, and the dependency-direction discipline (no SiCore-user concept; `sicore_user_id` removed) is the right call. A few **non-blocking** notes; one of them is worth a real-device check before merge.

### 1. Verify the channel picker actually matches stamped `channel_id` (please test before merge)
The picker is populated from `/api/v1/channels`, whose `id` is the top-level `channels.id` (a `crypto.randomUUID()`). But the value stamped on the session comes from the gateway:
- `lark.ts:348-349`: `personalChannelId = personalBot.channel_id ?? channelId`, `groupChannelId = group_channel_id ?? channelId`

For a two-path channel (open group + dedicated-app personal bot), the session is stamped with the **group/personal sub-channel id**, not necessarily the parent `channels.id` shown in the picker. The filter only works if those sub-channel ids are themselves real `channels.id` rows that appear in `/channels`. Logic looks consistent, but this is exactly the Feishu two-path area that has bitten us before — please point the channel dropdown at a real two-path group on siclaw-inner and confirm the filter returns rows.

### 2. `actorUserColumn` doesn't scope the sender by `channel_id` — fine, but add a comment
`CASE WHEN origin='channel' THEN sender_external_id ELSE user_id END` groups/filters by `sender_external_id` alone. This is **semantically safe** (Lark `open_id` is per-app unique and DingTalk `staffId` has a different shape — distinct people never collide, and the same person across channels of the same app collapsing into one actor is actually correct). But a future reader may think the channel dimension was dropped by mistake. Worth a one-line note on `actorUserColumn` stating the sender namespace doesn't collide across origins, so channel scoping is intentionally left to the explicit `channel_id` filter.

### 3. NULL-sender channel sessions silently drop from sender-grouped views
When an event omits the sender, `sender_external_id` is NULL → `actorUserColumn` yields NULL → excluded from `COUNT(DISTINCT)` and from any sender filter. This is the documented design choice, so no change requested — just flagging that "filter by sender" is not the same as "all channel activity", which operators should understand.

Nothing here blocks the design; (1) is the only item I'd want confirmed on a real two-path channel before merging.

chent1996 (commented):
## Review

Strong, architecture-aware change. The headline decision — auditing channel sessions by the siclaw-native sender id (Lark `open_id` / DingTalk `staffId`) and **removing** the earlier draft's `sicore_user_id` / `senderUserId` — is the right call: it keeps the dependency direction intact (SiCore depends on siclaw, never the reverse), so no SiCore identity concept leaks into the runtime. The tests lock this in (a `sicore_user:` session key still audits by the raw `open_id`). The schema, channel, and metrics layers are consistent with the existing patterns.

### What I verified

- **Clean against latest `main`.** The branch is behind `main` by several commits and touches three files that `main` also moved (`migrate.ts`, `siclaw-api.ts`, `adapter.ts`), but a real trial merge onto current `main` is conflict-free — `migrate.ts` edits are in different regions (`model_entries` on main vs `chat_sessions` here).
- **Schema.** `sender_external_id VARCHAR(128)` + `channel_id CHAR(36)`, added both in the `chat_sessions` CREATE TABLE and via idempotent `safeAlterTable`, both nullable, no forbidden types (TIMESTAMP(3)/ON UPDATE/JSON) — dialect-safe for MySQL + SQLite. `channel_id CHAR(36)` correctly matches `channels.id CHAR(36)`. Keeping these off `chat_messages` (derivable from the per-sender session) is the right normalization.
- **Security.** Every new query is parameterized; the only interpolated fragment is `actorUserColumn(alias)` where `alias` is hardcoded by callers (`""` / `"s"`), never user input — no injection. `channel-senders` is `requireAdmin`-gated.
- **Centralization.** `actorUserColumn()` as the single source of truth for the "who" axis, applied uniformly via `userChanCond`, is a good DRY move — it removes ~10 scattered `s.user_id = ?` sites and makes attribution consistent.

### Findings

**1. The audit response `userId` field is now polymorphic (design, non-blocking).**
`audit/sessions` and `audit/timing` now project `actorUserColumn("s") AS userId`, so `userId` holds a portal user UUID for web/api/a2a but a Lark `open_id` for channel rows. A separate `channelId` field is added, but the actor is overloaded onto `userId` rather than a distinct `actorId`/`senderId`. The only consumer today is the origin-aware Metrics UI, so this is safe now — but it creates a field whose meaning depends on a sibling column (`origin`). Any future consumer that treats `userId` as a portal user id (joins `siclaw_users`, links a profile) will mishandle channel rows. Worth a conscious decision: keep overloading, or return owner `userId` + a separate `senderId`?

**2. `distinctUsers` mixes id namespaces and drops NULL senders (minor metric semantics).**
For `entry="all"`, `COUNT(DISTINCT actorUserColumn(""))` counts UUIDs and `open_id`s in one set. No collision in practice (disjoint formats), but channel sessions with a NULL sender are excluded from `distinctUsers` (COUNT DISTINCT skips NULL) while still counting in `totalSessions` — a small inconsistency between the two tiles. Acceptable, but worth a sentence in the design doc so it isn't read as a bug later.

**3. Redundant sender filters (trivia).**
In `audit/sessions`, both `query.userId` (via `actorUserColumn`) and `query.senderExternalId` filter channel rows by sender. Both parameterized and harmless; the comment explains `senderExternalId` is the explicit path. Just flagging the overlap.

### Process

Behind `main` (clean merge verified). Please rebase onto latest `main` before merge; `mergeStateStatus` is BLOCKED.

---

No blockers from me — findings 1–3 are design/clarity points, not correctness bugs. Not approving here; leaving as a comment. The one I'd most like an explicit answer on is finding 1 (overloading `userId`).

jacoblee-io:
Thanks both — addressed in `5fa73c0`.

**@chent1996 #1 (overloaded `userId`) — fixed.** This was the right call. Row payloads (`audit/sessions`, `metrics/audit`, snapshot) now return `userId` = the owner `user_id` (always a portal user) **plus a separate `senderId`** = `sender_external_id` (channel sender, null off-channel). Nothing is overloaded by `origin` anymore: a consumer joining `siclaw_users` on `userId` stays correct for every row, and the UI renders `senderId` for channel rows, `userId` otherwise. `actorUserColumn` is now used **only** for the actor filter + distinct-actor count, never projected.

**@LikiosSedo #2 (comment on `actorUserColumn`) — added.** Its doc now states the sender namespace doesn't collide across origins (Lark `open_id` per-app-unique, DingTalk `staffId` distinct shape), so channel scoping is intentionally left to the explicit `channel_id` filter, not folded into the expression.

**@chent1996 #2 (`distinctUsers` NULL-sender + mixed namespace) — documented.** Added a "Metric semantics" note to the design doc: NULL-sender channel sessions are excluded from `distinctUsers`/sender views but still counted in `totalSessions` (so the tiles can legitimately differ), and the `all`-entry distinct set mixes UUIDs + sender ids without collision. Behavior intentional, now written down.

**#3 from both (redundant sender filters) — acknowledged, left as-is.** Both paths are parameterized; `senderExternalId` is the explicit channel filter and `userId` is hidden on the channel entry, so they don't actually fire together from the UI. No correctness issue.

**@LikiosSedo #1 (two-path `channel_id` — verify before merge):** In **standalone** siclaw there's no `personal_bot` / `group_channel_id` (those are SiCore-config only), so `groupChannelId === channelId === channels.id` — the picker value and the stamped value are the same row, filter matches. The two-path case only arises in a **SiCore-configured** deployment, where the stats UI is SiCore's own dashboard, not this Portal. So there's nothing to verify on a standalone two-path channel because standalone can't produce one. If we later wire a SiCore-config two-path channel through this Portal, that's where the sub-channel-id↔`channels.id` assumption should be re-checked — flagging it for that integration rather than blocking this PR.

jacoblee-io:
Updated + rebased onto latest `main` (now includes #373). History tidied to two commits (feat + docs). Since the last review round:

- **Sender picker is now a proper combobox** (dropdown of senders — open_id/staffId with message/session counts + relative last-seen, recency-ordered — plus free input and a clear button), replacing the browser-inconsistent `<input list>`+`<datalist>`.
- **Fixed a mapping bug**: the audit/sessions, metrics/audit, and snapshot handlers re-map DB rows into an explicit shape and had dropped the new `senderId` (channel rows showed "—"). Now passed through all three; regression test added.
- Verified end-to-end locally with Playwright against seeded channel data: dropdown lists senders, selecting one filters the sessions, and rows render the sender open_id (owner is a separate field). tsc + suites green (290 tests in the touched areas).

src/portal/siclaw-api.ts
```diff
@@ -3029,9 +3040,13 @@ export function registerSiclawRoutes(router: RestRouter, config: SiclawConfig, c
...
     const entry = normalizeEntry(query.entry ?? query.source ?? query.origin);
     const msg = entryMessagePredicate(entry);
+    const channelId = entry === "channel" ? (query.channelId || null) : null;
     const conds: string[] = ["m.role = 'tool'", "m.created_at BETWEEN ? AND ?", msg.predicate];
     const params: unknown[] = [from, to];
-    if (query.userId) { conds.push("s.user_id = ?"); params.push(query.userId); }
+    // Origin-aware actor: channel rows filter by the actual sender, never owner.
+    if (query.userId) { conds.push(`${actorUserColumn("s")} = ?`); params.push(query.userId); }
+    if (channelId) { conds.push("s.channel_id = ?"); params.push(channelId); }
+    if (query.senderExternalId) { conds.push("s.sender_external_id = ?"); params.push(query.senderExternalId); }
```
LikiosSedo: **Delegation sub-agent tool calls drop out of sender/channel-filtered views.**

`entryMessagePredicate` inherits delegation rows: a sub-agent session has `origin='delegation'` with `parent_session_id` → the parent channel session, and the predicate counts its tool rows under the parent's entry (via the `parent_s` LEFT JOIN). So the unfiltered **channel** Tools tab *does* show these sub-agent tool calls.

But `sender_external_id` / `channel_id` are only stamped on the parent channel session — the delegation child session has them NULL. The filters here anchor on the child alias `s`:

```
actorUserColumn("s")          // child origin='delegation' → resolves to s.user_id, not the sender
s.channel_id = ?              // NULL on the child
s.sender_external_id = ?      // NULL on the child
```

Net effect: as soon as an operator picks a channel or a sender, the delegated sub-agent tool calls silently disappear from the filtered list — even though the unfiltered channel view includes them. That partially defeats the PR's goal of isolating one person's activity, since sub-agent tool calls are often the most audit-relevant rows.

Suggested fix: mirror the predicate's parent inheritance in the filters, e.g. `COALESCE(s.sender_external_id, parent_s.sender_external_id)` and `COALESCE(s.channel_id, parent_s.channel_id)`, and a parent-aware variant of `actorUserColumn`. The session-level endpoint (~line 3207) is unaffected — it uses the session-level predicate, so delegation children are never counted as channel sessions there.

jacoblee-io:
@LikiosSedo good catch — fixed in `504e3f7`.

The Tools entry predicate counts a delegation child's tool rows under the parent's channel entry (via `parent_s`), but `channel_id` / `sender_external_id` are only on the parent, so the child-anchored filters dropped those rows the moment a channel/sender was picked. The `channelId` and `senderExternalId` filters (and the projected `senderId`) now `COALESCE(s.x, parent_s.x)`, so delegation children inherit the parent channel session and stay in the filtered view — attributed to the inherited sender. `user_id` is set on the child, so the actor (`userId`) filter needs no fallback. The session-level endpoint is untouched (its predicate never counts delegation children as channel sessions). Regression test asserts the COALESCE fragments. tsc + suites green.

jacoblee-io:
Deep self-review pass (multi-angle) — one more real (latent) issue found and fixed in `00a9bde`:

**summary + timing had the same delegation-undercount that `metrics/audit` was just fixed for, minus the COALESCE.** All three run the same message-level (parent_s-joined) count queries, but summary/timing filtered channel on bare `s.channel_id`, so a channel filter would drop delegation-child rows there too. It's latent (the Dashboard never passes `channelId` today), but it diverged from audit and was a landmine. Fixed by **centralizing** the parent-aware column into `channelColExpr(col, alias, parentAlias?)` (metrics-entry.ts) and using it at every site — audit (replaces the inline COALESCE), summary (parent_s for message-level calls, bare for session-level), timing (always message-level). No site can silently forget the inheritance now; unit test covers both forms. Also cleared the sender filter when the channel changes in the UI.

Checked & clean: `parent_s` is always joined where referenced; param/placeholder order; migration dialect-safety; the removed `sicore_user_id`/`senderUserId`/message-level `sender_external_id` have zero lingering refs. Full portal + channels suites green (1391).

Remaining are quality nits, not correctness (left out to keep this focused): `relTime` duplicates an existing relative-time ladder used in ~4 other components; the 3 session-projection SELECT lists repeat the same columns. Happy to fold those into a follow-up.

LikiosSedo (approved):
LGTM — approving.

Re-reviewed at head \`00a9bde3\`. The delegation finding from the earlier inline comment is properly resolved: the parent-aware \`channelColExpr(col, alias, parentAlias?)\` helper centralizes the \`COALESCE(child, parent)\` form, and every message-level filter/projection (audit/tools, summary prompt/tool/skill/daily counts, timing) now inherits \`channel_id\` / \`sender_external_id\` from the parent channel session, so delegation sub-agent rows no longer vanish when a channel/sender filter is applied. Session-level endpoints correctly stay on the bare column. The frontend follow-through is right too (\`AuditTable\` displays on \`senderId\` presence rather than \`origin === "channel"\`; clearing the sender on channel change). Good unit + integration test coverage on the new helper.

Non-blocking follow-ups (fine to defer): no composite index on \`chat_sessions(channel_id, sender_external_id)\` for the channel-senders GROUP BY; multi-level nested delegation only inherits one parent level (a pre-existing limitation of the delegation predicate, not introduced here).

PR #373, opened 2026-06-30
feat(api-keys): owner-scoped visibility — admins see all, users see own

jacoblee-io:
## Summary

API key management was admin-only. This makes it **self-service** while keeping
admin oversight: a regular user can create and manage their own keys; an admin
still sees and controls every key.

## Changes (backend only — `src/portal/agent-api.ts`)

| Endpoint | Before | After |
|---|---|---|
| `GET .../api-keys` | `requireAdmin` → all keys | `requireAuth`; **admin → all** keys for the agent, **regular user → only their own** (`created_by = self`) |
| `POST .../api-keys` | `requireAdmin` | `requireAuth`; key stamped `created_by = caller` |
| `DELETE .../api-keys/:kid` | `requireAdmin` → any | `requireAuth`; **admin → any**, **regular user → only their own** (else `403`, before any DELETE is issued) |

Follows the existing "admin-or-self" pattern (see the password-reset route in
`auth.ts`).

## Not changed

- Auth verification is unaffected — it uses `key_hash`, never `key_plain`.
- `key_plain` stays viewable to its owner (and to admins). **No show-once** in
  this change (deliberate — discussed and deferred).
- No schema / migration / dependency changes.
- No frontend change needed: the API Keys tab is not admin-gated and `/agents`
  is reachable by regular users, so the panel works once the backend allows them.

## Testing

- `npx tsc --noEmit` clean. `agent-api.test.ts` green (47 tests), with new cases:
  admin lists all (no owner filter) / user lists only own (created_by filter) /
  user can create (created_by = caller) / user deletes own (200) / user cannot
  delete another's key (403, no DELETE issued).

### DIFF
--- a/src/portal/agent-api.test.ts
+++ b/src/portal/agent-api.test.ts
@@ -704,16 +704,7 @@ describe("registerAgentRoutes", () => {
   // ── API Keys ─────────────────────────────────────────────
   describe("API keys", () => {
     describe("GET /api/v1/siclaw/agents/:id/api-keys", () => {
-      it("requires admin", async () => {
-        const { status } = await runRoute(router, fakeReq({
-          url: "/api/v1/siclaw/agents/a1/api-keys",
-          method: "GET",
-          headers: { authorization: `Bearer ${USER_TOKEN}` },
-        }));
-        expect(status).toBe(403);
-      });
-
-      it("lists keys for admin", async () => {
+      it("admin lists all keys for the agent (no owner filter)", async () => {
         query.mockResolvedValueOnce([[{ id: "k1", name: "test" }], []]);
         const { status, body } = await runRoute(router, fakeReq({
           url: "/api/v1/siclaw/agents/a1/api-keys",
@@ -721,6 +712,23 @@ describe("registerAgentRoutes", () => {
         }));
         expect(status).toBe(200);
         expect(body.data).toHaveLength(1);
+        // admin → filter on agent only
+        expect(query.mock.calls[0][1]).toEqual(["a1"]);
+        expect(query.mock.calls[0][0]).not.toMatch(/created_by = \?/);
+      });
+
+      it("regular user lists only their own keys (created_by filter)", async () => {
+        query.mockResolvedValueOnce([[{ id: "k1", created_by: "u1" }], []]);
+        const { status, body } = await runRoute(router, fakeReq({
+          url: "/api/v1/siclaw/agents/a1/api-keys",
+          method: "GET",
+          headers: { authorization: `Bearer ${USER_TOKEN}` },
+        }));
+        expect(status).toBe(200);
+        expect(body.data).toHaveLength(1);
+        // user → scoped to (agent, self)
+        expect(query.mock.calls[0][0]).toMatch(/created_by = \?/);
+        expect(query.mock.calls[0][1]).toEqual(["a1", "u1"]);
       });
     });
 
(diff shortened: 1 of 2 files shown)


chent1996 (commented):
## Review

The change itself is clean, correct, and well-tested. My substantive comments are about the **architectural consequences** of widening key creation from admins to all authenticated users — not about the diff in isolation.

### The core change is sound

- **Authorization is fail-closed.** `list` collapses any non-admin (including a missing role) to owner-only via `ownerOnly = auth.role !== "admin"`; `delete` enforces `auth.role !== "admin" && created_by !== auth.userId → 403`, and the 403 is returned **before any DELETE is issued** (the `query.mock.calls` length assertion locks this in).
- **Ownership is stamped correctly.** `agent-api.ts:606` binds `created_by` to `auth.userId`, so swapping `requireAdmin → requireAuth` is sufficient to attribute the key to the caller.
- **Consistent with existing patterns** — `requireAuth` + manual `sendJson(res, 401)` matches `siclaw-api.ts`, and the admin-or-self check mirrors the password-reset route in `auth.ts`.

**This does not constitute a capability-scope escalation.** In the current model, any authenticated user can already drive any agent — `/chat/send` and `/run` are `requireAuth` with no user↔agent binding (no `agent_members` table, no owner column on `agents`, "all agents globally visible"). A user who can already run an agent's tools interactively isn't gaining new reach by minting a programmatic key for the same agent. The creation gate is simply being aligned with the already-open usage gate.

### But it widens two latent gaps into real exposure

**🔴 1. Account-decoupled, never-expiring credentials.** Two pre-existing facts compound once *every* user can create keys:

- `agent_api_keys.created_by` has **no FK to users** (only `fk_ak_agent → agents`), and the user-deletion path (`auth.ts:280`) does not clean up keys. A self-issued key therefore **keeps working after its issuer's account is deleted**.
- `expires_at DEFAULT NULL` + `body.expires_at || null` means a key created without an explicit expiry **never expires**, whereas the issuer's JWT expires in 24h (`auth.ts:59`).

Before this PR, key creation was admin-only (bounded population, controlled lifecycle). After it, any user can manufacture permanent credentials decoupled from their own account lifecycle. This is a real escalation along the *durability* axis. Worth a conscious decision before merge — e.g. (a) cascade-clean keys on user deletion, (b) enforce a default/max expiry for self-service keys, or (c) explicitly accept and document the risk. I'd lean (a)+(b).

**🟡 2. `key_plain` is stored and returned in cleartext.** `list` returns the full plaintext key, and the column stores it alongside the hash. This is pre-existing, but the PR makes every regular user able to read back their own keys' plaintext at any time, expanding the population whose queries return usable at-rest credentials. Show-once is deliberately deferred per the description — not blocking, just flagging that the cost of deferral scales with the user count this PR unlocks.

### Architectural note (non-blocking)

This introduces a partial `created_by` ownership notion for API keys while the agent itself stays globally usable by anyone. It's self-consistent for key-management UX, but it's effectively a patch over the deeper absence of per-user agent access control. A line on intent — is this the seed of per-user agent authorization, or a one-off? — would keep the model coherent.

### Minor

- `POST` doesn't verify the agent `:id` exists; a non-existent id surfaces as an FK violation (500) rather than a 404. Pre-existing, non-blocking.

### Process

The branch is **behind `main` by several commits**. `agent-api.ts` is untouched in the intervening commits, so there's no semantic conflict, but please rebase onto latest `main` before merge.

---

Not approving — leaving this as a comment. The main ask is a conscious decision on finding 1 (credential durability / account decoupling) before this merges.

LikiosSedo (approved):
LGTM. Verified the premise: agents are globally visible to any authenticated user (GET /api/v1/agents is requireAuth with no per-user filter), so self-service key creation grants no privilege beyond existing interactive access. `created_by` is already stamped to the caller on INSERT, and there is no `INSERT INTO api_key_service_accounts` anywhere in src, so no service-account escalation path. DELETE returns 403 before issuing any DELETE. Clean.

PR #374, opened 2026-06-30
fix(agentbox): decouple language-following from memory

LikiosSedo:
## Summary

Following the user's input language was accidentally gated on memory, so memory-off agents drifted to the model's default language. This drops the `&& isMemoryEnabled()` guard from the `[System: respond in X]` injection so every agent follows the user's language.

### Problem

The runtime injects its `[System: respond in X]` language directive only when memory is enabled, so **memory-off agents get no language enforcement** and drift to the model's default — e.g. a Chinese-biased model (DeepSeek) answers an English question in Chinese. Surfaced on the GPU-cloud product/sales agent, which is intentionally memory-off (no-PVC, ephemeral).

### Solution

`git blame` points at `5d9a3ed` ("fix(memory): disable siclaw memory by default"), which added `&& isMemoryEnabled()` to the **wrong** block:

```ts
-    if (detectedLang !== "English") {
+    if (detectedLang !== "English" && isMemoryEnabled()) {
```

Following the user's language is a baseline behaviour independent of long-term memory. The block that legitimately depends on memory is the `PROFILE.md` Language persistence below it — and that already no-ops via `fs.existsSync` when there's no memory dir.

- Inject the directive **unconditionally** (drop `&& isMemoryEnabled()`).
- Leave `PROFILE.md` persistence untouched (still effectively memory-scoped via `fs.existsSync`).
- Drop the now-unused `isMemoryEnabled` import in `http-server.ts`.

This is the root-cause fix; a portal-side workaround in `siclaw-gpu-cloud` (re-injecting the directive for memory-off agents) becomes redundant once a runtime image with this change ships.

## Test Plan

- [x] `npm test` passes (199 files, 4035 passed, 2 skipped)
- [x] New regression test: a **memory-off** agent still gets `[System: respond in Chinese]` prepended (guards against re-coupling); companion test asserts English input is left untouched with memory off.

## Architecture Checklist

- [x] **Deployment mode**: No resource sync / skills / filesystem-write changes. Pure in-memory prompt-text adjustment; identical in LocalSpawner and K8sSpawner.
- [x] **Security model**: No new shell execution paths.
- [x] **DB parity**: No schema changes.
- [x] **Tool protocol**: No tool changes.

## General Checklist

### DIFF
--- a/src/agentbox/http-server.test.ts
+++ b/src/agentbox/http-server.test.ts
@@ -16,6 +16,7 @@ import type https from "node:https";
 
 const mockConfigState = vi.hoisted(() => ({
   modelRouting: undefined as unknown,
+  memoryEnabled: true,
 }));
 
 // Silence metrics auth side effects.
@@ -53,7 +54,7 @@ vi.mock("../core/config.js", () => ({
     },
     modelRouting: mockConfigState.modelRouting,
   }),
-  isMemoryEnabled: () => true,
+  isMemoryEnabled: () => mockConfigState.memoryEnabled,
 }));
 
 // Make sync-handlers a no-op registry.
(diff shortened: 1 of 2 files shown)


jacoblee-io (approved):
LGTM. Correct root-cause fix: the `[System: respond in X]` directive was wrongly gated on `&& isMemoryEnabled()` (collateral from the memory-off-by-default change), while the block that legitimately depends on memory — PROFILE.md persistence — is independently scoped via `fs.existsSync`, so dropping the guard does not over-reach. Language-following is detection-based (not hardcoded to Chinese; English is the no-op default), and the regression tests cover memory-off Chinese (injected) + memory-off English (untouched). tsc + full suite green.

PR #375, opened 2026-06-30
fix(agentbox): strip injected language directive from user-message events (live + synced)

LikiosSedo:
## Problem

The agentbox prepends `[System: respond in <Lang>]` to the user's prompt so the model follows the input language (made unconditional in #374). pi-agent records that as the user turn — so the directive surfaces in the displayed **user** message bubble on portals that render the live event stream (observed on sicore: a user bubble showing `[System: respond in Chinese] 你好`).

## Root cause (corrected after live debugging)

This is a **live-render** leak, not a persistence one. Verified against the live sicore DB: every persisted user row is the clean original text — the directive is **never** stored (the gateway persists the pre-injection text). The leak path is:

- pi-agent re-emits the user turn (with the injected directive) as a `message_start`/`message_end` brain event;
- the agentbox forwards every brain event to the frontend via `writeEvent` (and to portals like sicore as `chat.event`);
- the frontend renders that event as the user's message → the directive shows. A page reload clears it, because the DB is clean.

## Fix

- **Primary (`http-server.ts`):** strip the directive from the first text block of any user-role message event in `writeEvent`, before it leaves the agentbox. The model already received the directive; consumers (live + persisted) now see only the clean user text.
- **Defence-in-depth:** also strip user-role content at the two outbound append chokepoints — `session.ts` `persistAppendMessage` (delegation sync) and `gateway/chat-repo.ts` `appendMessage` (chat.appendMessage). These are no-ops on the already-clean persisted path but guard any future code that hands a brain-recorded user turn to persistence.

Shared `stripLanguageDirective` only strips a leading directive (optionally after a single Deep-Investigation marker line), so a user typing the phrase mid-message is untouched. The model still receives the directive — unchanged, covered by the existing `http-server.test.ts` cases.

## Tests / verification

- [x] Unit tests for `stripLanguageDirective`.
- [x] `npx tsc --noEmit`; full `npm test` (4040 passed) on the persistence-strip commit.
- [x] Live DB inspection on `sendong-sicore-test` proving the directive is never persisted (DB clean) — which is what redirected the fix to the live event path.
- [x] Live re-test on `sendong-sicore-test` with the agentbox rebuilt from this branch — confirmed: the injected directive no longer appears in the user bubble.

### DIFF
--- a/src/agentbox/__tests__/strip-user-directive-event.test.ts
+++ b/src/agentbox/__tests__/strip-user-directive-event.test.ts
@@ -1,0 +1,44 @@
+import { describe, it, expect } from "vitest";
+import { stripUserDirectiveFromEvent } from "../http-server.js";
+
+// The injected `[System: respond in X]` directive must not surface in the user
+// message rendered/streamed to consumers. These cover the live brain-event shapes
+// the agentbox forwards via writeEvent.
+
+describe("stripUserDirectiveFromEvent", () => {
+  it("strips the directive from a user message with array content", () => {
+    const out = stripUserDirectiveFromEvent({
+      type: "message_end",
+      message: { role: "user", content: [{ type: "text", text: "[System: respond in Chinese]\n你好" }] },
+    }) as any;
+    expect(out.message.content[0].text).toBe("你好");
+    expect(out.type).toBe("message_end");
+  });
+
+  it("strips the directive from a user message with string content", () => {
+    const out = stripUserDirectiveFromEvent({
+      type: "message_start",
+      message: { role: "user", content: "[System: respond in Chinese]\n你好" },
+    }) as any;
+    expect(out.message.content).toBe("你好");
+  });
+
+  it("preserves a leading Deep-Investigation marker, stripping only the directive", () => {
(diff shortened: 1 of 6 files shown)


jacoblee-io (approved):
LGTM. Correct chokepoint: stripping in the agentbox `writeEvent` covers every live consumer — siclaw frontend and portals like sicore both read the same agentbox SSE (via the gateway sse-consumer), so the upstream strip fixes the sicore leak shown in the report. The two persistence-path strips are sound defence-in-depth (the gateway already persists the pre-injection text). `stripLanguageDirective` only removes a leading directive (optionally after one DP marker line), is idempotent, and leaves a user-typed mid-message occurrence intact.

Two non-blocking follow-ups:
1. Add a direct test for `stripUserDirectiveFromEvent` itself (esp. the array/multimodal content branch) — right now only the shared helper is unit-tested, so a regression in event-shape handling would slip through.
2. Please tick the live re-test box once confirmed on a sicore agentbox built from THIS branch (the fix was already re-scoped once from persistence to the live path, so worth confirming the rebuilt image is clean).

PR #377, opened 2026-07-01
fix(portal-web): stop auto-title writing to the wrong chat session

silence8013:
## Summary

Fixes a race in the web chat where a newly-created (empty) session gets titled with a *different* conversation's first message, so the Recent Sessions list shows titles that don't match their sessions.

### Problem

`GET /agents/:id/chat/sessions` returns rows like an empty session (`message_count: 0`) whose `title` actually belongs to another, real session. Tracing every backend title-write path shows this row can only be produced by the `PUT /chat/sessions/:sid { title }` rename endpoint (it's the sole path that sets a title without appending a message or bumping `last_active_at`). The only caller that renames using message-derived text is the frontend auto-title effect in `AgentChat.tsx`.

Root cause is a stale-messages race: on a session switch `activeSessionId` updates synchronously, but `pilot.messages` still holds the **previous** session's messages until `usePilotChat`'s async history load resolves. In that window the auto-title effect derives a title from the previous session's first message and `PUT`s it onto the newly-active empty session.

### Solution

Track `messagesSessionId` in lockstep with every wholesale `setMessages` in `usePilotChat` (clear / cache-restore / load success / load error), expose it, and guard the auto-title effect on `pilot.messagesSessionId === activeSessionId`. A title is now only derived once the loaded messages actually belong to the active session — no rendering/UX change, just closes the write-timing gap.

Note: pre-existing mis-titled rows in the DB are not touched by this code fix and need a separate data correction (empty sessions with a non-default title).

## Test Plan

- [x] `npm test` (portal-web `vitest run`) passes — 18 files, 172 tests
- [x] Manual: create a session with a conversation, then click New Session; the new session no longer inherits the previous session's title

---
_Supersedes #376, which was auto-closed after a force-push that rebased this branch onto `main` to drop unrelated commits._

### DIFF
--- a/portal-web/src/components/AgentChat.tsx
+++ b/portal-web/src/components/AgentChat.tsx
@@ -266,6 +266,13 @@ export function AgentChat({ agentId, selectedSessionId, onSessionChange }: Agent
   useEffect(() => {
     if (!activeSessionId) return
     if (titledSessionRef.current === activeSessionId) return
+    // Guard against the stale-messages race: on a session switch `activeSessionId`
+    // updates synchronously, but `pilot.messages` still holds the PREVIOUS session's
+    // messages until usePilotChat's async history load resolves. Without this check we
+    // would derive a title from the previous session's first message and PUT it onto the
+    // newly-active (often empty) session — producing a session whose title belongs to a
+    // different conversation. Only proceed once the loaded messages match the active session.
+    if (pilot.messagesSessionId !== activeSessionId) return
     const userMsg = pilot.messages.find((m) => m.role === "user")
     const assistantMsg = pilot.messages.find((m) => m.role === "assistant")
     if (!userMsg || !assistantMsg) return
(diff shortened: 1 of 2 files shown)


jacoblee-io (approved):
LGTM. Correct fix for the stale-messages race: guarding the auto-title effect on `messagesSessionId === activeSessionId` ensures the title is only derived once the loaded messages actually belong to the active session. Verified every wholesale `setMessages` in the session-switch effect is paired with `setMessagesSessionId`, and the other wholesale `setMessages(pilotMsgs)` sites (recovery / background_turn_done refetch / subagent poll) all refetch the CURRENT `sessionId` under `cancelled` guards, so leaving them unpaired is harmless — messages and messagesSessionId stay consistent. Incremental `setMessages(prev => ...)` appends keep the same session, so they correctly do not touch it. No legit titling path is blocked (new session: empty load sets messagesSessionId, first turn appends incrementally, guard passes).

Nit (non-blocking): the return-type JSDoc says messagesSessionId is updated on every session switch/**load**, but the same-session refetch loads do not update it (harmless today since they are same-session). Consider tightening the wording to "maintained on the session-switch path", or colocating setMessagesSessionId with setMessages in a small helper so a future refetch that loads a different session can not silently desync the guard.

PR #378, opened 2026-07-02
KB compile as a first-class capability: BoxProfile + generic capability.* control plane

LikiosSedo:
## Summary

This branch makes **knowledge-base compilation a first-class runtime capability** instead of a compile-specific bolt-on, by introducing two generic layers:

1. **`BoxProfile`** — a declarative descriptor (image, env forwarding, HOME, volumes, **allowedTools** trust envelope) replacing the `isCompile` special-casing in the spawner. Adding a capability = registering a profile; `getBoxProfile` is **fail-closed** (an unknown or empty profile throws rather than silently handing out the all-tools agent shape). The profile is part of the box reuse key: a profile change stops and respawns the pod instead of silently reusing the old shape.

2. **The `capability.*` control plane** — one capability-agnostic wire between a consumer (sicore today) and the runtime:
   - consumer → runtime: `capability.start` (runtime **mints** the run id) / `message` / `cancel`
   - runtime → consumer: `capability.event` (log/turn/summary/lifecycle), `persistTurn`, `persistArtifact` (knowledge content goes to the **consumer's** store), `fetchInput` (raw sources + `ref=workspace` rehydration bundle), and the opaque run-state store ops `persistRunState`/`getRun`/`listActiveRuns`.

Ownership (option B): the runtime stays **stateless** but **owns execution** — run lifecycle, watchdog, reconciliation — persisting each transition to the consumer's dumb store so in-flight runs survive restarts. The consumer keeps domain storage, governance, and frontend. The legacy `compile.*` control plane (`compile-driver.ts`, 13 RPCs, state-machine coupling) is deleted.

## Load-bearing invariants (review anchors)

- **`contract.ts` IS the wire.** Field names are the exact snake_case keys; producers/handlers build payloads through the contract types. The terminal-status set is a shared constant mirrored by the Go consumer (`done`/`failed`).
- **Terminal outcomes are sticky** (first outcome wins), the watchdog **re-checks the store before reaping** (a stale recover snapshot can resurrect an already-ended run), and terminal persists that failed are retried by the reconcile loop with the *original* outcome — a successful `done` can never degrade to `failed`.
- **Reaping stops the box** (`onReap`); recovery **re-attaches relays to surviving boxes** (`onAdopt`, queued box events replay) and leaves dead boxes to lazy respawn with **workspace rehydration** (`/sources` success = fresh box signal; a live box 409s and its on-disk state is never overwritten).
- **Tiered staleness:** `running` silent past 10m = wedged turn → `failed`; `idle` is a conversation at rest, kept warm 2h then closed as `done`.
- `capability.message` refuses unknown/terminal runs (after a store `adopt` attempt) instead of spawning unmanaged boxes.

## Verification

- `npm test`: full suite green (4000+); capability suite 32 tests (lifecycle regimes, reconcile/adopt races, tiered reaping, rehydration gating, wire-shape pinning).
- Live e2e on sicore-siflow-test at every phase: chat streaming + persistence, run reuse, refresh recovery, profile-mismatch respawn, DB flip to the generic store, endpoint smoke.
- An independent adversarial review of the final commits; all confirmed findings fixed in-branch (terminal stickiness, reap store re-check, reuse/convergence winner agreement).

## Companion & deploy notes

- Companion sicore MR: `feat/kb-capability-platform` on k8s/sicore (consumer adapter + domain storage + frontend cleanup).
- Deploy pairing: apiserver must include the capability consumer handlers (B1+). The SSE path rename (`compile-runs` → `capability-runs`) ships apiserver+web together.

## Deferred (tracked in the workspace review doc)

- ~~kb-test read-only consumer flow~~ — **transport landed in this PR** (`b817e6d`: capability.testStart/testMessage/testClose on the run's LIVE box, stateless test relay, C4 snapshot path-confinement hook). Consumer adapter + test UI + judge follow on the sicore side; live e2e pending deploy.
- ~~kbc legacy `/compile`-era endpoints removal~~ — **done in this PR** (see the Addendum below; the blocking WIP turned out to be the box's test-session test suite and is adopted + committed).
- Inline-base64 size ceiling / blob-ref variant for very large bundles.
- Checkpoint-based resume (contract fields reserved; restart continuity is provided by relay re-attach + workspace rehydration).

## Addendum (2026-07-02): the kbc box source moves into this repo (`kbc/`)

- `def936f` **feat(kbc)** imports the compile-box source (aiohttp box server, one-shot agent, headless eval tools, constitution + playbook — 35 files) from its previously remote-less standalone repo. This realizes the ownership decision this PR is about: the capability platform owns box + tools + runtime end to end, and box↔runtime contract changes land as atomic PRs with shared CI from here on. (Full history stays in the original repo + a zip backup.)
- The "in-flight WIP" that deferred the legacy cleanup turned out to be the box's **test-session test suite** (snapshot parity with publish, read-only driver enforcement, HTTP open/close/cap/teardown, test-message) — adopted and committed first, all green.
- `c6502f9` **chore(kbc)** deletes the dead legacy compile control plane from the box (C5): `POST /compile`, `POST /rulings`, `POST /compile-turn`, BOX_COMPILE_TASK + nudge consts, park machinery, bundle fallback (−365 lines). The `KBC_SMOKE` wiring driver now speaks the capability-era vocabulary (summary + candidate write + turn_done) — the old one blocked on `/rulings`, which would deadlock with the endpoint gone. Live surface unchanged: `/sources /authoring /session /message /events` + the four `/test-*` endpoints + `/health`; the runtime keeps its defensive no-op for a vestigial `parked` frame (rolling upgrades against old box images).
- `9379e7d` **ci(kbc)** adds a path-filtered workflow (`kbc-ci.yml`) running the box suite only when `kbc/**` changes (the claude-agent-sdk dependency vendors a ~200MB engine binary), plus a CLAUDE.md Change Impact Matrix row tying both sides of the box↔runtime contract together.
- Image build is unchanged: `docker build -f platform/pod/Dockerfile .` with build ctx `kbc/`; select via `SICLAW_COMPILE_BOX_IMAGE`. Box tests: 9/9 green locally (protocol smoke rewritten to the live vocabulary).


chent1996 (commented):
## Review

Large, well-designed, architecturally significant change — and notably disciplined for its size. Reviewed against latest `main` (the branch is up to date). Net-vs-`main` this is **additive + small edits** to the spawner/manager/client (0 whole-file deletions); the "legacy `compile.*` deleted" framing in the description is branch-internal history — `compile-driver.ts` was never on `main`, so there's no risk of removing a control plane production depends on.

### What I verified

**BoxProfile (Layer 1).** `getBoxProfile` is genuinely fail-closed — an unknown non-empty name throws rather than silently degrading to the all-tools `agent` shape (`box-profile.ts` `getBoxProfile`). The normal agent path is preserved byte-for-byte: `AGENT_PROFILE` carries no `image`/`envForward`/`home`/`volumes`, so `image = boxConfig.image ?? profile.image ?? this.config.image` and the `?? []` guards all fall back to the pre-PR shape (`k8s-spawner.ts`). Profile enters the reuse key at both layers (`k8s-spawner.ts` pod-label check + `manager.ts` warm-reuse guard), and a pod with no `boxType` label defaults to `"agent"` — so pods spawned before this PR reuse correctly. The `terminating`/`deletionTimestamp` check is a nice robustness bonus. Trust envelope is expressed purely via `allowedTools`.

**Capability control plane (Layer 2).** I checked each load-bearing invariant the PR advertises against the code — all CONFIRMED:
- Terminal sticky (first outcome wins): `run-manager.ts` `endRun` guards `if (isTerminalCapabilityStatus(rec.status)) return;` with a synchronous check-and-set, and a failed terminal persist is retained + retried by `flushTerminal` with the *original* outcome.
- Reap re-checks the store before acting (`reapStale` re-fetches `getRun` before `onReap`/`endRun`).
- `capability.start` mints the run id in the runtime (`startRun` → `mintRunId`; `CapabilityStartRequest` has no `run_id`), persist-before-register with `failFast`.
- `capability.message` refuses unknown/terminal runs after an `adopt` attempt, before any box spawn (`server.ts`).
- Tiered staleness: `staleMs=10m` → `failed`, `idleTtlMs=2h` → `done`.
- `onReap` stops the box before the terminal mark; `onAdopt` re-attaches the relay, and `recover()` is called *after* `ensureCapabilitySession` is defined.

**Import hygiene (`kbc/`).** No committed secrets, no vendored binary in git history (the ~200MB engine is pulled at build/CI time via the pip dep), sane `.gitignore`s, non-root Dockerfile, and `src/` never imports/execs `kbc/` — the box is a separate image consumed by tag (`SICLAW_COMPILE_BOX_IMAGE`) over the HTTP+SSE contract. The CLAUDE.md matrix row tying both sides of the box↔runtime contract is exactly the right harness update.

### Architectural impact (beyond the diff)

1. **Singleton runtime is now correctness-bearing, not just a WS-map convenience.** The helm comment captures it: with relay re-attach + a shared adapter secret, two runtime pods share one `runtime_id`, both adopt the same runs, and split the box's single-consumer `/events` queue → silently lost artifacts/turns. So the capability platform is fundamentally single-runtime until runs are partitioned by runtime identity. Well-documented inline, but worth stating as a standing architectural ceiling.
2. **`contract.ts` is a cross-language, cross-repo coupling.** The terminal-status set, snake_case keys, and event vocabulary must stay in lockstep with the Go consumer (sicore) with no compile-time link. The companion MR + matrix row help, but drift between the two sides is a standing maintenance risk (a status one side treats as terminal becomes an immortal zombie on the other).
3. **Coordinated multi-repo deploy.** apiserver must ship the capability consumer handlers (B1+), and the `compile-runs → capability-runs` SSE rename ships apiserver+web together. Ordering-sensitive.

### Findings

The final commits already went through an adversarial pass that hardened terminal stickiness and the reap store re-check. These are residual narrow windows in that same logic:

- **A. `reapStale` doesn't re-validate freshness after its async gap (correctness, narrow).** Candidates are snapshotted synchronously, then processed with `await`s (`getRun` → `onReap` → `endRun`). The re-check covers the store's *terminal status* but never re-reads `lastActivityMs`. If a `capability.message` lands in the `getRun` window (it `touch`es the run and sets it `running`), the store re-check sees non-terminal and the reap proceeds — killing the box mid-turn and marking the freshly-active run terminal. Fix is small: re-check `lastActivityMs` against the threshold after the store re-check. This sits right on the "tiered staleness" invariant.
- **B. Store-outage window in the reap re-check can overwrite `done` with `failed` (correctness, narrow).** The re-check's `catch {}` proceeds to reap when `getRun` throws. A resurrection artifact (a run that ended `done` in the store but was re-added as non-terminal by a stale `recover()` snapshot) that ages out during a store outage gets reaped as `failed`, retained, and later flushed — overwriting the store's `done`. That violates the advertised "a `done` can never degrade to `failed`" in that window. The inline comment assumes "the store row is non-terminal anyway," which isn't true for a resurrection artifact. Fix: on store-unreachable during the re-check, skip this reap and defer to the next tick rather than proceeding.
- **C. `lastActivityMs` is authoritative execution state that isn't persisted (informational).** `recover`/`adopt` reset it to `now()`, so a restart gives every recovered run a fresh staleness window (a run silent 9 min pre-restart gets a fresh 10 min). Fail-open and probably fine, but it's genuine authoritative state that doesn't survive restart, contrary to the strict "stateless runtime" framing.
- **D. `setSessionRef` has no production caller (cleanup).** `session_ref` is therefore always persisted as `""`; recovery keys off `runId` via `getAsync`, so nothing breaks — but the setter is dead weight and the `contract.ts` "Reserved — no writer yet" comment is inaccurate (there is a setter).
- **E. Brittle 409 detection in `materialize.ts` (robustness).** Distinguishing a live box from a real error via `msg.includes("failed: 409")` is fragile; the test pins the same string, so a wording change in the client would slip through. Prefer matching a status field over the message text.
- **F. `deleteNamespacedPod(...).catch(() => {})` swallows all delete errors (fail-fast).** In `k8s-spawner.ts` the stale/mismatch delete now ignores every error, not just 404 — a real API error (e.g. RBAC) degrades into a `waitForPodDeleted` timeout instead of a clear failure. Narrow the catch to not-found, or at least log.
- **G. CI hardening (nit).** `kbc-ci.yml` has no `permissions:` block (add least-privilege `contents: read`); actions, base image, and pip deps are unpinned.

### Positives

Fail-closed profile abstraction with the normal path untouched; sticky terminals + reconcile/`flushTerminal` convergence; correct `recover`/`adopt`/re-attach ordering; zero net deletions from `main`; decoupled vendored import; documented singleton and box↔runtime invariants; solid test coverage (the specific invariant tests exist).

---

No hard blockers. A and B land exactly on the two invariants this PR advertises and both have small fixes, so I'd like an explicit decision on them before merge; C–G are cleanup/hardening. Not approving — leaving this as a comment.

LikiosSedo:
Thanks for the deep review — all findings addressed across two follow-up commits. Ready for another pass.

**A — reapStale freshness re-check (fixed, b2469581).** Moved the freshness re-derive to run *after* the store re-check's `await`, via a shared `stalenessOutcome()` helper used by both the initial snapshot and the post-await re-check. A `capability.message` landing in the `getRun` window now bumps `lastActivityMs` and the run is spared instead of reaped mid-turn. Regression test added.

**B — store-outage in the reap re-check (fixed, b2469581).** On store-unreachable during the re-check we now `continue` (defer to the next tick) instead of proceeding. A resurrection artifact that is `done` in the store can no longer be overwritten to `failed` via `flushTerminal` while the store is down. Regression test added.

**E — brittle 409 detection (fixed, b2469581).** The agentbox client now attaches `err.status`; `materialize` branches on the structured status (409 = box-already-live) with the message substring only as a fallback. A client message reword can no longer silently turn a 409 into a hard error. Test covers the structured path with a reworded message.

**F — swallowed pod-delete errors (fixed, b2469581).** Dropped the blanket `.catch(() => {})` on the stale/mismatch delete; errors now reach the outer 404-only catch (rethrow on real failures), consistent with the CA-mismatch delete below it. A real API error surfaces instead of degrading into a `waitForPodDeleted` timeout.

**G — CI hardening (fixed, b2469581).** Added `permissions: contents: read` and pinned the runner. I intentionally did **not** pin the SDK version in CI on this branch, since its Dockerfile is also unpinned here — pinning only CI would make CI test a different version than the image. The SDK pin lands on the branch that also pins the Dockerfile.

**D — dead `setSessionRef` (removed, fdddd7ac).** Decision: we do **not** need box-session-resume in the current design, so I removed the dead setter rather than wiring it. Rationale: the runtime routes/recovers a run by `run_id` and restores continuity by rehydrating the workspace **files** into a box, never by resuming the box's Claude Code session id; the box also uses an in-memory session store, so a session id wouldn't survive a container restart anyway. `session_ref` stays a reserved wire field (adopt/recover carry any stored value through) so a future persistent-session design has a slot without a protocol change, and the contract comment is corrected to say so.

**C — `lastActivityMs` restart-reset (kept, documented).** Left as-is: this is the deliberate "stateless runtime" simplification and is fail-open. Persisting it would add authoritative state to a runtime we intentionally keep stateless; not worth it.

On the two standing architectural notes: both understood and treated as hard constraints — (1) the capability plane is single-runtime until runs are partitioned by runtime identity (won't scale replicas), and (2) `contract.ts` ↔ the Go consumer must stay in lockstep (terminal set / snake_case keys / event vocab change together).

`npm test` for the touched suites and `tsc --noEmit` are green.

src/gateway/capability/run-manager.ts
```diff
@@ -0,0 +1,399 @@
...
+
+  // No setSessionRef: the runtime routes/recovers a run by runId and restores
+  // continuity by rehydrating the workspace FILES into a box, never by resuming
+  // the box's Claude Code session id. The box also uses an in-memory session
+  // store, so a session id wouldn't survive a container restart anyway. session_ref
+  // stays a reserved wire field (adopt/recover carry it through) for a future
+  // where the box persists sessions — until then there is deliberately no writer.
+
+  /** Move a live run to idle/running (non-terminal) + persist. */
+  async setStatus(runId: string, status: CapabilityLifecycleStatus): Promise<void> {
```
jacoblee-io, line 192: **[severe] `setStatus` lacks the terminal guard `endRun` has; a status write across the `capability.message` await can clobber a sticky terminal outcome.**

`setStatus` sets `rec.status` unconditionally, unlike `endRun` (`:203`, `if (isTerminalCapabilityStatus(rec.status)) return`). The `capability.message` handler calls `setStatus(runId, "running")` at `server.ts:513` **after** `await client.postJson('/message/...')` (`:509`). During that await, the detached relay task (`session-driver` on a `done`/`error` frame) or a concurrent `capability.cancel` can drive the run terminal via `endRun`. If that terminal persist fails (WS blip), `endRun`'s catch deliberately keeps the record in the map for `flushTerminal` to retry — and `setStatus` then flips it back to `running`. `flushTerminal` (`:346`) only processes terminal records, so the true outcome is lost and the run becomes an immortal non-terminal store row, later reaped as `failed` after `staleMs` — degrading a real `done`/`error`. The stickiness test only covers `endRun`↔`endRun`.

**Fix:** add the same `isTerminalCapabilityStatus` guard to `setStatus`.

src/gateway/capability/session-driver.ts
```diff
@@ -0,0 +1,131 @@
...
+        break;
+      case "done":
+        emit("lifecycle", { status: "done" });
+        await manager.endRun(runId, "done");
+        break;
+      case "error":
+        emit("lifecycle", { status: "failed", error: evt.error ?? "" });
+        await manager.endRun(runId, "failed");
+        break;
+      case "end":
```
jacoblee-io: **[severe] A clean box `end` never lands a terminal state; the run sits `idle` and the watchdog later mislabels it `done`.**

When the kbc box session ends cleanly (max_turns reached / SDK close) it emits `end` with no error and does not drop its `RUNS` entry. Here `case "end"` falls to `default: break` — neither `endRun` nor `setStatus` — so `driveCapabilitySession` resolves normally and `server.ts` only deletes the in-memory session map; `endRun` is never called. The run stays `idle`: subsequent `capability.message` hits `already_live`/409, and the idle-tier watchdog (`run-manager.ts:259`, idle past `idleTtlMs` → `done`) closes it as **`done`** ~2h later — persisting a dead/exhausted session as a success. The legacy single-shot compile always drove to an explicit terminal state.

**Fix:** treat a clean `end` without a prior `done`/`error` as an explicit terminal outcome (or have the box emit `done`/`error` before `end`).

src/gateway/capability/session-driver.ts
```diff
@@ -0,0 +1,131 @@
...
+        // at a time. Content is inline base64 (opaque to the transport). A failed
+        // persist must not fail the run (a transient WS blip would kill a healthy
+        // compile), but it IS real potential loss — the box's sync dedups by
+        // content hash, so this exact version is only re-sent if the file changes
+        // again. Hence one retry, then a loud log.
+        for (const a of evt.artifacts ?? []) {
+          const artifact: CapabilityPersistArtifactRequest = {
+            run_id: runId,
+            path: a.path,
+            content: { inline_base64: Buffer.from(a.content, "utf8").toString("base64") },
```
jacoblee-io: **[severe] Artifact sink is add/update-only; deleted or renamed KB pages leave orphan rows in the consumer store.**

`persistArtifact` / `CapabilityPersistArtifactRequest` is purely additive per-path (`{run_id, path, content}`) — no tombstone, no full-set/manifest field. The box's `_sync_workspace` only emits *changed* files and never prunes; a removed/renamed candidate page just stops appearing, so the consumer never learns to delete it and can't diff (it never receives the authoritative set). The deleted `compile.done` shipped a full `bundle_b64` tar snapshot that replaced the candidate set atomically. Net: renamed/removed pages become orphan rows in the published KB — a data-correctness regression.

**Fix:** carry a per-run authoritative path set (or a delete/tombstone) so the consumer can reconcile removals.

kbc/platform/pod/compile_box.py
```diff
@@ -0,0 +1,1064 @@
...
+    buildPublishBundleFromCandidates so the test reads BYTE-IDENTICALLY to what a
+    publish would serve. Returns (sha256 over sorted relpath+content, page_count).
+    Raises FileNotFoundError if there are no candidate pages or no root index.md."""
+    candidate = Path(workdir) / "candidate"
+    kdir = dest / ".siclaw" / "knowledge"
+    kdir.mkdir(parents=True, exist_ok=True)
+    h = hashlib.sha256()
+    count = 0
+    has_index = False
+    for f in sorted(candidate.rglob("*")) if candidate.is_dir() else []:
```
jacoblee-io, line 209: **[severe / security] Snapshot pinner follows symlinks; a compile agent can bake out-of-workspace file content into the read-only test snapshot.**

`_pack_candidates_to_wiki` walks `candidate.rglob("*")` + `is_file()` + `read_bytes()` with only a `".." in rel.parts` guard — no `is_symlink()`/`realpath` confinement. The compile session allows `Write`+`Bash`, so the agent can `ln -s /etc/passwd /work/candidate/leak.md` (relative, no `..`, `is_file()` true) and the pinner copies the target's content into the immutable snapshot, readable via the read-only test session. `_test_path_escape` only guards the test-session inputs (Read/Glob/Grep) and never sees the pin step, so C4 does not cover this vector.

**Fix:** skip `f.is_symlink()` (or realpath-confine each file under `candidate/`) before `read_bytes()`.

src/gateway/agentbox/box-profile.ts
```diff
@@ -0,0 +1,110 @@
...
+   */
+  envForward?: string[];
+  /** HOME override (e.g. "/work" when rootfs is read-only and the image HOME isn't writable). */
+  home?: string;
+  /** Extra writable emptyDir volumes. */
+  volumes?: BoxProfileVolume[];
+  /** Tool/trust profile: allowed tool names (null = all). Enforced end-to-end in A.3. */
+  allowedTools?: string[] | null;
+  /** Resource requests/limits override. */
+  resources?: { cpu?: string; memory?: string };
```
jacoblee-io, line 42: **`BoxProfile.resources` is a dead field; setting it silently does nothing and the pod can OOM.**

`resources?: {cpu?, memory?}` is declared but read nowhere in `src/`; `k8s-spawner.ts` sizes pods only from `boxConfig.resources` (default limit 2000m/4Gi). The spawner reads `profile.volumes` at the same site but never `profile.resources`. A memory-hungry profile that sets `resources` gets default limits and can OOM.

**Fix:** either delete the field or wire it as a fallback (`boxConfig.resources ?? profile.resources`).

src/gateway/capability/run-manager.ts
```diff
@@ -0,0 +1,399 @@
...
+  private watchdogTimer?: ReturnType<typeof setInterval>;
+  private reconciling = false;
+
+  constructor(
+    private readonly backend: RunStateBackend,
+    opts: CapabilityRunManagerOptions = {},
+  ) {
+    this.now = opts.now ?? Date.now;
+    this.staleMs = opts.staleMs ?? 10 * 60_000; // 10 min
+    this.idleTtlMs = opts.idleTtlMs ?? 2 * 60 * 60_000; // 2 h
```
jacoblee-io, line 100: **Abandoned capability box has only a hardcoded 2h reaper; no per-agent idle window is honored.**

`idleTtlMs` (2h) and `staleMs` (10m) are hardcoded and nothing forwards a per-agent idle setting into the manager. The kbc box has no idle self-destruct of its own, and the relay holds the box's `/events/:runId` SSE open for the run's whole life, so the manager's 2h `idleTtlMs` (via `onReap`→stop) is the *only* thing that reclaims an abandoned box — it can occupy a pod for up to ~2h. Consider making the idle window configurable / forwarded from the agent or profile.

src/gateway/capability/run-manager.ts
```diff
@@ -0,0 +1,399 @@
...
+      // there, just forget it (no box stop, and NEVER overwrite a successful
+      // "done" with "failed").
+      try {
+        const req: CapabilityGetRunRequest = { run_id: runId };
+        const row = (await this.backend.request(CAPABILITY_GET_RUN, req)) as CapabilityRunRow | null;
+        if (row?.status && isTerminalCapabilityStatus(row.status)) {
+          this.runs.delete(runId);
+          continue;
+        }
+      } catch {
```
jacoblee-io, line 313: **`reapStale`'s bare `catch { continue }` defers forever on a permanent per-run `getRun` error.**

A thrown `getRun` skips reaping with no give-up/backoff. The common case (store outage, or an unknown `run_id` returning `null`) is handled correctly, but a permanent per-run throw (e.g. a poison row that never deserializes) would defer on every tick and never `onReap` — a wedged box runs indefinitely. Low likelihood; worth a bounded give-up counter so a single poisoned run cannot pin a pod forever.

src/gateway/server.ts
```diff
@@ -356,6 +372,237 @@ export async function startRuntime(opts: StartRuntimeOptions): Promise<RuntimeSe
...
+    const profile = req.profile?.trim();
+    if (!profile) throw new Error("profile is required");
+    // Fail-closed BEFORE minting the run: an unknown profile must never fall
+    // back to some other box shape (that would hand out the wrong tool/trust
+    // envelope), and we don't persist runs we can't spawn.
+    getBoxProfile(profile);
+    const orgId = req.org_id;
+    const instruction = req.input?.instruction as string | undefined;
+    // siclaw mints the runId + persists a running state (the run is siclaw-owned).
+    const rec = await capabilityRunManager.startRun({ profile, orgId: orgId ?? "", correlationId: req.correlation_id });
```
jacoblee-io, line 480: **`runtime_id` is never written; cross-runtime reap hazard if the gateway is ever scaled past one replica.**

`startRun` omits `runtimeId`, so `runtime_id` is always `""`, and `recover()`'s `listActiveRuns({})` adopts every active row with no owner filter — a second replica would adopt and reap another's runs. This is bounded today by the ⚠️ CORRECTNESS-BEARING `replicas: 1` note in `gateway-deployment.yaml`, so it is an acknowledged constraint, not a live bug — but the reserved `runtime_id` field should be wired (with run partitioning) before that note is ever relaxed.

jacoblee-io (commented):
## Review: KB compile as a first-class capability

Solid architectural direction — `BoxProfile` cleanly generalizes the old `isCompile` special-casing, and the `capability.*` plane's core invariants (terminal stickiness, watchdog re-check, reconcile retry) hold as designed; the prior adversarial-review fixes check out. The findings below are the gaps that survived verification: 4 confirmed correctness/data/security defects, then lower-severity and altitude notes. Inline comments carry the specifics.

**Altitude / conventions (not inline):**

- **The "capability-agnostic" wire isn't actually capability-agnostic.** `CapabilityTestStartResponse.snapshot_hash`/`pages`, the `/sources`+`/authoring` semantics, and the `candidate/*.md` artifact shape are all KB-document-specific; the "generic" driver switch matches the kbc box's private SSE vocabulary (`summary`/`turn_done`/`syncArtifacts`/`parked`) via an untyped local `interface`, so a second capability's events fall through `default` and vanish; `persistArtifact` hard-codes `Buffer.from(content, "utf8")`, baking markdown-text into the format-neutral `inline_base64`. A second capability would fork the contract.
- **The wire contract is enforced in one direction only.** `contract.ts` says handlers MUST read payloads through the types, but every inbound `capability.*` handler uses `params as unknown as X` — a renamed/camelCase field from the Go consumer reads as a zero value with no compile error (the exact silent-drop the rule warns against).
- **CLAUDE.md Change Impact Matrix gap.** The PR adds the whole `src/gateway/capability/*` + `box-profile.ts` control plane (with documented deploy-ordering / single-replica correctness hazards) but the only new matrix row covers `kbc/platform/pod/*`. The Development Protocol requires a row for new high-blast-radius surfaces.
- **`runId` overloaded as the AgentBoxManager `agentId`** (per-pod reuse key): every run mints a unique pod identity that warm-reuse can never hit, the profile-mismatch stop+respawn is dead code for capability runs, and capability pods pollute agent-lifecycle bookkeeping. A per-run job wants the job-registry / debug-pod refcount model, not the per-agent singleton key.
- **Dead code:** `session-driver` `case "parked"` (self-described vestigial, no producer), `CapabilityRunRecord.sessionRef` (no writer; the reserved contract field already carries it), and a stale `contract.ts` comment ("runtime only calls persist + listActiveRuns" — adopt/reapStale both call getRun).

LikiosSedo:
Thanks — sharp review, especially the snapshot-pinner symlink vector. The four correctness/security/cleanup items are fixed in 64d68e9c; the altitude notes I've triaged below.

**Fixed (64d68e9c):**

- **[security] Snapshot pinner followed symlinks.** `_pack_candidates_to_wiki` now resolves each file's real path and skips anything that escapes `candidate/` — covers both a file symlink (`ln -s /etc/passwd candidate/leak.md`) and a symlinked directory. Test added with both vectors asserting no host content reaches the snapshot.
- **reapStale defers forever on a poison row.** Added a bounded give-up: after `MAX_REAP_DEFERRALS` (5) consecutive failed `getRun` re-checks the reap proceeds, so a permanently-throwing row can't pin its box. The counter clears on a successful re-check or on `touch()` (real activity), so a transient store outage never accrues toward the cap. Two tests.
- **`BoxProfile.resources` dead field → OOM.** Wired as `boxConfig.resources ?? profile.resources`, same precedence as `image`/`volumes`.
- **Dead code.** Removed the `session-driver` `parked` case (a stray one now falls through harmlessly — test asserts it), removed the `CapabilityRunRecord.sessionRef` mirror (no writer; `session_ref` stays a reserved wire field persisted as `""`), and corrected the stale `contract.ts` comment (`getRun` is called by `adopt` + the reap re-check).

**Altitude notes — triaged, not in this commit:**

- **The wire isn't truly capability-agnostic (accepted; framing corrected in follow-up).** You're right — `snapshot_hash`/`pages`, `/sources`+`/authoring`, `candidate/*.md`, the untyped SSE-vocabulary switch, and the `utf8` `persistArtifact` are all KB-specific; a second capability would fork the contract. I'll narrow the doc framing to "KB-shaped capability plane" rather than claim genericity, and treat true multi-capability generalization as the explicit work it is (typed event vocabulary + format-neutral artifact) when a second consumer actually lands. Not blocking this PR.
- **Inbound `params as unknown as X` (one-way contract).** Real silent-drop risk vs the Go consumer. The right fix is boundary validation on every inbound `capability.*` handler; that's a focused follow-up (backlog), tracked with the `contract.ts` lockstep constraint.
- **`runId` used as the `agentId` reuse key.** Agreed it's a per-agent singleton model applied to per-run work — warm-reuse never hits and the profile-mismatch respawn is dead for capability runs. Reworking onto the job-registry / debug-pod refcount model is a larger refactor; backlog.
- **`runtime_id` unwired / single-replica.** Acknowledged and already fenced by the ⚠️ CORRECTNESS-BEARING `replicas: 1`; wiring `runtime_id` + run partitioning is a prerequisite before that note is ever relaxed, tracked with it.
- **Idle window hardcoded 2h.** Reasonable to forward from the agent/profile; backlog enhancement.
- **CLAUDE.md Change Impact Matrix row.** Fair — I'll add a row for `src/gateway/capability/*` + `box-profile.ts` (deploy-ordering + single-replica hazards) in the follow-up.

jacoblee-io:
Thanks for the fast turnaround — confirmed at `64d68e9`: the symlink pin confinement, the `reapStale` bounded give-up, `profile.resources` fallback, and the dead-code removals all look correct and are covered by the new tests. No regressions in that commit.

One gap before this is mergeable, though: **the three findings I ranked highest are neither in `64d68e9` nor in the triage list above.** They aren't altitude notes — they're data/correctness defects, and I've re-confirmed all three are still present at `64d68e9`. Flagging so they don't fall through the crack between "fixed" and "backlog":

**1. `setStatus` lacks the terminal-sticky guard (`run-manager.ts:192`) — one line.**
`endRun` guards with `if (isTerminalCapabilityStatus(rec.status)) return` (`:203`); `setStatus` writes `rec.status` unconditionally. `capability.message` calls `setStatus(runId, "running")` *after* `await postJson('/message/...')`, so a relay `done`/`error` or a `capability.cancel` landing in that window (whose terminal persist failed and is being retained for `flushTerminal` retry) gets flipped back to non-terminal — the sticky outcome is lost and the run becomes an immortal store row, later reaped as `failed`. Mirror the guard:
```ts
async setStatus(runId, status) {
  const rec = this.runs.get(runId);
  if (!rec) return;
  if (isTerminalCapabilityStatus(rec.status)) return; // terminal is sticky — mirror endRun
  rec.status = status;
  ...
}
```

**2. A clean box `end` never terminalizes → run lingers `idle`, watchdog mislabels it `done` (`session-driver.ts`, `end` falls to `default`).**
On a clean stream close / `max_turns` exhaustion the box emits `end` with no error and doesn't drop its `RUNS` entry; the driver neither `endRun`s nor `setStatus`es, so the run sits `idle`, the next `capability.message` hits `already_live`/409, and the idle-tier watchdog closes it as **`done`** ~2h later — persisting a dead/exhausted session as success. Stopgap (same switch you just edited to drop `parked`):
```ts
case "end":
  // Clean close / max_turns: the box sends no done/error. Terminalize so the run
  // doesn't linger idle and get mislabeled `done` by the 2h watchdog. Sticky, so a
  // prior done/error still wins.
  await manager.endRun(runId, "done");
  break;
```
Cleaner fix is box-side: have `_run_wrapper` emit an explicit `done`/`error` *before* `end`, so the outcome reflects why the session ended rather than defaulting to `done`.

**3. Artifact sink is add/update-only — deleted/renamed KB pages leave orphan rows (`session-driver.ts:94` + `CapabilityPersistArtifactRequest`).**
Purely additive per-path, no tombstone / authoritative-set field; `_sync_workspace` emits only *changed* files and never prunes, so the consumer never learns to delete a removed page and can't diff (it never sees the full set). The deleted `compile.done` shipped a full bundle that replaced the candidate set atomically — this generalization dropped that. Needs a protocol addition (a per-run authoritative path set, a delete/tombstone, or a full-set manifest on `done`), not a one-liner.

#1 and #2 are cheap; #3 is the real design item. Happy to open a PR against the branch with #1 + #2 if that's useful. The altitude triage above all looks reasonable to defer.

kbc/platform/pod/office_ingest.py
```diff
@@ -0,0 +1,128 @@
...
+
+def xlsx_to_md(path: Path) -> str:
+    from openpyxl import load_workbook
+
+    out: list[str] = []
+    wb = load_workbook(str(path), data_only=True, read_only=True)
+    try:
+        for ws in wb.worksheets:
+            out.append(f"## Sheet: {ws.title}")
+            rows = [["" if c is None else str(c) for c in row] for row in ws.iter_rows(values_only=True)]
```
jacoblee-io, line 57: **No bound on parsed size / row count — a large or zip-bomb office file in a source bundle can OOM the box.**

`.pptx/.xlsx/.docx` are zip archives, so the on-disk `max_unpacked_bytes` cap in the `/sources` tar extraction doesn't bound what these parsers expand in memory. `xlsx_to_md` materializes **every** row into a list here (`rows = [... for row in ws.iter_rows(...)]`), defeating the `read_only=True` streaming, and `pptx_to_md`/`docx_to_md` load the whole document tree. A crafted decompression bomb — or just a legitimately huge spreadsheet — expands unbounded; the per-file `except Exception` catches parse errors but not an OOM-kill.

Blast radius is one run's own box (one pod per run), and the source comes from the KB author/uploader, so this is DoS-hardening rather than escalation — but worth a cap. Note `kbc/tools/ingest.py` in this same PR already bounds this with `MAX_TABLE_ROWS = 200`; mirroring that in `xlsx_to_md` (and a row/element ceiling for the others) is a cheap fix.

kbc/platform/pod/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+#     kbc-compile-pod
+FROM python:3.11-slim
+
+RUN apt-get update && apt-get install -y --no-install-recommends git ca-certificates \
+    && rm -rf /var/lib/apt/lists/*
+
+# claude-agent-sdk + aiohttp = the box runtime; python-pptx/openpyxl/python-docx
+# let office_ingest pre-render .pptx/.xlsx/.docx sources to markdown (pdf/text/
+# images are read natively by the agent's Read tool, so no parser is needed there).
+RUN pip install --no-cache-dir claude-agent-sdk aiohttp python-pptx openpyxl python-docx
```
jacoblee-io, line 20: **Office deps are unpinned.**

`pip install ... python-pptx openpyxl python-docx` with no version pins makes the box image non-reproducible and means the XML-parser hardening (XXE posture of the zip/XML backends) is whatever the latest release ships on build day. Modern versions mitigate XXE so this is informational, but pinning `==` versions (as elsewhere) removes both the reproducibility and the drifting-security-posture concern.

jacoblee-io (dismissed):
## Approving — all correctness/security findings resolved

Re-reviewed through `1db5584`. The three data/correctness defects I ranked highest are fixed correctly and with tests (`1a2fc64`): `setStatus` now mirrors `endRun`'s terminal-sticky guard; a bare clean box `end` terminalizes the run (runtime `case "end"` + box-side explicit `done`→`end`, sticky, lifecycle-frame deduped, old-image-safe); and workspace sync now emits deletion tombstones (box judges deletion by `is_file()` not collection-absence, `sent`-scoped, safe iteration). The earlier batch (`64d68e9`) — symlink-pin confinement, `reapStale` bounded give-up, `profile.resources` fallback, dead-code removal — also verified. The Forgejo removal has no dangling references, and the TS changes since are purely cosmetic (comment genericization).

**Approving on the code.** Two minor, non-blocking hardening notes on the new `office_ingest` inline.

**Merge/deploy precondition (operational, not a code blocker):** the PR's own cross-repo deploy ordering must hold — the consumer side (`feat/kb-capability-platform`: capability handlers B1+, the `compile-runs`→`capability-runs` SSE rename with apiserver+web) and, for the tombstones, `feat/kb-artifact-tombstone` must land **before** this box/runtime rolls out. An older consumer that doesn't understand `deleted` would treat a tombstone as an empty-content upsert and blank the page. Please confirm those are in place before merge-to-deploy.

**Still backlogged (acknowledged, fine to defer):** `idleTtlMs` hardcoded / not forwarded from the agent/profile; `runtime_id` unwired (fenced by the ⚠️ `replicas: 1` note); the "capability-agnostic" contract still carrying KB-specific fields + one-directional handler casts; the CLAUDE.md matrix row for `capability/*`.

jacoblee-io (approved):
Re-approving on `a94d3945`. The only change since my prior approval (`1db5584`) is a benign test-assertion update (`a94d394`, zh constitution now asserts `铁则` after the zh-playbook rewrite) — no production-code change. All prior correctness/security findings remain resolved; CI green (KBC Box / Portal Web / Test / Type Check), no conflicts. Reminder on record: the cross-repo deploy pairing (`feat/kb-capability-platform` + `feat/kb-artifact-tombstone` on the consumer side) must be honored at rollout — merging the code here does not trigger deploy, so that ordering is enforced at your manual rollout step.

PR #379, opened 2026-07-03
feat(subagent): batch map-reduce orchestration in spawn_subagent

chent1996:
## Summary

Delegating the same diagnostic across N targets (e.g. "find out why each of these 18 pods crashed, then summarize") previously required the model to hand-write N separate `spawn_subagent` prompts, flooded the parent context with every child's report, and cost N approvals. This PR turns `spawn_subagent` into a declarative 1..N batch tool: one call carries a task template plus an items list (and an optional reduce prompt), and the runtime executes the fan-out and synthesis deterministically.

### Solution

- **Schema**: `task_template?` + `items[1..50]` + `reduce_prompt?` replace the old `prompt` param. A single-item call with no reduce collapses to the exact legacy single-spawn path (same events, same work card); multi-item calls run through a group engine — bounded worker pool over the shared concurrency limiter, completion-order circuit breaker, scaled map-phase backstop, and a reduce child that synthesizes per-item results so the parent context only receives the final summary (6000-char cap) instead of N capsules.
- **Defaults**: single item runs foreground, multi-item runs background (per-call override available). Background groups reuse the existing job/notification machinery; live `group_progress` events plus persisted terminal delegation events drive a group card with per-item status and drill-in.
- Chosen over a separate batch tool (one tool = one mental model, no dual-default confusion) and over a YAML plan file (schema validation at the call boundary, no file lifecycle).

**Breaking (model-facing only, pre-release)**: the `prompt` param is removed and results are always `item_results[]`; default subagent concurrency raised 2 → 4 (`SICLAW_SUBAGENT_CONCURRENCY` to revert). `SUBAGENT_GROUP_ENABLED=false` degrades the tool to single-item-only as an ops lever.

## Test Plan

- [x] Backend: 4118 passed / 2 skipped (`npm test`), `tsc --noEmit` clean — covers worker-pool share, circuit breaker under concurrency (incl. in-flight aborts), collapse-path equivalence, reduce truncation from `fullSummary`, notification dedup
- [x] Frontend: 184 passed (`portal-web` vitest), `tsc --noEmit` clean — group card fold/rebuild, paged-history refetch merge, Jobs bar folding, legacy tool-name rendering
- [x] Manual (local `siclaw local` against a real cluster): foreground batch (3 pods, live per-item convergence + combined summary), single-task collapse (classic work card), background batch (launch → live progress → completion notification with full summary → card fold)


src/tools/workflow/spawn-subagent.ts
```diff
@@ -98,80 +175,182 @@ export function createSpawnSubagentTool(
...
+      details: {
+        ...modelVisible,
+        ...("childSessionId" in result ? { child_session_id: result.childSessionId } : {}),
+      },
     };
+  }
+
+  if ("itemResults" in result) {
+    // ── Batch (map→reduce) report ──
+    const hasReduce = typeof result.reduceSummary === "string";
```
jacoblee-io, line 322: **[severe] A failed reduce discards every per-item summary from the model-visible result — and overall status can still read `done`.**

When the reduce child fails, `runSubagentGroup` still sets `reduceSummary` (either the failed child's error capsule, or `Reduce stage failed: …` from the catch). `hasReduce = typeof result.reduceSummary === "string"` is then true, so `item_results` are stripped to `{item, status}` — the parent model receives zero content from N successful map children and can only re-run the entire batch. `buildGroupNotificationSummary` relays the same finding-free digest for background groups. Meanwhile the status ladder computes `doneCount === total → "done"` independently of the reduce outcome, so the tool can report `status: "done"` + `reduce_summary: "Reduce stage failed: …"`.

Suggestion: on reduce failure, fall back to emitting the per-item summaries (treat it as `hasReduce = false` with the failure noted in `reduce_summary`), so one transient LLM error doesn't throw away the whole map phase.

src/agentbox/session.ts
```diff
@@ -410,39 +424,520 @@ export class AgentBoxSessionManager {
...
+   * parent on completion. Background work blocks session release until it finishes.
+   */
+  private startBackgroundSubagentGroup(request: SpawnSubagentGroupRequest): SubagentGroupResult {
+    const jobId = request.spawnId;
+    const controller = new AbortController();
+
+    // Stop latch: the user pressed Stop before this group launched → register it terminal
+    // ("stopped") and skip the run. suppressNotifyTurn keeps the model from waking on its
+    // own cancellation.
+    if (this.sessions.get(request.parentSessionId)?._aborted) {
```
jacoblee-io, line 923: **[severe] Group stop latch persists no terminal delegation_event — the group card is stuck "Running" forever after reload, with the Stop button permanently armed.**

This latch registers the stopped job and fires a live `notifyParent`, but unlike `startBackgroundSubagent`'s latch (which persists a terminal delegation_event with a comment saying exactly why: "without these the launch card re-paints as Running… forever on reload"), it never calls `persistGroupTerminalEvent`. On reload, `annotateGroupCompletions` finds a launched background group with no bare-groupId terminal event → `groupStatus` stays undefined → `hasActiveBackgroundGroup()` returns true permanently. The live `subagent_done(is_group)` only triggers a refetch, which reads the same DB state.

Root cause worth fixing at the right altitude: `startBackgroundSubagentGroup` copies `startBackgroundSubagent`'s entire ~80-line launch skeleton (stop latch, register, `_backgroundWorkCount`, then/catch/finally notify) and has already drifted on this piece. Extracting a shared `launchBackgroundJob(...)` helper with a `persistTerminal` callback would fix the bug and prevent the next drift.

portal-web/src/components/chat/PilotArea.tsx
```diff
@@ -2288,6 +2331,282 @@ function AgentWorkBatchRow({ task }: { task: AgentWorkBatchTask }) {
...
+      ? (details.items as Array<Record<string, unknown>>)
+      : []
+
+  const total = Math.max(argItems.length, inlineResults.length, folded.length, liveItems.length)
+  const items: GroupItemView[] = []
+  for (let i = 0; i < total; i++) {
+    const inline = inlineResults[i]
+    const fold = folded.find((f) => numberValue(f.index) === i)
+    const live = liveItems.find((l) => numberValue(l.index) === i)
+    const rawItem = inline?.item ?? argItems[i]
```
jacoblee-io, line 2402: **[severe] Skipped items render as "Running" forever after reload.**

Skipped items (circuit-break / group timeout) are by design never persisted as child delegation events, and `persistGroupTerminalEvent` carries no per-item statuses — only `{status, capsule, reduceChildSessionId}`. For a *background* group the persisted tool row is just `{status:"launched"}` (no `item_results`). So after reload, every skipped index has no inline, no fold, and no live entry, and falls through to this `(background ? "running" : "pending")` fallback: a 50-item group that circuit-broke at 5 failures renders ~45 items with a blue "Running" pill under an overall "Failed" badge, permanently.

Related: the live path has the same blind spot — `runOne`'s skip branch sets `state.status = "skipped"` but returns without `emit("map")`, so live frames never show skipped either.

Suggestion: include a compact per-item status snapshot in the group terminal event (index → status is enough), and emit one final progress frame after the pool drains.

src/agentbox/session.ts
```diff
@@ -410,39 +424,520 @@ export class AgentBoxSessionManager {
...
+        // `#` is not touched by delegation-id validation (verified): children of a group are
+        // tied to it by this prefix, which the UI groups on.
+        spawnId: `${groupId}#${i}`,
+      };
+      let res: SpawnSubagentResult;
+      try {
+        // Submit INTO the global limiter (lazy child creation happens inside runSpawnedSubagent).
+        // Mark `running` + emit only once the slot is acquired (inside the thunk) — an item that
+        // is still waiting for a limiter slot must stay `queued`, not report as running.
+        res = await this.subagentLimiter.run(() => {
```
jacoblee-io: **Items queued in the shared limiter when the group aborts still spawn full child sessions and land as `partial` instead of `skipped`.**

`runOne` checks `mapAbort.signal.aborted` only *before* `subagentLimiter.run(...)`; the thunk doesn't re-check. When the breaker trips or the group backstop fires while items are waiting for a slot (limiter saturated by another session), each queued item still acquires a slot and `runSpawnedSubagent` performs `createSiclawSession` + persists a chat session and user message before its `stopRequested` check throws — returning `partial` with "Sub-agent cancelled by job_stop before producing a report." Cost: a wasted child session + persisted transcript per aborted-but-queued item, and a wrong status classification (see the reduce-gate comment below for where that classification bites).

Suggestion: re-check `mapAbort.signal.aborted` as the first statement inside the limiter thunk and return `skipped`.

src/agentbox/session.ts
```diff
@@ -410,39 +424,520 @@ export class AgentBoxSessionManager {
...
+    let reduceSummary: string | undefined;
+    let reduceChildSessionId: string | undefined;
+    let reduceTruncated = false;
+    // Run reduce when: a reduce_prompt was given, at least one item produced usable output
+    // (done or partial), the user didn't cancel, AND the circuit breaker did not trip. A
+    // map-phase timeout does NOT block reduce (the reduce child keeps its own 600s backstop),
+    // but a tripped breaker signals a systematic template/environment error: an in-flight child
+    // that mapAbort turned into a `partial` must NOT lift usableCount over the gate and drag a
+    // doomed batch into a wasted reduce. Circuit break ⇒ skip reduce, overall status failed.
+    if (request.reducePrompt && usableCount > 0 && !userAbort.signal.aborted && !breaker.tripped) {
```
jacoblee-io: **After a map-phase timeout, cancellation stubs pass this reduce gate — the reduce child runs a full LLM synthesis over pure cancellation notices.**

`usableCount` counts `partial`, and aborted in-flight/queued children come back as `partial` with the capsule "Sub-agent cancelled by job_stop before producing a report." (see comment on the limiter thunk above). The gate excludes `userAbort` and `breaker.tripped` but not the timeout path — the comment above it says a map timeout deliberately doesn't block reduce, but that assumes partials carry usable output; cancellation stubs violate that assumption. A group that times out with zero real completions still launches a reduce child to "synthesize" N cancellation notices, and the model-visible `reduce_summary` is fabricated from nothing while status reads `timed_out`.

Suggestion: give cancellation stubs their own status (or exclude them from `usableCount`).

src/agentbox/session.ts
```diff
@@ -410,39 +424,520 @@ export class AgentBoxSessionManager {
...
+      const reduceReq: SpawnSubagentRequest = {
+        description: `${request.description} — summary`,
+        prompt: buildReduceInput(request.reducePrompt, outcomes),
+        subagentType: request.subagentType,
+        runInBackground: false,
+        parentSessionId: request.parentSessionId,
+        parentAgentId: request.parentAgentId,
+        userId: request.userId,
+        taskListId: request.taskListId,
+        spawnId: `${groupId}#reduce`,
```
jacoblee-io: **`delegation_id` columns are `CHAR(36)` — the `#`-suffix scheme has zero headroom and can overflow on production MySQL.**

`groupId` is the raw provider tool-call id (`spawnId: toolCallId` in spawn-subagent.ts), and child ids are `${groupId}#${i}` / `${groupId}#reduce` with no length validation anywhere in the persistence chain. `src/portal/migrate.ts` defines `chat_sessions.delegation_id` and `chat_messages.delegation_id` as `CHAR(36)`. Anthropic's documented example id is 29 chars — `+ "#reduce"` = exactly 36, so a single extra character in the id (other providers, repair-generated ids) means `ER_DATA_TOO_LONG` under MySQL strict mode: the reduce child's session row and every delegation event for it fail to persist. node:sqlite ignores `CHAR` length, so `npm test` and `siclaw local` can never catch this — it's prod-MySQL-only.

The PR comment verified `#` passes delegation-id *character* validation but not column *width*. Suggestion: widen to `VARCHAR(64)` in the same PR.

jacoblee-io (commented):
## Review: batch map-reduce orchestration

Overall the architecture is sound: map/reduce children go through the existing `runSpawnedSubagent` executor and the shared `subagentLimiter`, the job/notification machinery (`claimNotification` dedup, `_backgroundWorkCount` accounting) is reused correctly, and the single-item collapse path is equivalent to the legacy single spawn. The findings below fall into two groups: **(a)** terminal-state paths where the group engine copied the single-spawn skeleton but dropped pieces the original deliberately has, and **(b)** re-implementations of helpers that already exist — one of which has already drifted into a real bug.

Inline comments are ranked in each thread. Additional non-inline notes:

**Status ladder** — a user Stop that lands *after* all N items finished `done` reports overall `partial` (ladder checks `userAbort.signal.aborted` before `doneCount === total`), and nothing in the result or notification indicates the reduce was skipped due to cancellation — `circuitBroken` gets an explicit flag, cancellation doesn't. With no `reduce_prompt` at all, a post-completion Stop still downgrades a fully-`done` batch to `partial`, which is a plain mislabel.

**Docs drift (in this PR)**
- `docs/design/2026-07-subagent-group.md` says `src/core/prompt.ts` "still teaches the old pattern" as a known residual — but this same PR rewrites prompt.ts to teach the items pattern. Stale on day one.
- The doc states the reduce runs when "`reduce_prompt` given, ≥1 usable item, user did not cancel", but the code additionally gates on `!breaker.tripped`. The doc contract and the enforced one diverge.
- The CLAUDE.md matrix row for `subagent-registry.ts` still says "Two independent master switches; flipping off hides the params/tools" — this PR adds a third switch whose OFF semantics deliberately do *not* hide the tool.

**Efficiency (non-blocking)**
- `annotateGroupCompletions` unconditionally sets `changed = true` and rebuilds the launch card on every annotation pass, defeating the identity-preserving contract every other annotate stage keeps — one background group in history means every poll re-renders it forever.
- The foreground group progress path forwards every per-item transition as a full N-item `tool_execution_update` frame with no throttle (the background path has the 700 ms throttle); a 50-item batch emits ~100 frames × 50-entry arrays.
- `groupWorkSummary` does `folded.find`/`liveItems.find` inside the 0..total loop (O(n²)) and is recomputed unmemoized on every card render.

portal-web/src/hooks/usePilotChat.ts
```diff
@@ -704,6 +746,131 @@ function hasActiveBackgroundSubagent(messages: PilotMessage[]): boolean {
...
+      metadata: {
+        ...(m.metadata ?? {}),
+        groupBackground: true,
+        ...(groupItems.length > 0 ? { groupItems } : {}),
+        ...(term
+          ? {
+              groupStatus: term.status,
+              // A non-empty terminal child_session_id means a reduce child ran → its capsule IS the
+              // reduce summary; an empty id means no reduce ran (the capsule is only a status digest).
+              ...(term.childSessionId
```
jacoblee-io: **The circuit-breaker reason is unreachable in every UI surface for a no-reduce batch.**

The backend deliberately writes `breaker.reason` into the terminal capsule ("so the user learns WHY the batch stopped"), but this gate only attaches `term.summary` when `term.childSessionId` is non-empty — and a circuit-broken group never ran a reduce, so `persistGroupTerminalEvent` wrote `childSessionId: ""`. The card shows `Failed` + per-item rows with no explanation; the background completion notification also only carries `reduceSummary` (absent here). The reason exists solely in the persisted event content that this line discards.

Suggestion: attach the capsule as a group-level summary unconditionally, and only the `groupReduceChildSessionId` drill-in behind the `childSessionId` check.

src/tools/workflow/spawn-subagent.ts
```diff
@@ -98,80 +175,182 @@ export function createSpawnSubagentTool(
...
-      // tool_execution_update the AgentWorkCard renders (status, tool trace, activity).
+      const items = (p.items ?? []) as Array<string | Record<string, string>>;
+      const reducePrompt = p.reduce_prompt?.trim() || undefined;
+
+      // Ops rollback lever (design decision #20): with the batch capability OFF, spawn_subagent
+      // degrades to a pure single-task tool — a multi-item plan or a reduce_prompt is rejected and
+      // the item cap is forced to 1. This is a behaviour switch, not a compatibility shim.
+      if (!SUBAGENT_GROUP_ENABLED && (items.length > 1 || reducePrompt)) {
+        return errorResult(
+          "spawn_subagent batch mode is disabled (SUBAGENT_GROUP_ENABLED=false): pass a single item and " +
```
jacoblee-io: **The rollback lever contradicts all model-facing guidance, and isn't actually flippable by ops.**

`SUBAGENT_GROUP_ENABLED=false` rejects multi-item calls here, but `buildDescription()`, `src/core/prompt.ts` ("don't split one batch into many single-item calls"), and `deep-investigation.ts` ("put them all in ONE spawn_subagent call's items list… Do not emit multiple spawn_subagent calls") are all ungated — under rollback the model is instructed to do exactly what the tool rejects, and the error text doesn't tell it to fall back to N separate calls. Also, the flag is a compile-time `export const` in subagent-registry.ts, not an env read — so the PR description's "`SUBAGENT_GROUP_ENABLED=false` degrades the tool … as an ops lever" doesn't hold: ops can't flip it without a rebuild.

Suggestion: read it from env like the sibling knobs, gate `buildDescription()` on it, and/or extend this error with the explicit fallback instruction.

src/agentbox/subagent-group.ts
```diff
@@ -0,0 +1,335 @@
...
+  const raw = typeof item === "string" ? item : JSON.stringify(item);
+  const oneLine = raw.replace(/\s+/g, " ").trim();
+  return oneLine.length > 120 ? `${oneLine.slice(0, 117)}…` : oneLine;
+}
+
+/** Stable key for duplicate detection (object key order does not matter). */
+function dedupeKey(item: string | Record<string, string>): string {
+  if (typeof item === "string") return `s:${item}`;
+  const sorted = Object.keys(item).sort().map((k) => `${k}=${item[k]}`);
+  return `o:${sorted.join("")}`;
```
jacoblee-io: **Ambiguous dedupe key: `join("")` makes distinct object items collide.**

`{a: "1b=2", b: "3"}` and `{a: "1", b: "2b=3"}` both serialize to `o:a=1b=2b=3` (identical key set, so both pass the template-key checks), and the whole valid batch is rejected with "Item N duplicates item M". Any value embedding `<otherKey>=` at a boundary triggers it — plausible with CLI args / selectors / log lines as item values. Low severity (clear error, no data loss) but trivially fixed: `JSON.stringify` the pairs or join with an escaped separator.

src/agentbox/session.ts
```diff
@@ -410,39 +424,520 @@ export class AgentBoxSessionManager {
...
+      if (breaker.tripped) {
+        // Stop the whole map phase: skip everything not yet started + abort in-flight.
+        mapAbort.abort();
+      }
+      emit("map");
+    };
+
+    // Worker pool: `getGroupWorkerShare()` workers drain a shared index counter, so at most
+    // that many children are ever submitted to the global limiter at once.
+    let nextIndex = 0;
```
jacoblee-io: **[reuse] Hand-rolled worker pool — `ConcurrencyLimiter` (already imported in this file) is exactly this.**

`let nextIndex = 0` + N async worker loops + `Promise.all`, plus a re-implementation of the limiter's own ≥1 clamp. `const pool = new ConcurrencyLimiter(workerShare); await Promise.all(tasks.map((_, i) => pool.run(() => runOne(i))))` is equivalent, and keeps one concurrency primitive (with one queueing/fairness semantics) in this file instead of two. Any future limiter fix (fairness, activeCount instrumentation) won't reach this copy.

portal-web/src/hooks/usePilotChat.ts
```diff
@@ -704,6 +746,131 @@ function hasActiveBackgroundSubagent(messages: PilotMessage[]): boolean {
...
+}
+
+/** The groupId for a group launch card (= the launched job_id == the spawn toolCallId). */
+function groupLaunchId(m: PilotMessage): string | undefined {
+  const parsed = m.content ? tryParseJson(m.content) : undefined
+  return messageString(parsed?.job_id) ?? messageDelegationId(m) ?? m.id
+}
+
+/** Map a delegation status/event string to a group item status (mirrors annotateSubagentCompletions). */
+function normalizeGroupStatus(raw?: string): string {
```
jacoblee-io: **[reuse] Second copy of the delegation-status normalizer — already drifting.**

The doc comment admits this "mirrors annotateSubagentCompletions", which carries the same five regex chains inline; this copy already adds `running`/`skipped` branches the single-subagent chain lacks. Next status token added to the backend will predictably be applied to one copy and not the other, and a group card and a single-subagent card will render the same delegation_event with different statuses. Same story one layer down: `groupLaunchId` is byte-identical to the jobId derivation inside `annotateSubagentCompletions`, and `isGroupForm` is duplicated verbatim in PilotArea.tsx (`isGroupFormMessage`). Suggestion: one shared `normalizeDelegationStatus` / `launchJobId` / `isGroupForm`, consumed by both folds and the renderer.

src/agentbox/subagent-group.ts
```diff
@@ -0,0 +1,335 @@
...
+  get reason(): string {
+    return `Circuit breaker: the first ${this.window} sub-agents all failed with no success — likely a template or environment error. Remaining items were skipped.`;
+  }
+}
+
+/**
+ * Bound the model-visible reduce summary to `maxChars` (design decision #7). A group
+ * summary synthesises N items, so it gets a larger budget than a normal capsule (1800).
+ */
+export function truncateReduceSummary(
```
jacoblee-io, line 329: **[reuse] Third char-budget truncation implementation — `truncateAtBoundary` in `delegation-summary.ts` (one file away) already does this, and does it better.**

`truncateAtBoundary` clips at a newline/sentence boundary and appends a suffix — exactly this job. This new one slices mid-word, making the group's reduce summary the only model-visible capsule in the system that can end mid-token. Suggestion: parameterize and reuse `truncateAtBoundary` (same for `truncateBody` above).

src/core/subagent-registry.ts
```diff
@@ -126,3 +136,113 @@ export function getSubagentType(name?: string): SubagentType | undefined {
...
+
+/**
+ * Max items allowed in a single `spawn_subagent` batch, from
+ * `SICLAW_SUBAGENT_GROUP_MAX_ITEMS` (default {@link DEFAULT_MAX_GROUP_ITEMS}). Bounds a
+ * fan-out the model can request in one turn; past the cap the tool fails fast with a
+ * "split into batches" hint before any child is started. Invalid / non-positive values
+ * fall back to the default. (When {@link SUBAGENT_GROUP_ENABLED} is off the tool clamps the
+ * effective cap to 1 regardless of this value.)
+ */
+export function getMaxGroupItems(env: NodeJS.ProcessEnv = process.env): number {
```
jacoblee-io, line 177: **[reuse] 4th–6th verbatim copies of the positive-int env parser in this same file.**

`getMaxGroupItems` / `getGroupItemBudgetMs` / `getGroupHardCapMs` repeat the parse-clamp-default pattern already present in `getBackgroundBashConcurrency`, `getSubagentConcurrency`, `getSubagentMaxRuntimeMs`. A single `parsePositiveIntEnv(raw, fallback, {unitMs?})` collapses all six and makes the seconds-vs-count distinction explicit at the call site.

src/tools/workflow/spawn-subagent.ts
chent1996: Fixed in 4e80e497. A failed or cancel-skipped reduce no longer sets `reduceSummary` — it is recorded in a local `reduceError` instead (`runSubagentGroup`), so `hasReduce` here is true only when the reduce actually completed, and `item_results` keeps every per-item summary for the model to self-synthesize. The failure reason still reaches the model via the new `group_summary` field (never alongside `reduce_summary`). Overall status: an all-done batch with a failed or cancel-skipped reduce now reports `partial`, not `done` (missing synthesis = not fully successful). Covered by new session.test.ts + spawn-subagent.test.ts cases.

src/agentbox/session.ts
chent1996: Fixed in 4e80e497. The group stop latch now mirrors the single-spawn latch: it persists a bare-groupId terminal delegation_event (status `partial` — the delegation enum has no `stopped` — with an all-`skipped` `itemStatuses` snapshot) in addition to the live `notifyParent`, so the card renders a terminal state after reload and the Stop button disarms. See `startBackgroundSubagentGroup` in session.ts.

portal-web/src/components/chat/PilotArea.tsx
chent1996: Fixed in 4e80e497. `persistGroupTerminalEvent` now carries an `itemStatuses` snapshot (`[{index, status}]` for every item), threaded additively through `DelegationEventPayload` → internal-api / chat-repo → message metadata as `item_statuses`. On reload `annotateGroupCompletions` consumes the snapshot, so never-persisted (skipped) items render as skipped instead of perpetual Running. Covered in usePilotChat.group.test.ts.

src/agentbox/session.ts
chent1996: Fixed in 4e80e497. `runOne` is now two-phase: the pre-slot check settles the early skip, and the limiter thunk re-checks `mapAbort.signal.aborted` after acquiring the slot (`skippedInSlot`), so items still queued when the breaker or backstop fires settle as `skipped` without spawning a child session. Regression test in session.test.ts.

src/agentbox/session.ts
chent1996: Fixed in 4e80e497. The gate is now `doneCount > 0`: a group map child reports `partial` only as a cancellation stub (that equivalence is asserted in tests), so stubs can no longer trigger a reduce. When user cancellation skips a reduce that had usable input, a `reduceSkippedForCancel` flag records it — the status ladder downgrades to `partial` and `group_summary` tells the model why the synthesis is missing.

src/agentbox/session.ts
chent1996: Fixed in 4e80e497. `delegation_id` is now `VARCHAR(64)` in both CREATE TABLEs and the `safeAlterTable` defs; because `safeAlterTable` only ADDs missing columns, existing MySQL deployments are widened by a new idempotent MySQL-only `widenColumn` MODIFY (checks information_schema first; sqlite no-op). 64 chars leaves `${providerToolCallId}#reduce` ample headroom. We did not add an explicit length clamp: provider tool-call ids are bounded far below 57 chars in practice, and if that assumption ever breaks, MySQL strict mode fails loudly rather than silently truncating. Covered by migrate-sqlite / schema-invariants tests.

portal-web/src/hooks/usePilotChat.ts
chent1996: Fixed in 4e80e497. The backend now computes a `groupSummary` (circuit-break reason / reduce failure / cancel note) whenever there is no reduce summary and persists it as the terminal capsule (`reduceSummary ?? groupSummary ?? status digest`). This gate no longer requires `childSessionId` — `term.summary` is attached unconditionally as `groupSummary`, the group card renders it, the task notification body uses it, and the model sees it as `group_summary`. The breaker reason is now reachable on every surface for a no-reduce batch.

src/tools/workflow/spawn-subagent.ts
chent1996: Fixed in 4e80e497. The switch is now a runtime env read — `isSubagentGroupEnabled()` reading `SICLAW_SUBAGENT_GROUP_ENABLED` (default ON) — so ops can actually flip it without a rebuild. OFF now also gates `buildDescription()` down to the single-task text, on top of clamping the item cap to 1 and rejecting `reduce_prompt`. We deliberately did not fork the static prompt.ts fan-out guidance on the flag: rollback-off is a rare emergency state, and the tool error text tells the model exactly how to proceed single-item — forking the system prompt for it would couple prompt assembly to a runtime env read.

src/agentbox/subagent-group.ts
chent1996: Good catch on readability — though the collision itself could not occur: the separator in the reviewed code was an invisible \x01 control byte, which GitHub renders exactly like the ambiguous bare join you quoted (it fooled our own re-verification too). That invisibility is a defect in its own right, so 4e80e497 replaces it with visible JSON-encoded sorted `[key, value]` pairs — collision-proof for any value content — and adds your exact example as a regression test in subagent-group.test.ts.

src/agentbox/session.ts
chent1996: Fixed in 4e80e497 as suggested — the hand-rolled pool is gone. The worker pool is now `new ConcurrencyLimiter(Math.min(getGroupWorkerShare(), total))` whose workers submit children into the shared `subagentLimiter`, keeping one concurrency primitive.

portal-web/src/hooks/usePilotChat.ts
chent1996: Fixed in 4e80e497. Both chains now consume a single `normalizeCompletionStatus` from the new `portal-web/src/lib/group-form.ts` (which also hosts the shared `isGroupForm` predicate). The group-only `running`/`skipped` tokens live in the shared helper with an ordering note, so the next backend status token lands in exactly one place.

src/agentbox/subagent-group.ts
chent1996: Fixed in 4e80e497. `truncateAtBoundary` is now exported from delegation-summary.ts with a `suffix` parameter, and both `truncateReduceSummary` and `truncateBody` delegate to it — the reduce summary clips at a newline/sentence boundary like every other model-visible capsule.

src/core/subagent-registry.ts
chent1996: Fixed in 4e80e497. A single `parsePositiveIntEnv(raw, fallback, {unitMs?})` now backs all six getters (counts and millisecond durations), with unit tests on the shared parser in subagent-registry.test.ts.

chent1996:
Thanks for the thorough review — all 13 inline threads are addressed in 4e80e497 (replies on each thread). Responses to the non-inline notes:

**Status ladder** — fixed. `doneCount === total` is now checked before `userAbort`/`timedOut`, so a Stop landing after every item finished reports `done`, not `partial`. Cancellation is no longer silent: a reduce skipped by user cancellation sets `reduceSkippedForCancel`, which downgrades an all-done batch to `partial` and surfaces the reason via the new `group_summary` field (same channel the circuit-breaker reason now uses). Cross-tested in session.test.ts (timeout / circuit / userAbort × doneCount ladder cases).

**Docs drift** — all three fixed in `docs/design/2026-07-subagent-group.md` + the CLAUDE.md matrix: the stale prompt.ts "known residual" note is gone, the documented reduce-gate contract now matches the enforced one (`doneCount > 0`, no cancellation-stub input, breaker excluded), and the matrix row documents the third switch's non-hiding OFF semantics (env-read `SICLAW_SUBAGENT_GROUP_ENABLED`: cap→1 + `reduce_prompt` rejected + description gated).

**Efficiency** — all three fixed: `annotateGroupCompletions` is identity-preserving (per-field compare; a settled background group in history no longer re-renders every poll), `groupWorkSummary` builds a Map and is memoized (O(n)), and the foreground progress path now has a trailing throttle that is cancelled on completion (the return value stays authoritative).

Verification on 4e80e497: backend `npm test` 4134 passed / 2 skipped, portal-web vitest 185 passed, `tsc --noEmit` clean in both.

One correction from re-verification worth flagging: the dedupe-key collision (inline #9) turned out not to be reproducible — the join separator was an invisible `\x01` control byte that renders as an empty string on GitHub. Since it fooled both the review and our first re-check, we replaced it with visible JSON-encoded pairs anyway and kept your example as a regression test.

chent1996:
Follow-up 059a3c32 — real-machine Stop testing on the review-fix build surfaced two more frontend gaps (both fixed and re-verified live):

1. The background-subagent history poller only covered the single-spawn form; a background group had no poll fallback. A user Stop suppresses both live convergence signals (job_stop claims the notification → no `subagent_done`; no synthetic turn → no `background_turn_done`), so a stopped group's card stayed "Running" until a manual reload even though the terminal event was already persisted.
2. The poller's stop-condition judged fold-ness on the freshly fetched page even when the post-abort suppression window skipped applying it — a terminal event persisting inside that ~1.5s window ended polling while the on-screen card never received the fold. (Same latent race existed for stopped single background spawns.)

Verified end-to-end on a live deployment: Stop mid-map on a 6-item batch → running children abort (`partial`), queued item lands `skipped`, reduce skipped with the cancel reason in `group_summary`, and the card + jobs bar converge to terminal within one poll cycle (~2.5s) with no reload.

jacoblee-io (commented):
## Review: deeper pass on the batch orchestration

Overall this is high quality — the core map→reduce engine, the two-abort-scope design, the completion-order circuit breaker (with late-`done` release), the idempotent `widenColumn` migration, and the collapse-path byte-equivalence are all correct and well documented. The read-only drill-in authorization (`resolveReadableSession`: one hop to the parent, gated on same-agent + owner + not-deleted, read endpoints only) is safely scoped. The findings below are what's left; only the first is a concrete, user-visible defect.

### 1. [Medium — bug] Paged-back + Stop leaves a background group card stuck on "Running" and leaks the poll

`portal-web/src/hooks/usePilotChat.ts` (background-subagent poll, ~L1296-1322).

Repro: a background batch is running → the user scrolls up and loads older history (`pageRef.current > 1`) → the user hits Stop. Stop suppresses both live convergence signals (`job_stop` claims the notification → no `subagent_done`; no synthetic turn → no `background_turn_done`), so the SSE `scheduleRefetch` never fires and this poll is the only safety net. But the poll is a wholesale page-1 replace gated on `applied = pageRef.current === 1 && …`. When paged back, `applied === false`, so it never `setMessages` (card never folds) and the stop condition `if (applied && !hasActiveBackgroundSubagent(...) && !hasActiveBackgroundGroup(...)) return` can never be reached → it re-schedules every 2.5s indefinitely.

Two symptoms together: the group card stays "Running" until a manual reload, and the poll `setTimeout` chain never stops (wasted page-1 fetches for as long as the user stays scrolled up).

This is the same paged-back blind spot that was fixed for the SSE path in 059a3c32 (`refetchHistory` uses `mergePage1IntoHistory` when `pageRef.current !== 1`), but the poll fallback kept the `pageRef === 1` gate. Suggested fix: in the poll, when `pageRef.current > 1` and not streaming, merge via `mergePage1IntoHistory` (same as the SSE path) and judge the stop/fold condition on the merged result rather than gating it behind `applied`.

Note: this also subsumes the "per-item rows retain a stale `running` pill after Stop" concern — the backend snapshot is fine (in-flight items are recorded as `partial` in the terminal event's `item_statuses`, and the fold overwrites live state), so the only way a stale `running` survives is when the fold never lands, i.e. exactly this bug.

### 2. [Medium — design] Multiple concurrent batches can starve an interactive spawn

`getGroupWorkerShare() = max(1, concurrency-1)` with the global `subagentLimiter` sized at `concurrency` (default 4). The "always leaves ≥1 slot for an interactive single spawn" guarantee holds for a single active group, but two batches in the same agentbox process (local multi-user, or multiple sessions of one agent in k8s) together demand `2*(concurrency-1)` global slots, saturating the limiter and leaving 0 for a foreground single spawn — which then blocks up to ~600s behind a child. Not a correctness bug (no deadlock; the orchestrator holds no slot), but the documented guarantee silently assumes one group. Worth a decision: reserve a dedicated foreground slot, or make the worker share adaptive to the number of live groups.

### 3. [Low — perf] `annotateGroupCompletions` identity-preservation is ineffective on the live poll path

The folded group fields (`groupBackground` / `groupStatus` / …) are never persisted; they are re-derived from the delegation_event on every fetch. So for the launch row the fresh input always has `prev.groupBackground === undefined`, `groupMetaUnchanged` is always false, and the row is rebuilt (new object) every poll — `SubagentGroupCard`'s `useMemo([message])` then recomputes every 2.5s. The identity-preservation only takes effect on the merge path's `older` bucket. Correctness is fine; the "no needless re-render" contract just isn't met on the main live path.

### 4. [Nit] Two small items

- `GROUP_REDUCE_INPUT_MAX_CHARS = 150_000`: a 50-item batch (≤1800 chars each ≈ 90K) fits, but this is a single prompt (~40K tokens) fed to the reduce child. Just confirm the reduce child's model context comfortably holds it plus its system prompt.
- `summarizeItemStatuses` — the `order` array lists only `done/partial/failed/timed_out/skipped`. If an item were ever `running`/`queued` at report time it would be silently dropped and the per-status counts wouldn't sum to `items.length`. The `await Promise.all` guarantees all items are terminal, so this is latent, but adding the two states (or asserting all-terminal) would harden it.

chent1996:
Thanks for the deeper pass — all four findings addressed in 95602d74.

**1. Paged-back + Stop (bug)** — fixed as suggested, with one refinement. When `pageRef.current > 1` (and not streaming/abort-suppressed) the poll now merges the fresh page 1 into the loaded history via `mergePage1IntoHistory`, same anti-clobber policy as the SSE path. For termination we deliberately did NOT judge the stop condition on the merged result inside the state updater (side-effects in a React updater are impure and StrictMode-hostile): the paged-back branch simply keeps polling, and once the merge folds the card the effect's `hasActiveBgSubagent` dependency flips false and the cleanup cancels the timer — so both symptoms (stuck card AND the leaked `setTimeout` chain) end through one mechanism. The page-1 fast-path stop stays, still gated on an applied page. Verified live: batch running → load older history → Stop → card folds to `partial` (with the cancel reason) within one poll cycle, scrollback intact, no reload.

**2. Multi-batch starvation (design)** — decided and implemented: a manager-wide `groupChildLimiter` of `max(1, concurrency-1)` that ALL group-spawned children (map workers and the reduce child) acquire BEFORE the global `subagentLimiter` slot. Groups collectively can never hold more than `concurrency-1` global slots, so an interactive single spawn always finds a free one, regardless of how many batches run. The strict acquisition order keeps the nested semaphore deadlock-free (group-slot holders only ever wait on the global limiter; global holders never wait on a group slot). We chose the reserved-slot invariant over an adaptive share because it is a hard guarantee rather than a rebalancing heuristic. Regression test: two concurrent 4-item groups at concurrency 4 → max observed in-flight children is exactly 3 (`session.test.ts`). The design doc's "no second, nested group semaphore" paragraph is updated to the new contract with the deadlock-freedom argument.

**3. Ineffective identity preservation on the poll path (perf)** — fixed via `preserveUnchangedRows` inside a shared `reconcileRefetchedPage1(prev, fresh)` used by every wholesale page-1 replace: rows whose role/content/status/metadata are unchanged get their PREVIOUS object back, so `SubagentGroupCard`'s `useMemo([message])` skips recomputation between polls. It composes with the live-frame carry (the carry re-attaches `groupProgress` first, which is exactly what makes the row compare equal). A false mismatch merely rebuilds a row — never stale data.

**4. Nits** — (a) reduce-input budget documented at the constant: 50×1800 ≈ 90K chars, the 150K ceiling ≈ 40K tokens, comfortable inside a 128K-token context plus the reduce child's system prompt; (b) `summarizeItemStatuses` got a catch-all bucket so the counts always sum to `items.length` even if the (currently terminal-only) status union widens — the `Promise.all` + `GroupItemStatus` type remain the primary all-terminal assertion.

Verification on 95602d74: backend 4135 passed / 2 skipped (+ the new collective-cap test), portal-web 189 passed (+2 reconcile tests), `tsc --noEmit` clean in both, and the paged-back Stop scenario re-verified end-to-end on a live deployment.

jacoblee-io:
### Merge readiness

- **CI**: green (Portal Web Test / Test / Type Check all pass).
- **Conflicts**: none — `MERGEABLE` against `main`.
- **Blocked on**: `REVIEW_REQUIRED` — no approving review yet (both reviews are `COMMENTED`).

No blocking correctness or data issue found. My recommendation before approving: address **finding #1** (paged-back + Stop → background group card stuck on "Running" + the poll never stops). It's a real, user-visible defect in the exact Stop-convergence path this PR has been hardening, and the fix is small — mirror the SSE path's `mergePage1IntoHistory` in the background-subagent poll and judge the fold/stop condition on the merged result. Narrow trigger (background batch + scrolled-back history + Stop) and self-heals on reload, so if you prefer, it's also fine to merge and fast-follow with a tracked issue.

Finding #2 (multiple concurrent batches can starve an interactive spawn) is a design call for the author, not a blocker. #3/#4 are perf/nits and don't block.

jacoblee-io (approved):
Verified the second-review follow-ups in 95602d7 — all four points are addressed correctly:

1. **Paged-back + Stop stranding/leak** — the bg-subagent poll now merges the fresh page 1 via `mergePage1IntoHistory` when paged back, and termination rides the effect's `hasActiveBgSubagent` dependency (folds → dep flips → cleanup cancels the timer). Fixes both the stuck "Running" card and the endless reschedule.
2. **Cross-group starvation** — the new manager-wide `groupChildLimiter (max(1, concurrency-1))` acquired before the global `subagentLimiter` collectively caps all group children at one below the global limit. Deadlock-free: strict acquisition order (group-slot holders only ever wait on the global limiter; global holders never wait on a group slot), and the collapse/interactive single-spawn path takes only the global limiter, so it always finds the reserved slot.
3. **Re-render** — `preserveUnchangedRows` restores row identity across refetches.
4. **`summarizeItemStatuses`** — catch-all bucket keeps the counts summing to `items.length`.

CI green (Test / Portal Web / Type Check), with new coverage for the collective cap and row identity.

Two non-blocking notes: the paged-back poll-merge branch itself isn't directly unit-tested (it mirrors the already-tested SSE-refetch path); and `preserveUnchangedRows` compares metadata via `JSON.stringify` per row per poll — negligible for a ≤50-item group card, not worth changing.

LGTM — approving.

PR #380, opened 2026-07-03
feat(channels): cache and display the bound group's chat title

jacoblee-io:
## Summary

Channel bindings currently show only the raw route key (`oc_xxx` chat id), so the Portal bindings list can't tell which Feishu group is which. This PR caches the group's chat title on the binding and shows it in the UI.

- `channel_bindings` gains a **`display_name`** column (display-only cache; identity stays `route_key`, never the name)
- **PAIR seeds the name**: the Lark gateway fetches `GET /im/v1/chats/:chat_id` (only needs `im:chat:readonly`, bot must be in the group) and passes it as `route_display_name` on `channel.pair`; a re-pair that fails to fetch keeps the existing name
- **Lazy backfill for existing bindings**: on a group message, the gateway refreshes the cached title at most once per process per binding via the new `channel.updateBindingMeta` RPC (WS handler + internal HTTP mirror) — also picks up group renames after a gateway restart
- Everything is best-effort: name fetch/update failures are logged and never affect message handling
- Portal Web bindings list shows the group title as the primary line, with channel name + route key as the secondary line

## Compatibility

No deploy-ordering constraint: an old frontend without `channel.updateBindingMeta` just makes the backfill log a warning; an old gateway leaves `display_name` NULL and the UI falls back to `route_key`.

## Tests

- `adapter-rpc.test.ts`: pair seeds/keeps `display_name`, `channel.updateBindingMeta` success/not-found/empty-name, handler count 48→49
- `lark.test.ts`: PAIR passes the fetched title; backfill fires once and honors the once-per-process guard
- `npm test` + `tsc --noEmit` clean (one pre-existing flake in `restricted-bash.test.ts` caused by stale `.claude/worktrees/` copies being collected by vitest — unrelated, fails on clean main too)

### DIFF
--- a/portal-web/src/components/AgentSettings.tsx
+++ b/portal-web/src/components/AgentSettings.tsx
@@ -773,7 +773,7 @@ function ChannelsTab({ agentId, selectedChannelIds, setSelectedChannelIds }: {
   agentId: string; selectedChannelIds: Set<string>; setSelectedChannelIds: (v: Set<string>) => void
 }) {
   const toast = useToast()
-  const [bindings, setBindings] = useState<{ id: string; channel_id: string; channel_name: string; channel_type: string; route_key: string; route_type: string }[]>([])
+  const [bindings, setBindings] = useState<{ id: string; channel_id: string; channel_name: string; channel_type: string; route_key: string; route_type: string; display_name?: string | null }[]>([])
   const [allChannels, setAllChannels] = useState<{ id: string; name: string; type: string }[]>([])
   const [loading, setLoading] = useState(true)
   const [pairingCode, setPairingCode] = useState<string | null>(null)
@@ -875,8 +875,8 @@ function ChannelsTab({ agentId, selectedChannelIds, setSelectedChannelIds }: {
             {bindings.map(b => (
               <div key={b.id} className="flex items-center justify-between px-3 py-2 rounded-md border border-border/50">
                 <div>
-                  <p className="text-[12px] font-mono">{b.channel_name || b.channel_id}</p>
-                  <p className="text-[10px] text-muted-foreground">{b.route_type}: {b.route_key}</p>
+                  <p className="text-[12px]">{b.display_name || <span className="font-mono">{b.route_key}</span>}</p>
+                  <p className="text-[10px] text-muted-foreground font-mono">{b.channel_name || b.channel_id} · {b.route_type}: {b.route_key}</p>
                 </div>
                 <button onClick={() => handleUnbind(b.id)} title="Unbind" className="p-1 rounded-md hover:bg-destructive/20 text-muted-foreground hover:text-red-400">
                   <Trash2 className="h-3.5 w-3.5" />
--- a/src/gateway/channel-manager.ts
+++ b/src/gateway/channel-manager.ts
@@ -37,6 +37,8 @@ export interface ResolvedChannelBinding {
   sessionKey?: string | null;
   createdBy: string | null;
   routeType: "group" | "user";
+  /** Cached chat title from the platform; null/absent until backfilled. */
+  displayName?: string | null;
 }
 
 /**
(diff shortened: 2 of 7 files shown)


LikiosSedo:
Reviewed together with the sicore companion (`k8s/sicore` MR !575, commit `cd3fff9f`). Backend semantics line up on both sides — `channel.updateBindingMeta` (normalize → empty rejects → `UPDATE ... WHERE channel_id=? AND route_key=?` → not-found errors), the conditional keep-existing-name on re-pair, and the `VARCHAR(255)` column. Clean, best-effort, no deploy-ordering constraint. LGTM; two low-severity nits, neither blocking:

**1. Backfill guard can be poisoned by a transient fetch failure** — `src/gateway/channels/lark.ts`, `backfillBindingDisplayName`. `bindingNameRefreshed.add(bindingId)` runs *before* `fetchLarkChatName`, so if the first group message's fetch fails transiently (rate-limit / blip), that binding is never retried until a gateway restart and its cached name stays NULL. Narrow blast radius (new bindings are seeded at PAIR, so this only bites pre-feature bindings on their first post-deploy message) and it's display-only, hence not blocking. If you want it tighter, mark the guard only after a definitive outcome (success or a non-retryable error), optionally with a per-binding attempt cap so a persistently-failing API isn't hammered.

**2. Truncation differs from the sicore side (parity nit)** — here `normalizeBindingDisplayName` uses `trimmed.slice(0, 255)` (255 UTF-16 code units); sicore caps with `[]rune(...)[:255]` (255 code points, matching MySQL's `VARCHAR(255)` char semantics). On a pathological non-BMP name (e.g. many emoji) `.slice` can split a surrogate pair. Group titles are short/BMP in practice, so cosmetic — noting only for cross-side consistency.

Nice touch keeping identity on `route_key` and treating the name as a pure display cache throughout.

jacoblee-io:
Both nits addressed in 461d99bf:

1. **Backfill guard** — replaced the fire-once `Set` with a per-binding attempt counter (cap 3). A transient fetch failure leaves room to retry on later messages; a success pins the guard for the process lifetime. The attempt is still counted up front so concurrent messages can't stampede the API. Covered by two new tests (transient-failure retry, persistent-failure cap).

2. **Truncation** — `normalizeBindingDisplayName` now caps at 255 code points (`[...s].slice(0, 255)`), matching MySQL `VARCHAR(255)` char semantics; can no longer split a surrogate pair. Test added with a 300-emoji name.

LikiosSedo (approved):
Reviewed both commits, including the follow-ups in 461d99bf. LGTM.

**Follow-up verification**

- Retryable backfill guard: attempts are counted per binding (cap 3) and incremented up-front so concurrent messages can't stampede the API; a successful fetch pins the guard for the process lifetime. The three new tests assert actual API call counts (once-per-process, transient-failure retry, cap enforcement), not just outcomes. ✓
- Code-point truncation: `[...trimmed].slice(0, 255)` iterates by code points, so surrogate pairs are never split — matches MySQL `VARCHAR(255)` char semantics; the 300-emoji test pins it down. ✓

**Cross-checked outside the diff**

- The bindings list endpoint (`GET /agents/:id/channel-bindings`) uses `SELECT cb.*`, so `display_name` reaches the UI with no further backend change.
- Excluding `display_name` from the upsert's update clause on a name-less re-pair is correct for both dialects (`ON DUPLICATE KEY UPDATE` / `ON CONFLICT ... DO UPDATE`), and the test asserts both forms.
- The `affectedRows`-based "not found" check is safe on both drivers: mysql2's default flags include `FOUND_ROWS` (matched rows), and SQLite's `changes` also counts same-value updates.
- An unknown WS method gets an error response (`runtime-connection.ts`), so the no-deploy-ordering claim holds — an old frontend just produces one logged warning per binding per process.
- Migration covers both fresh installs (`CREATE TABLE`) and upgrades (`safeAlterTable`).

**Non-blocking observations**

1. The retry budget only protects the fetch leg: if the chat-name fetch succeeds but `channel.updateBindingMeta` fails transiently (e.g. a WS blip), the guard is already pinned and the name won't land until a gateway restart. Given display-only semantics — and that pinning avoids repeated warnings against an old frontend — this reads as an intentional trade-off; just noting it.
2. A group title longer than 255 code points re-pushes one idempotent update per process lifetime (the raw fetched name never equals the truncated stored one). Harmless and rare.
3. For integrated deployments, the consumer's adapter surface needs the matching `route_display_name` param on `channel.pair` plus a `channel.updateBindingMeta` handler (and the column) before this is end-to-end there; degradation until then is graceful (a single warning).

PR #381, opened 2026-07-03
feat(tracing): OpenTelemetry agent-behavior tracing with DB-backed platforms

chent1996:
## Summary

Adds end-to-end OpenTelemetry tracing of agent behavior: an in-process recorder captures prompts, tool calls, and per-call LLM token usage as OpenInference-style spans, and exports them to external analysis platforms (Langfuse / Phoenix / generic OTLP) configured entirely from the DB — with live hot-reload, no restart required.

### Solution

- **Trace recorder** (`src/shared/tracing/`): OpenTelemetry provider + agent-behavior recorder emitting OpenInference span attributes; per-call LLM tokens, root prompt input, and a content-size cap; root model identity derived from `model_route_success`.
- **DB-backed platforms**: analysis targets live in a `tracing_exporters` table; `buildTracingConfig()` assembles OTLP exporters from the DB. On/off is derived from the configured platform set — there is no separate global enable switch.
- **Live hot-reload**: a Portal config change broadcasts `tracing.reloadAll` → Gateway fans out to every running box → `/api/reload-tracing` → `reinitTracing`, so targets can be added/removed without a restart. Tracing config never lands on disk.
- **Portal UI**: a Tracing tab on the Metrics page (`TracingPlatforms` + `useTracingPlatforms`) with masked secrets and an SSRF-guarded test probe.
- **migrate**: create the a2a composite indexes before dropping the FK-backed index (ordering fix).

## Test Plan

- [x] `npm test` — 204 files / 4088 passed, 0 failed

## Notes for reviewers

Rebased onto latest `main` (the branch had diverged 73 commits behind) and squashed 8 commits into one. Conflict resolution touched 6 files — Metrics tabs (merged into main's `TAB_ORDER`/`TAB_LABEL` structure), the gateway RPC method set (kept only the new `tracing.reloadAll`; main already refactored `metrics.live`), the agentbox shutdown hook, and three import blocks — plus realigning the tracing-reload test mock to main's current `AgentBoxManager` interface (`setSpawnEnvResolver` / `setPersistenceResolver` / `getAsync`).


src/shared/tracing/openinference-attrs.ts
```diff
@@ -0,0 +1,232 @@
...
+ * model-side sanitized output (the "model never reads unsanitized output"
+ * invariant), so re-running redaction would be redundant.
+ *
+ * Every return path is length-capped (capContent) as the LAST step — after any
+ * redaction — so no path can emit unbounded content and capping never bisects a
+ * secret.
+ */
+export function redactForExport(value: unknown, kind: ContentKind): string {
+  const text = typeof value === "string" ? value : safeStringify(value);
+  if (kind === "tool_result") return capContent(text);
```
jacoblee-io: **[security, high] `tool_result` is exported without redaction, but MCP-tool and query-tool results are NOT model-side sanitized — so their secrets ship raw to the external platform.**

`redactForExport` skips `redactSensitiveContent` for `kind === "tool_result"` (other kinds are redacted on the next line), on the stated premise that "the model-side pipeline already sanitized it." That holds only for **cmd-exec / script-exec** (which run `postExecSecurity` / `redactSensitiveContent`). **MCP tools** (`mcp-client.ts` → `mcpContentToAgentContent`) and **query tools** (`src/tools/query/*`) return their results with no sanitizer call. Proof the field isn't trusted-clean elsewhere: the delegation-persistence path runs `redactText(event.result?.content)` on the very same event field. So once an admin enables `sendContent`, any secret in an MCP/query result (API responses, DB rows, credentials in a fetched payload) is shipped verbatim to the configured Langfuse/Phoenix/OTLP endpoint. Default `sendContent=false` is the only thing gating this.

**Fix:** run `redactSensitiveContent` on `tool_result` unconditionally, or restrict the trusted-skip to the cmd-exec/script-exec tool names only.

src/agentbox/http-server.ts
```diff
@@ -1385,6 +1406,36 @@ export function createHttpServer(
...
+    if (!client) {
+      // Local mode without a Gateway URL: tracing is config-file driven, there
+      // is nothing to pull. Treat as a clean no-op rather than an error.
+      console.warn("[agentbox-http] No SICLAW_GATEWAY_URL configured, skipping tracing reload");
+      sendJson(res, 200, { ok: true, skipped: true });
+      return;
+    }
+    try {
+      const tracing = await client.fetchTracingConfig();
+      await reinitTracing({ tracing } as SiclawConfig);
```
jacoblee-io: **[correctness, med-high] Hot-reload rebuilds the provider without `deployment.environment.name`, so every trace falls into the `default` environment after any exporter change.**

The reload path does `reinitTracing({ tracing } as SiclawConfig)` where `tracing = fetchTracingConfig()` (DB only). But `environment` is set only in `loadConfig()` from the pod env `SICLAW_TRACING_ENVIRONMENT` at startup — and `buildTracingConfig()` (adapter.ts) emits only `serviceName`/`sendContent`/`exporters`, never `environment`. So after ANY exporter CRUD (add/edit/toggle/delete → `tracing.reloadAll` → `/api/reload-tracing`), the provider is rebuilt with no environment attribute and every span silently lands in Langfuse's `default` environment until the pod cold-restarts. The startup path applies the env merge; the reload path drops it. **Fix:** have `buildTracingConfig()` carry the environment (or re-merge the pod-env value in `reinitTracing`) so reload matches startup.

src/portal/tracing-exporters.ts
```diff
@@ -0,0 +1,233 @@
...
+}
+
+/**
+ * Lightweight SSRF guard for the admin test probe. Blocks cloud metadata and
+ * link-local / loopback addresses when the host is a LITERAL IP, while leaving
+ * RFC1918 private ranges open (internal backends like `phoenix:6006` live
+ * there). Hostnames are not resolved — this is a cheap, admin-only guardrail,
+ * not a full anti-rebinding defence.
+ */
+export function tracingTestSsrfGuard(rawUrl: string): { ok: boolean; error?: string } {
```
jacoblee-io, line 190: **[security, med] The advertised "SSRF-guarded test probe" is hostname-blind and doesn't cover the real export path.**

`tracingTestSsrfGuard` only inspects IP literals (`isBlockedIpLiteral`) — the comment itself says "Hostnames are not resolved." A DNS name resolving to `169.254.169.254` (cloud metadata) or `127.0.0.1` sails through, and there's no re-resolution to defend DNS rebinding. Worse, the **actual** exporter path (`otel-provider.ts:65` `buildSpanProcessors`) applies NO guard at all — just a URL parse — so an admin can save a private/internal/metadata endpoint, skip the Test button, and the live span stream flows there. So the guard provides little real protection for what it's named. It's admin-only (which bounds severity), but either strengthen it (resolve + block private/link-local, and enforce on save, not just the probe) or stop presenting it as an SSRF control.

src/gateway/internal-api.ts
```diff
@@ -195,6 +195,30 @@ export async function handleSettings(
   }
 }
 
+/**
+ * GET /api/internal/tracing-config
+ *
+ * Proxies to Portal's config.getTracingConfig RPC — the GLOBAL tracing config
```
jacoblee-io, line 201: **[security, med] `config.getTracingConfig` returns fully-assembled plaintext exporter auth headers, global and unscoped.**

Unlike the REST list/get routes (which `maskExporterAuth`), this internal RPC hands back the real assembled auth headers to any Gateway/Runtime-OU cert holder, for all exporters, with no tenant scoping. A compromised agentbox of *any* tenant can call it and read every third-party exporter credential (Langfuse Basic material, Phoenix bearer key, arbitrary OTLP headers). Consider scoping what a box can fetch, or at least documenting this as an accepted trust-domain assumption — but as written it widens third-party-credential exposure to the whole box fleet.

src/shared/tracing/otel-provider.ts
```diff
@@ -0,0 +1,215 @@
...
+export function getTracer(): Tracer {
+  return (provider ?? trace.getTracerProvider()).getTracer(TRACER_NAME);
+}
+
+/**
+ * Flush + shut down the provider. forceFlush is raced against a 3s timeout so a
+ * dead in-network backend (BatchSpanProcessor OTLP retry default is 10s+) cannot
+ * hold the process past the K8s 30s grace period.
+ */
+export async function shutdownTracing(): Promise<void> {
```
jacoblee-io, line 154: **[correctness, med] `reinitTracing`/`shutdownTracing` orphans in-flight spans on hot-reload.**

`shutdownTracing()` nulls/replaces the provider but nothing tells the recorder to abort its live `traces`/open spans, which live in a separate module and still reference the now-dead provider. A `POST /api/reload-tracing` (fired by any exporter CRUD) landing during an in-flight prompt loses that trace, and later events for it split onto a dead ROOT context. **Fix:** on reinit, flush/close the recorder's open spans (or version the provider so in-flight spans finish on the old one) before swapping.

jacoblee-io (commented):
## Review: OpenTelemetry agent-behavior tracing

The infrastructure is generally solid — CRUD and both test-probe routes are `requireAdmin`, the new endpoints carry the same mTLS OU check as their neighbors, the `tracing_exporters` DDL is MySQL+SQLite compatible (TEXT not JSON, no `ON UPDATE`/`TIMESTAMP(3)`), reads route through `dialect-helpers`, the recorder is fail-open and a true no-op when disabled (subscription isn't even wired), export is async/bounded (`BatchSpanProcessor`), the provider is a process singleton, and the reload fan-out is parallel + per-box try/catch. Efficiency is well-handled.

The findings below are the gaps. This is a "ship prompts + tool output to an external server" feature, so the sanitization one (#1) is the headline. Inline.

**Lower-severity / cleanup (not inline):**
- **token/cost likely double-counted:** written on BOTH the per-call `llm` spans and the aggregate ROOT `agent` span — Langfuse/Phoenix roll trace totals up from all observations, so trace-level totals double. Relatedly, the ROOT post−pre delta sums tokens from failed model-route candidates plus the winner (over-attributes to the delivered answer).
- **`tool_execution_end` mis-pairing:** with no `toolCallId` and >1 open tool span, the handler early-returns, dropping result/error status; the span is later closed OK by `endPrompt`, so an errored concurrent tool records as a success.
- **dead code / stale citations:** `TracingConfig.requiresTrustedBackend` (config.ts) is declared but never read; stale `tracing.enabled` master-switch comments persist (siclaw-api.ts ~3705, config.ts ~70) though `ALLOWED_CONFIG_KEYS` rejects the key and `enabled` is derived; `tracing-exporters.ts` cites a `tracing-platforms-DESIGN.md (module 1/2)` and the recorder cites "bench's phoenix-tracer" — neither exists in the repo; `sendContentEnabled()` is a pure pass-through wrapper (and one call site bypasses it); `chat_messages.trace_id` is written but never SELECTed back by any app code — confirm a consumer is planned or it's dead plumbing.
- **reuse:** the recorder re-snapshots `getSessionStats()` and recomputes the per-prompt token/cost delta that the existing `prompt_complete` diagnostic already carries (metrics.ts) — two delta paths for one prompt boundary, a divergence risk.

src/shared/tracing/agent-trace-recorder.ts
```diff
@@ -0,0 +1,715 @@
...
+  if (typeof usage.input === "number") attrs[Attr.tokenPrompt] = usage.input;
+  if (typeof usage.output === "number") attrs[Attr.tokenCompletion] = usage.output;
+  if (typeof usage.totalTokens === "number") attrs[Attr.tokenTotal] = usage.totalTokens;
+  if (isRecord(usage.cost) && typeof usage.cost.total === "number") {
+    attrs[Attr.cost] = usage.cost.total;
+  }
+  return attrs;
+}
+
+function computeTokenDelta(prev: BrainSessionStats, curr: BrainSessionStats): TokenDelta {
```
jacoblee-io: **[correctness, plausible] The ROOT token/cost delta re-derives per-prompt stats and sums failed model-route candidates.**

`endPrompt` snapshots `getSessionStats()` and computes a post−pre delta here, independently of the `prompt_complete` diagnostic that already carries the same prev/curr/model/outcome (metrics.ts computes the delta) — two snapshot+delta paths for one prompt boundary (divergence risk). And the post−pre delta spans the whole turn, so it includes tokens spent by failed model-route candidates, not just the delivered answer — over-reporting the ROOT's token/cost. Consider subscribing to `prompt_complete` (the event bus the recorder already models) and, if per-answer cost is intended, excluding rolled-back candidates.

src/shared/tracing/agent-trace-recorder.ts
```diff
@@ -0,0 +1,718 @@
...
```
jacoblee-io: **[still open — priority] Token/cost is double-counted: written on both the per-call `llm` spans and the aggregate ROOT `agent` span.**

`computeTokenDelta` populates the ROOT with a post−pre session delta while each `llm.call` span also carries its own token/cost. Langfuse/Phoenix roll trace totals up from all observations, so the trace-level totals are ~doubled. For an observability feature, its own headline numbers being wrong is worth fixing before merge. Relatedly, the ROOT post−pre delta spans the whole turn, so it also includes tokens spent by failed model-route candidates (over-attributing to the delivered answer). **Fix:** put token/cost on ONE level (per-call spans, and let the backend aggregate — or ROOT only), and exclude rolled-back candidates from the ROOT figure.

src/gateway/internal-api.ts
jacoblee-io, line 201: **[still open — priority] `config.getTracingConfig` returns fully-assembled plaintext exporter auth headers, global and unscoped.**

Unlike the REST list/get routes (which `maskExporterAuth`), this internal RPC hands the real auth headers for ALL exporters to any Gateway/Runtime-OU cert holder, with no tenant scoping — so a compromised agentbox of any tenant can read every third-party exporter credential (Langfuse Basic material, Phoenix bearer key, arbitrary OTLP headers). **Fix:** scope what a box may fetch to its own run/tenant, or at minimum make this trust-domain assumption explicit and narrow the surface.

src/portal/tracing-exporters.ts
jacoblee-io, line 190: **[still open] The advertised "SSRF-guarded" test probe is hostname-blind and doesn't cover the real export path.**

`tracingTestSsrfGuard` only inspects IP literals (the comment itself says "Hostnames are not resolved"), so a DNS name resolving to `169.254.169.254`/loopback bypasses it, and there's no re-resolution against rebinding. The actual exporter (`otel-provider.ts` `buildSpanProcessors`) applies no guard at all — an admin can save a private/metadata endpoint and skip the probe entirely. Admin-only bounds the severity, but as written the control does little. **Fix:** resolve + block private/link-local and enforce on save (not just the probe) — or stop presenting it as an SSRF guard.

src/shared/tracing/otel-provider.ts
jacoblee-io, line 154: **[still open] `reinitTracing`/`shutdownTracing` orphans in-flight spans on hot-reload.**

`shutdownTracing()` swaps out the provider but nothing tells the recorder to close its live `traces`/open spans, which reference the now-dead provider. A `POST /api/reload-tracing` (fired by any exporter CRUD) during an in-flight prompt loses that trace and splits later events onto a dead ROOT. **Fix:** flush/close the recorder's open spans on reinit (or version the provider so in-flight spans finish on the old one before it's shut down).

jacoblee-io (commented):
## Follow-up: the two highest findings are fixed; four remain open

Confirmed at the current head — the two I'd have blocked on are addressed correctly:

- **[fixed] `tool_result` is now redacted like every other kind** (`openinference-attrs.ts` — the `if (kind === "tool_result") return capContent(text)` skip is gone), so MCP/query tool results no longer ship raw once `sendContent` is on. This was the headline security gap. 
- **[fixed] Hot-reload re-merges `deployment.environment.name`** via the new `resolveTracingEnvironment()` single source, so an exporter CRUD no longer drops every span into Langfuse's `default`.
- **[bonus] `tool_execution_end` id-less pairing** now closes the oldest open span with the real status (an errored id-less tool is no longer dropped then silently closed OK).

The other four from the prior review are **not addressed** (their files are untouched in this revision). Re-flagging with current line numbers, in priority order. #6 and #4 are the ones I'd prioritize before merge.

src/portal/tracing-exporters.ts
jacoblee-io, line 190: **[still open] The "SSRF-guarded" test probe is hostname-blind and the real export path has no guard at all.**

Unchanged in this revision: `tracingTestSsrfGuard` only inspects IP literals (the comment still reads "Hostnames are not resolved — a cheap, admin-only guardrail"), so a DNS name resolving to `169.254.169.254`/loopback passes, with no re-resolution against rebinding; and `otel-provider.ts` `buildSpanProcessors` (the path that actually ships spans) applies no guard — an admin can save a private/metadata endpoint and skip the probe entirely.

It's admin-only, so if you're comfortable accepting it, the minimum ask is to **not present it as an SSRF control** (rename / soften the "SSRF-guarded" wording, as was done honestly for the plaintext-config trust-domain note). The stronger fix is to resolve + block private/link-local and enforce on save, not just in the probe. Either is fine — but right now the name over-promises what the code does.

jacoblee-io (commented):
## Status: 5 of 6 resolved — only the SSRF guard remains

Re-checked at the current head. Since my prior review:

- **[fixed] token/cost double-count** — `computeTokenDelta` / `tokenCountAttributes` / `TokenDelta` are gone; the ROOT span no longer carries token/cost (per-call `llm` spans do, and the backend aggregates). This also removes the failed-model-route-candidate over-attribution. 
- **[fixed] in-flight span orphaning on reload** — new `tracingRecorder.abortAll()` closes every live trace with a `tracing_reloaded` event, and it's correctly called in the reload path (`http-server.ts` reload → `abortAll()` **before** `reinitTracing`), so spans end on the current provider.
- **[accepted, documented] the internal `config.getTracingConfig` plaintext fetch** — now carries an explicit "trust-domain assumption (deliberate, not a leak)" note: the box needs the real headers to export directly (same model as `providers.apiKey`), tracing is global with no per-tenant concept, boundary is the caller cert. Reasonable resolution, with a pointer to scope it if exporters ever become tenant-scoped.
- (earlier) `tool_result` redaction and the hot-reload `environment` merge were already fixed.

That clears the two I'd flagged as priorities (#6 token accuracy, #4 credential exposure). **One remains** — inline below.

jacoblee-io (dismissed):
Approving on `49b2e89`. Of the six findings from my reviews, five are fixed or resolved and the last is an accepted, documented, admin-only non-blocker:

- **[fixed] `tool_result` redaction** — all content kinds now run through `redactSensitiveContent`; MCP/query tool results no longer ship raw once `sendContent` is on (this was the headline security gap).
- **[fixed] hot-reload `environment`** — `resolveTracingEnvironment()` is re-merged on reload, so an exporter CRUD no longer drops every span into Langfuse's `default`.
- **[fixed] token/cost double-count** — the ROOT span no longer carries token/cost (`computeTokenDelta`/`tokenCountAttributes` removed); per-call spans carry it and the backend aggregates, and failed-model-route candidates are no longer folded into the ROOT total.
- **[fixed] in-flight span orphaning on reload** — `tracingRecorder.abortAll()` closes live traces (with a `tracing_reloaded` event) and is called before `reinitTracing`, so spans end on the current provider.
- **[accepted, documented] internal `config.getTracingConfig` plaintext fetch** — now explicitly a trust-domain assumption (box needs the real headers to export directly, same model as `providers.apiKey`; boundary is the caller cert; scope it if exporters ever become tenant-scoped). Reasonable.
- **[accepted, non-blocking] SSRF test probe** — still hostname-blind and the real export path has no guard, but it's admin-only. Fine to ship. Follow-up ask (not blocking): either strengthen it (resolve + block private/link-local, enforce on save) or soften the "SSRF-guarded" wording so the name doesn't over-promise — mirrors how the plaintext-config trust-domain note was handled honestly.

CI green (Portal Web / Test / Type Check). Nice iteration through the security review.

jacoblee-io (approved):
Re-approving on `5350088`. The head moved only by a rebase over main (picking up #388 Feishu card feedback) plus a re-squash of the tracing commit — the tracing code is byte-identical to `49b2e89` (verified: empty diff across `src/shared/tracing/*`, `tracing-exporters.ts`, `internal-api.ts`, `http-server.ts`, `config.ts`). So my prior approval stands: 5 of 6 review findings fixed or resolved, and the SSRF test-probe (#3) accepted as an admin-only, non-blocking follow-up (strengthen the guard or soften the "SSRF-guarded" wording). CI green.

PR #382, opened 2026-07-04
kbc: box protocol v3 for the guided linear wizard

LikiosSedo:
Three additive behaviors for the guided wizard, all on existing artifact channels (no new events, no new protocol):

1. **Brief capture** — when the opening compile message carries the 「我的定调标签」 block, the handler parses it deterministically into `authoring/BRIEF.json` (code-written, not model-transcribed); BOX_ROLE instructs the agent to fold it into INTENT.md and compilation choices.
2. **Unified issue queue** — framing questions discovered mid-compile (audience scope, sanitization, process-data handling) are filed into the same `CONTRADICTIONS.json` ticket queue as source conflicts: best-guess + ⚠️ + ticket, never blocking.
3. **Earlier question proposals** — new `propose_questions` MCP tool; merge-with-dedup into `QUESTIONS_PROPOSED.json` (append-only across rounds).

never-block invariant, turn_done semantics, and the S1 self-check contract are untouched. Tests: compile_box 15 (11+4 new), selfcheck 5 — all green. Verified live on sicore-siflow-test (image `protov3-f2571f6`): BRIEF.json parsed verbatim, planted contradiction filed as ticket with options, proposals delivered post-compile.

**Follow-up commit `0171abe`** — stable proposal ids: `merge_proposed_questions` stamps `id = "q-" + fnv1a32(normalized_question)[:8]` (normalize = strip all whitespace, trim edge punctuation, lowercase) on every merged proposal, so the console's `proposal-mark` call always has an id (fixes the accept-500 where W3's tool path dropped it). The web side derives the same id as a fallback for pre-fix rows; the formula is locked by a shared test vector (`"A?" → "q-e40c292c"`). Deployed as image `protov3-0171abe`.


kbc/platform/pod/compile_box.py
```diff
@@ -506,6 +505,139 @@ def _install_authoring_bundle(bundle: bytes, workdir: str, expected_sha256: str
...
+            "styles": styles, "custom": custom, "raw": raw}
+
+
+def _write_brief(workdir: str, brief: dict) -> None:
+    path = Path(workdir) / _BRIEF_PATH
+    path.parent.mkdir(parents=True, exist_ok=True)
+    path.write_text(json.dumps(brief, ensure_ascii=False, indent=2) + "\n", "utf-8")
+
+
+def _capture_brief(run: "CompileRun", text: str) -> bool:
```
jacoblee-io, line 587: **[correctness] BRIEF.json is re-captured and overwritten on *every* message, so any later message that merely contains the phrase `我的定调标签` clobbers the real brief — often with an empty record.**

`_capture_brief` runs on every `/message` (`handle_message` → `_capture_brief(run, text)`), and `parse_brief_block` returns a dict whenever the marker *substring* is present (`:535`) — even with zero bullet lines it returns `{audience:"", styles:[], custom:[], raw:…}`, not `None`. `_write_brief` then unconditionally overwrites. There is no first-message / already-exists / has-at-least-one-parsed-field guard. So an owner who mentions the phrase mid-compile ("关于我的定调标签我想补充…"), or a relayed/quoted message, silently replaces the captured brief with a near-empty one — and BOX_ROLE tells the agent to read and follow BRIEF.json, so it then compiles against a blank brief. The test only exercises a no-marker message, so this is uncaught.

**Fix:** only capture on the first message of the run (or when BRIEF.json doesn't yet exist), and/or return `None` from `parse_brief_block` when no bullet field parsed.

kbc/platform/pod/compile_box.py
```diff
@@ -583,7 +715,29 @@ async def resolve_ticket(args):
...
+        path = Path(run.workdir) / QUESTIONS_PROPOSED_PATH
+        try:
+            existing = json.loads(path.read_text("utf-8")) if path.exists() else []
+            if not isinstance(existing, list):
+                existing = []
+        except Exception:
+            existing = []  # a corrupt/half-written prior file must not lose this round
+        merged, added, skipped = merge_proposed_questions(existing, incoming)
+        path.parent.mkdir(parents=True, exist_ok=True)
+        try:
```
jacoblee-io, line 774: **[correctness/robustness] Non-atomic write + corrupt-read-to-`[]` fallback silently drops all prior proposal rounds, contradicting the append-only guarantee.**

`QUESTIONS_PROPOSED.json` is written with a bare `write_text` (no temp + `os.replace`), and the read path falls back to `existing = []` on any parse error (`:729-730`, the comment acknowledges "a corrupt/half-written prior file must not lose this round" — but it does exactly that). A torn write (SIGTERM/OOM/disk-full mid-`write_text` during a long compile) leaves a truncated file; the next `propose_questions` reads `[]`, merges only the new round, and rewrites — every earlier question is gone (including ids the frontend already holds as `proposal_id`), while the call reports success. This directly breaks the documented "append-only across rounds, never wipes prior picks."

**Fix:** write atomically (temp file + `os.replace`). This is one of four inline non-atomic `authoring/*.json` writers (`_write_brief` `:559`, `propose_plan`, `resolve_ticket`, here) — extract a shared `_write_json_atomic(path, obj)` and route all four through it.

kbc/platform/pod/compile_box.py
```diff
@@ -506,6 +505,139 @@ def _install_authoring_bundle(bundle: bytes, workdir: str, expected_sha256: str
...
+        - 给谁看:内部工程师
+        - 内容倾向:详尽百科、保留内部信息、只留最新版本
+        - 自定义:偏排障场景
+        请按这些标签作为编译 brief 执行。
+
+    Parsing it in code (rather than trusting the model to transcribe it) keeps the
+    durable brief verbatim regardless of how the agent paraphrases downstream."""
+    if not message or _BRIEF_MARKER not in message:
+        return None
+    raw = message[message.find(_BRIEF_MARKER):].strip()
```
jacoblee-io: **[correctness/DoS] The brief `raw` field is sliced from the first marker occurrence to end-of-message — unbounded and verbatim into a file the agent is told to follow.**

`raw = message[message.find(_BRIEF_MARKER):].strip()` takes the *first* marker occurrence through EOF, with no size cap. Two consequences: (1) a marker mentioned in earlier prose locks the slice onto that point and pulls the whole unrelated tail into the durable `raw`; (2) an arbitrarily large opening message becomes an uncapped `raw` stored in BRIEF.json and folded into the agent's context (BOX_ROLE says to follow it) — a context-bloat channel. The sibling `_install_wiki_snapshot` already caps its input; this path doesn't. **Fix:** bound `raw` (e.g. slice to the block's own extent / a max length) and prefer the last marker occurrence, or reuse the structured fields and drop the raw tail.

kbc/platform/pod/compile_box.py
```diff
@@ -506,6 +505,139 @@ def _install_authoring_bundle(bundle: bytes, workdir: str, expected_sha256: str
...
+        _write_brief(run.workdir, brief)
+        return True
+    except Exception:
+        return False
+
+
+def _normalize_question(q: str) -> str:
+    """Dedup key for a proposed question: whitespace-stripped, trailing
+    punctuation removed, case-folded — so trivially-reworded repeats collapse."""
+    return re.sub(r"\s+", "", str(q or "")).strip("?？。.!！,，、").lower()
```
jacoblee-io: **[parity, plausible] `_normalize_question` can diverge from the frontend's normalization, re-opening the adopt-500 for exactly the pre-fix non-ASCII rows the id was meant to fix.**

For a *new* proposal the box stamps the id and the frontend uses it verbatim — parity is moot. But for pre-fix / id-less legacy rows the frontend *re-derives* the id (the documented fallback), and those are precisely the Chinese-question rows that hit the 500 before. This function uses Python `.lower()` (the docstring says "case-folded" — `.lower()` ≠ `.casefold()`), Python `re.sub(r"\s+")` (Unicode `\s` differs from JS `\s` on U+0085 / U+001C–U+001F / U+FEFF), and no NFC normalization. Any of these diverging from the JS side yields a different `q-<hash>` → the frontend POSTs a `proposal_id` the box never wrote → adopt/dismiss 500 recurs for non-ASCII questions.

**Fix:** pin cross-language parity with a shared test vector that includes non-ASCII + exotic-whitespace inputs (not just `"A?"`), or have pre-fix rows backfilled server-side rather than re-derived on the frontend. Also align the docstring with the actual `.lower()`.

kbc/platform/pod/compile_box.py
```diff
@@ -506,6 +505,139 @@ def _install_authoring_bundle(bundle: bytes, workdir: str, expected_sha256: str
...
+        qid = str(q.get("id", "")).strip() or _question_id(key)
+        merged.append({**q, "id": qid, "question": text})
+    added = skipped = 0
+    for item in incoming:
+        if not isinstance(item, dict):
+            skipped += 1
+            continue
+        text = str(item.get("question", "")).strip()
+        key = _normalize_question(text)
+        if not key or key in seen:
```
jacoblee-io, line 670: **[low] Dedup silently drops questions that differ only by trailing punctuation/case, and any all-punctuation question.**

`_normalize_question` strips edge punctuation and lowercases, so `删除?` and `删除。` collapse to one key — the second is counted `skipped` and its distinct `reference`/`source` are lost. An all-punctuation/whitespace question normalizes to `""` and is dropped by the `if not key` guard. Mostly acceptable as dedup, but for questions that legitimately differ only by trailing punctuation this is silent data loss rather than intended collapse — worth a comment or a secondary tiebreak if that case is real.

jacoblee-io (commented):
## Review: box protocol v3 (guided wizard)

The contract layer is clean: `propose_questions` schema ↔ handler match, the tool is registered in both `zh`/`en` tools.json and the allowed-tool set, `selfcheck`'s raw inventory only walks `raw/` (so `authoring/BRIEF.json` / `QUESTIONS_PROPOSED.json` / new CONTRADICTIONS entries aren't miscounted by the coverage ledger), never-block / turn_done / the S1 self-check are genuinely untouched, and the FNV-1a math is correct (`"A?" → q-e40c292c` verified). The findings below are on the Python box. Two are worth blocking on (both easy data-loss fixes). Inline.

**Lower-severity notes (not inline):** the four `authoring/*.json` writers (`_write_brief`, `propose_plan`, `resolve_ticket`, `propose_questions`) each inline a non-atomic `write_text` — the finding on QUESTIONS_PROPOSED is the sharpest instance, but the real fix is one shared `_write_json_atomic(path, obj)` (temp + `os.replace`) that all four route through.

kbc/platform/pod/compile_box.py
LikiosSedo: Fixed in c53c42cd. `_capture_brief` is now first-capture-wins — it returns early when `authoring/BRIEF.json` already exists — and `parse_brief_block` returns `None` when the marker is present but no tag field parses. A later message that merely quotes the phrase can no longer clobber the real brief with a near-empty record. New tests cover marker-without-fields (no write) and second-capture (does not overwrite the first).

kbc/platform/pod/compile_box.py
LikiosSedo: Fixed in c53c42cd. Extracted `_write_text_atomic` (same-dir temp file + `os.replace`, temp cleanup on failure) and routed all four authoring writers through it (`_write_brief`, `propose_plan`, `resolve_ticket`, `propose_questions`), so a torn write can no longer leave a half-file that the next read collapses to `[]`. Added a test for the atomic writer.

kbc/platform/pod/compile_box.py
LikiosSedo: Fixed in c53c42cd. The slice now anchors on the LAST marker occurrence (the wizard appends the brief block at the end of the message, so a marker mentioned in earlier prose no longer locks the slice) and is capped at `_BRIEF_RAW_MAX` (4000 chars). Both behaviors are covered by tests.

kbc/platform/pod/compile_box.py
LikiosSedo: Fixed in c53c42cd. `_normalize_question` no longer relies on `\s`: it strips an explicitly enumerated whitespace class (matching ECMAScript `\s`, i.e. including U+FEFF, excluding U+0085 / U+001C–U+001F), with a comment pinning it in lockstep with the frontend mirror. The docstring now says lowercase (not case-folded) to match the actual `.lower()`. Added a cross-language parity test vector (CJK + non-ASCII + BOM + exotic whitespace); the frontend side gets the same enumerated class plus the mirrored vector in its own repo, and I verified the derived ids are byte-identical across all vector entries via node.

kbc/platform/pod/compile_box.py
LikiosSedo: Documented as intentional in c53c42cd: the docstring now states that trailing-punctuation/case collapse is by design (questions differing only by trailing punctuation are treated as the same question) and that an all-punctuation question normalizes to the empty string and is dropped by the caller.

jacoblee-io (approved):
Approving on `c53c42c`. All findings from my review are fixed correctly and with tests:

- **[correctness] BRIEF.json overwrite** — `_capture_brief` is now first-capture-wins (writes only when BRIEF.json is absent) and `parse_brief_block` returns `None` when no tag field parses, so a later message that merely mentions the marker phrase no longer clobbers the real brief.
- **[correctness] unbounded brief raw** — now the LAST marker occurrence, capped to `_BRIEF_RAW_MAX` (4000).
- **[robustness] non-atomic writes** — extracted `_write_text_atomic` (temp + `os.replace`, cleanup-on-failure); all four `authoring/*.json` writers route through it, so a torn write can't drop prior proposed-question rounds. (`os`/`uuid` imports confirmed.)
- **[parity] question id** — `_normalize_question` replaces `\s` with an enumerated whitespace class that matches ECMAScript `\s` (includes U+00A0/U+1680/U+2000–200A/U+2028/U+2029/U+202F/U+205F/U+3000/U+FEFF, excludes the U+0085/U+001C–1F that Python `\s` would wrongly strip), docstring corrected to `lowercase`, and a cross-language id vector (CJK + non-ASCII + BOM/whitespace) is pinned by a test the frontend mirrors. Verified the class matches JS `\s`.
- **[low] dedup** — the intentional punctuation/case collapse is now documented.

No regressions in the fix commit; CI green (KBC Box / Portal Web / Test / Type Check). Only residual is the theoretical NFC precomposed-vs-decomposed edge in id normalization, fine to defer. No cross-repo deploy gate here (box-internal changes).

PR #383, opened 2026-07-04
kbc: swappable engine seam + Codex (subscription) as a second compile-box engine

LikiosSedo:
## What

A second, swappable engine for the KB compile box: **OpenAI Codex CLI (subscription-backed)** behind a new `EngineSession` seam. The Claude engine remains the default and is **behavior-identical** — selecting it means nothing changed (all pre-existing suites pass with untouched assertions).

Design: `DESIGN-kb-box-codex-engine-2026-07-02.md` (workspace doc). Stacked on `feat/kb-box-protocol-v3` → `feat/kb-test-box` → `feat/kb-capability-platform` (#378).

## Shape

```
kbc/platform/pod/
  compile_tools.py        # the 4 signal-tool BODIES, engine-neutral (moved from compile_box)
  engines/
    base.py               # SessionSpec + EngineSession protocol + pure C4 guard predicate
    claude.py             # pure relocation of the SDK code (default engine)
    codex.py              # NEW: codex exec --json driver, auth bootstrap, guardrails
  mcp_compile_server.py   # NEW: stdlib stdio MCP server → loopback callback → same tool bodies
  transform.py            # NEW: PDF → per-page PNG at /sources install (flag-gated, off for claude)
  Dockerfile.codex        # NEW: per-engine image (pinned codex 0.139.0 musl binary, no node)
src/gateway/agentbox/box-profile.ts   # NEW kb-compile-codex profile (additive)
```

- **Wire contract unchanged**: no engine field anywhere; `/session` body, SSE vocabulary, allowedTools passthrough all identical. A consumer switches engines by requesting the `kb-compile-codex` BoxProfile instead of `kb-compile` — profile difference → respawn, which "changing engines" should be. sicore needs zero changes until it wants to offer the choice.
- **Engine selection is an image property** (`KBC_ENGINE` baked into the per-engine Dockerfile); adapters import lazily so the codex image ships no claude-agent-sdk and vice versa.

## Codex adapter decisions

- **`codex exec --json` / `exec resume <thread> -` (one subprocess per turn, prompt over stdin)** rather than the beta Python SDK: exec is the surface the desktop spike validated against the pinned CLI (0.139.0); the SDK wraps the same event vocabulary, so swapping later is contained to `engines/codex.py`.
- **Auth invariant**: the box holds ONLY a short-TTL `CODEX_ACCESS_TOKEN` (spawn-time env via the profile). `auth.json` is written with an **empty refresh_token** + `last_refresh=now`, so the box can never rotate the subscription's refresh token out from under sicore (the single refresher). Token failure = explicit `error` event, never a silent retry burning subscription quota. `CODEX_API_KEY` wins when present (official automation path; zero architecture change when a key materializes).
- **Turns serialize**: codex has no mid-turn steering; a `/message` landing during a live turn queues as the next turn. Never-blocks semantics hold — the reply rides the following turn.
- **Guardrails for the missing `max_turns`**: per-turn wall-clock (`KBC_CODEX_TURN_TIMEOUT_SECS`, 3600) + event cap (`KBC_CODEX_MAX_TURN_EVENTS`, 10000); every path (ok/fail/timeout) still closes the turn — the never-stuck invariant holds.
- **Signal tools byte-identical across engines**: bodies live in `compile_tools.py`; claude wraps them in-process, codex reaches them via a stdio MCP server whose `tools/call` POSTs back to a loopback-only, per-session-token-gated listener running the same bodies with the live run's emit.
- **`POST /test-session` on a codex box → 501**: codex's sandbox limits writes, not reads — there is no hook equivalent for the C4 snapshot-read confinement. Refusing loudly beats serving an unguarded test session; the structural fix (dedicated snapshot-only kb-test box) is the planned P4 slice.
- **Multimodal leveler**: codex cannot read PDFs → `/sources` renders each PDF page to `<file>.pages/page-NNN.png` (pypdfium2, codex image only; `.png` is not a TEXT_SOURCE_EXT so the coverage ledger is untouched; one corrupt PDF skips itself, reported in the response, never blocks the corpus).

## Verification

- `test_compile_box.py` (14 + protocol smoke), `test_selfcheck.py`, `test_redblue.py`: **green, assertions untouched** — the claude-path no-regression evidence.
- NEW `test_engines_codex.py`: 11 tests against a **fake codex binary** (no LLM, no real CLI) — auth materialization invariants, JSONL event mapping + thread resume, stdin prompt delivery, failure/timeout guardrails, callback listener auth, MCP stdio protocol, full `run_session` e2e on `KBC_ENGINE=codex`, the 501, and render gating (real pypdfium2 render exercised).
- `box-profile.test.ts` 5/5, `tsc --noEmit` clean.
- Both images build: `Dockerfile` (claude, updated COPY set) and `Dockerfile.codex` (pinned binary fetch verified).
- CI: `kbc-ci.yml` now runs the codex suite too.

## Not in this PR (known follow-ups)

- **P3 auth wiring**: sicore-side short-TTL token minting → runtime envForward is consumer-side; the box-side contract (`CODEX_ACCESS_TOKEN`, expiry = explicit error + respawn) is final here.
- **P4 kb-test on codex**: structural snapshot-box isolation.
- **P5 quality re-validation**: en/zh × authoring scoring matrix + prompt-dialect tuning (incl. ledger-contract fidelity on codex) before production traffic.
- Codex sandbox on cluster kernels without Landlock: `KBC_CODEX_SANDBOX=danger-full-access` knob documented (pod remains the sandbox — same trust model as claude's bypassPermissions).


LikiosSedo:
Rebased this branch onto the deduplicated chain: it now sits on `feat/kb-box-protocol-v3` → `feat/kb-test-box` → `feat/kb-capability-platform`, so the former sibling topology (a duplicated `capability.test*` commit) is gone. The system-prompt half adapted cleanly — `run_session`/`test_session_driver` build box_role/test_role from the locale packs (`prompts/<locale>/*.md`) and pass the text through `SessionSpec.system_prompt`. Two follow-ups landed on top of the rebase:

1. **Stable proposal-id, ported into the engine-neutral tool layer (correctness, not style).** `compile_tools.merge_proposed_questions` now stamps every entry with `id = "q-" + fnv1a32(normalize(question))[:8]` (normalize = strip all whitespace + strip leading/trailing punctuation + lowercase), the same formula the frontend derives — a missing id makes adopt/dismiss POST an empty `proposal_id` and sicore returns 500. Locked against the canonical vector `"A?" → "q-e40c292c"` in the tests.
2. **Codex cancel-path subprocess reap.** `_run_turn`'s finally nulled `self._proc` before `close()`'s kill check ran, so a torn-down turn could orphan a live `codex exec` process burning subscription quota; the finally now terminates + bounded-waits before nulling.

**Two things to know before merging:**

- **locale-pack alignment for the tools is a deliberate pre-merge TODO.** This was a *mechanical* rebase: `compile_tools.py`'s tool descriptions + result strings and the read-confinement guard's deny message are kept **zh-only inline**, NOT wired to `prompts/<locale>/tools.json` the way the system prompts are. An English KB served by the Codex engine would get Chinese tool docs. There's a `TODO(locale)` block at the top of `compile_tools.py` spelling out the fix (a shared pack loader that both `compile_box.py` and `compile_tools.py` import to avoid the import cycle, plus threading the run's locale — including via an env var at spawn time to reach the separate `mcp_compile_server` process). Deferred because the Codex engine is experimental and not end-to-end verifiable here (no CLI/subscription).
- **Merge-order dependency with #384 (batch compile).** This branch predates batch mode, so its refactored engine-event relay has no `_suppress_turn_done` branch. If #383 merges **after** #384, the relay must re-add that branch (batch mode's single-`turn_done`-per-run contract: per-batch turns park their reply, and one `turn_done` fires after the cross-batch final pass) or a batch compile on the codex/claude seam will emit a `turn_done` per internal batch session.

Verification: `test_compile_box.py` / `test_engines_codex.py` / `test_selfcheck.py` / `test_redblue.py` green; `box-profile.test.ts` green; `tsc` clean.

PR #384, opened 2026-07-05
kbc+gateway: batch compile for large corpora (bounded sessions, file handoff)

LikiosSedo:
> **Review guide.** This branch carries the KB compile pipeline's evolution as one integration unit — the subsystems interleave in `compile_box.py` (routing, seam ordering, shared session driver) and share one test suite, so they are presented per-subsystem below for review, with commits, tests, and live evidence each. The #378/#385/#382 content visible in the diff arrived via merge commits from `main` and is already reviewed/merged there.

## 1. Batch compile for large corpora (`batching.py` + orchestrator)

**Problem.** Autocompact is disabled (the model gateway rejects the `context_management` field), so a large-corpus compile accumulates every Read result in one session and deterministically overflows the model window mid-run. Large inputs are a product requirement, so sessions must be bounded by construction rather than saved by compaction.

**Mechanism.**
1. Pre-flight (pure code): corpus bytes < `KBC_BATCH_THRESHOLD_BYTES` (400KB default) → the single-session path, byte-identical. Small KBs never batch.
2. Planning: deterministic baseline (top-level-dir grouping, greedy packing within `KBC_BATCH_BUDGET_BYTES` on *effective* sizes — images flat-cost, PDFs per page; oversized files solo); a bounded planner session may regroup topically but only replaces the baseline if code validation passes (every source exactly once, unique batch ids, budgets respected) and a fragmentation guard holds.
3. Batch loop: each batch is a fresh session over the same workspace — same role/tools/conventions. The handoff is the workspace itself (exact files), never a prose summary: `unaccounted=0` from the coverage ledger is code-checkable, a compaction summary is not.
4. One logical turn: per-session `turn_done` is suppressed; the single `turn_done` fires after the cross-batch final pass, so the consumer sees an ordinary turn; the full-corpus SELFCHECK + bounded repair rounds run over the final state. An orchestrator error also closes the turn (resumable `turn_done`), keeping never-block symmetric.
5. Crash-resume at batch granularity: `BATCH_PLAN.json` carries done stamps (synced to the durable store per batch); resume prunes sources deleted from `raw/` since the plan was pinned. Mid-batch owner chat queues onto the next directive; notes arriving during the tail phases get a bounded digest session before the turn closes.

Commits: `c94167c3` `54eb3a8c` `7eadcbe3` `9fd267c3` `d08b3d80` `8ef7b3f7` `b8d07245`. Tests: `test_batching.py` (14), orchestrator contract + review-fix scenarios in `test_compile_box.py`. Live: a 139-file / 6.6MB corpus compiled end-to-end in 28 batches (later 10–14 with effective-size packing) and published on the test cluster; kill-pod resume exercised.

## 2. Gateway: two-clock watchdog + kb `/work` 4Gi

`streamPath` gains an opt-in `onComment` hook; box `: heartbeat` comments bump a SEPARATE heartbeat clock (`touchHeartbeat`), and `stalenessOutcome` reaps a running run when data has been silent past `dataStaleMs` (60min, heartbeats can't extend it) OR when there is no liveness at all past `staleMs` (10min). A healthy compile in a long read-only phase is no longer reaped at 10min, and a wedged box that only heartbeats still dies. `dataStaleMs` is clamped to `>= staleMs`; a non-positive env override falls back to the default. kb box `/work` cap 1Gi→4Gi (both kb profiles move together, preserving the identical-box-shape invariant).

Commits: `d02d427a` `4a409114` `b8d07245`. Tests: run-manager two-clock + clamp suite; client SSE routing (incl. spec-valid space-less `data:` frames).

## 3. Scoped incremental recompile (`incremental.py`)

A compile trigger accompanied by a machine-computed changeset (`authoring/RAW_CHANGES.json`, written by the consumer: added/modified/deleted + unified diffs) routes to a scoped path: affected pages are reverse-looked-up via `compiled_from`, `CHANGESET.json` is materialized, the directive touches ONLY those pages, and a post-turn byte-integrity guard forces a repair if any page outside the authorized set changed — "leave the rest untouched" is enforced, not hoped. Missing/corrupt input → full-compile fallback (backward compatible).

Commits: `7f4d9183` `6a073e92` `c544b466` `5db57e07` `ddc9d161` `cbf157a2`. Tests: `test_incremental.py` + route/guard scenarios in `test_compile_box.py`.

## 4. Blind image-claim verification (`mediaverify.py`)

Root cause: the only repeated real faithfulness failures in A/B evaluation were image-derived numbers (axis/column mix-ups). Same-session re-reading could not catch them (confirmation bias — the checker sees the page's claim). The fix is structural: a fresh-context transcription session sees ONLY the image (never the page), a text-only comparison session sees the page + transcript (never the image), and findings feed a bounded repair session. Transcripts are cached by image content hash; chunked ≤`KBC_MEDIA_VERIFY_MAX_IMAGES` per round to respect API limits.

Commits: `27bdbd12` `634da6d9` (also pins the root-caused `context_management` 400 kill switch: `CLAUDE_CODE_DISABLE_EXPERIMENTAL_BETAS=1`) `b64e4cbb`. Tests: `test_mediaverify.py` + wiring in `test_selfcheck.py`. Live: caught a cross-image hardware-model misattribution and a bar-color column swap on real corpora.

## 5. Red-blue PK wired into the compile flow (`redblue.py`)

Post-convergence quality gate, asynchronous and off the logical turn: once the ledger closes and media verification is clean on the settled draft, a bounded PK run generates questions from raw (scaled to corpus size, cap `KBC_PK_QUESTIONS_MAX`), answers them against the draft, judges with the reference in hand, and drives at most one repair round + one targeted re-test. Judges that omit ids get one re-judge; still-ungraded questions are reported separately and never trigger repair. Wall-clock timeout salvages partial results (`state=partial`). Ops kill switch `KBC_PK_MODE=off` outranks everything.

Commits: `b18b3977` `dd5bdc87`. Tests: `test_redblue.py` (12) + PK wiring in `test_selfcheck.py`. Live: full 24-question PK loop with repair + re-test on the test cluster.

## 6. Session resilience: model-stall watchdog, rate-limit backoff, shutdown flush

A black-holed model request (no stream activity, no tool pending) is interrupted and re-issued with bounded retries, then fails fast (`ModelStallError`) instead of hanging the box; 429/529 result-messages back off and re-issue instead of ending the turn; SIGTERM final-syncs active runs' unsynced workspace. A live-but-slow stream and a long-running tool are never mistaken for stalls.

Commits: `357c7357` `5220a91a`. Tests: six stall/rate-limit/flush scenarios in `test_compile_box.py`.

## 7. Pipeline quality pass: media ledger, lint, converge phase

The coverage ledger now accounts for EVERY raw file (media included: compile it or exclude it — no silent `derived:true` escapes); lint gains orphan-page (index BFS), uncited-body-source, and duplicate-candidate checks; U+FFFD charset corruption in candidate pages is flagged with path+context; the final directive carries the code-computed duplicate/orphan/uncertainty inventory. `converge_phase` (ledger → media verify → PK → settled) is a durable, code-owned signal — authoritative for verify-on AND verify-off, so the consumer gates its test step on `settled` with no config lookup.

Commits: `8f798889` `2f1e6b7d` `c851f8b5` `f07c7b54`. Tests: `test_selfcheck.py` (extensive) + charset/lint units.

## 8. Consumer-managed LLM credential + ops knobs + release train

The consumer can bind per-role model choices + credentials for the KB capability: they arrive on the `/session` body (`llm` + `settings`), are applied in-process via an env whitelist (`_apply_session_config`) — the token never lands in the PodSpec. `KBC_*` runtime env forwards into kb boxes by prefix for ops knobs (PK on/off, budgets, model tiers) WITHOUT rebuilding the box image; secret-shaped names are refused at the spawner (defense-in-depth; credentials never travel via env). Helm gains declarative `runtime.llm` (endpoint + token Secret) and the compile-box image joins `make docker` with the same tag convention. Instruction-less hosting starts begin `idle` so they are not reaped as stuck turns.

Commits: `2bc3b162` `075989bf` `1adaf1b7` `98223f94` `c2528fca`. Tests: `test_apply_session_config`, spawner forwarding/refusal suite, materialize/contract tests. Live: end-to-end on the test cluster — ten env keys verified inside the SDK subprocess, credential flowing from the consumer only.

---

**Verification (whole branch, post-merge with main):** all 7 kbc pod suites green, `tsc --noEmit` clean, gateway+portal vitest 4191 passed. Needs a `kbc-compile-box` image rebuild + runtime redeploy to take effect (deploy notes in the ops runbook).


LikiosSedo:
Rebased this branch onto the deduplicated chain: `feat/kb-test-box` → `feat/kb-box-protocol-v3` → this branch now sit on top of the `feat/kb-capability-platform` head, so the former sibling topology (a duplicated `capability.test*` commit) is gone and every downstream prompt edit was re-targeted from the old inline role strings into the locale packs (`prompts/{zh,en}/*.md` + `tools.json`). One behavioral fix on top of the heartbeat commit: the previous `onComment → manager.touch(runId)` made a wedged box immortal (heartbeats fire exactly when no data flows), so it is replaced by a two-clock watchdog — `touch()` tracks the last data event, `touchHeartbeat()` a separate heartbeat clock, and a RUNNING run is reaped when data is silent past `dataStaleMs` (60 min, env `CAPABILITY_DATA_STALE_MS`) OR when there is no liveness at all past `staleMs` (10 min). Heartbeats bridge a long read-only phase, but a box that never sends data again still dies at 60 min. Full `npm test` green (4094 passing) and `tsc` clean.

LikiosSedo:
Pushed 5 new commits (9fd267c3..b18b3977) extending this PR beyond batch mode into compile-quality hardening:

- `8ef7b3f7` / `661288a0` — SDK stdio buffer 16MB; dead autocompact guard + SDK pin + natural batch-resume trigger (already deployed as buffix2)
- `8f798889` — **pipeline quality pass**: media joins the coverage ledger (images/PDFs must be compiled-from or excluded; kills the `derived: true` escape), new lint rules (orphan pages unreachable from index.md, body `(source: X)` citations missing from compiled_from), deterministic dup-candidate worklist fed to the final batch pass (merge-or-exempt), and a one-shot image numeric re-verification turn after the draft settles (both real fidelity failures in the batch-vs-oneshot A/B were chart misreads)
- `d08b3d80` — batch triggers back to machine-canonical strings only (the natural-language "继续编译" prefix had a false-positive edge; the web resume button speaks the canonical directive)
- `b18b3977` — **red-blue PK (S2) wired into the compile flow**: async post-check over a pinned snapshot after ledger + image re-verify settle, one bounded repair turn + one targeted retest, wall-clock salvage to state=partial, question budget 1.0/page capped at 24, chart-transcribed pages steer the question officer. Gates: `KBC_PK_MODE=auto|off` (default auto), `KBC_PK_REPAIR_ROUNDS=1`.

All four pod suites green (`test_selfcheck` / `test_compile_box` / `test_batching` / `test_redblue`). Deployed to siflow-test as `kbc-compile-box:pk1-b18b3977` for live validation.

LikiosSedo:
Three more commits (b18b3977..b64e4cbb), all driven by live findings from the first full-pipeline runs on siflow-test:

- `27bdbd12` — **image-verify hardening**: one verify session reading ~35 images tripped the API image-processing limits (~15 images silently dropped), so verify rounds are now chunked (≤`KBC_MEDIA_VERIFY_MAX_IMAGES`=8 per round, remainder rolls to the next settled turn) with one-Read-per-image discipline; plus column-alignment rules in the transcription discipline.
- `634da6d9` — **root-caused the bedrock `context_management` 400** by reading the bundled CLI (2.1.191) request builder: the rejected field is the thinking-clear context edit riding the experimental context-management beta (massapi masquerades as first-party → beta auto-enabled; adaptive thinking decides per-turn → intermittent 400s). Kill switch: `CLAUDE_CODE_DISABLE_EXPERIMENTAL_BETAS=1`, set in compile_box + the PK engine adapter + image-level ENV. `DISABLE_AUTOCOMPACT` (the old mitigation) was the wrong knob and stays only as belt-and-braces.
- `b64e4cbb` — **blind transcription + mechanical comparison (图像复核 v2, new `mediaverify.py`)**: claim-in-context re-reading proved to be a confirmation check — it passed 'GPU-Util 87-96%' (the MEM column) and a cross-image 'H20' model attribution (pixel-contradicted). Now a fresh vision session transcribes each image seeing ONLY the image (cached per content hash in `authoring/MEDIA_TRANSCRIPTS.json`), and a text-only session compares page claims against transcripts without ever opening the image, reporting 不一致 (misreads) and 超出转写范围 (inference disguised as a sourced fact). Findings come back as one repair turn. Also: the PK judge dropping ids from a verdict chunk (live: 4/24, twice) now gets one dedicated re-judge; persistently missing ids are 'ungraded' — excluded from failures, the repair queue and the pass-rate denominator.

Five pod suites green (incl. new `test_mediaverify`); deployed to siflow-test as `kbc-compile-box:pk4-b64e4cbb`. Full-pipeline live evidence: 139-source KB compiled with ledger closed (133 cited + 6 excluded), zero lint violations, chunked image verify, PK 24 questions with repair→targeted-retest closing 5 real failures (all 20 graded questions passing).

LikiosSedo:
Added one commit: `fix(kbc): flag U+FFFD charset corruption in candidate pages (path + body)`.

**Why here:** it extends `selfcheck.py`'s `lint_candidate` (the media-ledger + body-citation lint this PR introduced), so this branch is its only clean base.

**What it fixes:** an upstream lossy UTF-8 decode (a multibyte char split at a stream chunk boundary in the massapi gateway SSE) can write U+FFFD into emitted pages. The coverage ledger already flags a corrupted *path* (dangling + unaccounted), but the same corruption in *body prose* is invisible to the ledger and shipped silently. `lint_candidate` now scans each page's full text for U+FFFD and emits a `charset_corruption` violation, riding the existing lint→repair→converge machinery: `lint.ok=False` blocks `state=passed` so a corrupted draft can't settle/publish; the bounded repair turn gets a targeted 'restore from raw' directive.

**Verified on siflow-test 2026-07-07:** 需/础/成 each mangled to 3× U+FFFD across three pages (one path, two prose). Guard detects all; after a box respawn on the guarded image, the repair round healed both prose pages (converge settled, 0 U+FFFD). Unit + real-data tested.

Root cause is upstream (gateway SSE per-chunk decode should buffer across chunks) — tracked separately; this is the box-side integrity guard.

LikiosSedo:
Added two more commits — the box-side (write half) of the in-box verify converge (校对收敛门):

- `c851f8b5` feat(kbc): durable converge-phase signal for in-box verify (kills the phantom)
- `f07c7b54` feat(kbc): converge_phase settles when nothing pending (verify-off authoritative)

**What it does:** the box writes a durable `converge_phase` (verifying → revising → settled) into `SELFCHECK.json` around the post-turn verify loop (the red-blue PK / media re-check already wired in this PR). It's the authoritative, reload-surviving signal the frontend reads to keep the generate step 'in progress' through 校对/修订 and unlock the test step only once converged — replacing run_status, which was the root of the 'box looks idle yet still working' phantom. Additive: a field write + hook calls, no control-flow change (the never-stuck turn/repair logic is untouched); an absent phase reads as converged, so it's backward-compatible.

**Why here:** the converge hooks attach to this PR's verify flow (`_run_pk_flow` / `_run_media_verify_flow` / `_maybe_start_*`), so this branch is the clean base. Cherry-picked cleanly; unit + async wiring tests green (`test_pk_wiring`, `test_seam_settles_when_nothing_pending`, `test_converge_phase_helper`).

**Paired with:** the sicore read half is a separate MR (`feat/kb-verify-converge-gate` → develop) which gates the test step on `convergeSettled`. The two halves are backward-compatible independently. Deployment-proven on siflow-test (converge_phase=settled observed on a live compile).

LikiosSedo:
Added one more commit — the red-blue verdict stall fix (orphaned bug fix for this PR's red-blue PK, previously in no PR):

- `dd5bdc87` fix(kbc): red-blue verdict stall — de-agentify judge + contain per-item failures

**The bug (DB forensics on a real run):** PK_RESULT showed 24 questions all answered by blue, 0 verdicts. The judge was agentic (roots=[wiki, raw], max_turns 40), re-reading raw → it stalled on the gateway; and `asyncio.gather` without `return_exceptions=True` meant the first chunk's timeout aborted every sibling and discarded already-completed work.

**Fix:** de-agentify the judge — grade ONLY from the inlined `expected` rubric, fenced out of raw/wiki, gate tier; per-item containment (each blue/judge wrapped in try/except → recorded as ungraded, both gathers use `return_exceptions=True`); a `_finalize` helper that salvages graded results into passed/partial/unconverged instead of losing them to one timeout. Tests green (`test_redblue.py`: wall-timeout partial salvage, per-item containment, de-agentified judge).

Belongs here because it fixes redblue.py, which this PR introduced (`b18b3977`). Only touches redblue.py + its test — no overlap with the converge/charset commits.

LikiosSedo:
Added the box-side (executor) half of scoped incremental recompile (4 commits — DESIGN-kb-incremental-recompile-v2-2026-07-07):

- `7f4d9183` M1+M2: `incremental.py` deterministic core (reverse-look-up affected pages via compiled_from = the dependency index; git-style 3-category CHANGESET with unified-diff passthrough; byte-integrity guard) + protocol (sicore's `authoring/RAW_CHANGES.json` → enriched `authoring/CHANGESET.json`) + `compile_box` routing (`请增量重编` + a changeset → scoped path, precedence over the batch/full route).
- `6a073e92` M3: post-turn byte-integrity guard — pages changed outside the authorized set (affected ∪ declared added-targets ∪ index) force a repair even with a clean ledger. 'Leave the rest untouched' enforced by sha256, not the model's word.
- `c544b466` M4: incremental contract in box_role (zh/en).
- `5db57e07` fix: defensive getattr (full-suite regression).

**Fixes** the siflow-test finding that `请增量重编` routes to a full-corpus batch re-plan (not incremental). **Backward compatible**: no RAW_CHANGES → full compile fallback, so this is dormant until the sicore half lands. Ownership: 'can it compile' is sicore's (single-flight lock, separate track); 'how to incrementally compile' is the box's (this). All 6 box test files green (test_incremental 8, test_compile_box 27, + selfcheck/redblue/batching/mediaverify).

LikiosSedo:
Integrated `main` (post #378/#385/#382) via a merge commit (`80bdbd50`) — 34 commits on this branch made a rebase replay impractical, so the reconciliation is a single resolution point. What the merge does:

- **compile_box.py**: takes main's #382 review fixes verbatim (first-capture-wins brief + zero-field parse guard, `_write_text_atomic` on all four authoring writers, last-marker + capped brief raw, enumerated-whitespace id-parity class); keeps this branch's batch/incremental routing; two genuine fusions — the repair-prompt composition (branch's ledger + integrity parts now pass `locale`) and the post-repair seam (main's bilingual error message + branch's converge-phase chain).
- **run-manager.ts**: the branch's two-clock watchdog (heartbeats bridge data silence up to `dataStaleMs`, full silence dies at `staleMs`) is folded INTO main's `stalenessOutcome` helper, so the poison-row deferral machinery from the #385 review judges staleness by exactly the same rule. Adopted main's `setSessionRef` removal (workspace-file continuity; no callers remain).
- **Dockerfile**: pinned SDK (0.2.110) + `*.py` glob COPY kept; main's office-ingest deps (python-pptx/openpyxl/python-docx) added.
- **tests**: pure union — both sides' suites run together (main's #382 review-fix suite + snapshot-installer size guards + pack-hash ordering, alongside this branch's session-config/incremental/batch/model-stall/media-verify suites).

A follow-up commit (`d6b344ee`) genericizes the consumer references this branch's comments/docstrings had introduced ("the consumer" / 消费方), matching the repo convention; no pre-existing main lines and no wire tokens touched.

Verified locally: all 7 kbc pod suites green (incl. the #382 review-fix vector), `tsc --noEmit` clean, full vitest 4187 passed.

kbc/platform/pod/batching.py
```diff
@@ -0,0 +1,275 @@
...
+    b = budget if budget is not None else batch_budget_bytes()
+    errors: list[str] = []
+    sizes = {i["path"]: _eff(i) for i in inventory}
+    seen: dict[str, str] = {}
+    batches = plan.get("batches")
+    if not isinstance(batches, list) or not batches:
+        return ["plan has no batches"]
+    for batch in batches:
+        sources = batch.get("sources")
+        bid = str(batch.get("id", "?"))
```
jacoblee-io, line 223: **[correctness] `validate_plan` doesn't enforce batch-id uniqueness, so a model-proposed plan with two batches sharing an `id` silently drops an uncompiled batch.**

Coverage here is checked per *source path* (via `seen`), never per batch `id`. Two batches with the same `id` both pass validation. Downstream, `stamp_done` (compile_box.py:1711) marks *every* batch with the matching id as done, so completing one stamps both — the second, uncompiled batch disappears from `pending_batches` and is never run. It's only caught later by the coverage ledger (its sources come back `unaccounted` → a repair round), i.e. after wasted work and a confusing state. **Fix:** add an id-uniqueness (and non-empty-id) check to `validate_plan` — reject a plan with duplicate batch ids the same way it rejects a duplicated source.

src/gateway/agentbox/k8s-spawner.ts
```diff
@@ -248,11 +248,25 @@ export class K8sSpawner implements BoxSpawner {
...
+    // them via the runtime deployment env instead of rebuilding the box image).
+    const forwarded = new Set(env.map((e) => e.name));
+    const forwardOne = (name: string, value: string | undefined) => {
+      if (value !== undefined && value !== "" && !forwarded.has(name)) {
+        forwarded.add(name);
         env.push({ name, value });
       }
+    };
+    for (const name of profile.envForward ?? []) {
+      if (name.endsWith("*")) {
```
jacoblee-io, line 269: **[security] The new `KBC_*` prefix env forwarding widens the trust boundary via trust-by-naming.**

The trailing-`*` form forwards *every* runtime env var with that prefix into the box PodSpec (`Object.entries(process.env)` → `env.push({name, value})`). Any value that happens to be `KBC_`-prefixed in the runtime env — a credential parked there, or an over-broad var — lands in the pod spec in cleartext, readable by anyone with pod-get in the box namespace. The box-side whitelist only guards the `/session` body, not this env path, so it doesn't cover the leak. **Suggestion:** forward an explicit allowlist of `KBC_` knob names rather than a blind prefix glob, or document that no secret may ever be `KBC_`-prefixed on the runtime.

(Note: the `ANTHROPIC_*` forwarding a few lines up is pre-existing and intended — the box runs the Claude Agent SDK and the pod is the sandbox — so this comment is only about the new prefix-glob.)

kbc/platform/pod/batching.py
```diff
@@ -0,0 +1,275 @@
...
+
+
+def scan_sources(raw_dir: Path) -> list[dict[str, Any]]:
+    """Inventory of raw sources: repo-relative path + size, sorted for stable
+    plans. Hidden files and empty files are skipped (they carry no compile
+    content; the installer never writes them, but a defensive skip is cheap)."""
+    items: list[dict[str, Any]] = []
+    if not raw_dir.is_dir():
+        return items
+    for p in sorted(raw_dir.rglob("*")):
```
jacoblee-io, line 105: **[defense-in-depth] `scan_sources` follows symlinks (no `is_symlink()`/realpath confinement).**

`raw_dir.rglob("*")` + `p.is_file()` + `p.stat()` all follow symlinks, so a symlink under `raw/` pointing outside would be inventoried (and its target size/bytes read), and a directory-symlink cycle could hang/error `rglob`. Largely closed upstream today — `raw/` is the read-only frozen snapshot and the source-bundle installer rejects non-file/non-dir tar members — so a symlink shouldn't exist in `raw/`. But this is the same class the S1 pinner was hardened against (`f.resolve().relative_to(...)`); mirroring that realpath-confinement here keeps the two scan paths consistent and robust if `raw/` ever becomes writable.

jacoblee-io (commented):
## Review: batch compile — please split this PR

**Scope is the headline concern.** The description covers three things (batch compile, the gateway heartbeat fix, `/work` 1Gi→4Gi), but the branch diff is a ~17k-line omnibus. Beyond the already-reviewed stacked work (#378/#379/#385/#382, present as merge commits), it also bundles **at least these undescribed subsystems**:

- **scoped incremental recompile (M1–M4)** — `incremental.py` (new, ~269L) + a second compile route in `compile_box.py`
- **media / image-claim verification** — `mediaverify.py` (new, ~238L)
- **red-blue PK rework wired into the compile flow** — `redblue.py` (~697L, `KBC_PK_MODE=off` kill switch)
- **model-stall watchdog** and **graceful-shutdown flush**
- **consumer-managed LLM credential + helm `runtime.llm` + `KBC_*` env forwarding**

A 3-line description over 7+ substantial subsystems can't be reviewed honestly as one unit. **Recommend splitting** into per-subsystem PRs (batch / incremental / mediaverify / redblue / LLM-credential+helm), each reviewable on its own.

**Coverage disclosure:** I reviewed only the *described* scope — the batch-compile mechanism (`batching.py` + the orchestration in `compile_box.py`) and the gateway heartbeat. **`incremental.py`, `mediaverify.py`, and the `redblue.py` rework are NOT reviewed here.**

Two notes on the described parts that checked out:
- **Gateway heartbeat is well-designed** — it uses a *separate* clock (`touchHeartbeat` / `lastHeartbeatMs`, not `touch`), and `stalenessOutcome` still reaps a `running` run after `dataStaleMs` (60min) of *data* silence regardless of heartbeats, so a heartbeat-only wedged box is still reaped. SSE comment-vs-data routing and cross-chunk buffering are correct.
- **`batching.py` core is sound** — greedy packing never exceeds budget, oversized-solo works, and `validate_plan`'s coverage check is set-based (catches the drop-A/add-B-twice swap). Deterministic (`sorted(rglob)`), so crash-resume matches.

Inline: the two described-scope issues worth fixing + lower-severity notes.

**Lower-severity (not inline):**
- **Batch crash-resume robustness cluster:** the `stamp_done` write reaches the durable store only via the periodic/next-batch sync, so a crash in that window *re-runs* an already-done batch (safe direction — never skips a not-done one); resume reuses the plan without rescanning `raw/`, so a source *deleted* between runs leaves a batch directive pointing at a missing file (added sources are caught by the coverage ledger, deletions aren't); mid-run owner notes arriving during the tail phases (ledger-repair / media-verify) are acked as "will pass to later batches" but then abandoned when `turn_done` fires — contrary to the "rides the next directive" contract.
- **never-block asymmetry:** on an orchestrator error the batch path emits `error`, not `turn_done` — a consumer that ends the logical turn only on `turn_done` could hang (the runtime's `endRun(failed)` still terminates the run, but the single-logical-turn signal is separate).
- **gateway `dataStaleMs` (run-manager.ts:119):** `Number(...) || 60min` swallows an explicit `0`, and the documented `dataStaleMs >= staleMs` invariant has no runtime guard — an operator setting it below `staleMs` silently disables the heartbeat bridge. Also `streamPath` now generalizes to arbitrary SSE producers but requires `data: ` with the space; a space-less (SSE-valid) `data:` line would be misrouted as a comment (latent — today's box always sends the space).

kbc/platform/pod/batching.py
LikiosSedo: Fixed in b8d07245. `validate_plan` now rejects duplicate batch ids and empty/missing ids the same way it rejects a duplicated source, with a comment explaining the stamp_done twin-stamp hazard. Unit test covers both cases.

src/gateway/agentbox/k8s-spawner.ts
LikiosSedo: Fixed in b8d07245, taking the belt-and-braces middle: the prefix glob now refuses secret-shaped names (`TOKEN|SECRET|PASSWORD|CREDENTIAL|API_?KEY|PRIVATE`, case-insensitive, matched on the part after the prefix) with a warning log, and the invariant is documented at the forwarding site — forwarded prefixes are ops knobs only; credentials reach the box via the `/session` body, never env. Explicit (non-glob) names remain the profile's deliberate declaration and are untouched. Spawner test asserts `KBC_PK_MODE` still forwards while `KBC_MASSAPI_TOKEN` / `KBC_WEBHOOK_SECRET` / `KBC_SIGNING_API_KEY` are refused.

(A strict allowlist was considered and rejected: the point of the glob is adding an ops knob via deployment env without a runtime code change; the deny-shape keeps that while closing the parked-credential path.)

kbc/platform/pod/batching.py
LikiosSedo: Fixed in b8d07245. `scan_sources` now skips symlinks and realpath-confines every inventoried path to `raw/` — the same hardening as the S1 snapshot pinner, so the two scan paths agree. Test covers a file symlink pointing outside and an escaping directory symlink.

LikiosSedo:
Thanks — all three inline findings were real; fixed in `b8d07245` (replies on each thread). Your lower-severity list is addressed in the same commit:

- **Resume vs deleted sources**: resume now prunes sources gone from `raw/` out of pending batches (an emptied batch is stamped done) and says so in a summary — no directive points at a missing file. *(Added sources stay the coverage ledger's job, as you noted.)*
- **stamp_done durability window**: the done-stamp (and the batch's pages) now sync to the durable store immediately after each batch, not just on the periodic tick — the re-run-a-finished-batch crash window is gone in practice.
- **Tail-phase owner notes**: notes arriving during ledger-repair/media-verify now get a bounded "note digest" session before the turn closes (two passes, then ledger re-check) — the "rides the next directive" contract holds instead of being silently dropped at `turn_done`.
- **never-block asymmetry**: an orchestrator error now also emits a resumable `turn_done` after the `error` event, so a consumer gating on `turn_done` can't hang.
- **`dataStaleMs`**: clamped to `>= staleMs` (the invariant is now enforced, not documented), and a non-positive env override falls back to the default instead of an accidental 0.
- **Space-less `data:`**: both SSE parsers (fetch + https) now take the value after the colon minus one optional space, per spec.

New tests for each (batching id/symlink/prune units; an orchestrator resume-prune / error-turn_done / tail-note-digest scenario; spawner secret-refusal; run-manager clamp + env fallback; client tight-`data:` frame). Whole branch re-verified: 7 kbc suites, `tsc`, vitest 4191.

**On splitting:** we're keeping this as one PR, deliberately. The subsystems interleave in `compile_box.py` (routing, seam ordering, one shared session driver) and share one test suite — a 5-way surgical split would re-run the stacked-PR conflict churn this branch just came out of, and the split surgery itself is a fresh source of risk. What we've done instead is fix the real defect you named — the 3-line description — by rewriting the PR body into a **per-subsystem review guide** (mechanism, commits, tests, and live evidence for each of the 8 subsystems). Your batch + heartbeat rounds are done; the remaining units (`incremental.py`, `mediaverify.py`, `redblue.py`, LLM-credential/helm) are each self-contained by file and documented in the body — happy to take them as separate review rounds on this PR at whatever pace works, and to fix findings per round as above.

jacoblee-io (commented):
## Review: incremental / mediaverify / redblue (the previously-unreviewed subsystems)

Per the request to review the rest of the PR (not split it): I've now covered the three subsystems the description omitted — scoped incremental recompile, media/image-claim verification, and the red-blue PK rework. The batch-compile + gateway-heartbeat scope from the earlier passes is clean and fixed.

**Recurring theme:** these three are parallel implementations that re-open bug classes already solved elsewhere in this same PR. Two of the high-severity findings are mechanisms *this PR itself adds* (the model-stall watchdog; the converge_phase signal) being bypassed / not handled in a new path. Details inline.

**Lower-severity (not inline):**
- **incremental guard is single-shot:** `_incr_pending` is consumed/cleared before the repair prompt is injected, so the byte-integrity guard's own repair turn runs unguarded — a second out-of-scope edit made during repair ships silently.
- **mediaverify sync path has no per-turn round cap:** unlike `KBC_L1_REPAIR_ROUNDS`, the `while True` verify/repair loop is bounded only by per-page one-shot marking; repair turns that add new `(source: img)` citations re-enter `pending` and can blow up cost/time on a large or adversarial corpus.
- **basename collision drops a claim from media enumeration** (`selfcheck.py` body-citation resolver accepts only a unique basename match) — a page whose only image citation collides never enters `pending` and is never verified.
- **redblue de-agentification is convention, not a hard cap:** the verdict/blue calls are still agent loops (`max_turns=40`), bounded only by per-call timeouts — the stall the PR fixed can recur under a backed-up massapi.

kbc/platform/pod/compile_box.py
```diff
@@ -952,9 +1409,530 @@ async def run_session(run: CompileRun):
...
+    once wired, the byte-integrity guard) apply unchanged."""
+    cs = incremental.materialize_changeset(run.workdir)
+    if cs is None:
+        # Race / empty after the route check → fall back to a normal turn.
+        await run.client.query(text)
+        return
+    run._incr_pending = {"before": incremental.page_hashes(run.workdir), "changeset": cs}
+    await run.emit({"type": "summary",
+                    "text": f"增量重编:{len(cs['affected_pages'])} 页受影响,只改这些,其余不动。"})
+    await run.client.query(incremental.build_scoped_directive(cs))
```
jacoblee-io: **[incremental, high] The incremental turn bypasses the model-stall watchdog this PR adds.**

`_start_incremental` calls `run.client.query(...)` directly (here and at :1492) without `run._begin_turn(directive)`. Every other model-turn injection site arms the watchdog first (the normal message path, the batch directive at :1536, the planner at :1641), and the stall watchdog gates on `run._turn_active` / retries on `run._last_directive`. So an incremental turn runs with stale watchdog state: a black-holed incremental request is either never reaped (the ~1h wedge the watchdog exists to prevent) or, if `_turn_active` is stale-true, a stall-retry re-issues the *previous* turn's `_last_directive` instead of the scoped incremental one. **Fix:** route the incremental kickoff through `run._begin_turn(...)` like the other sites.

kbc/platform/pod/compile_box.py
```diff
@@ -819,6 +934,318 @@ def _l1_repair_rounds() -> int:
...
+        if sent is not None:
+            try:
+                await _sync_workspace(run, sent)
+            except Exception:
+                pass
+        await run.emit({"type": "summary", "text": _pk_narration(summary)})
+        if summary.get("state") == "repairing":
+            await _set_converge_phase(run, "revising")
+            await run.inject_user_message(redblue.build_pk_repair_prompt(summary))
+        elif summary.get("state") in ("passed", "partial", "unconverged"):
```
jacoblee-io: **[redblue, high] A PK `failed` state leaves converge_phase stuck at "verifying" forever — the exact stall class this PR claims to fix.**

`run_pk` is fail-open and *returns* `state == "failed"` rather than raising (line :1208 explicitly preserves `"failed"` through the retest merge, so it's a real returned value). But the terminal dispatch handles only `repairing → revising` (:1227) and `passed/partial/unconverged → settled` (this line) — there is **no `failed` branch**, and the `except` (:1234) only fires on an actual raise. So when PK fails, `_set_converge_phase` is never advanced past the `"verifying"` set at the top of `_run_pk_flow`, wedging the frontend test-step gate indefinitely. **Fix:** add `failed` to the settle branch (a fail-open PK must still terminalize converge_phase), or have the except path own the converge_phase reset.

kbc/platform/pod/compile_box.py
```diff
@@ -819,6 +934,318 @@ def _l1_repair_rounds() -> int:
...
+    """Kick the blind transcribe+compare flow as a background task (图像复核 v2).
+    Marks the chunk verified up-front (one-shot semantics, same as v1) so a
+    crash costs one chunk's verification, never a loop."""
+    try:
+        chunk = _media_verify_due(run)
+    except Exception:
+        return False  # a broken due-check must never break the turn seam
+    if not chunk:
+        return False
+    selfcheck.mark_media_verified(run.workdir, list(chunk))
```
jacoblee-io: **[mediaverify, high] Pages are marked verified BEFORE verification runs, so a total transcription failure ships image claims unchecked and never self-heals.**

`selfcheck.mark_media_verified(run.workdir, list(chunk))` runs here, before `run_blind_verify` at :1012 (same order at :1753/:1755). If every image on a page fails to transcribe (corrupt/unsupported format, LLM/timeout error, or unparseable JSON after the single retry), `run_blind_verify` hits `if not ts: return` (mediaverify.py:206) and skips the page with no findings — but the page is already permanently marked verified. Its image claims (the MEM→GPU-Util / H20 class this feature targets) ship unverified, indistinguishable from a clean pass, and a recompile won't retry. **Fix:** mark verified only on a completed verification (or re-queue pages whose transcription failed), so fail-open doesn't equal silent false-pass.

kbc/platform/pod/incremental.py
```diff
@@ -0,0 +1,269 @@
...
+    对齐 selfcheck 正文引用的宽松度,防两侧路径前缀写法不一致漏配)。"""
+    if not sources:
+        return set()
+    src_basenames = {posixpath.basename(s) for s in sources}
+    hit: set[str] = set()
+    for rel, page in pages.items():
+        if rel == INDEX_PAGE or "error" in page:
+            continue
+        cf = set(page.get("sources", []))
+        if cf & sources or {posixpath.basename(c) for c in cf} & src_basenames:
```
jacoblee-io: **[incremental, medium] Basename fallback in `_pages_citing` cross-contaminates scope AND weakens the byte-freeze guard across same-named files in different dirs.**

`{posixpath.basename(c) for c in cf} & src_basenames` treats `snap/config.md` and `vendor/config.md` as the same source. Changing one pulls the other's page into both `affected_pages` (needless recompile) and `authorized_pages` — and the latter is the problem: a genuine *out-of-scope* drift into the collided page is now authorized, so the post-turn byte-integrity guard won't flag it. The byte-freeze guarantee silently fails whenever basenames collide across directories (common with per-topic folders + generic filenames like `config.md`/`README.md`). **Fix:** match on full raw-relative paths; drop the basename fallback or only apply it when unambiguous.

kbc/platform/pod/redblue.py
```diff
@@ -392,69 +511,93 @@ async def _body() -> dict:
...
+        #    grades from the inlined `expected` raw-truth rubric, gate-tier model,
+        #    NO raw root (roots=[wiki_dir]) so it can't loop re-reading raw. The
+        #    old agentic verdict (raw root + "回原始语料核对" + max_turns 40) is
+        #    what hung and, via fail-fast gather, vaporized all 24 answered blues.
+        async def _judge(chunk: list[dict]) -> None:
             qa = "\n\n".join(
                 f"[{q['id']}] 问题: {q['question']}\nraw 真值要点: {q.get('expected', '-')}\n"
                 f"raw 出处: {q.get('source_ref', '-')}\n"
-                f"蓝队回答: {detail['answers'].get(q['id'], {}).get('answer') or '(蓝队无回答)'}\n"
+                f"蓝队回答: {detail['answers'].get(q['id'], {}).get('answer') or '(蓝队无回答/未测)'}\n"
```
jacoblee-io: **[redblue, medium] A blue-team infrastructure failure is graded as a KB signal and inflates pass_rate.**

When the blue call crashes/times out, no answer is recorded and the judge is handed the literal `(蓝队无回答/未测)` to grade. If that scores as `正确标未覆盖` (honest-uncovered = a pass), an infrastructure outage enters the graded denominator as a pass — a systemic blue failure silently inflates `pass_rate` and defeats the PK's whole purpose. The per-item error record is then popped (erasing observability of the outage). **Fix:** treat a blue infra failure as ungraded/excluded (not fed to the judge as an answer), and keep the error record so a blue outage is visible rather than scored.

kbc/platform/pod/mediaverify.py
```diff
@@ -0,0 +1,238 @@
...
+{{"chart_type": "表格/柱状图/折线图/界面截图/流程图/其他",
+  "title_or_header": "图内可见的标题或表头,没有则 null",
+  "legend": {{"系列名": "颜色描述", ...}} 或 null,
+  "axes": "坐标轴含义与单位,没有则 null",
+  "facts": [{{"label": "行/列/系列定位", "value": "可见值"}}, ...],
+  "illegible": ["读不清的部分", ...],
+  "notes": "其他可见要点(仍然只写图里有的)"}}"""
+
+COMPARE_SYSTEM = """你是知识库图像断言的【比对员】。你看不到图片本身——只有另一位转写员产出的
+结构化转写(它是图内可见信息的完整清单)和引用了该图的知识页正文。
```
jacoblee-io, line 103: **[mediaverify, medium] The comparer is told the transcripts are the "complete inventory," but a partially-failed transcription is silently short — producing both false-passes and false-fails.**

`COMPARE_SYSTEM` asserts the transcript set is the 完整清单 (complete inventory). When one of several cited images fails to transcribe, it's simply absent with no signal: a wrong claim about the missing image has nothing to contradict it (false-pass), while a *correct* claim about it is flagged 超出转写范围 → a spurious ⚠️存疑 downgrade (false-fail). **Fix:** pass a per-image status (transcribed vs failed) so the comparer distinguishes "not in the image" from "couldn't read the image," and don't assert completeness when transcription was partial.

kbc/platform/pod/compile_box.py
```diff
@@ -1240,6 +2257,35 @@ async def handle_message(request: web.Request):
...
+        if _is_compile_trigger(text):
+            # A second trigger mid-batch changes nothing the plan doesn't know.
+            await run.emit({"type": "summary", "text": "分批编译进行中,本轮结束后再发起。"})
+        else:
+            # Owner chat mid-batch: queue it — the next batch/终审 directive
+            # relays it, so nothing lands inside a half-read internal session.
+            run._batch_notes.append(text)
+            await run.emit({"type": "summary", "text": "已收到,会带给后续批次一并考虑。"})
+        return web.json_response({"ok": True, "queued": True})
     await run.client.query(text)
```
jacoblee-io, line 2568: **[high] The owner `/message` turn (this line) never arms the model-stall watchdog — nor does the incremental path — so the two most common owner-initiated turns run unprotected against the exact black-hole the watchdog exists for.**

The watchdog only judges a turn when `run._turn_active` is set, which happens in `_begin_turn`. That is called from `inject_user_message` (~:294, internal self-check/PK repair injects), the batch directive (~:1536), and the planner (~:1641) — but **not** here (the owner's direct `/message`) and **not** in `_start_incremental` (~:1492/:1497), both of which call `run.client.query(...)` directly. So a model request that black-holes on an owner message or an incremental compile freezes for ~1h with no recovery — precisely what L1 was built to prevent. (Corrects my earlier comment that "the normal message path arms it"; that was `inject_user_message`, not this handler.) **Fix:** route both call sites through `_begin_turn(text)`.

kbc/platform/pod/compile_box.py
```diff
@@ -952,9 +1409,530 @@ async def run_session(run: CompileRun):
...
+
+
+async def _model_stall_watchdog(run: CompileRun) -> None:
+    """Reap a turn wedged on a black-holed model request. Interrupt the attempt;
+    _consume_turn_stream then re-issues it (or fails). Only judges an ACTIVE turn,
+    and relaxes to the tool bound while a tool is pending — never false-kills a
+    live turn or a long tool (I4)."""
+    while not run.done:
+        await asyncio.sleep(_MODEL_WATCHDOG_POLL_S)
+        if not run._turn_active or run._stall_retrying:
```
jacoblee-io: **[med, plausible] Even when armed, the watchdog can wedge permanently on a true black-hole.**

Once it fires, the watchdog sets `_stall_retrying=True` and calls `client.interrupt()`, and this gate then skips the run (`... or run._stall_retrying: continue`) until `_consume_turn_stream` clears the flag on the interrupted ResultMessage. But a genuinely black-holed upstream request is exactly the case where `interrupt()` may not force the frozen stream to yield anything — so the flag stays set, the receive loop blocks forever in `receive_messages()`, and the watchdog never fires for this run again. The ~1h freeze recurs with the recovery latch stuck. **Fix:** bound the post-interrupt wait (`asyncio.wait_for`/`timeout`) and hard-fail the turn if the interrupted result doesn't arrive within a deadline, rather than waiting unconditionally.

kbc/platform/pod/compile_box.py
```diff
@@ -1405,8 +2451,30 @@ async def handle_close_test(request: web.Request):
...
+    for run in runs:
+        try:
+            n = await _sync_workspace(run, run._sync_sent)
+            if n:
+                await run.emit({"type": "summary", "text": f"[shutdown] 落盘 {n} 个改动文件"})
+        except Exception:
+            pass
+    deadline = time.monotonic() + _SHUTDOWN_DRAIN_MAX_S
+    while runs and time.monotonic() < deadline:
+        if all(run.events.empty() for run in runs):
```
jacoblee-io: **[med] The graceful-shutdown drain completes on `events.empty()`, which flips true when the relay *dequeues* the last item — before it's written to the socket — so the final `syncArtifacts` can still be lost.**

`asyncio.Queue.empty()` becomes true the moment the SSE relay pops the last event, not when `await resp.write()` has flushed it. So the drain loop can break and aiohttp force-closes the connection with the final workspace-sync frame still in flight — the exact "last ≤interval of work lost" that F3 exists to prevent, just with a smaller window. There's no `task_done()`/`join()` gate tying drain-complete to *delivery*. **Fix:** gate the drain on the relay confirming the write completed (queue `join()` after `task_done()`, or an explicit acked-flush), not on `empty()`.

kbc/platform/pod/compile_box.py
```diff
@@ -1144,6 +2124,39 @@ async def _teardown_test_session(run: "TestRun"):
...
+
+def _apply_session_config(body: dict) -> None:
+    """Apply consumer-managed llm/settings from the /session body to os.environ.
+    Never log the token; whitelist settings keys to the box's own vocabulary."""
+    llm = body.get("llm")
+    if isinstance(llm, dict):
+        if llm.get("base_url"):
+            os.environ["ANTHROPIC_BASE_URL"] = str(llm["base_url"])
+        if llm.get("auth_token"):
+            os.environ["ANTHROPIC_AUTH_TOKEN"] = str(llm["auth_token"])
```
jacoblee-io, line 2419: **[med, latent] `_apply_session_config` writes the consumer credential and model tiers into process-global `os.environ`, so a second concurrent run clobbers the first's token/tier.**

`ANTHROPIC_AUTH_TOKEN` / `ANTHROPIC_BASE_URL` / all `KBC_*` settings are set on `os.environ` (process-global). The box is "usually" one run per pod (per the note at ~:83), but it *can* host more — and if a second `/session` arrives, it overwrites the first run's credential and model tier globally, so the earlier run's later turns silently use the newer run's token/tier. Bounded today by one-run-per-pod, latent otherwise. **Fix:** carry llm/settings per-run (on the run object / per-session env passed to that run's SDK client) rather than mutating the shared process environment.

jacoblee-io (commented):
## Follow-up review: model-stall watchdog / graceful-shutdown / config (the residual net-new subsystems)

Covering the parts of this omnibus I hadn't yet audited (batch, gateway-heartbeat, incremental, mediaverify, redblue were earlier). Focus: the model-stall watchdog (L1), graceful-shutdown flush (F3), and the consumer session-config path.

**Correction to my earlier watchdog comment.** In the earlier review I said "the normal message path arms `_begin_turn`." That was wrong — line ~294's `_begin_turn` is inside `inject_user_message`, which serves *internal* injected turns (self-check / PK repairs). The actual owner `/message` HTTP handler ends in a bare `run.client.query(text)` and does **not** arm the watchdog, same as the incremental path. So the set of watchdog-protected turns is: internal injects + batch/planner. The set that is **un**protected is: the owner's direct `/message` turns and incremental turns — i.e. the most common owner-initiated ones. Details inline.

**Lower-severity (not inline):**
- **stall-fatal ends with `error`, not `turn_done`** — a consumer that ends the logical turn on `turn_done` hangs (the batch path deliberately emits `turn_done` even on error; the stall path doesn't mirror that).
- **retry has no rollback / can double-apply** — a mid-turn stall that already wrote some candidate pages is retried by re-`query`-ing `_last_directive` on the same session with no rollback → duplicated/half-merged pages; and a real ResultMessage landing on the same poll tick as the interrupt is mis-classified as interrupted (wipes the completed reply, re-queries).
- **rate-limit backoff resets the idle clock** — a request that black-holes right after a 429 restarts stall detection from zero each cycle (masks the stall up to 5×); `_rate_retries` also leaks across the unarmed turns.
- **RUNS is never GC'd**, so on shutdown a finished run with an orphaned non-empty queue makes the drain's `all(empty())` never true → every shutdown burns the full 8s drain deadline, starving active runs' grace window.
- **engine.py tiers are otherwise clean** — every tier has a served-id default and is plumbed consistently; `effort` is genuinely unused.

LikiosSedo:
One more commit on top of the review fixes: `03a79156` extends the repo's locale contract (prompt packs + `selfcheck._is_en`, platform default **English**, zh only when the consumer declares `locale=zh`) to every subsystem this branch adds — it had only covered the #382-era pipeline.

- **User-facing narration**: all batch/incremental/image-verify/PK/stall/rate-limit/shutdown summaries, logs, `turn_done` texts, session labels and reply prefixes now render through a `_loc(run, en, zh)` gate. An English-locale KB no longer receives Chinese progress lines.
- **Session directives**: the batch directive, final-review checklist, planner role + budget line, note-digest directive, the scoped-incremental directive and integrity-repair, the blind-transcription/comparison prompts and the PK stage prompts are all bilingual now — so agent replies (which surface in `turn_done`) come back in the KB's language. The red-blue blue-team persona now uses the locale'd test-role pack (the en pack existed; PK wasn't threaded).
- **Deliberately NOT localized**: wire/data tokens — compile trigger phrases, the brief marker + field labels (frontend wire format), finding kinds (`不一致`/`超出转写范围`), PK score/category tokens, dup reasons. Prompts instruct the model to emit the exact stored token (with an English gloss); render sites translate for display only. This keeps stored artifacts and cross-repo contracts byte-stable.
- **Tests**: default-locale assertions switched to English across the suites; each module gains a zh-branch check; the orchestrator suite gains a full zh-locale scenario.

Known residual (pre-existing on main, unchanged): selfcheck lint-violation `detail` strings are stored in SELFCHECK.json in Chinese and render verbatim inside English repair prompts — flagging for a separate pass since it predates this branch.

Verified: all 7 kbc suites, `tsc`, vitest 4191 green.

kbc/platform/pod/compile_box.py
LikiosSedo: Fixed in 1234829e. `_start_incremental` now arms `run._begin_turn(directive)` before the query — and so does its empty-changeset fallback path (which also queried bare). The incremental-route test asserts `_turn_active` + `_last_directive` carry the scoped directive after kickoff.

kbc/platform/pod/compile_box.py
LikiosSedo: Fixed in 1234829e. `failed` joined the settle branch (with a comment explaining why: run_pk is fail-open and RETURNS failed), and the exception path also terminalizes the gate. Test covers both shapes — returned `state=failed` and an outright raise — asserting `converge_phase == "settled"` with `pk.state == "failed"` recorded.

kbc/platform/pod/compile_box.py
LikiosSedo: Fixed in 1234829e. Verified marks now land only AFTER a completed verification. Failed pages (transcription/comparison error) get `media_verify.attempts` bumps and retry on later triggers, bounded by `KBC_MEDIA_VERIFY_ATTEMPTS` (default 2); past the budget they ship marked but VISIBLY flagged in `media_verify.exhausted` (plus a summary line) — fail-open no longer reads as a clean pass. An in-flight set replaces the up-front mark as the loop/crash guard. Test walks retry → exhaust and asserts a failed page is never falsely in `verified_pages`.

kbc/platform/pod/incremental.py
LikiosSedo: Fixed in 1234829e. The basename fallback is gone entirely: both sides pass through a `_norm_source` (posix separators, normpath, strip one leading `raw/`/`drop/` segment) and match FULL paths only — which covers the prefix-form mismatch that motivated the fallback without any cross-directory collision. Tests: same-basename different-dir sources no longer cross-match (and the byte-freeze guard flags drift into the collided page again); `raw/`-prefixed compiled_from still matches its unprefixed changeset form.

kbc/platform/pod/redblue.py
LikiosSedo: Fixed in 1234829e. Judge chunks are built only from ids with a recorded answer; a blue infra failure flows through the existing ungraded machinery with reason `blue infrastructure failure: <err>` (out of the denominator, never triggers repair) and its error record is KEPT. Verdict acceptance is also fenced to the chunk's ids, so an echoed/hallucinated id can neither grade an unasked question nor erase the outage record. A genuine honest-uncovered reply still grades normally. Test asserts denominator/pass_rate exclusion, no judge exposure, and the preserved error record.

kbc/platform/pod/mediaverify.py
LikiosSedo: Fixed in 1234829e — folded into the same mechanism as the mark-before-verify fix: a page with ANY untranscribed image is never compared (it fails whole, with an explicit error line, and retries under the attempt budget). The comparison prompt's complete-inventory premise is therefore always true when a comparison actually runs. Test: a two-image page with one failed transcription is skipped by the comparer while its fully-transcribed sibling completes.

kbc/platform/pod/compile_box.py
LikiosSedo: Fixed in 1234829e. The owner `/message` handler arms `run._begin_turn(text)` before the query, and the incremental path (plus its fallback) does the same — `_begin_turn`'s docstring now describes reality. Arming every turn also stops `_rate_retries` leaking across previously-unarmed turns.

kbc/platform/pod/compile_box.py
LikiosSedo: Fixed in 1234829e. The watchdog stamps `_stall_interrupted_at` when it fires; while `_stall_retrying` it now bounds the wait (`KBC_STALL_INTERRUPT_DEADLINE_S`, default 120s). Past the deadline it closes the turn honestly (error + a resumable `turn_done`), clears the latch, and disconnects the client to unblock the wedged receive loop. Test drives a client whose `interrupt()` is accepted but never yields — asserting the turn closes and the disconnect fires.

kbc/platform/pod/compile_box.py
LikiosSedo: Fixed in 1234829e. The relay `task_done()`s each event AFTER the socket write, and the drain gates on `queue.join()` (delivery, not dequeue) under the same deadline. The join scope is also filtered to runs that are non-ended AND have a live relay — which fixes the adjacent lower-severity item too: a finished run with an orphaned queue (or a run nobody is streaming) no longer burns the whole drain window.

kbc/platform/pod/compile_box.py
LikiosSedo: Acknowledged — deferring the per-run carry, and 1234829e documents the invariant at `_apply_session_config`: it leans on the platform's one-pod-per-run spawn (the gateway creates a dedicated box per runId), and if multi-run pods ever become real the llm/settings move to per-run SDK-client env instead of process-global mutation. Bounded today exactly as you noted.

LikiosSedo:
Rounds 2+3 addressed in `1234829e` (replies on all ten threads). Your "recurring theme" diagnosis was exactly right — the fixes are mostly re-wiring the new subsystems INTO the mechanisms this branch added: every model-turn kickoff (owner `/message`, incremental + fallback) now arms the stall watchdog; PK's fail-open `failed` terminalizes `converge_phase`; media verification marks only after completion.

Your lower-severity lists, item by item:

**Fixed in the same commit:**
- *incremental guard single-shot* → the byte-integrity guard re-arms for its own repair turn (judged against the ORIGINAL baseline; consumed on the clean pass). Test updated to the new semantics.
- *mediaverify sync-path round cap* → `KBC_MEDIA_VERIFY_ROUNDS` (default 3) caps the batch-tail verify/repair loop, on top of the per-page attempt budget.
- *stall-fatal ends with error, not turn_done* → generalized: a driver that dies mid-turn (`_run_wrapper`) closes the logical turn with error + a resumable `turn_done` — never-block symmetry with the batch path.
- *rate-limit `_rate_retries` leaking across unarmed turns* → gone as a side effect of arming every turn.
- *RUNS drain burn* → the shutdown drain's join scope is filtered to live runs with a live relay, so orphaned queues can't starve the grace window (full RUNS GC still deferred — one-pod-per-run keeps the map tiny in practice).

**Deferred, deliberately (each with rationale):**
- *retry double-apply / same-tick interrupt-vs-result race* — accepted residual of retry-on-same-session; a rollback layer is heavier than the failure it prevents (re-compiled pages are idempotent-ish merges; the coverage ledger catches divergence). Happy to revisit if it shows up live.
- *selfcheck body-citation basename collision (media enumeration)* — pre-existing resolver leniency on main, same bucket as the stored-zh lint details; queued for a follow-up pass rather than this PR.
- *redblue max_turns=40 convention* — per-call wall-clock timeouts already bound each stage; tightening to a hard cap is a tuning question we'd rather do against live traces.
- *rate-limit backoff resetting the idle clock* — masking is bounded by `_MODEL_RATE_MAX_RETRIES`; left as-is.
- *`engine.py` `effort` unused* — already acknowledged on the backlog.

Verified: all 7 kbc suites green (new coverage: watchdog arming on incremental, interrupt-deadline with a swallowing client, driver-crash turn closure, PK failed/raise → settled, media retry→exhaust visible flag, partial-transcription skip, blue-infra ungraded, basename isolation + prefix normalization), `tsc` clean, vitest 4191.

kbc/platform/pod/compile_box.py
```diff
@@ -952,9 +1514,672 @@ async def run_session(run: CompileRun):
...
+        if run._stall_retrying:
+            # Interrupted, waiting for the interrupted result. A true black-hole
+            # can swallow interrupt() too — then the latch stays set and the
+            # receive loop blocks forever. Bound the wait; past the deadline,
+            # close the turn honestly and disconnect to unblock the loop.
+            if time.monotonic() - run._stall_interrupted_at > _STALL_INTERRUPT_DEADLINE_S:
+                run._stall_retrying = False
+                run._turn_active = False
+                await run.emit({"type": "error",
+                                "error": f"model stall: interrupt produced nothing within {_STALL_INTERRUPT_DEADLINE_S}s"})
```
jacoblee-io, line 2134: **[regression, med-high] The stall-deadline recovery disconnects the persistent session and leaves `run.client=None` with no reconnect — so the "send the message again to retry" it just promised 409s until the runtime reaps the box.**

When the post-interrupt deadline fires, the branch emits `turn_done` ("…nothing was applied; send the message again to retry.") and then `await client.disconnect()`. But `run_session` is a single `connect()` + one persistent `_consume_turn_stream(stop_on_result=False)` with no reconnect loop: disconnecting ends that loop, and its `finally` sets `run.client = None` permanently. The next `/message` then hits `if not run.client: 409 "session is not live"` — so the retry the box explicitly instructs is impossible on this box, and recovery only happens later via the runtime's stale reap + a fresh respawn (with rehydration). Net: the fix traded a wedged-but-alive session for a cleanly-ended-but-dead one that lies about being retryable.

**Fix:** after the deadline disconnect, re-establish the session (reconnect → new `run.client`) so the promised resend actually works — or change the `turn_done` text to reflect that a respawn is required.

kbc/platform/pod/compile_box.py
```diff
@@ -850,7 +1392,18 @@ async def _post_turn_selfcheck(run) -> str | None:
     await run.emit({"type": "summary", "text": selfcheck.narration(report, locale)})
     if report["state"] == "repairing":
         run._l1_repairs_used += 1
-        return selfcheck.build_repair_prompt(report, locale)
+        if incr and incr_violations:
```
jacoblee-io: **[regression, med] The byte-integrity guard is only re-armed on `incr && incr_violations`, so an in-scope-but-ledger-unclean incremental repair turn runs unguarded.**

`_incr_pending` is consumed to `None` at the top of the post-turn seam (~:1363) and only re-armed here, gated on `incr and incr_violations`. An incremental turn that stayed in scope (no byte violations) but failed coverage/lint enters `repairing` with `incr_violations == []` → the guard is not re-armed → the ledger-repair turn runs with `_incr_pending=None`, so an out-of-scope edit made during that repair ships without a byte-freeze check. This re-opens the "unguarded repair turn" class the incremental guard was meant to close. **Fix:** re-arm whenever `incr` is set and the turn is going to repair (`if incr:`), not only when this turn already had violations.

kbc/platform/pod/selfcheck.py
```diff
@@ -524,3 +708,107 @@ def build_repair_prompt(report: dict, locale: str | None = None) -> str:
...
+    mv = report.get("media_verify") or {}
+    mv["verified_pages"] = sorted(set(mv.get("verified_pages") or []) | set(pages))
+    if exhausted:
+        mv["exhausted"] = sorted(set(mv.get("exhausted") or []) | set(pages))
+    mv["at"] = datetime.now(timezone.utc).isoformat(timespec="seconds")
+    report["media_verify"] = mv
+    write_selfcheck(workdir, report)
+
+
+def bump_media_attempts(workdir: str, pages: list[str]) -> dict[str, int]:
```
jacoblee-io, line 785: **[low] `bump_media_attempts` never resets a page's count on success and never prunes the map.**

A page that fails once then succeeds keeps its accrued attempt count (it becomes verified, but the entry lingers), and the attempts map grows unbounded across a long run. Not a false-pass, but if that page is later re-cited and re-enters `pending`, the residual count can push it to "exhausted" after fewer real failures than `KBC_MEDIA_VERIFY_ATTEMPTS` implies — shipping it with the visible-unverified marker prematurely. **Fix:** clear a page's entry when it completes verification (and/or prune on re-entry).

jacoblee-io (commented):
## Follow-up: round-2/3 fixes verified — all addressed; 3 new regressions from the fix commit

Commit `1234829e` correctly fixes **all 10** inline findings from my earlier passes (watchdog now arms on the owner `/message` + incremental turns; the post-interrupt wait is deadline-bounded; redblue `failed` now settles converge_phase; media marks only completed pages and retries failed ones with a visible exhausted marker; incremental matches by full normalized path, not basename; a blue infrastructure failure is recorded ungraded instead of fed to the judge; the shutdown drain gates on `events.join()` = delivery, not `empty()`; the global-`os.environ` assumption is documented as one-pod-per-run). It even picked up the lower-severity "stall-fatal must still emit turn_done" note. Nice.

The large fix commit (+1289) introduced a few new issues, ranked below. The i18n commit `03a79156` is clean (no machine-parsed token/JSON-key was localized). Inline for the top two + one low.

**Also (low, not inline):** `_apply_session_config` still mutates process-global `os.environ`; it's now documented as relying on one-pod-per-run, which is fine as a bounded assumption, but the honest long-term fix is per-run SDK-client env (noted here only so it isn't lost).

kbc/platform/pod/compile_box.py
LikiosSedo: Fixed in 48dd55d5, taking your second option: the turn_done text now promises exactly what happens — "The compile session will be recreated automatically on your next message" — and a comment at the branch pins the recovery-ownership rationale: run_session deliberately has no reconnect loop; this path fires only on a double black-hole (model wedged AND interrupt swallowed), and recovery belongs to the platform's terminal-run → fresh-spawn → workspace-rehydration machinery (live-proven), which the `end` this disconnect produces triggers immediately. Building an in-box reconnect loop into the every-session driver for that rarity seemed like the wrong trade. The deadline test now asserts the honest wording.

kbc/platform/pod/compile_box.py
LikiosSedo: Fixed in 48dd55d5: the re-arm is now gated on `if incr:` alone — any incremental turn entering repair keeps the guard for its repair turn, whether it got there via byte violations or via an unclean ledger. New test: an in-scope incremental turn with an unaccounted source enters repairing with `out_of_scope_pages == []` and the guard is re-armed for the ledger-repair turn.

kbc/platform/pod/selfcheck.py
LikiosSedo: Fixed in 48dd55d5: a COMPLETED verification clears the page's `media_verify.attempts` entry (in `mark_media_verified`'s non-exhausted path), so a later re-entry starts from a clean count and the map stays bounded. Exhausted pages keep their count as forensics — they are marked verified and never re-enter pending. New test: fail-once-then-succeed ends verified, un-exhausted, with the attempts entry gone.

jacoblee-io (approved):
Approving on `48dd55d`. Across four review rounds the whole net-new surface of this PR was covered — batch compile, the gateway heartbeat, incremental recompile, media/image-claim verification, the red-blue PK rework, the model-stall watchdog, graceful-shutdown flush, and the config/env + helm/spawn-env plumbing — and every substantive finding is now fixed with tests:

- round-1 (batch): duplicate batch-id validation, KBC_* env-forward secret-shaped refusal, scan_sources symlink confinement, resume prune / stamp sync / tail-note digest / error-turn_done.
- round-2 (incremental/media/redblue): watchdog arming on owner+incremental turns, redblue `failed` settles converge_phase, media marks only completed pages (failed retry + visible exhausted marker), incremental full-path (not basename) matching, blue-infra failure recorded ungraded, partial-transcription no longer lies to the comparer.
- round-3 (residual): owner `/message` + incremental now arm the stall watchdog, the post-interrupt wait is deadline-bounded (and now emits turn_done), the shutdown drain gates on delivery (`events.join()`), global-`os.environ` documented as one-pod-per-run.
- round-4 (regressions from the round-2/3 fix commit): the stall-deadline turn_done text is now honest about platform respawn-recovery (no false in-place-retry promise), the byte-integrity guard re-arms for ANY incremental repair turn, and a completed media verification clears its attempt count.

The round-4 commit is small and clean, no new regression. CI green (KBC Box / Portal Web / Test / Type Check).

Two standing notes (non-blocking): (1) the PR still ships far more than its description says — worth a heads-up for future readers, though it's been reviewed subsystem-by-subsystem here; (2) taking effect requires a `kbc-compile-box` image rebuild + runtime redeploy, and the consumer-side (sicore) LLM-binding config must be in place for the kb-llm path.

PR #385, opened 2026-07-05
kbc: S1 compile self-check + published-version test sessions + box protocol hardening

LikiosSedo:
## What

Rebases the compile-box authoring work onto the deduplicated `feat/kb-capability-platform` head and hardens the box protocol. This branch was previously a **sibling** of the capability-platform work, so its first commit re-added `capability.test*` with divergent wording; that duplicate is dropped and the whole series now replays on top of the platform head (which already carries `capability.test*` plus the locale-parameterized prompt packs). All downstream prompt edits were re-targeted from the old inline Python role strings into `kbc/platform/pod/prompts/{zh,en}/*.md` + `tools.json`, with `zh` kept as the source of truth and faithful English translations added.

## Highlights

- **L1 deterministic compile self-check.** Completion moves from model self-certification to code verification: a coverage ledger derived from each candidate page's `compiled_from` (plus an explicit `authoring/EXCLUSIONS.json`) is checked against the raw text inventory, link/provenance lint runs alongside, and any gap triggers **one bounded repair round** injected after `turn_done`. The never-stuck invariant is preserved — a repair is just the next turn. Engine-neutral analysis lives in `selfcheck.py` (pure stdlib).
- **L2 red-blue PK calibration framework.** One writer, asymmetric examiners (judge reads raw + snapshot and writes questions; a blue team using the read-only consumer persona answers with the raw mechanically invisible; the judge grades with four-category attribution). **CLI-only — not wired into the main compile loop.** The consumer persona is single-sourced from the locale packs, so the test session and the blue team can't drift.
- **Published-version test sessions.** `capability.testStart` gains an optional `bundle_base64` / `bundle_sha256`: the box installs that tar.gz as the pinned snapshot instead of packing the run's `candidate/` draft, using the **same content-hash formula** for both sources so gradings on (question × snapshot) stay comparable across draft and published versions.
- **Deterministic plan-approval signal.** `propose_plan` writes `authoring/PROPOSED_PLAN.json` from the tool argument (code, not model formatting); the owner's approve control keys off that artifact. `PLAN.md ## Next Pages` checkbox maintenance stays an advisory nudge — the UI is never held hostage by how the model formatted its notes.
- **Box protocol hardening.** Resume-after-fixup discipline (a repair dispatch that lands mid-compile returns to the interrupted task); an independent auditor re-review pass after the index is written; `resolve_ticket` echoes the dispatch nonce so a receipt matches the exact dispatch round; test-session answers end with a machine-readable `SOURCES:` line for provenance chips; the workspace is synced **before** `turn_done` so consumers refetch durable files.
- **massapi/Bedrock `context_management` 400 mitigation.** massapi proxies to Bedrock, which rejects the `context_management` field Claude Code attaches once a turn passes the autocompact buffer. `DISABLE_AUTOCOMPACT` is set (module `setdefault` + Dockerfile `ENV`) so the field is never sent.
- **Model pins.** `compile = claude-opus-4-6`, read-only test session = `claude-sonnet-4-6` (the gate/consumer tier), both overridable via `KBC_COMPILE_MODEL` / `KBC_TEST_MODEL`; the box talks to massapi, which serves specific ids.

## Verification

- `kbc/platform/pod` suites green: `test_compile_box.py`, `test_selfcheck.py`, `test_redblue.py`.
- Gateway `npm test` green; `npx tsc --noEmit` clean.


kbc/platform/pod/compile_box.py
```diff
@@ -985,28 +1070,77 @@ async def handle_health(request: web.Request):
     return web.json_response({"status": "ok", "runs": len(RUNS), "test_sessions": len(TEST_SESSIONS)})
 
 
+def _install_wiki_snapshot(bundle: bytes, dest: Path, expected_sha256: str | None = None) -> tuple[str, int]:
```
jacoblee-io, line 1073: **[security / DoS] `_install_wiki_snapshot` drops both size guards its sibling installers enforce.**

`_install_source_bundle` (`:337`) and `_install_authoring_bundle` (`:435`) each check `max_bundle_bytes` on the compressed input **and** accumulate `total_bytes` against `max_unpacked_bytes` during extraction. This new installer does neither, and additionally buffers every file's full bytes into an in-memory `pages` list for the content hash. A gzip-bomb `bundle_base64` — a small compressed body that passes the 768 MB HTTP request cap but decompresses to gigabytes of `.md`/`.json` — OOM-kills the box, which also takes down the live parent authoring run sharing the pod.

Tempered by: the caller is the trusted runtime and `bundle_base64` has no frontend producer yet — so it's a latent DoS on a not-yet-exposed path. But it's exactly the hardening the two sibling paths already have. **Fix:** share one guarded extractor (or mirror the `max_bundle_bytes` + `max_unpacked_bytes` accounting here).

kbc/platform/pod/selfcheck.py
```diff
@@ -0,0 +1,444 @@
...
+    entries, errors = [], []
+    for i, item in enumerate(data):
+        if not isinstance(item, dict) or not item.get("pattern") or not item.get("reason"):
+            errors.append(f"{EXCLUSIONS_PATH}[{i}] needs {{pattern, reason}}")
+            continue
+        entries.append({"pattern": str(item["pattern"]), "reason": str(item["reason"])})
+    return entries, errors
+
+
+def _matches(path: str, pattern: str) -> bool:
```
jacoblee-io, line 208: **[correctness] The L1 coverage gate's matcher is unreliable in both directions — undermining the "deterministic verification" premise.**

`_matches` uses `fnmatch`, whose `*` crosses `/`. Two failure modes:

1. **Over-exclusion → false PASS (ships an incomplete KB):** an exclusion `notes/*` (intended for the files directly under `notes/`) also silently matches `notes/sub/secret.md`; that raw file drops out of `unaccounted`, and L1 reports `passed` while the content was never compiled.
2. **Under-exclusion → false GAP (a complete KB reported unconverged):** the exclusion namespace is stripped-only, but `compiled_from` tolerates *both* `raw/`-prefixed and stripped entries (lines 111-113 strip `raw/`/`drop/`). A model that writes `{"pattern": "raw/live.csv"}` — natural, since the adjacent `compiled_from` field accepts that prefix — matches nothing, so the file stays `unaccounted`, the run reports `unconverged` on a complete KB, and it burns a repair round the model can't satisfy (it already declared the exclusion).

**Fix:** path-segment-aware matching (or require `**` for cross-segment), normalize/accept both prefix forms on the exclusion side to match `compiled_from`, and validate each pattern matches ≥1 inventory path (warn on a no-op exclusion).

kbc/platform/pod/selfcheck.py
```diff
@@ -0,0 +1,444 @@
...
+    _pack_candidates_to_wiki — hashed in rel_posix order) — the self-check
+    idempotency key. None when there is no candidate tree."""
+    cand = Path(workdir) / "candidate"
+    if not cand.is_dir():
+        return None
+    entries: list[tuple[str, bytes]] = []
+    for f in cand.rglob("*"):
+        if not f.is_file() or f.suffix not in (".md", ".json"):
+            continue
+        entries.append((f.relative_to(cand).as_posix(), f.read_bytes()))
```
jacoblee-io: **[correctness] `candidate_tree_hash` reads files with no guard, and it runs *before* the self-check's fail-open — so one unreadable candidate file silently disables coverage verification.**

`candidate_tree_hash` does `f.read_bytes()` with no try/except, and it's called by `state_key`, which runs before `run_layer1`'s fail-open. The compile session has Write+Bash, so it can leave a dangling symlink, a FIFO, or a perm-denied file under `candidate/` — `read_bytes` then raises in `state_key`. `turn_done` still fires (not hard-stuck), but the self-check no-ops for as long as that file exists, so the deterministic coverage gate is silently off. `candidate_pages` (`:131-136`) guards its `read_text` per-file; this path should mirror that guard (record an error entry instead of raising).

jacoblee-io (commented):
## Review: S1 self-check + published-version test sessions + protocol hardening

The TS/contract layer is clean: `bundle_base64`/`bundle_sha256` field names match the box, the rebase re-introduced no duplicate `capability.test*` handlers, transport is a direct gateway→box HTTP POST (768 MB cap), the draft-vs-published content-hash formula is genuinely identical for valid flat-root bundles, sha256 is verified before install, the new installer rejects non-file tar members, and the read-only test persona is single-sourced from `prompts/*/test_role.md`. Model pins are valid massapi ids.

The findings below are on the Python box. Two are worth blocking on (a DoS-hardening gap the sibling installers already close, and the L1 coverage gate being unreliable in both directions — which undermines the "deterministic verification" premise of the feature). Inline.

**Design / lower-severity notes (not inline):**

- **"One bounded repair round" is per gap-episode, not per run** (`compile_box.py`, `_l1_repairs_used`). The counter resets on each pass, so a model that oscillates closed↔open (e.g. a page merge that drops a `compiled_from`) re-arms a fresh repair every episode. Never-stuck holds and each episode terminates, but the README wording overstates the guarantee — it's one round per distinct gap-episode, unbounded in count over a long restructuring compile.
- **The content-hash formula is triplicated** (`_install_wiki_snapshot`, `selfcheck.pack_candidates_to_wiki`, `selfcheck.candidate_tree_hash`) with only a hand-written "keep in sync" comment and no cross-module test. If one copy drifts (a field added to the digest, a different sort), draft and published `snapshot_hash` silently diverge for identical content and the (question × snapshot) grading comparability breaks. Extract one shared `content_hash(pages)` helper.
- **Dead/pending code:** `redblue.py` ships in the image but the box runtime never imports it (CLI-`__main__` only; `build_pk_repair_prompt` has zero callers) — acknowledged as wiring-pending, so just a bit-rot tracking note against `compile_box`'s inject seam. `effort` is a reserved param threaded through the PK engine stack but never set or read. `DISABLE_AUTOCOMPACT` is applied in `compile_box.py` + `engine.py` + Dockerfile ENV but not the `compile_agent.py` debug entrypoint (container ENV covers the shipped path, so low-severity).

kbc/platform/pod/selfcheck.py
```diff
@@ -0,0 +1,444 @@
...
+            break
+        if re.match(r"^[A-Za-z_][A-Za-z0-9_-]*\s*:", line):  # top-level key
+            in_list = False
+            key = line.split(":", 1)[0].strip()
+            rest = line.split(":", 1)[1].strip()
+            if key == "compiled_from":
+                has_key = True
+                # block form (`compiled_from:` alone) opens the list; inline
+                # `compiled_from: []` means key present with zero entries
+                in_list = rest == ""
```
jacoblee-io: **[correctness] An inline flow-list `compiled_from: [a.md]` parses to zero sources, triggering a spurious repair on a correctly-cited page.**

`in_list = rest == ""` only opens the block form; an inline YAML flow list (`compiled_from: [raw/a.md]`) leaves `in_list` false, so `has_compiled_from` is set True (the lint sees provenance and stays silent — no `no_provenance` violation) but `sources` stays `[]`. The cited raw files are then falsely `unaccounted`, and L1 fires a repair round asking the model to compile what it already compiled. **Fix:** parse the inline `[...]` form too (or reject it in the lint so the mismatch is visible).

LikiosSedo:
Thanks — sharp review, all four are real. Fixed at `2fa07f9e`.

**[security/DoS] `_install_wiki_snapshot` size guards (compile_box.py).** Added the `max_bundle_bytes` (compressed) + accumulated `max_unpacked_bytes` checks the two sibling installers already enforce (`KBC_MAX_SNAPSHOT_BUNDLE_BYTES` 64 MB / `KBC_MAX_SNAPSHOT_UNPACKED_BYTES` 256 MB), so a gzip-bomb `bundle_base64` under the 768 MB HTTP cap is rejected before it buffers into memory. Test asserts both caps + a within-limits install.

**[correctness] L1 matcher (selfcheck.py `_matches`).** Both directions:
- **Over-exclusion:** replaced `fnmatch` with segment-aware globbing — `*`/`?` match within one segment, only `**` (or a trailing `/`) crosses `/`. `notes/*` now excludes the files directly under `notes/` but not `notes/sub/secret.md`.
- **Under-exclusion:** the exclusion pattern is now normalized through the same `raw/`/`drop/`-stripping as `compiled_from` (shared `_strip_source_prefix`), so `{"pattern": "raw/live.csv"}` matches inventory `live.csv`.
- Added a **no-op-exclusion warning** (`coverage.noop_exclusions`, surfaced in the narration) so a pattern that matches nothing is flagged as a likely typo — non-blocking.

**[correctness] `candidate_tree_hash` unguarded read.** Wrapped `read_bytes` in try/except — since `state_key` runs *before* `run_layer1`'s fail-open, a perm-denied/FIFO candidate file previously disabled the gate silently. Now it's hashed with empty bytes (so the tree hash still rotates and re-triggers) and `candidate_pages` surfaces the real read error as an `unreadable` lint violation.

**[correctness] inline `compiled_from: [a.md]`.** `parse_compiled_from` now parses the inline flow form too — no more spurious repair on a correctly-cited page.

**Design notes:**
- **Hash triplication → fixed.** Extracted `selfcheck.content_hash(pages)` as the single formula; `pack_candidates_to_wiki`, `candidate_tree_hash`, and `_install_wiki_snapshot` all call it, and a cross-module test asserts they agree (incl. a nested tree where Path-sort vs string-sort diverge).
- **"One bounded repair round" wording → fixed** in the README: clarified it's per gap-episode (resets on close), not a per-run bound.
- **Dead/pending code — acknowledged, tracking:** `redblue.py` / `build_pk_repair_prompt` are wiring-pending (S0 calibration gate) and `effort` is a reserved PK knob — both intentional bit-rot-risk to revisit when the auto-trigger lands. `DISABLE_AUTOCOMPACT` on `compile_agent.py` (the local debug entrypoint) — low, since the container ENV covers the shipped path; will add for consistency.

Stdlib suites green (`test_selfcheck` incl. the new A/B/C/D + content-hash cases, `test_redblue`, `test_office_ingest`); `test_compile_box` (incl. the new size-guard test) runs in CI.

jacoblee-io (approved):
Approving on `2fa07f9`. All four findings from my review are fixed correctly and with tests:

- **[security/DoS]** `_install_wiki_snapshot` now enforces `max_bundle_bytes` + accumulated `max_unpacked_bytes`, mirroring the sibling installers — the gzip-bomb OOM path is closed.
- **[correctness]** `_matches` is now segment-aware (`_seg_glob`: `*` no longer crosses `/`, `**`/trailing-`/` for subtrees) and normalizes a `raw/`/`drop/` prefix on the exclusion side, closing both the over-exclusion false-PASS and the prefix-mismatch false-GAP; no-op patterns are surfaced as a non-blocking warning. Verified the recursion against `notes/*`, `*.md`, and `**/x.md`.
- **[correctness]** `candidate_tree_hash` guards `read_bytes` (empty-bytes fallback, no raise before the fail-open), so an unreadable candidate file no longer silently disables the coverage gate.
- **[correctness]** `parse_compiled_from` handles the inline flow list `compiled_from: [a.md]`, ending the spurious repair on a correctly-cited page.
- **[design]** the digest formula is extracted to one `content_hash` shared by all three callers (cross-module consistency test added); README corrected to "per gap-episode" repair budget.

No regressions in the fix commit; CI green. Remaining items are the low/acknowledged ones (redblue.py wiring-pending dead code, the reserved `effort` param, `compile_agent.py` autocompact — covered by the container ENV), all fine to defer. No cross-repo deploy gate here since `bundle_base64` has no frontend producer yet and the path is box-internal.

PR #386, opened 2026-07-06
feat(kbc): compile-box robustness — model-stall watchdog, rate-limit backoff, graceful shutdown

LikiosSedo:
## What

Three box-side robustness fixes for the KB compile pod (`kbc/platform/pod/compile_box.py`), **stacked on #384 (batch compile)** because they harden its batch / persistent / planner drivers. No runtime or contract changes.

**L1 — model-stall watchdog.** A compile turn can wedge on a black-holed model request (massapi/upstream accepts the connection but never streams a response). The bundled `claude` CLI's own request timeout is ~60min and not tunable from the box, so a routine upstream hiccup froze a live run for an hour (observed: a batch-1 first request sat silent 63 min before the CLI timed out and recovered on its own). A turn-scoped watchdog reaps the wedged request and re-issues the turn in seconds. Idle is measured as model-response latency — a pending tool relaxes the bound, and `include_partial_messages` gives fine-grained liveness — so a live-but-slow generation is never false-killed. Exhausted retries fail the run with a clear reason instead of hanging.

**C2 — rate-limit resilience.** Under 5-10 concurrent boxes (× the red-blue PK) massapi can 429 / 503 / 529. Past the CLI's internal retries the turn ends with `is_error` + `api_error_status` (CLI ≥ 2.1.110). We back off (exponential, capped) and re-issue rather than failing the run; exhaustion ends the turn with an owner-facing note (run idle), not a crash. massapi's limits are not ours to fix — this is graceful handling.

**F3 — graceful shutdown flush.** The box had no SIGTERM handling: on a pod kill (delete / eviction / OOM after a reap) the last ≤`SYNC_INTERVAL_SECS` of compiled work — living only on the emptyDir `/work` — was lost, because `_run_wrapper`'s final sync never runs on signal death. An aiohttp `on_shutdown` handler final-syncs every active run's workspace and gives the SSE relay a bounded window to drain. Best-effort within the grace period.

## Coverage

The watchdog runs for the whole run (`_run_wrapper`) and monitors shared `CompileRun` state, so it covers every model call — the persistent driver, each batch session, and the batch planner. The stall/rate retry seams live in a shared `_consume_turn_stream`; `_emit_message` is unchanged. New events (`turn_stalled`, `rate_limited`, `[shutdown]` summary) are forward-compatible — the runtime default-ignores them.

## Tests

Box suite green (protocol smoke + all unit tests), including 7 new: black-hole → interrupt+retry recovers; live-but-slow stream not reaped; long tool not reaped; stall exhaustion → error; 429 backoff → completes; 429 exhaustion → graceful; shutdown flush. Env knobs: `KBC_MODEL_IDLE_TIMEOUT_S` (90), `KBC_MODEL_TOOL_IDLE_TIMEOUT_S` (660), `KBC_MODEL_MAX_RETRIES` (3), `KBC_MODEL_RATE_MAX_RETRIES` (5), `KBC_MODEL_RATE_BACKOFF_BASE_S`/`_CAP_S` (2/30), `KBC_SHUTDOWN_DRAIN_S`/`_MAX_S` (0.5/8).

## Deploy

Box behavior change → needs a `kbc-compile-box` image rebuild (runtime env `SICLAW_COMPILE_BOX_IMAGE`); the runtime is unchanged and default-ignores the new events. Runbook: `RUNBOOK-kb-compile-deploy-2026-07-06.md`.

## Scope (deliberately out of this PR)

Assessed against an explicit "basic usability, skip low-probability" bar: **F1** (artifact-persistence ack/reconcile) and **C1** (concurrency memory tuning) were analyzed and intentionally not done (F1 is self-mitigating via periodic re-sync + coverage ledger + recovery self-heal; C1 is low-probability on the large GPU nodes and the per-box 4Gi limit is empirically adequate). **L4** (turn-delivery durability) and the **two-clock-watchdog deploy** remain, coupled to the run-manager review on #378. Full analysis: `REVIEW-kb-compile-robustness-audit-2026-07-06.md`.

> Stacked on #384 — GitHub retargets the base to `main` automatically when #384 merges.

### DIFF
--- a/kbc/platform/pod/compile_box.py
+++ b/kbc/platform/pod/compile_box.py
@@ -170,6 +170,32 @@ _MODEL_TOOL_IDLE_TIMEOUT_S = float(os.environ.get("KBC_MODEL_TOOL_IDLE_TIMEOUT_S
 _MODEL_MAX_RETRIES = int(os.environ.get("KBC_MODEL_MAX_RETRIES", "3"))
 _MODEL_WATCHDOG_POLL_S = float(os.environ.get("KBC_MODEL_WATCHDOG_POLL_S", "10"))
 
+# ── Rate-limit resilience (C2) ───────────────────────────────────────────────
+# massapi under concurrency (5-10 boxes × the red-blue PK) can return 429/503/529.
+# The bundled CLI retries internally a few times; past that the turn ends with
+# is_error=True and api_error_status set (CLI >= 2.1.110). We back off and re-issue
+# rather than failing the run — massapi's limits are not ours to fix (out of
+# scope), so this is graceful handling, not a fix. Exhaustion ends the turn with a
+# clear owner-facing note instead of a crash.
+_MODEL_RATE_STATUSES = frozenset({429, 503, 529})
+_MODEL_RATE_MAX_RETRIES = int(os.environ.get("KBC_MODEL_RATE_MAX_RETRIES", "5"))
+_MODEL_RATE_BACKOFF_BASE_S = float(os.environ.get("KBC_MODEL_RATE_BACKOFF_BASE_S", "2"))
+_MODEL_RATE_BACKOFF_CAP_S = float(os.environ.get("KBC_MODEL_RATE_BACKOFF_CAP_S", "30"))
+
+# ── Graceful-shutdown flush (F3) ─────────────────────────────────────────────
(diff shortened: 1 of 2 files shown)


LikiosSedo:
Folded into #384 (feat/kb-batch-compile) to reduce the open-PR count — these commits now live on that branch (compile-box production hardening lands together with batch compile). Closing; the branch is deleted.

PR #387, opened 2026-07-07
feat(capability): consumer-managed LLM credential + KBC settings for kb boxes

LikiosSedo:
Stacked on #384 (DESIGN-kb-llm-binding-v2). The consumer (sicore) becomes the single source of truth for the KB boxes' LLM endpoint/token and behavior knobs; helm stays minimal (cluster fallback + ops kill switch only).

- `contract.ts`: `CapabilityFetchInputResponse` gains optional `llm{base_url, auth_token}` + `settings{KBC_*}` — same channel and shape as the `locale` precedent. Absent fields ⇒ the runtime-env (helm Secret) fallback.
- `materialize.ts` / `server.ts`: opaque passthrough into the box `/session` body; never logged; token never lands in the pod spec (the box spawns before fetchInput, so this is in-process config by construction).
- `compile_box`: `_apply_session_config()` applies `llm` + whitelisted `KBC_*` keys to `os.environ` before any SDK/engine session exists. Precedence: consumer config > runtime env forward > image defaults — with ONE exception: a runtime-level `KBC_PK_MODE=off` is the ops kill switch and outranks consumer config. A stale forwarded `ANTHROPIC_API_KEY` is cleared when a consumer token arrives.
- Compatible both ways: old sicore → env fallback; new sicore against an old box → unknown body fields ignored.

sicore counterpart: `feat/kb-llm-binding` on GitLab (maintainer config table + resolve + maintainer dialog).

Tests: `_apply_session_config` unit (llm/whitelist/kill-switch), capability vitest 39, five kbc pod suites, tsc clean.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,7 @@ RUNTIME_IMAGE  = $(REGISTRY)/siclaw-runtime:$(TAG)
 AGENTBOX_IMAGE = $(REGISTRY)/siclaw-agentbox:$(TAG)
 PORTAL_IMAGE   = $(REGISTRY)/siclaw-portal:$(TAG)
 OCR_IMAGE      = $(REGISTRY)/siclaw-ocr:$(TAG)
+KBC_IMAGE      = $(REGISTRY)/kbc-compile-box:$(TAG)
 
 # ── OCI labels injected into every image ──
 DOCKER_LABELS = \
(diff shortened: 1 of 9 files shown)


LikiosSedo:
Folded into #384 (feat/kb-batch-compile) to reduce the open-PR count — these commits now live on that branch (compile-box production hardening lands together with batch compile). Closing; the branch is deleted.

PR #388, opened 2026-07-09
feat(channels): Feishu reply feedback + per-agent bot config on the Agent Channels tab

jacoblee-io:
## Summary

Two related Feishu-channel features in one branch.

### 👍/👎 reply feedback

The finalized answer card now carries a locale-aware feedback row. Clicks arrive as `card.action.trigger` callbacks over the **same long connection** as messages — no public webhook.

- The button's `action.value` is **self-contained** (`session_id + card_id + channel_id + locale`), so persistence needs no Feishu-message-id → session mapping; the post-click echo rebuilds the row from that payload (the in-memory map keeps only the card's sequence counter).
- New `message_feedback` table: one row per (reply, person), upsert on `(message_ref, sender_external_id)` so a re-vote flips the rating. `message_ref` stores the CardKit `card_id`; FK to `chat_sessions` cascades.
- New `chat.recordFeedback` RPC (WS handler + internal HTTP mirror). The callback validates, **races the persist against a 2.5s deadline** (Feishu's callback window is seconds; the WS RPC timeout is 30s), toasts, and best-effort highlights the chosen button. A DB failure (e.g. clicking a card whose session was deleted) returns `{success:false}` instead of throwing.
- Feedback is attached **only to a real answer** — never under an error or empty-result notice.
- Buttons are appended **after streaming mode is off** (structural element ops on a streaming card risk rejection); non-zero CardKit codes (the SDK does not throw on them) are treated as failures.

### Per-agent Feishu bot on the Agent Channels tab

Mirrors the sicore design: the primary way to connect an agent to Feishu is a dedicated app configured on the agent's own **Channels tab**; the shared-app + manual-PAIR flow becomes a collapsed **Advanced** section (hidden once the dedicated bot is active). Channel config stays **admin only**.

- `GET/PUT/DELETE /api/v1/siclaw/agents/:id/personal-bot`. The bot is a regular `channels` row whose config carries `personal_bot { agent_id, access_mode:'open', owner_user_id, group_auto_bind }` — the adapter's `channel.list` / `resolvePersonalChannelBinding` already understand that shape, so the gateway serves it with no extra plumbing. GET never returns the secret; PUT keeps the stored secret when blank and rejects a resolved-empty secret, and preserves the stored owner on update; DELETE disables (keeps credentials).
- **Every channel write** (shared-app CRUD + personal-bot upsert/disable) pushes a best-effort `channel.reload` to connected runtimes via the shared `broadcastChannelReload`, so changes go live without a restart — fixes deleted channels lingering on stale WS connections.
- Agent Channels tab reworked to three sections: collapsible bot card (admins edit in place, others read-only), Active groups, Advanced shared-app pairing.
- Settings → Channels lists only shared apps; the list endpoint flags personal-bot rows server-side (config never leaves the handler on the list path) and the page filters them out.

## Deploy notes

- Feishu console: the app's **event callbacks must use the long-connection subscription** (same place as event subscription). No new permission scopes.
- Standalone stores channel `app_secret` in the `channels.config` JSON as today's shared-channel path does (no behavior change). Secret encryption for K8s/sicore is a separate, Go-side concern.

## Tests

`npm test` (208 files, 4227 passed) + `tsc --noEmit` + portal-web `vitest`/`tsc` all clean. New coverage: feedback button build/echo/failure paths, card-action callback (persist, malformed, deadline-race, DB-error → success:false), personal-bot routes (admin-gate, secret redaction, config shape, channel.reload push), channel list personal-bot flagging.


src/gateway/channels/lark.ts
```diff
@@ -322,6 +335,82 @@ function backfillBindingDisplayName(
...
+    channelId: value.channel_id ?? null,
+    source: "lark",
+  }).catch((err): { success: boolean; error?: string } => {
+    console.error(`[lark] Feedback persist failed card=${value.card_id}:`, err);
+    return { success: false, error: err instanceof Error ? err.message : String(err) };
+  });
+  const result = await Promise.race([
+    persist,
+    new Promise<null>((resolve) => {
+      const timer = setTimeout(() => resolve(null), FEEDBACK_PERSIST_DEADLINE_MS);
```
LikiosSedo: Nit (non-blocking): this deadline timer is never `clearTimeout`-ed after the race settles. When `persist` wins quickly (the common case), the 2.5s timeout still stays armed until it fires. `unref()` correctly keeps it from holding the event loop open, so there's no correctness or shutdown impact — just a dangling timer per click for up to `FEEDBACK_PERSIST_DEADLINE_MS`. If you want it clean, capture the timer and `clearTimeout` it once `Promise.race` resolves.

src/gateway/channels/lark.ts
```diff
@@ -322,6 +335,82 @@ function backfillBindingDisplayName(
...
+ * RPC client (same path as message persistence); the button-state echo on the
+ * card itself is best-effort (needs this process to still know the card's
+ * sequence).
+ */
+export async function handleLarkCardAction(
+  data: any,
+  larkClient: any,
+): Promise<{ toast: { type: string; content: string } } | undefined> {
+  // EventDispatcher flattens `event.*` onto the top level (same as messages).
+  const value = data?.action?.value as Partial<FeedbackActionValue> | undefined;
```
LikiosSedo: This assumes Feishu delivers the callback `action.value` as an already-parsed object (we read `value.kind` / `value.rating` directly). That matches the CardKit `behaviors: [{ type: "callback", value: <object> }]` shape and your deploy notes show this was validated against live long-connection callbacks, so it's very likely fine. Flagging only because some Feishu/CardKit versions serialize `action.value` to a JSON string — if that ever happens here, `value.kind` is `undefined` and the handler silently `return undefined`s (feedback quietly no-ops, no crash). Cheap insurance would be a `typeof value === "string" ? JSON.parse(value) : value` guard before the discriminator check. Worth a quick confirm across the Feishu client versions you target.

LikiosSedo (commented):
Reviewed both features end to end (feedback chain + per-agent bot) with full context, and ran the affected suites locally — all green: `lark-card`, `lark` (96), `adapter-rpc` (140), `channel-api` (14), `siclaw-api.personal-bot` (10), `migrate-sqlite`.

Looks solid and mergeable — no blocking issues. A few things I verified hold up well: the `config.personal_bot` shape the new PUT writes matches exactly what the pre-existing `resolvePersonalChannelBinding` / `channel.list` expect (so the "zero extra plumbing" claim is accurate); secret redaction is airtight on all three paths (GET, list config-strip, create/update select-back); and the FK-failure → `{success:false}` degradation is correct and idempotent.

Two non-blocking nits inline — neither needs to gate the merge.

jacoblee-io:
Both non-blocking nits addressed in d74015ad:

1. **Dangling deadline timer** — the `Promise.race` timer is now captured and `clearTimeout`-ed once the race settles, so the common fast-persist path no longer leaves it armed for the full 2.5s.

2. **Stringified `action.value`** — added the `typeof value === "string" ? JSON.parse(value) : value` guard before the discriminator check, so feedback works whether a CardKit version delivers the value as an object or a JSON string (instead of silently no-oping on the string case). Covered by a new test.

Thanks for the thorough review.

LikiosSedo (approved):
Both nits from the previous pass are addressed correctly, each with a test:

- `action.value` now accepts the JSON-string delivery form (`try/catch → undefined` keeps the malformed case as a silent no-op, object path unchanged) — covered by the new `accepts action.value delivered as a JSON string` test.
- The deadline timer is captured and `clearTimeout`-ed after the race settles; the executor runs synchronously during `Promise.race([...])` construction, so the timer is always assigned before the clear — no race.

Re-ran the affected suites locally: `lark.test.ts` (97) + `lark-card.test.ts` all green (128 passed). Verified earlier that the `config.personal_bot` shape matches the pre-existing resolver, secret redaction holds on all three paths, and the FK-failure degradation is correct/idempotent.

LGTM 🚀

PR #389, opened 2026-07-09
feat(runtime): advertise KB-compile capability to the consumer on connect

LikiosSedo:
## What

The Runtime now advertises its capabilities (currently `{ compile }`) to the consumer over the existing `/ws/runtime` connection, via a best-effort `runtime.register` RPC sent on every (re)connect.

`compile` is `true` iff a compile-box image was configured for this Runtime — i.e. helm `agentbox.compileBoxEnabled` set `SICLAW_COMPILE_BOX_IMAGE`. This is exposed as `isCompileCapable()` (single source of truth, next to where the compile-box profiles resolve that image).

## Why

Lets the consumer route KB-compile runs to a compile-capable Runtime **from the Runtime's own config**, with zero consumer-side routing configuration.

Every connected Runtime sends `runtime.register`, but only one with `compileBoxEnabled` advertises `compile: true` (the rest send `compile: false`) — so in a multi-Runtime deployment the consumer picks the capable one with no per-Runtime env/values on its side. A non-capable Runtime explicitly reporting `compile: false` is intentional: the consumer learns "not me" rather than inferring it from silence. Enabling/moving compile becomes a one-line change on the Runtime's helm.

## Compatibility

Advertisement is advisory and non-fatal. A consumer that predates `runtime.register` replies `unknown method` and keeps the connection; the Runtime ignores the failed ack (verified: the consumer's WS dispatch returns an error response for unknown methods without dropping the connection). New-Runtime ↔ old-consumer therefore stays safe — normal chat is unaffected.

## Tests

`frontend-ws-client.test.ts`: advertises on open, does not advertise when capabilities are empty (the generic-client guard; the production caller always passes `{ compile }`), re-advertises on every reconnect, survives an `unknown method` ack. `tsc` clean, 20/20 pass.

Pairs with the consumer-side change that stores the advertisement and resolves the compile Runtime from it.

### DIFF
--- a/src/gateway/agentbox/box-profile.ts
+++ b/src/gateway/agentbox/box-profile.ts
@@ -48,6 +48,19 @@ export const AGENT_PROFILE: BoxProfile = {
 };
 
 /**
+ * Whether this Runtime can host KB compile/test boxes — i.e. it was given an
+ * explicit compile-box image (helm `agentbox.compileBoxEnabled` ⇒ the
+ * SICLAW_COMPILE_BOX_IMAGE env). This is the single source of truth the Runtime
+ * advertises to its consumer on connect so the consumer can route compile runs
+ * here WITHOUT any consumer-side config. The bare `kbc-compile-box:latest`
+ * fallback in the profiles above is deliberately NOT treated as capable: an
+ * unset env means KB stays dark (fail-closed), so we must not claim capability.
+ */
+export function isCompileCapable(): boolean {
+  return !!process.env.SICLAW_COMPILE_BOX_IMAGE?.trim();
+}
+
+/**
  * kb-compile — a KB compile box. Reproduces the pre-refactor compile special-case
  * DECLARATIVELY: dedicated image, Anthropic-compatible LLM env (the lean box does
  * not phone home for settings), a writable /work with HOME pointed at it.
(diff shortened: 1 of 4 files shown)


jacoblee-io (approved):
Approving on `4120c04f`. Clean, well-scoped, well-tested. Verified:

- `isCompileCapable()` reads the same `SICLAW_COMPILE_BOX_IMAGE` env the kb-compile/kb-test profiles resolve their image from, so the advertised capability matches the actual image source. Env-unset ⇒ advertises `compile:false` despite the profile's `kbc-compile-box:latest` fallback — the intended fail-closed (KB stays dark unless explicitly enabled), and the comment says so.
- Only one `FrontendWsClient` constructor (bootstrap-runtime), so the wiring is complete; `capabilities` defaults to `{}`, so nothing else is affected.
- `advertiseCapabilities()` is fire-and-forget with a `.catch`, re-fires on every reconnect, and the `unknown method` ack from an older consumer can't break the connection or `connect()` — matching the stated compatibility contract. Tests cover open-advertise / empty-no-advertise / reconnect / unknown-method-survival.

One benign note (non-blocking): bootstrap always passes `capabilities: { compile: <bool> }`, which is never empty, so the "empty ⇒ don't advertise" guard never triggers in production and a non-capable Runtime advertises `{compile:false}` rather than staying silent. That's harmless (arguably better — the consumer explicitly learns "not me"), but the PR description's "only the Runtime with compileBoxEnabled advertises compile" is slightly imprecise: every Runtime sends `runtime.register`; only a capable one advertises `compile:true`. No change needed.

PR #390, opened 2026-07-09
fix(capability): malformed artifact entry must not kill the relay; stop the box when it does die

LikiosSedo:
## Problem (seen live on siflow-test, 2026-07-09)

A relay crash in `driveCapabilitySession` right after `turn_done` (a runtime predating the tombstone branch threw `ERR_INVALID_ARG_TYPE` in artifact construction) permanently killed the run's event pipe. Everything the box emitted afterwards was lost:

- the final SELFCHECK sync → the store kept a mid-repair state (a dangling-citation warning shown on the publish page forever),
- `converge_phase: settled` → the consumer's converge hook never fired, so the incremental baseline never advanced,
- deletion tombstones → deleted pages resurrected from the store on the next respawn,
- and the box pod leaked (the relay-failure path only `endRun(failed)`s — 4 live boxes observed for one repo).

## Fixes (three boundaries)

1. **session-driver**: artifact construction moved inside a per-entry guard. A malformed entry — still reachable today via a non-string `content` — is skipped with a loud log instead of killing the stream and everything after it.
2. **server relay catch**: stop the box before terminalizing the run. A failed run can never be re-adopted (`adopt` refuses terminal runs), so its box is unreachable garbage; without this every relay crash leaks a pod.
3. **capability.cancel**: the swallowed `stop` failure now logs loudly — it is exactly how pods leak silently.

## Tests

`session-driver.test.ts`: malformed entry skipped + later artifacts still land + run not failed; deletion tombstone persists as `{deleted:true}`. 11/11 pass.

### DIFF
--- a/src/gateway/capability/session-driver.test.ts
+++ b/src/gateway/capability/session-driver.test.ts
@@ -128,6 +128,57 @@ describe("driveCapabilitySession — box event → capability wire mapping", () 
     expect(artifactCalls).toHaveLength(2);
   }, 10_000);
 
+  it("a malformed artifact entry is skipped loudly — the relay and later artifacts survive", async () => {
+    // Live 07-09: a runtime predating the tombstone branch threw in artifact
+    // construction (Buffer.from(undefined)) and the WHOLE relay died right after
+    // turn_done — the final SELFCHECK sync, settled, and end were lost, every
+    // incremental round wedged DIRTY, and the box pod leaked. Construction is
+    // still throwable today (a box bug sending a non-string content), so the
+    // guard is a boundary rule, not a legacy patch: skip the bad entry, keep
+    // the stream.
+    const fe = fakeFrontend();
+    const mgr = fakeManager();
+    const errSpy = vi.spyOn(console, "error").mockImplementation(() => {});
+    try {
+      await driveCapabilitySession({
+        client: fakeClient([
+          {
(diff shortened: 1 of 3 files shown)


jacoblee-io (approved):
Approving on `4ddc4b4e`. Clean, on-point root-cause fix — all three boundaries check out:

- **session-driver**: artifact construction is now inside a per-entry try/catch, so a non-string `content` (Buffer.from throw) skips just that entry with a loud log + `continue` instead of killing the whole loop. This is the root cause — the downstream losses the PR describes (final SELFCHECK sync, `converge_phase: settled`, deletion tombstones) were all just consequences of the relay dying after turn_done, so keeping the stream alive fixes them together. No throw source remains in the loop that can escape (persist requests keep their own retry/log catch).
- **server relay catch**: `stop(runId)` before `endRun(failed)` — a terminal run can't be re-adopted, so its box is unreachable garbage; ordering (stop then terminalize) matches the onReap invariant, and `agentBoxManager.stop` here is the same call/​pattern used at onReap. Plugs the observed pod leak.
- **capability.cancel**: the swallowed stop failure now logs loudly — observability only, behavior unchanged.

Tests cover malformed-entry-skipped + later-artifact-still-lands + run-not-failed, and tombstone persistence. CI green.

PR #391, opened 2026-07-09
feat(kbc): deterministic convergence — code-restore out-of-scope edits, gate settle on dangling citations, escalate residuals as tickets

LikiosSedo:
## Principle

What code can detect and fix, **code fixes**; what needs the model, the **bounded repair loop** handles; what neither resolves becomes a **ticket** in the owner's question queue. The publish page only *displays* residual state — it must never be where the owner discovers work.

Live context (siflow-test, 2026-07-09): a lone dangling citation settled straight through to the publish page as owner homework ("1 处引用指向不存在的源，建议回生成页让 AI 核对"), and 3/3 scoped incremental rounds landed `unconverged` because the byte-integrity guard asked the *model* to restore pages byte-exact from a sha256 — structurally impossible.

## L0 — incremental byte guard restores by code

The pre-turn snapshot now keeps page **bytes** alongside hashes. Out-of-scope violations are restored mechanically (modified pages written back, deleted pages recreated, byte-exact) **before** the coverage ledger runs, so the ledger judges the restored tree. The repair prompt survives only as a fallback for a snapshot miss. `SELFCHECK.incremental` gains `restored_pages` for transparency and a summary line narrates the restore.

## L1 — dangling citations gate settle

`coverage.closed` now requires consistency in **both** directions: every source accounted *and* every citation real. The repair prompt already carried the dangling-fix instructions — the gate (`ledger_clean`) just never fired on dangling alone. Narration names the dangling count so a dangling-only repair round is explained.

## L2 — residuals become a ticket, not publish-page homework

When the repair budget is spent with residuals (`state=unconverged`), the driver files **one** contradiction ticket by code — the model's own schema, a stable content-fingerprint id (repeat unconverged rounds with identical residuals never duplicate it), never clobbering the model's tickets or an unreadable ledger. It rides the existing rule → dispatch → `resolve_ticket` loop; incremental and full rounds share it.

`box_role.md` (zh/en): the out-of-scope clause now tells the model the system auto-restores drift — and not to lean on that backstop, since a restore also discards whatever improvement it slipped in.

## Tests

- `test_incremental.py`: `page_bytes` + `restore_pages` (modified written back, deleted recreated, created skipped) — 9/9.
- `test_selfcheck.py`: dangling alone blocks `closed` + narration names it; `file_residual_ticket` files/dedupes/preserves/bails — all green.
- `test_compile_box.py`: violation auto-restored (no repair turn, byte-exact, reported, guard consumed); unconverged files exactly one ticket — full suite green, incl. the pre-existing guard tests (which now exercise the snapshot-miss fallback path).
- `test_redblue` / `test_batching` / `test_mediaverify` / `test_office_ingest`: all green.


jacoblee-io (commented):
## Review: deterministic convergence (L0 code-restore / L1 dangling gate / L2 residual ticket)

Good direction and the principle is right (code fixes what it can, residuals become a ticket, the publish page only displays). But four issues — two of them are prior latent hazards (#384 path normalization, #382/#384 non-atomic write) that this PR *escalates* into harder failures now that dangling gates settle and a ticket is code-filed. Inline.

**Lower-severity / by-design notes (not inline):**
- **L2 fingerprint collision (low):** `file_residual_ticket` fingerprints the *truncated* residual display list (`[:10]` items / `[:80]`-char slices), so two genuinely different residual sets that share the first 10 items in a category collide to the same ticket id and the second is silently deduped away. Fingerprint the full residual set (sorted) before truncation.
- **L0 restore discards in-scope improvements (by-design, worth watching):** restoring an out-of-scope page to baseline also drops any legitimate improvement the model bundled into it (box_role admits this). If that improvement had fixed a coverage/lint issue, the ledger re-flags it, the model is told to redo it, but the page still isn't authorized next turn (the repair_pages widening only covers lint/dangling targets, not arbitrary in-scope facts) → the "improvement undone → unconverged" loop can reappear, relocated. Acceptable tradeoff, but the widening doesn't fully close it.

kbc/platform/pod/selfcheck.py
```diff
@@ -434,7 +434,12 @@ def coverage(workdir: str, pages: dict[str, dict], exclusions: list[dict]) -> di
         "unaccounted": unaccounted,
         "dangling_citations": dangling,
         "noop_exclusions": noop_exclusions,
-        "closed": not unaccounted,
+        # closed = the ledger is consistent in BOTH directions: every source
+        # accounted AND every citation real. dangling used to be display-only —
+        # the repair prompt listed the fix, but the gate (ledger_clean) never
+        # fired on it, so a lone dangling citation sailed through settle and
+        # surfaced as owner homework on the publish page.
+        "closed": not unaccounted and not dangling,
```
jacoblee-io, line 461: **[L1, med-high] Now that dangling gates `closed`, a non-normalized `compiled_from` path wedges convergence — the exact publish-page homework this PR set out to kill.**

`_norm_source_entry` (line ~93, unchanged) strips `raw/`/`drop/` prefixes and quotes but does **not** `posixpath.normpath` — whereas intra-wiki link resolution (lines ~278/~329) does. So a citation like `compiled_from: - ./live.csv` (or `sub/../x.md`) stays literal in `cited`, while `source_inventory` yields `live.csv`. The same file is then flagged **both** `unaccounted` (`live.csv` ∈ source_set, ∉ cited) **and** `dangling` (`./live.csv` ∈ cited, ∉ source_set). With the new gate `closed = not unaccounted and not dangling`, that's `closed=False` — a permanent block: the model gets contradictory repair orders, burns the (default 1) repair round, and every round L2-files a ticket for *correct* work. In #384 this was a cosmetic double-report (dangling was display-only); gating settle on dangling turns it into a convergence wedge. **Fix:** `posixpath.normpath` in `_norm_source_entry` so citation paths canonicalize the same way link targets do.

kbc/platform/pod/selfcheck.py
```diff
@@ -710,6 +719,95 @@ def build_repair_prompt(report: dict, locale: str | None = None) -> str:
...
+                     if en else "自检自动回修额度已用完,以下残留没有修完,要怎么处理?"),
+        "sources": [{"doc": "authoring/SELFCHECK.json", "quote": "; ".join(residuals)[:600]}],
+        "options": (["Run another repair round", "Accept as-is"] if en else ["再修一轮", "接受现状"]),
+        "current_value": "unresolved residuals" if en else "残留未处理",
+        "affected_pages": sorted(pages)[:20],
+        "status": "open",
+        "answer": None,
+    })
+    path.parent.mkdir(parents=True, exist_ok=True)
+    path.write_text(json.dumps(tickets, ensure_ascii=False, indent=2), encoding="utf-8")
```
jacoblee-io: **[L2, med-high] `file_residual_ticket` writes CONTRADICTIONS.json with a bare `write_text` — not the atomic helper — so a torn write corrupts the owner's whole ticket queue.**

This is the #382/#384 hazard recurring: `resolve_ticket` and `propose_questions` both write this same file via `_write_text_atomic` (temp + `os.replace`). This new writer does a read-modify-write (reads the model's tickets, appends the residual ticket, rewrites the array) with a plain `write_text`. A crash/interleave mid-write — right at the budget-exhausted `unconverged` turn seam where this fires — truncates the file, dropping the model's own tickets, and every later ticket read/write then bails on the JSON error. **Fix:** write via `_write_text_atomic` (import it into selfcheck, or replicate temp-file + `os.replace`).

kbc/platform/pod/incremental.py
```diff
@@ -141,6 +141,37 @@ def page_hashes(workdir: str) -> dict[str, str]:
...
+    本护栏;或快照缺失)防御性跳过,留给回修指令兜底。返回实际还原的页。"""
+    restored: list[str] = []
+    cand = Path(workdir) / "candidate"
+    for rel in pages:
+        original = before_bytes.get(rel)
+        if original is None:
+            continue
+        target = cand / rel
+        target.parent.mkdir(parents=True, exist_ok=True)
+        target.write_bytes(original)
```
jacoblee-io: **[L0, med] `restore_pages` has no per-page guard — one unrestorable page aborts the whole restore and silently disarms the guard for the round.**

The loop does `target.write_bytes(original)` with no per-page try/except. If a single page can't be written (its path is now a directory, a permission/OS error), the exception propagates out of `restore_pages`; it's caught only by the coarse fail-open around `_post_turn_selfcheck`, so the entire self-check (remaining restores + ledger + narration + re-arm) is skipped — and `_incr_pending` was already consumed, so the drift is neither restored nor re-flagged on a later turn. The docstring promises unrestorable pages are "left for the repair-prompt fallback," but a mid-loop throw skips the later violations too. **Fix:** wrap each page's restore in try/except (log + continue), so one bad page falls through to the ledger/repair fallback instead of disarming the whole guard.

kbc/platform/pod/compile_box.py
```diff
@@ -1367,17 +1367,34 @@ async def _post_turn_selfcheck(run) -> str | None:
     if not (Path(workdir) / "candidate" / "index.md").is_file():
```
jacoblee-io: **[L0, med-low] The `index.md`-missing early-return fires after `_incr_pending` is already consumed, so an incremental turn that deletes index.md escapes the byte-freeze guard.**

`_incr_pending` is read and cleared a few lines up (`run._incr_pending = None`), then this line returns `None` when `candidate/index.md` is absent. On an *incremental* turn index.md always pre-existed, so "index missing now" is itself an out-of-scope deletion the guard should catch/restore — but this early-return skips the guard entirely and the snapshot is already gone, so the deletion is never restored or re-flagged. The early-return is meant for the full-compile "index not written yet" case. **Fix:** when `incr` is armed, run the restore/guard before the index check (or don't early-return on missing index for an incremental turn).

kbc/platform/pod/selfcheck.py
LikiosSedo: Fixed in 6d4db64. `_norm_source_entry` now runs `posixpath.normpath` after prefix-stripping (with the `"" → "."` edge guarded), so citation paths canonicalize exactly like link targets. Agreed on the escalation framing — this was the same pattern the interlock bug had: a latent double-report turning into a convergence wedge once the gate tightened. Test: `test_citation_path_normalization` (`./snap/live.csv` + `snap/sub/../sub/x.md` → closed, no double-report).

kbc/platform/pod/selfcheck.py
LikiosSedo: Fixed in 6d4db64. selfcheck now carries a local `_write_text_atomic` (same-dir uuid temp + `os.replace`, cleanup-and-raise on failure — mirrors the driver's helper verbatim; importing it would cycle) and `file_residual_ticket` writes through it. Also took the review-note fix in the same commit: the ticket id now fingerprints the FULL sorted residual set — the `[:10]` caps are display-only (`test_residual_fingerprint_full_set`: two sets sharing a 10-item prefix file two distinct tickets).

kbc/platform/pod/incremental.py
LikiosSedo: Fixed in 6d4db64. Each page's restore is wrapped in `try/except OSError` (log + continue), so an unrestorable page stays an unrestored violation for the repair-prompt fallback while every other page is still restored and the ledger pass behind them still runs. Test: `test_restore_skips_unrestorable_page` (violation path became a directory → skipped, sibling restored byte-exact).

kbc/platform/pod/compile_box.py
LikiosSedo: Fixed in 6d4db64, with one addition beyond the suggested reorder: falling through alone wasn't enough — `_orphan_pages` returns `[]` with no root ("index missing is gated elsewhere"… the elsewhere being exactly this early-return), so an index-deleting turn would have settled green. Incremental turns now skip both early-returns (a byte-unchanged tree still short-circuits — no violations possible), and a new `index_missing` lint rule takes over the check: any tree that reaches lint must have its routing page (the full-compile mid-Execute case still early-returns, so no false positives). Test: `test_incremental_index_deletion_cannot_escape_guard` (index deleted + out-of-scope drift → drift restored, `index_missing` repair ordered, state=repairing).

LikiosSedo:
All four findings fixed + the fingerprint note, in 6d4db64 — replies inline. On the second by-design note (L0 restore discarding in-scope improvements that the widening doesn't cover): agreed it's a relocated-loop risk worth watching rather than fixing preemptively — the failure mode is bounded (unconverged → residual ticket, never silent loss), and today's live acceptance round exercised the widened path end-to-end clean (repair edited 3 out-of-scope charset pages, all survived, converged passed/settled, `restored_pages=[]`). If the relocated loop shows up in practice, the next widening increment is deterministic too (authorize pages the ledger's unaccounted-fix names via their compiled_from registration). All seven box test suites green.

jacoblee-io (approved):
Approving on `6d4db64`. All four review findings plus the fingerprint note are fixed correctly, with tests:

- **L1 citation normalization** — `_norm_source_entry` now `posixpath.normpath`s (returning "" for `.`), the same way link targets resolve, so a `./x`/`sub/../x` citation no longer double-reports as unaccounted AND dangling — closing the convergence wedge that gating `closed` on dangling would otherwise have created.
- **L2 atomic ticket write** — `file_residual_ticket` writes CONTRADICTIONS.json via a local temp+`os.replace` helper (selfcheck can't import the driver's without a cycle), so a torn read-modify-write on the shared ticket queue can't drop the model's tickets.
- **L0 per-page restore** — `restore_pages` guards each page (`try/except OSError` + log/continue); one unrestorable page no longer aborts the remaining restores and the ledger pass.
- **L0 index-deletion guard** — an incremental turn now falls through instead of early-returning after `_incr_pending` was consumed, and a new `index_missing` lint rule takes over the check, so deleting `candidate/index.md` can't settle green. The full-compile mid-Execute case still early-returns before lint, so no false positive.
- **L2 fingerprint** — the residual-ticket id now hashes the full sorted residual set (only the display quote is truncated), so two distinct residual sets sharing a prefix no longer collide to one ticket.

Read the whole +169 fix commit — surgical, no new regression; consumers stay consistent since `page["sources"]` is uniformly normpath'd. CI green (KBC Box / Portal Web / Test / Type Check).

PR #392, opened 2026-07-09
fix(lark): respond to card 👍/👎 callback immediately to avoid Feishu 200671

jacoblee-io:
## Problem

Clicking a Feishu answer-card 👍/👎 button intermittently pops **`code: 200671`** in the Feishu client — but the vote **is** recorded (Runtime logs `[lark] Feedback recorded …`, and the `message_feedback` row exists). `200671` is a Feishu **business error** meaning it did not receive a valid card-callback response in time.

## Root cause — timing, not the callback mechanism

Feishu enforces a hard **~3s budget on the callback response, measured end-to-end** (its edge → runtime pod → back). `handleLarkCardAction` **awaited the persist** before returning the toast, and persistence hops to Portal/sicore over the WS RPC. That latency + the two network legs intermittently exceeded 3s → Feishu rejected an otherwise-valid response with 200671, **after** the row had already been written. This is exactly why it's intermittent/load-correlated and why the data always lands.

The registration is **correct** and unchanged: the official Feishu Node SDK wires `card.action.trigger` through `EventDispatcher` + `WSClient.start({eventDispatcher})` and returns `{toast}` — identical to ours; SDK 1.24.0+ supports this over the long connection ([official docs](https://open.feishu.cn/document/feishu-cards/card-callback-communication?lang=zh-CN)). The SDK does base64 the handler's return value back as the callback response (verified in the 1.58.0 bundle). So the earlier hypothesis that we "must use a separate card handler / HTTP callback" is **not** the cause.

## Fix

Respond **optimistically and synchronously**: return the success toast the moment the click is validated, and run persistence + the button-highlight echo **fully detached**. The response now leaves in microseconds regardless of Portal/sicore latency, well inside the 3s budget. Feedback is best-effort (and reliable in practice); a rare persist failure is logged rather than surfaced — strictly better than showing 200671 on a click that did save. The 2.5s deadline-race is removed (too close to 3s once RTT is counted, and it still returned an error toast on the slow path).

## Verification

- Per the original report: after deploy, click 👍 once — Feishu should show the success toast (「已收到你的反馈，谢谢！」) and **no** 200671. Persistence was already fine and isn't the indicator.
- If any residual 200671 remains, pull the exact `msg` by Log ID on the open platform (200671 is Log-ID-diagnosable).

## Tests

`lark.test.ts` (97) + `lark-card.test.ts` (31) pass. Card-action suite rewritten for the synchronous optimistic-response contract: sync toast return, detached persist, slow-persist-does-not-block, persist-failure-swallowed, JSON-string `action.value`, locale/rating paths.

> Note: 8 unrelated suites fail locally on a missing `@arizeai/openinference-semantic-conventions` (tracing dep merged to main today, not resolvable in this env) — fails identically on clean `main`; CI/build-host with a fresh install is unaffected.

### DIFF
--- a/src/gateway/channels/lark.test.ts
+++ b/src/gateway/channels/lark.test.ts
@@ -1061,10 +1061,16 @@ describe("handleLarkCardAction — 👍/👎 feedback", () => {
     };
   }
 
-  it("persists the vote keyed by the clicker's open_id and returns a success toast", async () => {
-    recordChannelFeedbackMock.mockResolvedValue({ success: true });
-    const result = await handleLarkCardAction(makeCardAction(), makeLarkClient());
+  // Flush the detached persist/echo IIFE queued by the (synchronous) handler.
+  const flush = () => new Promise((r) => setImmediate(r));
 
+  it("returns the success toast synchronously and persists the vote detached", async () => {
+    recordChannelFeedbackMock.mockResolvedValue({ success: true });
+    // Synchronous return — the callback response must not await persistence.
+    const result = handleLarkCardAction(makeCardAction(), makeLarkClient());
+    expect(result).toEqual({ toast: { type: "success", content: expect.stringContaining("反馈") } });
+
+    await flush();
     expect(recordChannelFeedbackMock).toHaveBeenCalledWith({
       sessionId: "sess-1",
       messageRef: "CARD-1",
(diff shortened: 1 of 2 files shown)


chent1996 (approved):
LGTM

PR #393, opened 2026-07-09
fix(kbc): convergence hardening round 2 — eight findings from the full-implementation audit

LikiosSedo:
## Context

A four-perspective audit (convergence consistency / lifecycle robustness / contract coherence / scale & resources) was run over the merged #384 + #390 + #391 implementation. This PR carries **all three audit batches** plus two review-hardening rounds — it is materially larger than a box-only fix: the diff spans the kbc box (Python), the runtime capability gateway (TS), the K8s spawner/orphan GC, and the mTLS cert-Secret lifecycle.

## Batch A — the eight box-side convergence findings (`cdb6c2b4`)

1. **[HIGH] media-verify clean/failed paths hand back to the seam** — they used to just stop, parking `converge_phase` at `verifying` forever in the single-session case (test step never unlocked, PK never ran). The findings path re-enters the seam via its repair turn; clean/failed now chain PK, else settle.
2. **ADDED_TARGETS.json resets at round start** — a per-round declaration that nothing ever cleared: earlier rounds' merge targets stayed authorized forever, eroding the byte-freeze guarantee round over round.
3. **`dispatch_nonce` added to `resolve_ticket`'s schema** (+ zh/en tool descs) — the prompt ordered the echo and the consumer matches receipts to dispatch rounds by it, but a model follows the declared parameter list, so the echo was structurally always omitted.
4. **Failed incremental dispatch clears the arm** (kickoff `query` + repair-injection paths) — a turn that never started must not leave a stale snapshot; the next unrelated turn would be judged against it and the mechanical restore would silently revert the owner's edits.
5. **No-op repair turns reach the gate** — the `state_key` dedup early-return let a repair turn that changed nothing keep `state=repairing` forever with no residual ticket while the seam settled. Unchanged tree + last state `repairing` now falls through: budget spent honestly, ticket filed. Covers the batch `_run_ledger_repairs` loop too.
6. **`write_selfcheck` atomic** — SELFCHECK.json is the sole carrier of the converge signal, written exactly at the SIGTERM/OOM-prone turn-end seam.
7. **`KBC_L1_REPAIR_ROUNDS` default 1 → 2** — under the gateway's per-chunk SSE decoding bug, a long zh repair turn's own output can pick up fresh U+FFFD; one round was a coin-flip on large zh KBs. Drop back once the gateway does cross-chunk decoding.
8. **New `page_too_large` lint** — a page crossing the sync cap (1MB) is silently absent/stale in the store and the published version while every local check stays green; now a model-fixable violation (split the page).

## Batch B — box pod lifecycle GC (`d405dba6`)

- **Completed runs stop their box**: the relay's `finally` owns the stop for both close shapes (clean `end` and crash) — a normally-completed run used to leave its pod Running + cert Secret forever.
- **Periodic orphan sweep** (K8s spawner, duck-typed; wired with a run-liveness oracle in server): terminal-phase capability pods, Running pods whose run is no longer live, and cert Secrets whose pod is gone. Chat agent boxes are never touched.
- **Burstable resources**: `memoryRequest`/`cpuRequest` split request from limit; kb-compile requests 1Gi instead of the 256Mi default.
- **Verify-task teardown**: the run wrapper cancels detached media/PK tasks at run end.

## Batch C — prompt/schema/code contract alignment (`a0bc0a92`)

EXCLUSIONS segment-glob semantics documented + noop-exclusion signal reaches the repair prompt; system-filed residual tickets join the model's contract; CHANGESET biconditional wording matches the code + full-compile kickoff clears stale changesets; batch-final ledger pass requires the index; per-source diffs cap at 64KB; `run_layer1` carries `converge_phase`; zh gains the en language rule.

## Review hardening round 2 (`42714a4f` + `db994176`)

- Sweep oracle store-backed on memory miss and fail-safe (unknown ⇒ live); cert Secrets carry `boxType` + 10-min age guard for the Secret-before-pod spawn TOCTOU; sweep timers unref'd.
- Media hand-back mirrors the full seam (next chunk → PK → settle); `repairing` report stamps `converge_phase=revising`; `injected_repair` set only after a successful inject; materialize-race fallback clears stale CHANGESET.

## Review hardening round 3 (`13c5245c` + `7b1b841c`) — this review's findings

Box:
- **Settle exactly once** (med-high): a findings-path inject throwing after the primary settle used to re-settle the chunk with `result=None` — double-bumping failed pages and pushing just-verified pages toward a spurious `exhausted`.
- **Failed chunks no longer skip the rest of the drain** (med): chunk-1 failure now defers its pages (session-scoped) and chains on to the unattempted chunks; a later turn's fresh trigger retries the deferred pages.
- No-op gate falls through on a missing/corrupt SELFCHECK.json (heals the file); `_run_ledger_repairs` grants a fresh repair budget per episode; `page_too_large` measures raw on-disk bytes (CRLF divergence closed).

Runtime:
- **Setup-failure stop** (med): a pod spawned whose session setup then failed leaked until the sweep — the setup block now stops the box and rethrows.
- **Sweep oracle keyed by the RAW run id** (med): from the pod's `agent` label (stamped at spawn) instead of prefix-strip inversion, which was exact only for minted lowercase-UUID ids.
- Orphan-sweep timer cleared in `cleanup()` (was misplaced in `setSpawnEnvResolver`); requests clamp to limits (API-rejected-pod footgun); Secret age guard treats a missing `creationTimestamp` as young, not ancient.

## Tests

All seven box suites green; runtime `tsc` clean; full TS suite 213 files / 4334 tests green. New tests per round are listed in the respective commits.


LikiosSedo:
Batch B (box pod lifecycle GC) added in d405dba, per the same audit:

- **Completed runs now stop their box**: the relay's `finally` owns the stop for both close shapes (clean `end` and crash) — a normally-completed run used to leave its pod Running + cert Secret forever (the watchdog only reaps live-tracked runs; #390 only covered the crash path).
- **Periodic orphan sweep** (K8s spawner, duck-typed; wired with a run-liveness oracle in server): terminal-phase capability pods, Running pods whose run is no longer live, and cert Secrets whose pod is gone. Chat agent boxes are never touched.
- **Burstable resources**: new `memoryRequest`/`cpuRequest` split request from limit; kb-compile requests 1Gi instead of the 256Mi default (10 hot compiles bin-packed onto one node then bursting to 4Gi each = OOM territory).
- **Verify-task teardown**: the run wrapper cancels detached media/PK tasks — they kept burning model calls after run end.

Tests: sweep semantics (terminal + run-dead removed / live + chat kept / ghost Secret removed), request-limit split via the real kb-compile profile, wrapper task cancellation. 96/96 spawner, 225/225 agentbox+capability, seven box suites green.

LikiosSedo:
Batch C (contract alignment) added in a0bc0a9 — the audit's remaining drift findings. Highlights: EXCLUSIONS segment-glob semantics documented + the noop-exclusion signal now reaches the repair prompt (the model could never converge an exclusion it wrote with the wrong glob shape); system-filed residual tickets join the model's contract (no wholesale CONTRADICTIONS.json rewrites; rulings on them are actions, not values); the CHANGESET biconditional wording now matches the code and a full-compile kickoff clears stale changesets; the batch-final ledger pass requires the index (mid-Execute exemption scoped correctly); per-source diffs cap at 64KB (degrade to re-read) so CHANGESET.json can't outgrow the sync cap; run_layer1 carries converge_phase; zh gains the en language rule. The derived/media finding verified as already enforced via the body_source_uncited lint (noted in the commit). Four new tests; all seven box suites green.

With this, all three audit batches (A: convergence, B: pod lifecycle GC, C: contracts) are on this PR.

LikiosSedo:
Pre-merge adversarial review round (two independent reviewers over the full PR diff — box python and runtime TS — plus a self-review pass). Ten findings, all real ones fixed in 42714a4 + db99417:

**Runtime (42714a4):**
- The sweep's liveness oracle was memory-only: a helm upgrade restarting runtime+consumer together (boot recover() fails soft) would have let the first sweep delete every live idle box. Now store-backed on memory miss and fail-safe (unknown = live).
- Cert Secrets now carry boxType and the Secret pass is scoped to capability types (Secrets had no boxType — the pass could delete a chat box's cert), with a 10-min age guard closing the Secret-before-pod spawn TOCTOU.
- Sweep timers unref'd + cleared in cleanup().

**Box (db99417):**
- The media hand-back had traded the visible verifying-wedge for a silent skip: settling after chunk 1 skipped chunks 2..N and PK for image sets above the per-pass cap. The finally now mirrors the full seam (next chunk → PK → settle), releasing the single-flight ref first, and chains only after a CLEAN pass — failed pages keep their later-trigger retry cadence instead of hot-looping the attempt budget (caught by the existing retry-exhaust test).
- A repairing report stamps converge_phase=revising itself (no momentary repairing+settled sync window).
- injected_repair set only after a successful inject (a failed injection no longer wedges converge at revising).
- The materialize-race fallback clears stale CHANGESET too.

Accepted without change: dispatch_nonce as a required tool param (prompt orders it, handler defaults it, matches the four pre-existing required params).

Verified-clean lists from both reviewers cover: never-stuck boundedness of the no-op fallthrough, SELFCHECK RMW atomicity, seam ordering/PK gating, test-relay isolation (test sessions cannot stop the authoring box), tombstone propagation, teardown ordering. All seven box suites + 176 agentbox TS tests green.

kbc/platform/pod/compile_box.py
```diff
@@ -1093,6 +1105,7 @@ async def _run_media_verify_flow(run, chunk: dict[str, list[str]]) -> None:
                 f"自检(图像):{result['images']} 张图已核,{len(findings)} 条断言与图不符/超源,注入回修{tail}")})
             await _set_converge_phase(run, "revising")
             await run.inject_user_message(mediaverify.build_repair_prompt(findings, locale=getattr(run, "locale", None)))
+            injected_repair = True  # only after a SUCCESSFUL inject — a failed one must fall through to the chain below (review)
```
jacoblee-io, line 1121: **[media chain, med-high] If the findings-path `inject_user_message` throws, the run double-settles the media chunk — re-counting just-verified pages as failed and dropping the findings.**

`_settle_media_outcome(run, chunk, result)` already ran a few lines above with the REAL result (completed pages marked verified, failed pages bumped). Then on the findings path `run.inject_user_message(...)` runs; `injected_repair = True` is only set *after* it returns. If the inject throws, control jumps to the `except`, which calls `_settle_media_outcome(run, chunk, None)` a SECOND time — and `None` means "every page in the chunk is a failed attempt", so pages just marked verified get a spurious failed-attempt bump (can push them to `exhausted`), while the real image-contradiction findings are silently lost (the repair was never delivered and isn't retried). **Fix:** don't re-settle in the `except` when the primary settle already ran (guard on a "settled" flag), or wrap only the inject so its failure falls through to the seam chain without a second settle.

kbc/platform/pod/compile_box.py
```diff
@@ -1107,6 +1120,28 @@ async def _run_media_verify_flow(run, chunk: dict[str, list[str]]) -> None:
...
+        # chunk — settling after chunk 1 silently skipped the rest AND PK,
+        # review finding), then PK, else settle. Failed passes settle too
+        # (never-stuck — a later turn's seam retries while attempts remain).
+        if not injected_repair:
+            try:
+                # Chain the next chunk only after a CLEAN pass: failed pages
+                # retry on later triggers (the designed cadence) — chaining them
+                # here would hot-loop a transient outage straight through the
+                # attempt budget and ship an instant exhausted-pass.
+                if not ((chunk_clean and _maybe_start_media_verify(run)) or _maybe_start_pk(run)):
```
jacoblee-io: **[media chain, med] A transient failure on the FIRST chunk skips every downstream chunk and PK, then settles green.**

The self-drain chains the next chunk only on `chunk_clean`, and `_maybe_start_pk` is blocked while any media page is still pending. So a transient outage on chunk 1 (not clean) neither chains chunks 2..N nor starts PK — it falls straight to `_set_converge_phase("settled")`. The comment frames this as the "designed cadence" (failed pages retry on a later trigger), but the scope is broader than intended: it's not just chunk 1's failed *pages* that wait, it's chunks 2..N that were never attempted at all — and in a single session with no follow-up owner turn they're never verified while the phase reads settled. **Fix:** chain the next chunk (or re-enqueue the unattempted chunks) even on a failed pass, distinct from the per-page attempt-budget retry, so a chunk-1 blip doesn't silently drop downstream verification.

src/gateway/server.ts
```diff
@@ -457,19 +457,23 @@ export async function startRuntime(opts: StartRuntimeOptions): Promise<RuntimeSe
...
           .finally(() => {
             capabilitySessions.delete(runId);
+            // The relay ending — cleanly (`end`: the box's session coroutine
+            // exited and can never take another turn) or by crash (the catch
+            // above) — means this one-run pod is unreachable garbage either
+            // way. Stop it here, or every NORMALLY-completed run leaks a
+            // running pod + cert Secret forever (audit finding; the crash
+            // path was covered piecemeal before, this owns both). stop() is
+            // 404-tolerant, so the idle-reap double-stop stays quiet.
+            void agentBoxManager.stop(runId).catch((stopErr) =>
```
jacoblee-io, line 486: **[lifecycle GC, med] The box-stop here is unreachable if session setup fails after the pod is spawned — that path leaks a pod until the orphan sweep.**

The pod is created by `capabilityBoxClient(runId, ...)` at the top of the IIFE. This stop lives only in `driveCapabilitySession(...).finally(...)`, which is never reached if `materializeCapabilityInputs(...)` or `postJson('/session/...')` throws first — the IIFE rejects, `pending.catch` just deletes the session-map entry, and `capability.start`'s catch does `endRun('failed')` but never stops the box. So a spawn-succeeded-but-setup-failed run leaks its pod + cert Secret until the ~10-min orphan sweep. Ironic for the PR that adds "completed runs stop their box" — the setup-failure terminal path is the one it misses, and it makes the sweep load-bearing for it. **Fix:** stop the box in the setup-failure catch (or move the pod spawn+stop under one try/finally that covers materialize + `/session`).

src/gateway/server.ts
```diff
@@ -482,6 +486,33 @@ export async function startRuntime(opts: StartRuntimeOptions): Promise<RuntimeSe
...
   void capabilityRunManager.recover();
   capabilityRunManager.startWatchdog();
+  // Capability-box orphan GC: a box is live iff its run is tracked and
+  // non-terminal. Pod names are `agentbox-<runId>` (podName sanitizes the id;
+  // capability run ids are lowercase uuids, so the strip is exact).
+  // Optional-call: startRuntime tests inject minimal manager fakes that predate
+  // this method — the sweep is an ops concern, never a boot requirement.
+  agentBoxManager.startOrphanSweep?.(async (boxId) => {
+    // Prefix-strip inverts podName ONLY because capability run ids are minted
+    // 36-char lowercase UUIDs (sanitize + slice are no-ops). If run ids ever
```
jacoblee-io: **[orphan sweep, med] The sweep can kill a LIVE box if a runId isn't a lowercase UUID.**

The oracle inverts `podName` by prefix-strip (`boxId.slice("agentbox-".length)`), which is exact ONLY because minted capability run ids are 36-char lowercase UUIDs (the comment concedes this is an undefended invariant). But `podName` sanitizes/lowercases/truncates its input — an adopted or consumer-recovered `runId` that isn't already a clean UUID won't round-trip, so both the memory lookup (`capabilityRunManager.get(runId)`) and the store fallback key on the wrong id → miss → the box is classified orphan → a live compile is reaped. **Fix:** stamp the raw runId as a pod annotation/label and read it back for the oracle, rather than reconstructing it from the sanitized pod name.

src/gateway/agentbox/manager.ts
```diff
@@ -65,6 +87,7 @@ export class AgentBoxManager {
    * pays nothing. Currently supplies SICLAW_AGENTBOX_IDLE_TIMEOUT.
    */
   setSpawnEnvResolver(fn: (agentId: string) => Promise<Record<string, string> | undefined>): void {
+    if (this.orphanSweepTimer) { clearInterval(this.orphanSweepTimer); this.orphanSweepTimer = undefined; }
```
jacoblee-io: **[GC lifecycle, med-low] The orphan-sweep timer is cleared in `setSpawnEnvResolver`, not in `cleanup()` — contradicting the code's own comment and silently disabling GC if the resolver is re-set.**

`startOrphanSweep` sets `orphanSweepTimer` with a comment saying it "must be cleared in `cleanup()` — same discipline as the run watchdog," but the `clearInterval(this.orphanSweepTimer)` actually lives here in `setSpawnEnvResolver`. Two consequences: (1) `cleanup()` never clears it (a soft leak — it's `unref`'d so it won't block exit, but the interval keeps firing post-cleanup); (2) if `setSpawnEnvResolver` is ever called *after* `startOrphanSweep`, it silently cancels the orphan GC. **Fix:** move the clear into `cleanup()` (alongside the health-check timer) and drop it from `setSpawnEnvResolver`.

jacoblee-io (commented):
## Review: convergence hardening round 2

**Scope note (please fix the description).** The body says "This PR fixes the eight box-side findings; runtime lifecycle GC and remaining prompt-contract drifts land separately." But the branch actually contains, on top of the 8-findings commit (`cdb6c2b4`): **`d405dba6` box-pod lifecycle GC** (orphan sweep, clean-completion stop, burstable requests) and **`a0bc0a92` prompt-contract alignment** — i.e. exactly the two things the body says land separately, plus `42714a4f` (sweep hardening) and `db994176` (media chain). So this PR is materially larger than described (TS-gateway GC + capability sweep + mTLS-Secret scoping are all in the diff). Reviewers/mergers should know that.

**Security (positive, verified):** the capability-scoped Secret is the per-box **mTLS cert** Secret (not an LLM credential), scoped by `boxType` so the sweep can't delete a chat box's cert; the store-backed sweep oracle correctly **fails safe** (store error ⇒ treat as live ⇒ don't reap). No new exfil surface.

**8 box findings:** #2 (ADDED_TARGETS per-round reset), #3 (dispatch_nonce schema + zh/en), #4 (arm cleared on dispatch failure, both paths), #7 (rounds 1→2, no off-by-one) all verified correct. #1/#5 are right in direction but the media chain #1 touched introduced new issues (inline). 

**Lower-severity (not inline):**
- **no-op gate write-race (compile_box.py, ~`_post_turn_selfcheck`):** the fall-through reads `last_state` from a *second* `read_selfcheck`, not the report it's recomputing — a racing background write returning None can drop the gate → `repairing` forever with no ticket. Plausible.
- **`_run_ledger_repairs` doesn't reset `_l1_repairs_used`:** a batch ledger phase entered after the persistent budget was spent files a residual ticket with zero repair attempts.
- **`page_too_large` byte method (selfcheck.py):** lint measures `len(text.encode())` (from `read_text`, newline-translated) while the actual sync gate compares `f.stat().st_size` (raw); a CRLF page just over the cap passes lint but sync skips it (false green). One-directional, low (model pages are LF).
- **burstable requests have no request≤limit guard** (k8s-spawner/box-profile): a request above the default limit → unschedulable pod. Latent footgun on the new surface.
- **age guard fails open on a missing `creationTimestamp`** (createdMs=0 → looks ancient → TOCTOU bypass). Low.

kbc/platform/pod/compile_box.py
LikiosSedo: Fixed in 13c5245c. A `settled` flag is set right after the primary settle; the except settles only when it never ran, so a findings-path inject failure (or any throw after the settle) can no longer double-charge failed pages or push just-verified pages toward a spurious `exhausted`. New test: `test_media_inject_failure_does_not_double_settle` (inject raises with a 1-attempt budget — verified mark stands, no phantom attempts, drain still hands back to the seam).

kbc/platform/pod/compile_box.py
LikiosSedo: Fixed in 13c5245c. The drain now chains past a failed chunk: pages that failed THIS drain go into a session-scoped deferred set excluded from the due-check, so the chain moves on to the not-yet-attempted chunks instead of hot-looping the failed pages through the attempt budget. A fresh (seam) trigger clears the deferral — the per-page later-trigger retry cadence is unchanged and the existing retry-exhaust test still passes as-is. New test: `test_media_failed_chunk_still_drains_later_chunks` (chunk 1 fails → chunk 2 verifies in the same drain, failed page charged exactly one attempt, fresh trigger retries it).

src/gateway/server.ts
LikiosSedo: Fixed in 7b1b841c. The setup phase (materialize + POST /session) is wrapped in its own try/catch that stops the box and rethrows — once `driveCapabilitySession` starts, its `finally` takes over stop ownership as before, so there is exactly one owner per path and no double-stop (stop is 404-tolerant anyway). New test boots the real `startRuntime` with a postJson-throwing box client and asserts the spawned pod is stopped under the same runId it was spawned with (`server-capability-setup-failure.test.ts`).

src/gateway/server.ts
LikiosSedo: Fixed in 7b1b841c, via a channel that already existed: spawn stamps the raw agentId (= the runId for capability boxes) as the pod's `{labelPrefix}/agent` label, so the sweep now reads that label and hands the oracle the RAW id — no reconstruction from the sanitized pod name. The prefix-strip survives only as a fallback for label-less debris, where it is best-effort by definition. (An annotation would have worked too, but the label is already on every pod ever spawned, so legacy pods are covered without a migration.) New test: a pod whose label id (`Adopted_Run.7`) cannot round-trip through podName() is kept when the oracle knows it live.

src/gateway/agentbox/manager.ts
LikiosSedo: Fixed in 7b1b841c — the clear was simply misplaced. It now lives in `cleanup()` (alongside the health-check stop) and `setSpawnEnvResolver` no longer touches the timer. New fake-timer test: the sweep keeps ticking across a resolver re-set and stops after `cleanup()`.

LikiosSedo:
Review round 3 addressed in 13c5245c (box) + 7b1b841c (runtime). All five inline findings fixed (replied in-thread), plus the review body's lower-severity list — none of them survived verification as false positives:

- **No-op gate write-race**: the exact torn-read mechanism can't occur (`write_selfcheck` is temp+`os.replace` atomic), but the robustness gap is real — a missing/corrupt SELFCHECK.json on an unchanged tree (e.g. the model's own Bash damaging it) silently dropped the gate. `None` now falls through: the recompute heals the file and spends the budget honestly.
- **`_run_ledger_repairs` budget**: now resets `_l1_repairs_used` on entry — each ledger episode is its own bounded budget instead of inheriting a spent session counter and filing zero-attempt residual tickets.
- **`page_too_large` byte method**: `candidate_pages` stamps `stat().st_size` and the lint uses it — same byte method as the sync gate, CRLF false-green closed (test writes a CRLF page in the raw>cap>decoded window).
- **request≤limit guard**: requests clamp down to the limit with a warning (small K8s-quantity parser; unparseable values pass through — the API server stays the authority).
- **Secret age guard**: missing `creationTimestamp` now reads as YOUNG (skip this round), not infinitely old.

**Scope note**: the PR description has been rewritten to describe what the branch actually contains — all three audit batches (A convergence / B pod-lifecycle GC / C contracts) plus the two review-hardening rounds, spanning box Python, the TS capability gateway, the K8s spawner sweep, and the cert-Secret lifecycle.

Verification: all seven box suites green; `tsc` clean; full TS suite 213 files / 4334 tests green. Ten new tests across the two commits.

jacoblee-io (approved):
Round 3 verified — all ten findings from the previous review are correctly addressed and the fix code introduces no regressions.

KBC (13c5245c):
- #1 media double-settle: `settled` flag gates the except; a post-settle findings-inject throw no longer re-settles with result=None.
- #2 failed-chunk downstream skip: failed pages defer into a session-scoped `_media_deferred` (excluded from the due-check) and the drain chains on to the not-yet-attempted chunks; a fresh (drain=False) trigger clears the deferral. Verified loop-free: `_media_task` is released before chaining, and the candidate set strictly shrinks each iteration.
- no-op gate: None now falls through instead of dropping the gate.
- `_run_ledger_repairs`: fresh per-episode budget (`_l1_repairs_used = 0`).
- page_too_large: measures on-disk `stat().st_size`, matching the sync gate.

Capability GC (7b1b841c):
- #6 setup-failure stop: materialize + POST /session wrapped; box is stopped and the error rethrown.
- #7 sweep oracle: keyed off the raw run id from the pod `agent` label, pod-name strip is fallback only — no longer reaps live boxes with non-UUID ids.
- #8 sweep timer: cleared in cleanup(), removed from setSpawnEnvResolver.
- request>limit clamp and Secret age-guard fail-closed both landed.

CI green (Type Check / KBC Box / Test / Portal Web). Non-blocking: the PR description predates the GC + contract-alignment work now in the branch — please update it to match the actual scope before/after merge.

PR #394, opened 2026-07-09
refactor(tools): fold cluster connectivity probe into cluster_list as opt-in flag

jacoblee-io:
## What

Replaces the standalone `cluster_probe` tool with an opt-in `probe` flag on `cluster_list`.

- **Default (`probe` omitted):** unchanged — a cheap cached metadata read that touches no cluster.
- **`probe:true`:** runs `kubectl version` against each *returned* cluster in parallel (bounded, concurrency 8) and folds `reachable` / `server_version` / `probe_error` into every entry.

## Why

Checking reachability across N clusters previously meant `cluster_list` + N serial `cluster_probe` calls — each a separate agent turn, each a forced cache-bypass credential acquire. Folding it into one flag cuts that to a single call, and keeping it opt-in means the common "just list the clusters" path never probes clusters the agent won't use (the first real `kubectl`/script command already reveals an unreachable cluster).

## Changes

- **broker:** add `probeClusters(names, {concurrency:8})` batch (bounded parallel, per-cluster failures fold into unreachable results, never a rejected batch, order preserved). `probeCluster` now reuses the cached kubeconfig (`ensureCluster`) instead of a forced cache-bypass acquire — a reachability check needs no fresh credentials, so this saves a `credential.get` per probe and warms the cache for the `kubectl`/script calls that follow.
- **cluster_list:** add `probe` param + reachability merge; update tool description.
- **delete** `cluster-probe.ts` + its test; drop it from `all-entries.ts` and from both the backend and frontend `inspect_infra` capability groups.
- **prompt.ts:** fold reachability into the `cluster_list` guidance (drop the `cluster_probe` clause).

`AuditTable`'s historical tool filter keeps `cluster_probe` so past audit rows stay filterable.

## Test

- `cluster-list.test.ts`: 4 new cases (no-probe → no cluster contact / no reachability; `probe:true` → reachable+server_version folded in; name-filtered subset only; unreachable → `reachable:false`+`probe_error`, no `server_version`).
- Backend cluster/broker/capability/agent-api suites green (112 passed); `tsc --noEmit` clean for changed files. (The unrelated `@arizeai` tracing-dep failures are a local node_modules install gap on `main`, not from this change.)


LikiosSedo (commented):
## Review — LGTM, ready to merge

Reviewed the full diff line by line, ran the affected suites locally (cluster-list 11 + credential-broker 27, all green), `tsc --noEmit` clean on a fresh install (the `@arizeai` gaps the PR notes didn't even reproduce after `npm install`). CI is green.

### Verified

- **Deletion is migration-safe.** Grepped the whole tree for `cluster_probe`/`clusterProbe`/`cluster-probe`: the only survivors are intentional — the `cluster-list.ts` doc comment, the broker's `"cluster_probe"` purpose string, and `AuditTable`'s historical filter (kept by design). Crucially, agent tool grants store **capability-group keys** (`inspect_infra`), not tool names — `resolveCapabilities` derives names from the current `CAPABILITY_GROUPS`, and the group no longer lists `cluster_probe`, so no stored agent can reference the removed tool. No DB migration, no "unknown tool" path. Both backend and frontend capability groups were updated in lockstep.
- **Batch semantics are correct.** `probeClusters` bounds fan-out with `ConcurrencyLimiter(8)`; `probeCluster` folds every per-cluster failure into an unreachable `ProbeResult` and never throws, so `Promise.all` can't reject; order is preserved and `cluster_list` additionally keys results by name via a Map, so it's robust even if order weren't. The limiter releases its slot in `finally` (throw-safe), FIFO. This directly addresses the "many-bound-cluster agent shouldn't spawn an unbounded kubectl fan-out" concern.
- **Probe targets only the returned subset** (name filter applied before probing) — covered by the `name`-filtered test.
- **Field precedence:** probe fields are spread after `flattenClusterMeta`, so a probe result wins over any same-named admin meta key — correct ordering.

### Non-blocking notes

1. **(low · test-coverage)** The broker-level changes carry no direct broker test: `probeClusters`' batch contract (order preservation, per-cluster failure isolation, concurrency bound) and the `probeCluster` switch from `acquireCluster({bypassCache:true})` to `ensureCluster` (cache reuse) are only exercised through `cluster-list.test.ts`, which mocks `probeClusters` wholesale. The deleted `cluster-probe.test.ts` only tested the tool wrapper too, so this isn't a coverage regression — but since the batch + cache-reuse is the load-bearing new behavior, a small `credential-broker.test.ts` case (e.g. one reachable + one throwing cluster → both fold, order kept, ≤concurrency in flight) would lock the contract.

2. **(informational)** The cache-reuse switch is a deliberate, well-documented trade-off — it saves a `credential.get` per probe and warms the cache for the following `kubectl`/script calls, and arguably tests reachability of *exactly the credential subsequent calls will use*. One edge worth noting: a locally-cached-but-unexpired credential that was revoked server-side would report `reachable:true`, where the old forced acquire would have surfaced the failure. Given the probe is advisory (the first real kubectl reveals an unreachable cluster anyway), this is fine.

No correctness bugs and no security concerns. The opt-in fold is the right call: the common "just list the clusters" path stays a cheap cached read with zero cluster contact, and the reachability case collapses from `cluster_list` + N serial `cluster_probe` turns into one bounded-parallel call.

jacoblee-io:
Thanks for the thorough review.

- **Note 1 (test coverage):** addressed in `9ea1f6ef` — added broker-level `probeClusters` tests (stubbing `probeCluster` to avoid real kubectl) that lock the batch contract: input order preserved, each per-cluster result folded through so the batch never rejects, and in-flight probes bounded by the concurrency cap. `credential-broker.test.ts` now 29 green.
- **Note 2 (revoked-but-cached credential reads reachable):** acknowledged, keeping as-is by design — the probe is advisory (the first real kubectl surfaces an unreachable/denied cluster), and the reuse deliberately tests the exact credential subsequent calls will use while saving a `credential.get`.

LikiosSedo (approved):

PR #395, opened 2026-07-09
feat(kb): consumer meta — blind-generated KB summary + catalog progressive disclosure

LikiosSedo:
Knowledge bases get skill-style progressive disclosure: a consumer-facing summary generated at compile settle, delivered to consuming agents via the knowledge catalog.

## What
- **Box (kbc)**: at settle, a blind read-only engine pass — allowed roots contain only `candidate/`, so raw isolation is structural — generates `authoring/CONSUMER_META.json` (`summary`/`when_to_use`/`not_for`/`topics`/`entry_pages`, locale follows the compile locale). selfcheck validates it as a non-blocking lint fact; generation is fail-open and never blocks settle. Kill switch: `KBC_CONSUMER_META_MODE=off`.
- **Runtime catalog**: `buildKnowledgeWikiCatalog` reads `_consumer_meta.json` from each bound bundle root (injected by the consumer platform at publish) and renders a "Knowledge Bases" routing section (name + published version + summary + use-when/not-for) within the 4000-char budget; degradation drops `not_for`, then `when_to_use`, truncating the summary last. Missing/invalid meta → byte-identical pre-meta output.
- `_consumer_meta.json` is never installed or served as a wiki page; published test snapshots stay hash-identical to their drafts.

## Tests
- kbc: test_selfcheck 41 checks OK (5 new), test_compile_box 44 OK, incremental/redblue/batching/mediaverify/office_ingest all OK.
- TS: vitest 213 files / 4341 passed; overview-generator 30/30 (8 new incl. the consumer-platform wire format `published_version: "v1"`); tsc clean.

## Notes
- The consumer platform pins the (owner-editable) meta into the published bundle at publish; old images/bundles without meta keep exactly the current behavior.
- Live effect requires rebuilding the compile-box image and the runtime image.


LikiosSedo:
Lifecycle follow-up pushed in `20e04c49`:

- consumer meta now finishes before the final `settled` workspace sync, so it belongs to the same open staging mutation
- a `verifying` sync is emitted before bounded generation to keep the authoring read model busy during finalization
- the final `settled` edge is emitted last; no background writer mutates an already committed draft
- generation remains fail-open, bounded, tree-hash idempotent, and self-healing when the generated file is deleted

Validation: the full `kbc/platform/pod/test_selfcheck.py` suite passes, including consumer-meta generation, ordering, regeneration, self-heal, and failure paths.

PR #396, opened 2026-07-10
Harden knowledge compiler runtime sync

LikiosSedo:
## Summary

- checkpoint and restore the exact frozen consumer input revision across dead-box recovery
- persist workspace changes as retryable atomic batches and replay content/tombstones after relay recovery
- advance compile provenance only through an explicit, idempotent `commit_input` signal; replayed or rehydrated files never imply a completed compile
- downgrade a legacy/missing-revision commit to content-only persistence instead of retrying a deterministic rejection
- remove Bash from the KB compiler and confine compiler, batch-planner, and test-session file tools to their workspaces with `PreToolUse` guards

## Why

The previous relay could permanently drop artifacts, a recovered box could silently switch to current raw instead of its pinned input, and replaying an unchanged `candidate/index.md` could be mistaken for a completed compile. The compiler also exposed shell and unrestricted file paths while its model credential lived in the process environment.

## Rollout gate

This is a **hard runtime/consumer cutover**, not a backward-compatible runtime fallback. The runtime starts calling `capability.persistArtifacts`; an older Sicore consumer does not implement that method.

1. Merge and deploy paired Sicore MR !618 first.
2. Confirm the new consumer handles `capability.persistArtifacts`, `input_revision`, and `commit_input`.
3. Only then deploy this Siclaw PR.

Roll back in reverse order. Sicore !618 retains the singular artifact handler for older runtimes during the rolling window, but this runtime intentionally does not fall back to singular writes because that would lose atomic batch, tombstone, and provenance semantics.

## Validation

- `npm test -- --run` — 4,344 passed, 2 skipped
- `npx tsc --noEmit`
- `npm run build`
- focused capability tests — 56 passed
- `python test_compile_box.py` protocol smoke with fake SDK session
- `python test_batching.py`
- `python test_incremental.py`
- `python -m py_compile ...`
- `git diff --check`

### DIFF
--- a/kbc/platform/pod/compile_box.py
+++ b/kbc/platform/pod/compile_box.py
@@ -264,6 +264,12 @@ class CompileRun:
         self._stall_interrupted_at = 0.0
         self._batch_active = False
         self._batch_notes: list[str] = []
+        # Full compile provenance is an explicit commit, never inferred from a
+        # replayed or rehydrated candidate/index.md.
+        self._full_compile_pending = False
+        # Keep the last full-compile commit replayable across relay restarts.
+        # The consumer dedupes it by immutable input revision.
+        self._commit_input_replay = False
         # Scoped incremental (真增量): armed at kickoff with {before: page_hashes,
         # changeset}; the post-turn seam runs the byte-integrity guard against it,
         # then clears it. None = this turn is not a scoped incremental.
@@ -357,6 +363,7 @@ MAX_SYNC_FILE_BYTES = int(os.environ.get("KBC_MAX_SYNC_FILE_BYTES", str(1024 * 1
 # result (a Read of a big source file) kills the whole session with a fatal
 # "exceeded maximum buffer size" — seen live on a 139-source compile (2026-07-06).
 SDK_MAX_BUFFER_BYTES = int(os.environ.get("KBC_SDK_MAX_BUFFER_BYTES", str(16 * 1024 * 1024)))
+_SYNC_TOMBSTONE = "__deleted__"
 
 
 def _collect_workspace_artifacts(workdir: str) -> list[dict]:
@@ -382,19 +389,31 @@ def _collect_workspace_artifacts(workdir: str) -> list[dict]:
     return out
 
 
-async def _sync_workspace(run: CompileRun, sent: dict) -> int:
+def _workspace_sync_cursor(workdir: str) -> dict[str, str]:
+    """Hash the workspace state already installed by the consumer."""
+    return {
+        art["path"]: hashlib.sha256(art["content"].encode("utf-8")).hexdigest()
+        for art in _collect_workspace_artifacts(workdir)
+    }
+
+
+async def _sync_workspace(run: CompileRun, sent: dict, *, commit_input: bool = False) -> int:
     """Emit a syncArtifacts event for workspace files changed since `sent`
     (path → content sha), plus TOMBSTONES ({path, deleted: true}) for
-    previously-synced files that no longer exist on disk. Updates `sent`;
+    previously-synced files that no longer exist on disk. Updates `sent` after
+    enqueue and retains tombstone markers for reconnect replay;
     returns the number of changed entries."""
+    if commit_input and not (Path(run.workdir) / "candidate" / "index.md").is_file():
+        raise FileNotFoundError("cannot commit compile input without candidate/index.md")
     changed = []
+    next_sent = dict(sent)
     collected = set()
     for art in _collect_workspace_artifacts(run.workdir):
         collected.add(art["path"])
         sha = hashlib.sha256(art["content"].encode("utf-8")).hexdigest()
         if sent.get(art["path"]) == sha:
             continue
-        sent[art["path"]] = sha
+        next_sent[art["path"]] = sha
         changed.append(art)
     # Tombstones: a previously-synced file the agent deleted (page merge, rename,
     # restructure) must be deleted from the consumer's store too — otherwise the
@@ -408,11 +427,37 @@ async def _sync_workspace(run: CompileRun, sent: dict) -> int:
     for path in [p for p in sent if p not in collected]:
         if (wd / path).is_file():
             continue  # still on disk, just not collectable — keep the row
-        del sent[path]
-        changed.append({"path": path, "deleted": True})
-    if changed:
-        await run.emit({"type": "syncArtifacts", "artifacts": changed})
+        if sent.get(path) != _SYNC_TOMBSTONE:
+            next_sent[path] = _SYNC_TOMBSTONE
+            changed.append({"path": path, "deleted": True})
+    if changed or commit_input:
+        event = {"type": "syncArtifacts", "artifacts": changed}
+        if commit_input:
+            event["commit_input"] = True
+        await run.emit(event)
+        # Advance the dedup cursor only after the event is queued. Tombstone
+        # markers are retained so an SSE re-attach can replay deletions too.
+        if changed:
+            sent.clear()
+            sent.update(next_sent)
+        if commit_input:
+            run._commit_input_replay = True
     return len(changed)
+
+
+def _workspace_replay_artifacts(run: CompileRun, sent: dict) -> list[dict]:
+    """Authoritative workspace replay for a newly attached SSE relay.
+
+    A runtime can crash after dequeuing syncArtifacts but before the consumer
+    acknowledges it. Re-send every current file and every remembered tombstone
+    on attach; consumer upserts/deletes are idempotent.
+    """
+    artifacts = _collect_workspace_artifacts(run.workdir)
+    current = {item["path"] for item in artifacts}
+    for path, value in sorted(sent.items()):
+        if value == _SYNC_TOMBSTONE and path not in current:
+            artifacts.append({"path": path, "deleted": True})
+    return artifacts
 
 
 async def _sync_loop(run: CompileRun, sent: dict):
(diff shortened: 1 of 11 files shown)


kbc/platform/pod/compile_box.py
```diff
@@ -2515,6 +2544,29 @@ async def guard(input_data, tool_use_id, context):
     return guard
 
 
+def _make_compile_path_guard(root: Path, locale: str | None = None):
```
jacoblee-io: **Confinement is wired per-session, and a sibling session with `Write` was missed.**

This PR adds the PreToolUse guard to `_compile_session_opts` (main compile) and `_make_test_path_guard` is on the test session — but the batch-planner session in `_plan_batches` (around line 1998) runs:

```python
allowed_tools=["Read", "Write", "Glob"],
permission_mode="bypassPermissions",
```

with **no `hooks=` entry at all**. So the exact escape class this PR closes on the compile session (an absolute / `../` path in Read/Write/Glob reaching outside `/work`) stays open on a session that can *write*. The pod is still the OS sandbox, but the whole point of this change is defense-in-depth inside the pod, and the planner is left out.

Root cause is that `_make_compile_path_guard` and `_make_test_path_guard` are copy-pasted (identical except the Bash short-circuit). Collapse them into one `_make_path_guard(root, locale=None, deny_bash=False)` and attach it to `_plan_batches` too — one mechanism keyed by profile means the next new session can't silently miss the guard.

src/gateway/capability/run-manager.ts
```diff
@@ -210,6 +213,15 @@ export class CapabilityRunManager {
     }
   }
 
+  /** Checkpoint the exact input before the box session is allowed to run. */
+  async setInputRevision(runId: string, inputRevision: string): Promise<void> {
```
jacoblee-io, line 219: **`failFast` checkpoint persist adds a new abort point on the healthy fresh-run start path.**

`setInputRevision` → `persist(rec, { failFast: true })` rethrows on a WS failure, and it is `await`ed in `server.ts:450` inside the box-setup `try` — *before* the box takes its first turn. A rethrow there stops the box and fails `capability.start` (endRun `failed`). Before this PR that path had no such persist, so a transient consumer WS blip could not abort a run that was otherwise fine to start.

Failing closed is defensible (recovery can't reconcile artifacts without knowing the installed revision), but a *transient* blip now costs the entire run start. Consider retrying the checkpoint persist a couple of times (the same rationale by which non-failFast transitions swallow blips) before giving up and failing closed.

src/gateway/capability/contract.ts
```diff
@@ -46,6 +46,7 @@ export const CAPABILITY_TEST_CLOSE = "capability.testClose" as const;
 /** siclaw → consumer: live stream + content sink + input fetch. */
 export const CAPABILITY_EVENT = "capability.event" as const;
 export const CAPABILITY_PERSIST_ARTIFACT = "capability.persistArtifact" as const;
+export const CAPABILITY_PERSIST_ARTIFACTS = "capability.persistArtifacts" as const;
```
jacoblee-io, line 48: Nit / cleanup: with the relay switched to the batch verb, the singular `CAPABILITY_PERSIST_ARTIFACT` (line above) now has no producer or consumer left in this repo — grep finds only its own declaration. It reads like a live wire method. Keep `CapabilityPersistArtifactRequest` (the `Omit<>` in the new batch type reuses it) but delete or `@deprecated`-annotate the dead string constant so a consumer-side maintainer can tell the singular is retired.

(Also: the `session-driver.ts` file header docstring still lists `capability.persistArtifact` — update it to the plural + note the all-or-nothing batch semantics.)

kbc/platform/pod/compile_box.py
```diff
@@ -408,13 +411,33 @@ async def _sync_workspace(run: CompileRun, sent: dict) -> int:
...
     if changed:
         await run.emit({"type": "syncArtifacts", "artifacts": changed})
+        # Advance the dedup cursor only after the event is queued. Tombstone
+        # markers are retained so an SSE re-attach can replay deletions too.
+        sent.clear()
+        sent.update(next_sent)
     return len(changed)
 
 
+def _workspace_replay_artifacts(run: CompileRun, sent: dict) -> list[dict]:
```
jacoblee-io, line 448: Low / efficiency: `_workspace_replay_artifacts` calls `_collect_workspace_artifacts`, which re-`rglob`s and re-reads+base64-encodes every workspace file from disk on *each* SSE re-attach with `replay=1`. On the 139-source compile cited elsewhere, a relay flap re-encodes the entire workspace into one frame before the queue is even drained — cost scales with total workspace size per reconnect, not per delta. Consumer upserts are idempotent so this is correct; but since `sent` already holds the authoritative path→sha map, replay could stream the last-known set (or gate the full re-read on a cheap mtime/size check) instead.

jacoblee-io (commented):
### Review — Harden knowledge compiler runtime sync

Solid PR; the four changes (tombstone retention + SSE replay, atomic-batch persist with retry, input-revision checkpoint through recovery, drop Bash + PreToolUse guard) are well-constructed and the core mechanics hold up under scrutiny. I verified by reading the branch code, not the comments:

- `sent` is the same object as `run._sync_sent`, so replay reads the live dedup map; the `sent.clear(); sent.update(next_sent)` cursor advance correctly defers until after `run.emit` succeeds (emit-failure rollback intact).
- The `for(;;)` retry is bounded: `touchHeartbeat` bumps only the heartbeat clock, so the watchdog's `dataStaleMs` data clock still reaps a run wedged against a dead consumer (and `onReap` stops its box) — no permanent pod leak. `endRun` is idempotent, so the throw-on-terminal path doesn't double-process.
- Recovery/reattach: on a 409 `setInputRevision` is skipped so the recovered checkpoint is preserved; `recover()`/`adopt()` repopulate `inputRevision` from `checkpoint`. Ordering (checkpoint before box runs) is correct.

**Findings** (see inline):

1. **[med] `_plan_batches` planner session has no path guard** — it runs Read/**Write**/Glob under `bypassPermissions` with no PreToolUse hook, while this PR guards the compile + test sessions. Same write-escape class the PR closes stays open here. Root cause: the two guard factories are copy-pasted rather than shared, so the planner was missed. I'd fix this before merge — it's this PR's own subject.
2. **[med-low] `setInputRevision` failFast** aborts a healthy fresh run on a transient consumer WS blip (new abort point on the start-critical path). Fail-closed is reasonable, but consider a couple of retries first.
3. **[low, description] "additive" overstates it.** On the runtime side the switch to `capability.persistArtifacts` is a *hard* cutover — the singular caller is removed with no fallback or version negotiation. If sicore isn't deployed first with the batch handler, every compile wedges then gets reaped `failed` (~10–60 min) with the box leaking until reap. The deploy order is stated, but please correct the "additive" wording and confirm sicore ships first.
4. **[low] guard duplication** (root cause of #1) and **[low] dead singular constant + stale docstring** and **[low] replay full re-read** — all inline.

Dropped two candidates as non-issues: the `{..,.}/etc` Glob brace traversal (Claude Code's Glob doesn't do shell brace expansion) and the `content: null` skip-vs-empty-file semantics change (unreachable — the box never emits null content for a non-deleted artifact).

No merge-blocking correctness bug, but I'd want #1 addressed (and ideally #2/#3 answered) before approving.

jacoblee-io (approved):
Approving. The force-push addressed every finding from my previous review, and I re-reviewed the new full-compile provenance mechanism that came with it — both check out.

Findings resolved:
- #1 / #4 (planner guard + duplication): collapsed into a single `_make_path_guard(root, locale, *, deny_bash)` factory, and `_plan_batches` now attaches `_make_compile_path_guard` — the unguarded Write session is closed and the confinement is now one mechanism keyed by profile.
- #2 (`setInputRevision` failFast): now retries 4× with exponential backoff before failing closed, and rolls back the in-memory `rec.inputRevision` on total failure so an unacked checkpoint is never retained. Better than what I suggested.
- #5 (dead constant + stale docstring): singular `CAPABILITY_PERSIST_ARTIFACT` removed (interface kept for the `Omit<>`), session-driver header updated to the batch verb.

New provenance-commit logic reviewed (sound):
- `_workspace_sync_cursor` primes the diff cursor from the rehydrated workspace at run start, so the first periodic tick can't echo an unchanged `candidate/index.md` and impersonate a completed compile.
- `commit_input` is an explicit full-compile signal (`_full_compile_pending`), set exactly on a full-corpus recompile (single-turn fallback, batch orchestrator, or the incremental materialize-race fallback) — never on a true scoped incremental. It fails closed (requires `candidate/index.md`, blocks turn_done until durable) and replays across relay restarts via `_commit_input_replay`.
- session-driver downgrades `commit_input` to content-only with a loud user-visible warning when the run has no pinned input revision; recovery pins the exact frozen source manifest via `fetchInput(input_revision)`.

One behavior note, non-blocking: the batch path's completion sync is now fail-closed (the old `try/except pass` is gone), so a completed compile that somehow produced no `candidate/index.md` now surfaces as a turn error instead of a silent "complete" — intended and stricter, just flagging it.

CI green (Type Check / KBC Box / Test / Portal Web). Deploy-order dependency on sicore (batch `capability.persistArtifacts` handler + `fetchInput` input_revision pin) remains the operational gate.

jacoblee-io (approved):
Approving. The force-push addressed every finding from my previous review, and I re-reviewed the new full-compile provenance mechanism that came with it — both check out.

Findings resolved:
- #1 / #4 (planner guard + duplication): collapsed into a single `_make_path_guard(root, locale, *, deny_bash)` factory, and `_plan_batches` now attaches `_make_compile_path_guard` — the unguarded Write session is closed and the confinement is now one mechanism keyed by profile.
- #2 (`setInputRevision` failFast): now retries 4× with exponential backoff before failing closed, and rolls back the in-memory `rec.inputRevision` on total failure so an unacked checkpoint is never retained. Better than what I suggested.
- #5 (dead constant + stale docstring): singular `CAPABILITY_PERSIST_ARTIFACT` removed (interface kept for the `Omit<>`), session-driver header updated to the batch verb.

New provenance-commit logic reviewed (sound):
- `_workspace_sync_cursor` primes the diff cursor from the rehydrated workspace at run start, so the first periodic tick can't echo an unchanged `candidate/index.md` and impersonate a completed compile.
- `commit_input` is an explicit full-compile signal (`_full_compile_pending`), set exactly on a full-corpus recompile (single-turn fallback, batch orchestrator, or the incremental materialize-race fallback) — never on a true scoped incremental. It fails closed (requires `candidate/index.md`, blocks turn_done until durable) and replays across relay restarts via `_commit_input_replay`.
- session-driver downgrades `commit_input` to content-only with a loud user-visible warning when the run has no pinned input revision; recovery pins the exact frozen source manifest via `fetchInput(input_revision)`.

One behavior note, non-blocking: the batch path's completion sync is now fail-closed (the old `try/except pass` is gone), so a completed compile that somehow produced no `candidate/index.md` now surfaces as a turn error instead of a silent "complete" — intended and stricter, just flagging it.

CI green (Type Check / KBC Box / Test / Portal Web). Deploy-order dependency on sicore (batch `capability.persistArtifacts` handler + `fetchInput` input_revision pin) remains the operational gate.

PR #397, opened 2026-07-10
Improve context preflight error messages

LikiosSedo:
## Summary

- distinguish missing or non-positive context windows from windows that are too small for the configured compaction reserve
- include the actual context-window and reserve-token values in the error
- provide an actionable recommendation to increase the window or lower `compaction.reserveTokens`
- add regression coverage for both invalid-window cases

## Root cause

The preflight check previously collapsed `contextWindow <= 0` and `contextWindow - reserveTokens <= 0` into the same generic `invalid context window` message. A model configured with a 10,000-token window therefore gave no indication that Siclaw's default 16,384-token compaction reserve was the conflicting value.

## User impact

Users now see the exact configured values and a concrete remediation instead of a generic model error.

## Validation

- `npm test` — 4,337 passed, 2 skipped
- `npx tsc --noEmit`
- `npm run build`
- `git diff --check`

### DIFF
--- a/src/core/brains/pi-agent-brain.ts
+++ b/src/core/brains/pi-agent-brain.ts
@@ -303,12 +303,20 @@ export class PiAgentBrain implements BrainSession {
       Math.max(0, Math.floor(contextWindow * 0.02)),
     );
     const budget = contextWindow - reserveTokens;
-    if (contextWindow <= 0 || budget <= 0) {
+    if (contextWindow <= 0) {
       return {
         ok: false,
         compacted: false,
         contextWindow,
-        errorMessage: `Context preflight failed: invalid context window for ${model.provider}/${model.id}`,
+        errorMessage: `Context preflight failed for ${model.provider}/${model.id}: context window must be greater than 0 tokens (received ${contextWindow}). Configure a positive context window for this model.`,
+      };
+    }
+    if (budget <= 0) {
+      return {
+        ok: false,
+        compacted: false,
+        contextWindow,
+        errorMessage: `Context preflight failed for ${model.provider}/${model.id}: context window ${contextWindow} tokens must exceed compaction reserve ${reserveTokens} tokens. Increase the model context window or lower compaction.reserveTokens.`,
       };
     }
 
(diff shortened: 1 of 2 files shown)


src/core/brains/pi-agent-brain.ts
```diff
@@ -303,12 +303,20 @@ export class PiAgentBrain implements BrainSession {
...
     );
     const budget = contextWindow - reserveTokens;
-    if (contextWindow <= 0 || budget <= 0) {
+    if (contextWindow <= 0) {
       return {
         ok: false,
         compacted: false,
         contextWindow,
-        errorMessage: `Context preflight failed: invalid context window for ${model.provider}/${model.id}`,
+        errorMessage: `Context preflight failed for ${model.provider}/${model.id}: context window must be greater than 0 tokens (received ${contextWindow}). Configure a positive context window for this model.`,
```
jacoblee-io, line 311: Nit (non-blocking): `received ${contextWindow}` prints the *normalized* value, not what the user configured — `contextWindow` is `Math.max(0, Math.floor(model.contextWindow || 0))`, so a negative or non-numeric `model.contextWindow` all render as `received 0`. Since the whole point of this PR is showing the user the actual configured value, consider interpolating the raw `model.contextWindow` here (the `<= 0` branch is only reached when the normalized value is 0 anyway).

jacoblee-io (approved):
Approving — clean, correct, and well-tested.

The split is behavior-equivalent to the old `contextWindow <= 0 || budget <= 0`: the `contextWindow <= 0` branch returns first, so the `budget <= 0` branch is only reached when `contextWindow > 0` (i.e. `reserveTokens >= contextWindow`), and the ordering correctly prefers the "window must be > 0" message over the reserve message when the window is 0. Both messages preserve the `Context preflight failed` prefix that other suites assert on (`model-routing.test.ts:1071`, `sse-consumer.test.ts:161` use `toContain`), and `model-routing.ts:1014`'s fallback phrasing already matches the new `for provider/id` shape. Two new tests cover both cases. CI green.

Two optional nits, neither blocking:
1. The `received` value is the normalized `0` rather than the raw config — inline.
2. Stale fixtures: `model-routing.test.ts:1049` and `sse-consumer.test.ts:154` still hard-code the old full string `"...: invalid context window..."`. They pass (the assertions only check the `Context preflight failed` prefix), but they now model a message shape production no longer emits — worth updating to the new wording next time either file is touched.

PR #398, opened 2026-07-10
fix(agentbox): key multi-repo knowledge directories by repo identity

LikiosSedo:
## Problem

Multi-repo materialization previously used only a sanitized display name for `repos/<dir>/`. Distinct names can collapse to the same ASCII slug, and every CJK-only name collapses to `repo`, allowing one bound KB to overlay another.

## Fix

- derive every multi-repo directory as `<readable-slug>--<sha256(repo_id)[:12]>`
- reject duplicate repo ids or the astronomically unlikely derived-name collision before extraction
- generate the top-level index from the exact derived directory names
- keep single-repo layout unchanged

Using identity on every path is deterministic and avoids a branchy first-repo/fallback-repo layout. Materialization already swaps a complete staging directory into place, so the naming migration leaves no stale directories.

## Verification

- CJK-only repo regression proves distinct directories, isolated content, and matching index links
- `npx vitest run src/agentbox/sync-handlers.test.ts src/shared/knowledge-package.test.ts` (57 passed)
- `npm run build`
- `git diff --check`

### DIFF
--- a/src/agentbox/sync-handlers.ts
+++ b/src/agentbox/sync-handlers.ts
@@ -14,8 +14,8 @@ import path from "node:path";
 import { loadConfig, reloadConfig, writeConfig } from "../core/config.js";
 import {
   extractKnowledgePackageToDir,
+  knowledgeRepoDirName,
   replaceDirectoryContentsFromStaging,
-  sanitizeKnowledgeRepoDir,
 } from "../shared/knowledge-package.js";
 import type {
   GatewaySyncType,
@@ -258,12 +258,6 @@ interface KnowledgeSyncStatus {
     id: string; name: string; version: number; sha256: string;
     expectedSha256?: string | null; fileCount?: number | null; sizeBytes: number;
   }>;
-  // Populated when two distinct repos sanitize to the same directory name and
-  // one was relocated to its repo-id directory to avoid a silent overwrite.
-  dirNameConflicts?: Array<{
-    repoId: string; repoName: string; sanitizedDir: string;
-    conflictsWith: string; fallbackDir: string;
-  }>;
 }
 
 let lastKnowledgeSyncStatus: KnowledgeSyncStatus | null = null;
@@ -322,7 +316,6 @@ export const knowledgeHandler: AgentBoxSyncHandler<KnowledgeBundlePayload> = {
     const stagingDir = path.join(knowledgeDir, `.sync-staging-${Date.now()}-${process.pid}`);
     fs.mkdirSync(stagingDir, { recursive: true });
     const syncedRepos: KnowledgeSyncStatus["repos"] = [];
-    const dirNameConflicts: NonNullable<KnowledgeSyncStatus["dirNameConflicts"]> = [];
 
     try {
       if (repos.length === 1) {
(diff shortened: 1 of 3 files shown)


jacoblee-io (commented):
Reviewed — clean and correct; the fix is well-scoped and I verified each claim against the code:

- **Identity keying**: `knowledgeRepoDirName` = `sanitize(name).slice(0,66)` + `--` + `sha256(repoId.trim())[:12]`, so the max length stays at 80 (matching the old `sanitizeKnowledgeRepoDir` cap), all-CJK names no longer collapse to a shared `repo`, and `crypto` is imported.
- **Atomicity**: both new guards (duplicate `repo.id`, derived-name collision) throw *before* the staging swap; the `catch` removes the staging dir and rethrows, so a bad bundle leaves the existing knowledge dir intact.
- **Single-repo unchanged**: `repos.length === 1` uses the flat staging path and never calls `knowledgeRepoDirName`.
- **Index links** use the derived `dirName` with `name v{version}` display — matches the new regression test.
- **No stale dirs**: `replaceDirectoryContentsFromStaging` clears every target entry (except the staging dir) before moving the new contents in, so directories under the old sanitized-name scheme are wiped on the next materialize.
- `repo.id` is a required field on the payload type, so the "id required" throw in `knowledgeRepoDirName` is defensive-only, not a new failure mode.

One cosmetic nit inline; nothing blocking. CI green (Type Check / Test / Portal Web).

src/agentbox/sync-handlers.ts
```diff
@@ -330,8 +330,18 @@ export const knowledgeHandler: AgentBoxSyncHandler<KnowledgeBundlePayload> = {
         const repoRoot = path.join(stagingDir, "repos");
         fs.mkdirSync(repoRoot, { recursive: true });
         const indexLines = ["# Knowledge Index", "", "This index was generated from active knowledge repositories.", ""];
+        const seenRepoIds = new Set<string>();
+        const seenDirNames = new Set<string>();
         for (const repo of repos) {
-          const dirName = sanitizeKnowledgeRepoDir(repo.name);
+          if (seenRepoIds.has(repo.id)) {
```
jacoblee-io, line 336: Nit (cosmetic, non-blocking): the `seenRepoIds` guard dedups on the raw `repo.id`, but `knowledgeRepoDirName` hashes `repoId.trim()`. So two ids differing only in surrounding whitespace (`"abc"` vs `" abc"`) slip past this guard yet derive the *same* directory name — they're still caught, but by the `seenDirNames` collision check below with the less-specific "directory collision" error rather than "duplicate repository id". If you want the more precise message to win, dedup on `repo.id.trim()` here too. Both paths already throw, so this is purely about which error the caller sees.

jacoblee-io (approved):
Approving. Verified the fix against the code: multi-repo directories are keyed by `<slug>--sha256(repoId.trim())[:12]` (max length still 80), the duplicate-id and derived-name-collision guards both throw before the atomic staging swap, single-repo layout is untouched, the index links use the derived directory names, and `replaceDirectoryContentsFromStaging` fully clears the target so no old sanitized-name directories linger. `repo.id` is required by the payload type, so the id-required throw is defensive only. The CJK-only regression test covers distinct dirs + isolated content + matching index links. CI green.

Only a cosmetic nit left (trim in the `seenRepoIds` guard for a more precise error message) — not blocking.

PR #400, opened 2026-07-11
feat(kbc): align wiki output with OKF v0.1

LikiosSedo:
## Summary

Align newly authored Siclaw knowledge bundles with OKF v0.1 while keeping three concerns explicit: OKF core conformance, Siclaw's portable-output profile, and compatibility with existing legacy pages.

## What changed

- Parse concept frontmatter with `yaml.safe_load` and require a non-empty string `type`.
- Validate reserved `index.md` and `log.md` structure, including entries in every log date group.
- Treat file-relative Markdown links as a Siclaw producer profile rather than an OKF core rule; bundle-root links remain valid OKF input.
- Ignore fenced and inline code while linting links and body provenance markers.
- Preserve complete imported source filenames, including spaces, nested Chinese/ASCII parentheses, commas, semicolons, and enumeration punctuation.
- Resolve spaced page links in CommonMark angle form, URL-encoded form, and the tolerant raw form already emitted by existing agents.
- Capture format violations at incremental kickoff. Existing violations on pages unchanged by the round remain visible migration debt; violations on actually changed or new pages still fail the gate.
- Teach the runtime consumer to follow nested relative Markdown links while retaining legacy wikilink support.
- Share one Python dependency list between the compile image and KBC CI, including `PyYAML==6.0.3`.
- Keep safe YAML serialization and mandatory type checks in the legacy emitter.

## Review fixes

- A body citation such as `(source: 上线方案（终版）、GPU 调度，修订；v2.md)` no longer becomes an impossible `body_source_malformed` violation.
- Source examples inside fenced or inline code no longer become real provenance requirements.
- Removing the extension or citing a file absent from `compiled_from` remains fail-closed.
- Space-separated source locators such as `(source: report.pdf p.12)` are parsed without truncating filenames; malformed suffixes and uncited files remain fail-closed.
- Incremental format enforcement now keys off pages actually changed this round, so merely authorizing an untouched legacy `index.md` no longer triggers migration.
- Rebased onto the current `main` after #403; the current KBC mTLS test is included in validation.

## Behavior boundary

- Ordinary incremental compilation enforces current rules on pages actually changed or created while leaving unchanged legacy pages byte-stable.
- A non-incremental authoring round still evaluates the full tree and can surface legacy format debt. This PR does not add a separate manual migration command; reviewers should treat that as an explicit product boundary rather than an implicit migration workflow.

## Validation

- [x] `python test_compile_box.py`
- [x] `python test_office_ingest.py`
- [x] `python test_selfcheck.py`
- [x] `python test_redblue.py`
- [x] `python test_mtls_auth.py`
- [x] `python test_incremental.py`
- [x] `npm test -- --run` — 4,391 passed, 2 skipped
- [x] `npx tsc --noEmit`
- [x] `npm run build`
- [x] `docker build --check -f platform/pod/Dockerfile .` (from `kbc/`)
- [x] `git diff --check`

## Architecture checklist

- [x] **Incremental boundary**: unchanged historical pages remain byte-stable and are not added to repair scope merely because they were authorized.
- [x] **Consumer compatibility**: standard Markdown links and legacy wikilinks remain supported.
- [x] **Security model**: untrusted YAML uses `safe_load`; no shell execution path changes.
- [x] **Database parity**: not applicable; no database changes.


jacoblee-io (commented):
### Review — align wiki output with OKF v0.1

Solid, well-tested conformance work, and the security posture is clean (`yaml.safe_load` for tenant frontmatter, `safe_dump` on emit, no shell-path changes). The TS `overview-generator`/`agent-factory`/`test_role` edits are in-scope — the consumer side of the same `[[wikilink]]` → standard-markdown-link move, with legacy links still tolerated. I verified the incremental *format*-violation grandfathering logic is correct (new/editable/non-baseline all block; only baseline+untouched are grandfathered) and that `_markdown_prose` doesn't crash/loop on pathological backticks/fences/CRLF.

Findings (see inline). The two I'd fix before merge:
- **[med]** the new `body_source_malformed` check false-positives on a *space*-separated file+locator citation (`(source: a.md §3)`, `(source: report.pdf p.12)`) — reproduced against the exact regex;
- **[med]** and because non-format lint isn't grandfathered under incremental, a single such citation on an untouched legacy page can wedge every incremental round from converging. Fixing the citation parser resolves both, or widen the grandfathered set.

Lower severity (inline): reserved page whose body starts with `---` mis-parsed; `_MD_LINK_RE` drops a link with `)` in its title; >300-char citations silently dropped.

Two more, non-inline:
- **[low] index vs log fence-masking asymmetry**: `_okf_index_violations` validates the raw (unmasked) body while `_okf_log_violations` masks fences first, so the two reserved-page validators disagree on whether code counts as content — the index check is too lenient and the log check is prone to a false "empty date group" when the only entries sit in a fence.
- **[low, cleanup] `_markdown_prose`/`_mask_span` are duplicated** verbatim in `kbc/tools/emit.py` and `kbc/platform/pod/selfcheck.py` (~40 lines of hand-rolled CommonMark-ish masking that must stay byte-in-sync across two packages). At least cross-reference them so a future fix lands in both.

No security or data-loss issue; CI green. The incremental-wedge risk (first two findings) is the one worth confirming against real legacy citation forms before this reaches existing KBs.

kbc/platform/pod/selfcheck.py
```diff
@@ -245,18 +248,288 @@ def _matches(path: str, pattern: str) -> bool:
...
-# Body-level provenance mentions: (source: X) / (源:X) / (来源:X). The capture
-# is then mined for filename-looking tokens so locators ("§3", "p.12") and
-# prose sources ("内部访谈") never false-positive.
-_BODY_SOURCE_RE = re.compile(r"[（(]\s*(?:source|src|源|来源)\s*[:：]\s*([^）)]{1,300})[）)]", re.IGNORECASE)
-_FILENAME_RE = re.compile(
-    r"[^\s,;、；()（）'\"`]+\.(?:" + "|".join(sorted(e[1:] for e in KNOWN_SOURCE_EXTS)) + r")\b",
+_BODY_SOURCE_START_RE = re.compile(
+    r"(?P<open>[（(])\s*(?:source|src|源|来源)\s*[:：]\s*", re.IGNORECASE,
+)
+_SOURCE_FILE_END_RE = re.compile(
```
jacoblee-io, line 260: **[med] `body_source_malformed` false-positives on a space-separated file+locator citation.** The lookahead `(?=\s*(?:[,，;；、]|$))` only accepts a separator char or end-of-string after the extension — never whitespace-then-locator. I reproduced against this exact regex:

```
(source: a.md §3)            -> malformed ['a.md §3']
(source: 报告.pdf 第3节)        -> malformed ['报告.pdf 第3节']
(source: a.md, b.pdf p.5)     -> files ['a.md'], malformed ['b.pdf p.5']
(source: a.md, §3)            -> OK (comma form)
```

The function's own docstring says "Locator-only items such as §3 and p.12 are accepted after a file", but a *space* between the file and the locator breaks that. This is a new violation kind in this PR; the old `_FILENAME_RE.findall` extracted the filename regardless of what followed, so this is a tightening regression. Consider allowing `\s+<locator>` (or any whitespace) after the extension before requiring a separator/EOS.

kbc/platform/pod/compile_box.py
```diff
@@ -1765,8 +1766,25 @@ async def _post_turn_selfcheck(run) -> str | None:
                 after = incremental.page_hashes(workdir)
                 incr_violations = incremental.integrity_violations(incr["before"], after, editable)
     report = selfcheck.run_layer1(workdir)
+    grandfathered_format: list[dict] = []
     if incr:
-        report["incremental"] = {"out_of_scope_pages": incr_violations, "restored_pages": restored_pages}
+        blocking, grandfathered_format = selfcheck.filter_incremental_format_violations(
```
jacoblee-io, line 1772: **[med, related] Only format-policy violations are grandfathered under incremental; non-format lint (`body_source_malformed`, `broken_link`, orphan) is not.** `baseline_format_violations` (line ~2094) captures only `format_violation_keys` = `format_policy_violations`, so `filter_incremental_format_violations` can only grandfather those. `body_source_malformed` is added in `lint_candidate` directly, so it always stays in `blocking` — even on an untouched legacy page. Combined with the citation false-positive above, a single legacy `(source: x.md §3)` on any untouched page makes `report["lint"]["ok"]` false and can wedge every incremental round on that KB from converging. Either widen the grandfathered set to include the non-format lint kinds for untouched pages, or fix the citation parser (previous comment) so the false-positive can't arise.

kbc/platform/pod/selfcheck.py
```diff
@@ -245,18 +248,288 @@ def _matches(path: str, pattern: str) -> bool:
...
+    """Markdown text with YAML frontmatter, fenced code, and code spans masked.
+
+    Link lint operates on rendered prose, not raw source. A raw ``[[`` scan
+    rejects ordinary Bash conditionals and example links inside fenced/inline
+    code, both of which OKF explicitly permits. This deliberately implements
+    the code constructs the compiler emits without adding another production
+    markdown dependency.
+    """
+    _, body, error = parse_okf_frontmatter(md_text)
+    if error and not md_text.startswith("---"):
```
jacoblee-io, line 323: **[low] A reserved page whose body legitimately starts with `---` gets its head content mis-parsed as frontmatter.** `parse_okf_frontmatter` treats a leading `---` as a frontmatter opener; for a `log.md`/nested `index.md` that opens with a thematic break (or a table using `---`), the content up to the next `---` is consumed as "frontmatter", and `_markdown_prose` returns only the remainder — which can trip a spurious `okf_log_structure`/`okf_index_structure` violation. Reserved pages are documented as carrying no frontmatter, so consider not stripping a frontmatter block from them at all (pass the raw text through for reserved pages).

kbc/platform/pod/selfcheck.py
```diff
@@ -245,18 +248,288 @@ def _matches(path: str, pattern: str) -> bool:
     return _seg_glob(pattern.split("/"), path.split("/"))
 
 
-_MD_LINK_RE = re.compile(r"\]\(([^)#\s]+\.md)(?:#[^)]*)?\)")
+_MD_LINK_RE = re.compile(r"\]\(\s*(<[^>\n]+>|[^)\n]+)\)")
```
jacoblee-io, line 251: **[low] `_MD_LINK_RE` drops a link whose title text contains `)`.** `\]\(\s*(<[^>\n]+>|[^)\n]+)\)` stops the destination capture at the first `)`, so `[t](a.md "see (v2)")` isn't recognized as an out-link — a page reachable only via such a link would be reported as a false orphan. Low frequency, but the titled-link branch in `_markdown_link_targets` won't see it either.

kbc/platform/pod/selfcheck.py
```diff
@@ -269,36 +542,104 @@ def _strip_frontmatter(text: str) -> str:
...
+    return targets
+
+
+def _body_source_payloads(text: str) -> list[str]:
+    """Extract source-marker payloads while preserving nested filename pairs."""
+    payloads: list[str] = []
+    prose = _markdown_prose(text)
+    for match in _BODY_SOURCE_START_RE.finditer(prose):
+        stack = [")" if match.group("open") == "(" else "）"]
+        for pos, char in enumerate(prose[match.end():match.end() + 301], start=match.end()):
```
jacoblee-io, line 584: **[low] Citations longer than 300 chars are silently dropped.** `prose[match.end():match.end() + 301]` bounds the nested-paren scan to 300 chars; a longer `(source: ...)` never closes its paren stack within the window, so no payload is appended and a genuinely malformed/uncited long citation escapes the lint (false-negative). A long citation is unusual, but the cap silently narrows coverage — consider logging or widening it.

jacoblee-io (commented):
Re-review of aa29314c — the new commit is a correct, no-regression refinement (grandfather only pages that *actually changed* this round, via `incremental.changed_pages`, rather than any authorized page). Nice.

But it doesn't cover the two findings I'd want fixed before merge, both still open:
- **#1 (med)** `body_source_malformed` false-positives on a space-separated file+locator citation — the `_SOURCE_FILE_END_RE` regex is unchanged.
- **#2 (med)** non-format lint (`body_source_malformed`, `broken_link`, orphan) is still not grandfatherable, because only `format_policy_violations` keys enter the baseline. So a single legacy space-form citation on an untouched page can still wedge every incremental round.

Fixing #1 (allow whitespace before a locator) is the smallest change and neutralizes the #2 trigger in practice. The five lower-severity items from the previous review (reserved-page `---` mis-parse, `_MD_LINK_RE` title-`)`, >300-char citation drop, index/log fence asymmetry, `_markdown_prose` duplication) are also untouched — all low, at your discretion.

kbc/platform/pod/compile_box.py
```diff
@@ -1765,8 +1766,26 @@ async def _post_turn_selfcheck(run) -> str | None:
                 after = incremental.page_hashes(workdir)
                 incr_violations = incremental.integrity_violations(incr["before"], after, editable)
     report = selfcheck.run_layer1(workdir)
+    grandfathered_format: list[dict] = []
     if incr:
-        report["incremental"] = {"out_of_scope_pages": incr_violations, "restored_pages": restored_pages}
+        format_changed_pages = set(incremental.changed_pages(incr["before"], after))
```
jacoblee-io, line 1771: Good refinement — keying grandfathering on actually-*changed* pages (not merely-authorized ones) is correct and I confirmed no regression. But this is orthogonal to the earlier finding: `filter_incremental_format_violations` can still only grandfather violations whose key is in `baseline_format_violations`, and that baseline is `format_violation_keys` = `format_policy_violations` only. `body_source_malformed` (and `broken_link`, orphan) are emitted in `lint_candidate`, NOT in `format_policy_violations`, so they can never be grandfathered — a legacy one on an untouched page still lands in `blocking` and keeps `report["lint"]["ok"]` false. So the incremental-wedge path is still open. Either fold the non-format lint kinds into the grandfatherable baseline for untouched pages, or fix the citation parser below so the false-positive can't arise in the first place.

kbc/platform/pod/selfcheck.py
```diff
@@ -245,18 +248,289 @@ def _matches(path: str, pattern: str) -> bool:
...
-# Body-level provenance mentions: (source: X) / (源:X) / (来源:X). The capture
-# is then mined for filename-looking tokens so locators ("§3", "p.12") and
-# prose sources ("内部访谈") never false-positive.
-_BODY_SOURCE_RE = re.compile(r"[（(]\s*(?:source|src|源|来源)\s*[:：]\s*([^）)]{1,300})[）)]", re.IGNORECASE)
-_FILENAME_RE = re.compile(
-    r"[^\s,;、；()（）'\"`]+\.(?:" + "|".join(sorted(e[1:] for e in KNOWN_SOURCE_EXTS)) + r")\b",
+_BODY_SOURCE_START_RE = re.compile(
+    r"(?P<open>[（(])\s*(?:source|src|源|来源)\s*[:：]\s*", re.IGNORECASE,
+)
+_SOURCE_FILE_END_RE = re.compile(
```
jacoblee-io, line 260: Still unchanged — the med finding stands. `_SOURCE_FILE_END_RE`'s lookahead accepts only `[,，;；、]` or end-of-string after the extension, so a space-separated file+locator citation (`(source: a.md §3)`, `(source: report.pdf p.12)`, or the 2nd file in `(source: a.md, b.pdf p.5)`) is flagged `body_source_malformed`. Fixing this (allow whitespace-then-locator) is the cheapest way to also close the incremental-wedge, since it removes the false-positive that the un-grandfathered lint would otherwise block on.

jacoblee-io (approved):
Approving on b2e8a426 — the two medium findings are resolved and I verified the fix against the exact new regex (no regression).

**#1 (citation false-positive) fixed.** `_SOURCE_FILE_END_RE` now also accepts `\s+<locator>\s*(sep|$)` after the extension, and the scan loop consumes an optional locator + separator before advancing, so a locator binds to the file before it. Reproduced:

| citation | result |
| --- | --- |
| `(source: a.md §3)` | file `a.md` ✓ |
| `(source: 报告.pdf 第3节)` | file ✓ |
| `(source: a.md §3, b.pdf p.5)` | two files ✓ |
| `(source: 内部访谈)` | malformed ✓ (still fail-closed) |
| `(source: §3)` (no file) | malformed ✓ |
| `(source: a.md p.12 garbage)` | malformed ✓ (trailing junk still caught) |
| nested-paren full filename | preserved ✓ |

**#2 (incremental wedge) resolved in practice.** With #1 fixed, a legacy space-form citation no longer produces `body_source_malformed`, so the un-grandfathered lint can't wedge an incremental round on it. The earlier `changed_pages`-vs-`editable` refinement (aa29314c) is also correct and regression-free. The structural note — non-format lint kinds aren't grandfatherable — remains, but its only realistic false-positive trigger is gone, and a genuine `broken_link`/orphan on an untouched legacy page is worth surfacing rather than hiding.

Earlier lower-severity items remain open, all optional (inline on the first review): reserved page whose body starts with `---` mis-parsed; `_MD_LINK_RE` drops a link with `)` in its title; >300-char citations silently dropped; index-vs-log fence-masking asymmetry; `_markdown_prose`/`_mask_span` duplicated in emit.py and selfcheck.py (keep them in sync).

Security clean (`yaml.safe_load`/`safe_dump`, no shell path change). CI green (Type Check / KBC Box / Test / Portal Web).

PR #401, opened 2026-07-11
fix(runtime): harden capability box lifecycle boundaries

LikiosSedo:
## Summary

- fail capability setup closed with stage-specific materialization errors and metrics
- authenticate compile-box data routes with client certificates and isolate KB boxes with the Runtime-only NetworkPolicy
- preserve K8s boxes across Runtime rolling restarts; setup failure only cleans up a box created by that setup attempt
- treat consumer LLM config as one authority block and use Runtime Helm fallback only when the block is absent
- deliver LLM credentials through the mTLS `/session` payload and keep LLM/embedding secrets out of KB PodSpecs
- add bounded Prometheus metrics for capability start, materialization, relay failures, and active runs
- run the mTLS route suite in KBC CI

## Verification

- `npm test` (4361 passed, 2 skipped)
- `npm run build`
- 111 focused capability/manager/spawner tests
- full `kbc/platform/pod/test_compile_box.py` protocol suite
- `test_mtls_auth.py` (8 passed)
- Helm lint and Runtime/compile-box templates
- local merge simulation against current `main`
- `git diff --check`

## Operational note

A live box's already-connected SDK child keeps its session-start credential. Token rotation therefore still requires a grace window plus KB box respawn; this PR deliberately does not claim hot rotation.

## Dependency

Targets `main` directly.

### DIFF
--- a/kbc/platform/pod/compile_box.py
+++ b/kbc/platform/pod/compile_box.py
@@ -1701,6 +1701,13 @@ DEFAULT_COMPILE_ALLOWED_TOOLS = [
 ]
 
 
+def _compile_model() -> str:
+    """Resolve the model shared by every compiler-owned SDK session."""
+    return (os.environ.get("KBC_COMPILE_MODEL")
+            or os.environ.get("ANTHROPIC_MODEL")
+            or "claude-opus-4-6")
+
+
 def _compile_session_opts(run: "CompileRun", wd: str, system_prompt: str, session_id: str) -> "ClaudeAgentOptions":
     """One options builder for the persistent session AND every batch session —
     identical role/tools/model so a batch page is written under exactly the same
@@ -1717,9 +1724,7 @@ def _compile_session_opts(run: "CompileRun", wd: str, system_prompt: str, sessio
         # Pin the compile model explicitly: the box talks to massapi (Bedrock),
         # which serves specific ids — the SDK default may not be one, and the KB
         # compile default is opus by product decision. Overridable per-deploy.
-        model=(os.environ.get("KBC_COMPILE_MODEL")
-               or os.environ.get("ANTHROPIC_MODEL")
-               or "claude-opus-4-6"),
+        model=_compile_model(),
         max_turns=int(os.environ.get("KBC_MAX_TURNS", "150")),
         max_buffer_size=SDK_MAX_BUFFER_BYTES,
         session_id=session_id,
@@ -2042,7 +2047,7 @@ async def _plan_batches(run: "CompileRun", inventory: list) -> dict:
             permission_mode="bypassPermissions",
             hooks={"PreToolUse": [HookMatcher(hooks=[_make_compile_path_guard(Path(wd), run.locale)])]},
             setting_sources=[],
-            model=os.environ.get("KBC_COMPILE_MODEL", "claude-opus-4-6"),
+            model=_compile_model(),
             max_turns=8,
             max_buffer_size=SDK_MAX_BUFFER_BYTES,
             session_id=str(uuid.uuid4()),
--- a/kbc/platform/pod/compile_box.py
+++ b/kbc/platform/pod/compile_box.py
@@ -56,6 +56,10 @@ import batching
 import incremental
 import mediaverify
 import office_ingest
+from mtls_auth import (
+    client_certificate_error as _client_certificate_error,
+    server_ssl_context,
+)
 import redblue
 import selfcheck
 from engine import ClaudeEngine
@@ -3070,8 +3074,19 @@ async def _flush_on_shutdown(_app) -> None:
             pass
 
 
+@web.middleware
+async def _client_certificate_middleware(request: web.Request, handler):
+    error = _client_certificate_error(request)
+    if error is not None:
+        return web.json_response({"error": error}, status=403)
+    return await handler(request)
+
+
 def build_app() -> web.Application:
-    app = web.Application(client_max_size=_http_max_request_bytes())
+    app = web.Application(
+        client_max_size=_http_max_request_bytes(),
+        middlewares=[_client_certificate_middleware],
+    )
     app.on_shutdown.append(_flush_on_shutdown)
     app.add_routes([
         web.post("/sources", handle_sources),
@@ -3090,25 +3105,9 @@ def build_app() -> web.Application:
 
 
 def _ssl_context():
-    """Production mTLS: with certs, HTTPS + required client cert (only the
-    runtime/gateway can drive the box). No certs locally → plain HTTP."""
+    """Production mTLS, or intentional plain HTTP when no certs exist locally."""
     cert_dir = Path(os.environ.get("SICLAW_CERT_PATH", "/etc/siclaw/certs"))
-    crt, key, ca = cert_dir / "tls.crt", cert_dir / "tls.key", cert_dir / "ca.crt"
-    if not (crt.exists() and key.exists()):
-        return None
-    import ssl
-    ctx = ssl.SSLContext(ssl.PROTOCOL_TLS_SERVER)
-    ctx.load_cert_chain(str(crt), str(key))
-    if ca.exists():
-        ctx.load_verify_locations(str(ca))
-        # CERT_OPTIONAL, not CERT_REQUIRED: the k8s readiness/liveness probes hit
-        # /health over HTTPS WITHOUT a client cert; requiring one fails the TLS
-        # handshake so the probes never pass and kubelet kills the box. The box is
-        # only addressed by the runtime in-cluster, so requesting-but-not-requiring
-        # the client cert is acceptable for v1. (Production: exempt /health + check
-        # the cert per-route, like agentbox does.)
-        ctx.verify_mode = ssl.CERT_OPTIONAL
-    return ctx
+    return server_ssl_context(cert_dir)
 
 
 def main():
--- a/kbc/platform/pod/compile_box.py
+++ b/kbc/platform/pod/compile_box.py
@@ -15,7 +15,7 @@ structured signals:
 Surface (driven by the runtime):
   POST /sources            {run_id?, workdir?, bundle_base64, bundle_sha256?, locale?} install the frozen raw bundle → workdir/raw
   POST /authoring          {run_id?, workdir?, bundle_base64, bundle_sha256?, locale?} install authoring/candidate/eval/release assets → workdir/
-  POST /session/{run_id}   {workdir?, instruction?, allowed_tools?, locale?} start the run's persistent conversational session (waits for the first /message); idempotent
+  POST /session/{run_id}   {workdir?, instruction?, allowed_tools?, locale?, llm?, settings?} start the run's persistent conversational session (waits for the first /message); idempotent
   POST /message/{run_id}   {message} inject one user turn into the persistent session (prepare/compile/patch are all ordinary turns)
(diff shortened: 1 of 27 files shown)


src/gateway/agentbox/k8s-spawner.ts
```diff
@@ -506,16 +504,38 @@ export class K8sSpawner implements BoxSpawner {
...
-            livenessProbe: {
-              httpGet: { path: "/health", port: 3000 as any, scheme: "HTTPS" },
-              initialDelaySeconds: 10,
-              periodSeconds: 10,
-            },
+            // NetworkPolicy for capability boxes admits only Runtime ingress.
+            // Kubelet HTTP probes originate outside that podSelector on several
+            // CNIs, so use an in-container HTTPS check for KB profiles. The
+            // endpoint is deliberately the sole route that needs no client cert.
+            readinessProbe: profile.name === "kb-compile" || profile.name === "kb-test"
```
jacoblee-io, line 511: **[med] The KB exec probes set no `timeoutSeconds`, so kubelet's 1s default applies.** A cold `python -c` (interpreter start + `import ssl,urllib` + a loopback TLS handshake) can exceed 1s under node load, and the inner `urlopen(timeout=2)` never helps because kubelet SIGKILLs the exec at 1s first. With the default `failureThreshold: 3`, three clipped liveness checks restart the container mid-compile. Add `timeoutSeconds: 3` (or higher) to both the readiness and liveness exec probes for the kb-compile/kb-test profiles.

kbc/platform/pod/compile_box.py
```diff
@@ -1712,7 +1717,9 @@ def _compile_session_opts(run: "CompileRun", wd: str, system_prompt: str, sessio
         # Pin the compile model explicitly: the box talks to massapi (Bedrock),
         # which serves specific ids — the SDK default may not be one, and the KB
         # compile default is opus by product decision. Overridable per-deploy.
-        model=os.environ.get("KBC_COMPILE_MODEL", "claude-opus-4-6"),
+        model=(os.environ.get("KBC_COMPILE_MODEL")
+               or os.environ.get("ANTHROPIC_MODEL")
```
jacoblee-io: **[med-low] This new `ANTHROPIC_MODEL` fallback tier was added to the compile session but not to the batch planner.** `_plan_batches` still uses `model=os.environ.get("KBC_COMPILE_MODEL", "claude-opus-4-6")` (around line 2045), so a consumer that selects a model via `llm.model` (which `_apply_session_config` maps to `ANTHROPIC_MODEL`) with no `KBC_COMPILE_MODEL` set drives the compile session on the chosen model but the planner silently pins to opus — a cost/compliance divergence. Mirror this precedence chain there. (Same 'sibling session missed' shape as the path-guard gap in #396's `_plan_batches`.)

src/gateway/server.ts
```diff
@@ -1189,7 +1233,11 @@ export async function startRuntime(opts: StartRuntimeOptions): Promise<RuntimeSe
     async close() {
       metricsAggregator?.destroy();
       frontendClient.close();
-      await agentBoxManager.cleanup();
+      // Older embedded test/adapter managers may only implement cleanup(); the
+      // concrete manager's shutdown() preserves K8s boxes across Runtime rolls.
+      const manager = agentBoxManager as AgentBoxManager & { shutdown?: () => Promise<void> };
+      if (typeof manager.shutdown === "function") await manager.shutdown();
```
jacoblee-io, line 1239: **[low] The K8s box-preservation guarantee rests on a duck-typed `shutdown` presence check that silently degrades to a destructive path.** If a future wrapper/Proxy/decorator ever hands `startRuntime` a K8s-backed manager that doesn't surface `shutdown` as a callable, this falls to `manager.cleanup()`, and `K8sSpawner.cleanup()` deletes EVERY `app=agentbox` pod + cert Secret in the namespace (chat AND capability) on every graceful Runtime shutdown/roll — exactly what `shutdown()` exists to prevent. Prefer branching on the known K8s flag (or asserting `shutdown` exists in K8s mode) over silently choosing the destructive fallback.

src/gateway/agentbox/manager.ts
```diff
@@ -72,7 +79,8 @@ export class AgentBoxManager {
         console.warn("[agentbox-manager] orphan sweep failed:", err?.message ?? err));
     // unref'd + stored (review finding): the sweep must never pin the event
     // loop or outlive cleanup() — same discipline as the run watchdog.
-    (setTimeout(tick, 60_000) as any).unref?.();
+    this.orphanSweepInitialTimer = setTimeout(tick, 60_000);
```
jacoblee-io, line 82: **[low] `startOrphanSweep()` has no re-entry guard, unlike `startHealthCheck`/`startWatchdog`.** A second call overwrites `orphanSweepInitialTimer`/`orphanSweepTimer` without clearing the previous pair, leaving the first setTimeout+setInterval live-but-unreferenced — two sweep loops run (double K8s LIST load) and `stopBackgroundLoops()` only cancels the latest. Single call site today, so latent; add `if (this.orphanSweepTimer) return;` at the top to match the other loop starters.

src/gateway/capability/session-config.ts
```diff
@@ -0,0 +1,33 @@
...
+ * Consumer ownership is whole-block: even an empty consumer object is
+ * authoritative. Runtime's Helm credentials are consulted only when the
+ * consumer omitted the object entirely, which prevents a tenant-selected URL
+ * from accidentally receiving a Runtime-owned token.
+ */
+export function resolveCapabilitySessionLlm(
+  consumer: CapabilityLlmConfig | undefined,
+  env: RuntimeEnv = process.env,
+): CapabilityLlmConfig | undefined {
+  if (consumer !== undefined) return consumer;
```
jacoblee-io, line 21: **[low] Two fail-closed footguns in the whole-block authority.** (a) `consumer !== undefined` returns the object as-is, so a consumer sending `llm: {}` is authoritative-empty → the box's `_apply_session_config` pops all `ANTHROPIC_*` → no credentials → the SDK connect fails and the whole `capability.start` fails closed. The `undefined`-vs-`{}` distinction is fragile across JSON-RPC. (b) The Runtime fallback below emits `{ base_url }` when only `ANTHROPIC_BASE_URL` is set (empty token Secret) — an endpoint with no credential, again a cryptic SDK failure. Both are documented as intended whole-block behavior, but consider (a) treating an empty consumer object as absent, and (b) requiring `auth_token` XOR `api_key` before emitting a fallback, so a misconfig surfaces as a clear error rather than a token wipe.

jacoblee-io (commented):
### Review — harden capability box lifecycle boundaries

Large, security-sensitive PR; I reviewed it in depth (mTLS gate, box lifecycle, secret placement, LLM authority, fail-closed materialization, metrics) and the core mechanics are sound. Verified against the branch code, not the comments:

- **mTLS application gate** (`mtls_auth.py` + middleware): HTTPS-only when certs are present (no dual HTTP listener to bypass), `/health` is the sole exact-path exemption, `CERT_OPTIONAL` still rejects a no-cert caller (403) and fails the TLS handshake for any cert not signed by the CA, and OU ∈ {Runtime, Gateway} is the sole authorization. Confirmed the Runtime dials the box presenting its `issueServerCertificate` client cert (`OU=Runtime`), so the gate admits legitimate traffic and rejects the OU-less AgentBox cert. Partial TLS material fails startup loudly.
- **Secrets out of PodSpec**: `envForward` drops `ANTHROPIC_API_KEY`/`AUTH_TOKEN`; embedding creds forward only to the `agent` profile. Box startup doesn't read tokens from env (the SDK connects per-run after `/session`), so no regression.
- **Preserve-across-rolls**: K8s `shutdown()` no longer stops boxes; `this.boxes` is always empty in K8s so the clear is a no-op; the orphan sweep + watchdog (reading live pods, not the map) still reclaim truly-dead boxes. `created` disposition is correct incl. the profile-recreate path (`created:true`).
- **NetworkPolicy**: `boxType In [kb-compile,kb-test]` matches the spawner's stamped label; the runtime ingress podSelector matches the runtime deployment labels; both gated on `runtime.enabled`.
- **Fail-closed materialize + LLM whole-block authority + metrics**: empty-KB/empty-workspace return cleanly (no throw), 409 reattach short-circuits before the workspace stages, `created:false` preserves adopted boxes on transient failure, and the metrics are bounded (fixed `stage`/`outcome` enums, no run/repo ids).

**Dependency**: this branch stacks on #396 (its two commits are ancestors) — merge #396 first, or the diff-against-main carries it.

**Findings** (all non-blocking; see inline): [med] KB exec probes lack `timeoutSeconds` (liveness flap risk under load); [med-low] batch-planner model ignores the new `ANTHROPIC_MODEL` tier; [low] destructive `cleanup()` fallback in K8s behind a duck-typed check; [low] `startOrphanSweep` re-entry timer leak; [low] LLM empty-`{}` / base-url-only fail-closed footguns.

No merge-blocking correctness or security bug. I'd fix the exec-probe timeout (#1) and the planner model (#2) before shipping; the rest are hardening.

jacoblee-io (dismissed):
Approving — no merge-blocking correctness or security issue. I verified the security core against the branch code: the mTLS application gate is fail-closed and OU-aligned (Runtime dials with an `OU=Runtime` client cert; no-cert/unsigned-cert/wrong-OU are all rejected; `/health` is the only exemption and there is no dual HTTP listener to bypass), LLM/embedding secrets stay out of the KB PodSpec, materialization fails closed with typed stages, K8s box preservation across Runtime rolls is correct, the NetworkPolicy selectors align with the spawner's stamped labels, and the lifecycle metrics are bounded.

Please still address these as follow-ups (both cheap, both real):

- **[med] KB exec probes lack `timeoutSeconds`** (`k8s-spawner.ts`): kubelet's 1s default can clip a cold `python -c` + TLS handshake under node load, and with the default `failureThreshold: 3` the liveness probe can restart a box mid-compile. Add `timeoutSeconds: 3` (or higher) to both KB exec probes.
- **[med-low] batch-planner model ignores the new `ANTHROPIC_MODEL` tier** (`compile_box.py` `_plan_batches`, ~line 2045): a consumer selecting a model via `llm.model` drives the compile session on it but the planner silently pins to opus. Mirror the compile session's precedence chain.

The three low items (destructive `cleanup()` fallback behind a duck-typed check, `startOrphanSweep` re-entry guard, LLM empty-`{}`/base-url-only footguns) are hardening — fix at your discretion.

Dependency: this stacks on #396 — merge #396 first, or the diff-against-main carries its two commits.

jacoblee-io (approved):
Re-approving on 5e8f80cc — both follow-ups from the previous review are fixed, cleanly and at the right depth:

- **#1 exec-probe timeout**: `timeoutSeconds: 3` added to both KB readiness and liveness exec probes, so the inner `urlopen(timeout=2)` now resolves before kubelet's probe deadline instead of being SIGKILLed at the 1s default. Covered by a test asserting `timeoutSeconds === 3` on both.
- **#2 planner model**: extracted a shared `_compile_model()` helper (`KBC_COMPILE_MODEL → ANTHROPIC_MODEL → opus`) now used by BOTH `_compile_session_opts` and `_plan_batches` — one precedence policy instead of a duplicated chain, which fixes the root cause (the two sessions can't drift again). Test proves the planner honors a consumer-selected `ANTHROPIC_MODEL`.

The three low items remain optional hardening. CI green (Type Check / KBC Box / Test / Portal Web). Reminder: merge #396 before this one.

PR #402, opened 2026-07-11
feat(kb): add typed authoring command protocol

LikiosSedo:
## Summary

- add a generic `capability.command` envelope alongside conversational `capability.message`
- validate and execute typed KB authoring actions in the compile box
- keep routing, idempotency, operation fencing, and plan binding independent of display/content locale
- retain fixed-prefix message handling only as a rolling-upgrade adapter
- preserve the #401 LLM/session authority and adopted-box lifecycle boundaries after the stack rebase

## Verification

- Runtime command, setup, and session-config tests
- TypeScript `--noEmit`
- full `kbc/platform/pod/test_compile_box.py` protocol suite
- GitHub Type Check, KBC Box Test, and Portal Web Test

## Dependency

Targets `main` directly. Logical merge/review order remains #401 then #402; #403 stays stacked on this branch. The paired Sicore consumer MR is !622.


kbc/platform/pod/compile_box.py
```diff
@@ -2856,7 +3107,76 @@ async def handle_message(request: web.Request):
...
+        # action that has not actually been scheduled. Exact replays are handled
+        # above; every distinct concurrent command fails closed and can be
+        # retried as a new product action after the run returns idle.
+        if run._turn_active or run._batch_active:
+            raise CommandRejected("another authoring command is already running", 409)
+        if run._command_context is not None and run._command_context != context:
+            raise CommandRejected("run is pinned to another operation generation", 409)
+        _prepare_command(run, command)
+        text = _render_command(run, command)
+        run._command_context = context
```
jacoblee-io, line 3175: **[med — please confirm against the sicore contract] `_command_context` pins the run to one `(operation_id, generation)` for its whole life and is never cleared.** Any later command with a different context gets a hard 409 ("pinned to another operation generation"). Per design invariants #5/#9 this is intended (one run = one op/gen), so the v1 follow-up chain (submit_decisions → approve_plan → apply_rulings → repair_test → resume) is only safe if sicore sends the *same* generation for every follow-up on that run. If sicore ever treats a post-compile repair as a new write-epoch and bumps `generation`, the box 409s the legitimate follow-up with no way to re-pin short of terminating the run. Worth an explicit assertion/test on the sicore side (MR !622), and consider a clearer box-side signal for a generation mismatch than the generic pin error.

kbc/platform/pod/compile_box.py
```diff
@@ -2856,7 +3107,76 @@ async def handle_message(request: web.Request):
         if full_compile:
```
jacoblee-io, line 3112: **[low] `_turn_active` is not reset on a dispatch failure, unlike `_full_compile_pending`.** `_begin_turn` sets `_turn_active=True` before `await run.client.query(text)`; if `query` throws (e.g. a dead SDK subprocess), this `except` resets only `_full_compile_pending` and re-raises, leaving `_turn_active` True with no ResultMessage to clear it. Because the command path now uses `_turn_active` as its concurrency fence, every subsequent command 409s "another authoring command is already running" until the idle watchdog fires (~90s) — which may then re-issue the very directive the consumer already saw fail. Reset `_turn_active` here too, symmetric with `_full_compile_pending`.

kbc/platform/pod/compile_box.py
```diff
@@ -2856,7 +3107,76 @@ async def handle_message(request: web.Request):
...
+        # second command while a turn/batch is active would acknowledge an
+        # action that has not actually been scheduled. Exact replays are handled
+        # above; every distinct concurrent command fails closed and can be
+        # retried as a new product action after the run returns idle.
+        if run._turn_active or run._batch_active:
+            raise CommandRejected("another authoring command is already running", 409)
+        if run._command_context is not None and run._command_context != context:
+            raise CommandRejected("run is pinned to another operation generation", 409)
+        _prepare_command(run, command)
+        text = _render_command(run, command)
```
jacoblee-io, line 3174: **[low, latent] Render is not fail-closed.** `handle_command` catches only `CommandRejected`, but `_render_command` (`strings[action].format(...)`) and `_command_strings` (`FileNotFoundError`) can raise on a locale pack with a missing key or an extra placeholder — surfacing as an uncaught HTTP 500 instead of a typed 4xx. Both shipped packs (en/zh) are complete so there's no live bug, but a future/custom pack would 500. Consider wrapping render in a try that maps a pack error to a clear 500-with-context (or a 400).

kbc/platform/pod/compile_box.py
```diff
@@ -778,6 +807,197 @@ def _capture_brief(run: "CompileRun", text: str) -> bool:
...
+
+    if action == "compile.approve_plan":
+        plan_id = _bounded_string(parameters.get("plan_id"), "parameters.plan_id", required=True, limit=64).lower()
+        if not re.fullmatch(r"[0-9a-f]{64}", plan_id):
+            raise CommandRejected("parameters.plan_id must be a sha256")
+        parameters = {**parameters, "plan_id": plan_id}
+    elif action == "compile.submit_decisions":
+        decisions = parameters.get("decisions")
+        if not isinstance(decisions, list) or not decisions:
+            raise CommandRejected("parameters.decisions must be a non-empty array")
```
jacoblee-io, line 900: **[low] No length cap on `decisions[]` (and `rulings[]` at ~914).** Per-item fields are bounded, but the arrays themselves are only bounded by the shared `client_max_size` default (sized large for source-tar uploads). One pod per user keeps the blast radius small, but a dedicated `len()` cap (e.g. a few hundred) would be cheap defense.

jacoblee-io (dismissed):
Approving — clean, well-tested protocol addition; no merge-blocking bug. I reviewed the box command handling, the `handle_message → _dispatch_authoring_turn` refactor, and the `capability.command` RPC.

Verified sound:
- **Idempotency is race-free**: the `command_id in _accepted_command_ids` check and the `.add()` have no `await` between them, so under asyncio a concurrent duplicate always returns `duplicate` rather than launching a second turn.
- **Legacy message routing is byte-for-byte equivalent**: the `action is None` path through `_dispatch_authoring_turn` reproduces the old inlined flow exactly (`_capture_brief`, incremental/batch routing, the stale-changeset unlink, the mid-batch branch, and `_full_compile_pending`).
- **RPC is correct**: envelope validated then `command` forwarded verbatim; `get ?? adopt` + terminal check + attach; `duplicate` does not flip status to running (no phantom in-flight turn); `compile.scout` runs an ordinary turn that ends in `turn_done → idle`; a box 4xx rejects the RPC (status never flips).
- **`.format()` on the locale templates is injection-safe** (substituted values aren't re-parsed); `approve_plan` hashes the exact deterministic bytes the box wrote; the incremental/resume prepare-guards have no exploitable TOCTOU.

Findings (all non-blocking; inline): [med, confirm vs sicore MR !622] the `(operation_id, generation)` pin can 409 legitimate multi-step follow-ups if the consumer ever bumps generation mid-attempt; [low] `_turn_active` not reset on a dispatch throw; [low] render not fail-closed for a bad locale pack; [low] no array-length cap on `decisions[]`/`rulings[]`.

Stacking/merge order: this sits on the **pre-fix** #401 (commit 961c446a, before the probe/planner fix 5e8f80cc). Merge order #396 → #401 → #402; rebase #402 onto the merged #401 and confirm the probe-timeout + `_compile_model` fixes survive (this branch doesn't touch those lines, so it should be clean). #403 stacks further on this.

CI green (Type Check / KBC Box / Test / Portal Web).

jacoblee-io (dismissed):

jacoblee-io (dismissed):

jacoblee-io (approved):
Re-approving on 926a381f — this is a clean rebase onto the now-merged main (#396/#397/#398/#401 all landed). Verified:

- **#402's own typed-command code is byte-identical** to the version I approved (the old-vs-new delta is entirely the sibling PRs' content absorbed from the new base — nothing in the command-protocol paths changed).
- **#401's probe/planner fix survived the rebase**: `_compile_model()` is present and used by BOTH `_compile_session_opts` and the `_plan_batches` planner, so the model precedence stays consistent — the concern from my earlier stacking note is resolved.
- CI green (Type Check / KBC Box / Test / Portal Web).

My four earlier findings remain open (all non-blocking): [med, confirm vs sicore MR !622] the `(operation_id, generation)` pin can 409 legitimate follow-ups if the consumer bumps generation mid-attempt; [low] `_turn_active` not reset on a dispatch throw; [low] render not fail-closed for a bad locale pack; [low] no array-length cap on `decisions[]`/`rulings[]`. None block merge, but the generation-pin contract is worth confirming on the sicore side before relying on the multi-step command flow in production.

PR #403, opened 2026-07-11
fix(kb): harden typed authoring command delivery

LikiosSedo:
## Summary

- make Runtime and compile-box command acceptance durable and idempotent
- reject command ID reinterpretation and preserve lifecycle fencing
- carry immutable input_revision through start, materialization, and commit
- preserve structured command error code, status, and retriability across Runtime, AgentBox, and Portal boundaries
- keep locale and prompt prose out of routing, idempotency, and state transitions
- fix the Runtime SDK event classifier so stream activity refreshes liveness without clearing a pending tool result
- restore the exact Runtime lifecycle state when a recovered compile box acknowledges a replayed capability.message as duplicate

## Stack

- #401 and #402 are merged
- this PR is restacked directly on main at f1a8c1a7
- paired Sicore authoring delivery is tracked in !626
- broad or unscopable source drift uses explicit full regeneration; incremental commands never silently degrade to full rebuilds

## Validation for current head

- Runtime suite: 217 files passed; 4390 tests passed, 2 skipped
- Runtime TypeScript build passed
- Portal Web: 19 files and 189 tests passed
- KBC compile-box, office-ingest, self-check, red/blue, and mTLS suites passed
- duplicate-ACK and rejected-POST lifecycle regressions are covered in server-capability-message-id.test.ts
- clean restack and diff check against main f1a8c1a7
- GitHub CI pending after the restack

## Earlier integration evidence

The pre-restack stack completed a real full regeneration, deterministic self-check repair, red/blue test, publish, Raw stale/recovery validation, and target-CNI NetworkPolicy proof in sicore-siflow-test. The final replay-state fix in c04f93eb has not yet been deployed and must use a newly built Runtime image for live verification.

## Remaining follow-ups

- KBC Python boxes still return 404 for /api/internal/metrics-snapshot; noisy but not business-blocking
- relay replay/event sequencing and asynchronous start acceptance remain separate protocol follow-ups
- hot credential rotation remains unsupported for an already-connected SDK child


src/gateway/server.ts
```diff
@@ -657,18 +702,43 @@ export async function startRuntime(opts: StartRuntimeOptions): Promise<RuntimeSe
...
+        command: req.command,
+      });
+    } catch (err) {
+      await capabilityRunManager.setStatus(runId, previousStatus);
+      throw err;
+    }
+    // A box-level duplicate after runtime recovery has no future turn_done. Put
+    // the hosting run back exactly where it was before this replay.
+    if (accepted.duplicate) await capabilityRunManager.setStatus(runId, previousStatus);
+    await capabilityRunManager.rememberCommandReceipt(runId, commandId, digest);
```
jacoblee-io, line 741: **[med-low, ordering race] `rememberCommandReceipt` re-persists `rec.status`, which can reorder past the relay's `turn_done → idle` write.** `persist()` snapshots `rec.status` synchronously at call time; on the happy path the box `postJson` returns at turn *dispatch* (not completion), so this runs with `rec.status = "running"`. Meanwhile the SSE relay independently issues `setStatus(runId, "idle")` on `turn_done`. Both are unordered `frontendClient.request()` calls to the consumer's last-write-wins store, and running/idle are both non-terminal so the sticky-terminal guard doesn't apply. If the "running" checkpoint lands after the "idle" one, the durable row is stuck `running` with no future `turn_done`. In-memory `rec` is correct so the watchdog won't notice — but a Runtime restart rehydrates `running` via `recover()` and re-blocks find-or-start until the ~10-min stale reap. The `does not overwrite a fast turn_done` test forces the idle-persist to complete first, so it doesn't cover this reorder. Consider persisting the receipt/message-id checkpoint WITHOUT rewriting a possibly-stale status (decouple the two). Same shape at the `capability.message` `rememberMessageId` call (line 683).

kbc/platform/pod/compile_box.py
```diff
@@ -293,6 +293,10 @@ def __init__(self, run_id: str, workdir: str, round_: int, instruction: str = ""
         # Keep the last full-compile commit replayable across relay restarts.
         # The consumer dedupes it by immutable input revision.
         self._commit_input_replay = False
+        # Consumer-minted turn ids accepted by this live box. The runtime also
+        # checkpoints them; this local set closes the crash window where the box
+        # accepted a turn but the runtime died before persisting its ack.
+        self._message_ids: set[str] = set()
```
jacoblee-io, line 299: **[low] Box-side `_message_ids` (and `_accepted_commands`) are unbounded, unlike the Runtime's 128-entry cap (`MAX_MESSAGE_IDS`/`MAX_COMMAND_RECEIPTS`).** One pod per user and a bounded run lifetime keep the blast radius small, but a long-lived run accumulates every id forever. A matching cap (keep the most recent N) would make the two sides symmetric.

jacoblee-io (dismissed):
Approving — an exceptionally thorough, well-tested hardening pass; no merge-blocking issue. It makes my earlier #402 message-idempotency concern moot by doing it properly (durable, crash-ordered receipts with rollback).

Verified sound (read every line + an independent trace):
- **Durable idempotency**: box `_accepted_commands` (id→digest) + Runtime `commandReceipts` in the run checkpoint; same payload → duplicate ack, id reuse with a different payload → 409. `message_id` dedup mirrors it. Receipts persist only AFTER the box accepts, and roll back in-memory if the checkpoint write fails so the caller can safely retry.
- **Digest asymmetry is not a defect**: Runtime canonicalizes the raw command, the box digests the normalized command, nothing cross-compares them, and a Runtime-receipt hit short-circuits before the box. Stricter-at-Runtime on an extra field is intended fail-closed behavior.
- **Error envelope round-trips intact**: `AgentBoxClient` now throws `RpcResponseError` but still carries `.status`, so `materialize.isBoxAlreadyLive` keeps detecting the /sources 409 reattach; no retriable inversion (409→false, 503→true, explicit body wins); status survives frontend-ws → Portal `runtime-connection`.
- **input_revision pinning** is persisted before spawn/materialize (fail-closed) and the returned revision is validated against the pinned one.
- **SDK event-classifier fix**: assistant-tail StreamEvents no longer clear `tool_pending`, so a long-running tool isn't measured against the short model-idle bound.
- **Fast turn_done race**: the RPCs publish `running` before the box POST and restore `previousStatus` on rejection/duplicate — covered by tests.

Findings, all non-blocking:
- **[med-low] status-persist reorder** in `rememberCommandReceipt`/`rememberMessageId` — inline (server.ts). Self-heals via the stale reap; only bites after a Runtime restart in a narrow window.
- **[low] `_turn_active` not reset on a `run.client.query()` throw** (compile_box.py `_dispatch_authoring_turn`, the `except BaseException` after `run._begin_turn` — around line 3159, not in this diff's hunks so noted here): `_begin_turn` set `_turn_active=True` and the except clears only `_full_compile_pending`, so a submit-time SDK failure leaves the run 409-ing every later command ("another authoring command is already running") until the idle watchdog terminalizes the box. Reset `_turn_active` there too, symmetric with `_full_compile_pending`. (Still open from the #402 review.)
- **[low] box-side id sets unbounded** vs the Runtime 128 cap — inline (compile_box.py).

Still open from #402 (low, not re-raised): render-not-fail-closed for a bad locale pack, no command-array length cap, and the `(operation_id, generation)` pin — please confirm the generation contract against the paired Sicore MR !626 before relying on the multi-step command flow in production.

CI green (Type Check / KBC Box / Test / Portal Web). No stacking dependency — #401/#402 are merged and this is restacked on main.

jacoblee-io (approved):
Re-approving on 37bd97f8 — the status-persist reorder finding (#1) is fixed correctly and at the right depth.

`persist()` now chains per-run full-row writes through a `persistTails` map, so writes for one run land in mutation order. Combined with the two properties already in place — `state` is snapshotted synchronously at call time, and `setStatus` mutates `rec.status` synchronously before enqueueing its persist — the store now converges to the true final status: a late `rememberCommandReceipt`/`rememberMessageId` write either already carries `idle` (relay flipped first) or is serialized *before* the relay's `idle` write, so the phantom-`running`-after-restart window is closed. Verified the chain doesn't poison on a failed write (`previous.catch(() => undefined)`), the `failFast` rejection still reaches its awaiter, and tail entries are deleted on settle (guarded against a later replacement) so there's no unbounded growth. Different runs stay independent. New test covers ordered persists + failure isolation.

The two low items I left open remain optional hardening (not addressed here, which is fine): box-side `_turn_active` not reset on a `query()` throw, and the unbounded box-side id sets. CI green (Type Check / KBC Box / Test / Portal Web).

Deploy note (from the PR body): this replay/serialize work needs a freshly built Runtime image for live verification.

PR #404, opened 2026-07-14
feat(kb): replace hidden question proposals with guided testing

LikiosSedo:
## Summary

- remove the compile-time side effect that silently generated 3-5 test questions
- add explicit owner-triggered `capability.testRecommend`
- create exactly one grounded regression-test recommendation per invocation; later explicit invocations remain allowed
- run a one-shot read-only reviewer restricted to `raw/` and `candidate/`
- require authoritative `close_confirmed: true` before consumers release a test-session lease
- keep unchanged owner chat turns from migrating an inherited legacy draft; changed, compile, incremental, and repair turns remain strict
- bound Box, Runtime, and Sicore consumer timeouts at 180s, 210s, and 240s

Compilation and test authoring are separate. Siclaw does not mutate the question set; paired Sicore MR !626 adopts a recommendation only after its durable execution lease and stable-draft checks succeed.

## Security, lifecycle, and concurrency

- the recommendation session cannot read `authoring/`, `eval/`, `release/`, or arbitrary workspace paths
- no mutation tools are available
- unknown and root-wide path requests fail closed
- AgentBox enforces single-flight recommendation execution and returns 409 for a concurrent request
- Runtime confirms test-session closure only after the box acknowledges teardown or box discovery proves the pod is absent
- an uncertain close RPC fails closed and never returns `close_confirmed`
- ordinary chat skips the post-turn migration gate only when candidate content and exclusions are byte-identical
- candidate edits, explicit compilation, incremental rounds, unfinished repairs, and forced ledger passes still run the full self-check

## Verification

Rebased onto current `main` at `ce2d22e6`.

- all KBC protocol and self-check suites passed, including compile-box, selfcheck, batching, incremental, media, red-blue, office ingest, and mTLS
- targeted Runtime recommendation and close tests: 4 passed
- `npm test`: 219 files, 4,419 passed, 2 skipped
- `npx tsc --noEmit`
- `npm run build`
- `git diff --check`

## Paired delivery

- Sicore MR !626 at `d40ede4e`: durable recommendation lease, stable-draft rehydration and tree fencing, question adoption, and blind test execution
- This PR: Runtime and AgentBox recommendation contract, read-only execution boundary, authoritative session close, and legacy-safe turn gating
- Deploy the Runtime and AgentBox from this PR before deploying the paired Sicore consumer


kbc/platform/pod/compile_box.py
```diff
@@ -3402,6 +3559,29 @@ async def handle_open_test(request: web.Request):
...
 
+async def handle_test_recommendation(request: web.Request):
+    """Explicitly author one raw-grounded test question for a stable draft."""
+    run_id = request.match_info["run_id"]
+    parent = RUNS.get(run_id)
+    if not parent:
+        return web.json_response({"error": "unknown run"}, status=404)
+    if parent._turn_active or parent._batch_active:
+        return web.json_response({"error": "knowledge authoring is still running"}, status=409)
+    if run_id in RECOMMENDATIONS_ACTIVE:
```
jacoblee-io, line 3628: **[low] Single-flight is one-directional.** `handle_test_recommendation` 409s when a parent turn/batch is active, but `/command` and `/message` (`_dispatch_authoring_turn`) don't check `RECOMMENDATIONS_ACTIVE`, so a compile/chat turn can start *after* a recommendation is already running — two live SDK sessions over one pod workspace, and the read-only recommender can read a half-written `candidate/` page. Not corruption or an isolation break (the recommender can't write), but the "one at a time" invariant only holds in one direction. Consider a symmetric guard (reject a turn while a recommendation is in flight, or vice-versa).

jacoblee-io (dismissed):
Approving on 693c6816 — the rebase resolved the conflict (now MERGEABLE on current main incl. the merged #400), the core work survived intact, and I reviewed the three new commits.

Verified:
- **Core intact after rebase**: the hidden `propose_questions` removal and the isolated read-only recommendation session (`_recommendation_path_escape`, `_validate_test_recommendation`, `recommend_test_question`, single-flight `RECOMMENDATIONS_ACTIVE`) are unchanged from the version I reviewed; the security boundary (raw/+candidate/ only, symlink-safe via `resolve()`, fail-closed on unknown tools, reference answer never sent to the blind consumer) still holds.
- **New commits** extend #400's format-migration grandfathering to ordinary owner conversation, consistently:
  - `8d7e8e71` skips the post-turn gate when a turn leaves candidate/EXCLUSIONS byte-identical and isn't a compile/incremental/repair/full/ledger/repairing turn — a pure chat turn no longer conscripts an inherited legacy draft into an OKF migration.
  - `693c6816` page-scopes legacy format debt on an owner edit (same `changed_pages` logic as incremental): changed pages, new violations, non-format findings, repair, and full compiles stay strict.
  - `bdcf99a9` is test hardening.
- **Strictness is correct**: of the six `_begin_turn` call sites, ONLY `_dispatch_authoring_turn` passes `grandfather_legacy_format=not full_compile`; `_start_incremental`, the internal kickoff/verify sites, and the batch planner all default to strict — so internal repair/verify/batch/incremental/full-compile turns never grandfather (no false-negative on format debt).

Low, non-blocking (three carried over, one new): the single-flight asymmetry (inline); Grep secondary args (glob/type) aren't guard-inspected but ripgrep confines them to the search root; the box recommendation timeout is env-overridable above the hardcoded 210s Runtime bound (default ladder 180<210<240 is correct); and `_turn_format_guard` now recomputes `format_violation_keys` over all candidate pages at the start of every non-full owner turn (O(pages), fine for interactive turns).

CI green (Type Check / KBC Box / Test / Portal Web). Paired Sicore MR !626 owns the guided-first-question UI + adoption.

jacoblee-io (approved):
Re-approving on 7ba231b0 — this commit is a genuine isolation hardening of the recommendation session, verified correct with no regression.

- **Catalogue isolation (the real win)**: the session no longer uses the `claude_code` preset + `allowed_tools` alone. Under `bypassPermissions`, `allowed_tools` only skips permission prompts — it does NOT hide the default tool catalogue, so previously only the PreToolUse hook stood between the reviewer and Bash/Write/WebFetch. Now `system_prompt` is a plain reviewer prompt, `tools=[Read,Glob,Grep]`, `disallowed_tools=[Bash,Write,Edit,NotebookEdit,Agent,Task,WebFetch,WebSearch]`, `skills=[]`, `strict_mcp_config=True` — the catalogue itself is minimal, so it's triple defense (catalogue + disallow + hook).
- **Glob guard tightened, not loosened**: I traced the reworked branch — `path=cwd + pattern=raw/**/*.md` (Claude Code's common form) now resolves without a denial loop; `cwd + **/*` still denied; absolute `/work/raw/**` allowed but `/work/raw-secret/**` denied (the `base + "/"` prefix check blocks the sibling-prefix bypass); `/etc/**` denied; `static_base` keeps its empty-wildcard guard. No new escape.
- **New `recommend_guard_deny` prompt** exists in both en/zh with exactly the `{raw_root}/{candidate_root}/{offender}` placeholders the `.format(...)` call passes — no KeyError on the deny path.
- **Error envelope**: `handle_test_recommendation` now returns structured `{code,message,retriable}` (runtime_unavailable / test_session_failed / recommendation_invalid / turn_budget_exhausted / internal_error) with a catch-all 500, consistent with the #403 error-envelope work; `max_turns` 12→20 (env-tunable) with an explicit `error_max_turns → turn_budget_exhausted` (422, retriable) distinction.

Core question-removal + recommendation isolation (raw/+candidate/ only, symlink-safe resolve, evidence validation, reference answer never sent to the blind consumer) remains intact. MERGEABLE on current main; CI green (Type Check / KBC Box / Test / Portal Web).

One low item carried over (not addressed here, fine): the single-flight guard is still one-directional — `/command`/`/message` don't check `RECOMMENDATIONS_ACTIVE`, so a compile turn can start while a recommendation is in flight. Read-only, non-blocking.

PR #405, opened 2026-07-14
fix(channels): run sub-agents foreground in channel (Feishu/DingTalk) sessions + graceful 409

jacoblee-io:
## Problem (channel only)

In a **channel** session (Feishu/DingTalk), when the agent fans out a multi-item `spawn_subagent` batch it defaults to **background/detached**: the turn returns a "已开始…稍后汇总" ack and the real conclusion is delivered later via a best-effort, gateway-in-memory, no-retry path. A channel has no persistent client to receive that async conclusion, so the card shows the ack (not the answer), a follow-up while busy dumped a raw `409`, and feedback buttons landed on the interim card.

Scope note (thanks @LikiosSedo): `spawn_subagent`'s registration is `modes: ["web","channel","cli"]`, so `ToolRegistry.resolve` filters it out of `task`/`api`/`a2a` sessions entirely — those paths never expose the tool, so only **channel** (of the modes that *have* it and *lack* a persistent client) needs this. `web`/`cli` also have it but are persistent (keep background). Long-running API work streams via a2a (`message/stream`), which has no such issue.

## What changed

- **Foreground sub-agents in channel** (`agent-factory.ts`, `tool-registry.ts`, `spawn-subagent.ts`): `foregroundSubagentOnly = mode === "channel"`. `spawn_subagent` hides `run_in_background` and a multi-item batch blocks + runs in parallel + returns the real conclusion. `run_in_background` exec (within-turn perftest server→client) untouched.
- **Live progress on the card** (`lark.ts`): a foreground batch blocks the parent in one tool call, so surface `tool_execution_update` group progress as a ⏳ step, rendered from `details.items` in the channel locale (`子任务执行中… N/M 完成`).
- **Graceful 409** (`lark.ts`): `409 "session busy"` → queue-until-idle retry with backoff, friendly notice on timeout; sanitize all errors; the busy notice is also sent when CardKit creation failed (`sessionBusy` in the text fallback).

## Separate bug fixed here (needed to deploy)

**`fix(migrate)`**: `channel_bindings.created_by` was only in the CREATE TABLE — existing installs errored `Unknown column 'cb.created_by'` on every channel message. Added the missing `safeAlterTable` backfill.

## Testing

- Unit: `ToolRegistry.resolve` per session mode (spawn_subagent present in web/channel/cli, absent in task/api/a2a); foreground flag behavior; lark progress/409/sanitize/card-fallback; migrate. Full suite green, `tsc` clean.
- E2E on siclaw-inner (Feishu): "后台并发三个 subagent, sleep 10/15/20" → ran foreground in parallel (~20s), live "子任务执行中… N/M 完成", conclusion + 👍/👎 on the final card, no 409.

## Deployment

Touches **agentbox** (gate) + **runtime** (channel) + **portal** (migrate). Deploy all three; portal runs the migration on startup.

### DIFF
--- a/src/core/agent-factory.ts
+++ b/src/core/agent-factory.ts
@@ -403,6 +403,12 @@ export async function createSiclawSession(
       memoryDir: memoryEnabled ? memoryDir : undefined,
       sessionEventEmitter: opts?.sessionEventEmitter,
       spawnSubagentExecutor: opts?.spawnSubagentExecutor,
+      // Channel (Feishu/DingTalk) is the one entry that BOTH exposes spawn_subagent (see its
+      // registration `modes`) AND has no persistent client to receive a detached conclusion —
+      // so a detached batch would strand its result. Force sub-agents foreground there. web/cli
+      // also have spawn_subagent but are persistent (keep background); a2a/api/task don't expose
+      // spawn_subagent at all, so they're unaffected either way. run_in_background exec untouched.
+      foregroundSubagentOnly: mode === "channel",
       jobStopExecutor: opts?.jobStopExecutor,
       backgroundExecExecutor: opts?.backgroundExecExecutor,
       taskOutputReader: opts?.taskOutputReader,
--- a/src/gateway/channels/lark.ts
+++ b/src/gateway/channels/lark.ts
@@ -54,6 +54,19 @@ const QUEUE_FULL_NOTICE_BY_LOCALE = {
   "zh-CN": "⏳ 当前会话还有较多消息排队处理中，请稍后再发。",
   "en-US": "⏳ This channel session already has several messages queued. Please try again later.",
 } as const;
+// Session is single-threaded in AgentBox: after waiting out the busy window (queue-until-idle)
+// the session is still occupied (e.g. a long run_in_background exec job). Ask the user to retry
+// rather than clobbering the in-flight work or dumping the raw 409.
+const SESSION_BUSY_NOTICE_BY_LOCALE = {
+  "zh-CN": "⏳ 还在处理上一条，请稍候再发。",
+  "en-US": "⏳ Still working on the previous message — please try again shortly.",
+} as const;
+// Generic failure notice. The raw error can leak internal endpoints / infra to everyone in the
+// chat, so we log the real error and show this instead.
+const AGENT_ERROR_NOTICE_BY_LOCALE = {
+  "zh-CN": "❌ 处理时出错了，请稍后重试。",
+  "en-US": "❌ Something went wrong while processing this. Please try again later.",
+} as const;
 const NEW_SESSION_NOTICE_BY_LOCALE = {
   "zh-CN": "✅ 已开启新会话，此入口中的历史上下文已清空。",
   "en-US": "✅ Started a new session. Previous context for this channel entry has been cleared.",
@@ -928,11 +941,14 @@ async function processQueuedLarkMessage(ctx: QueuedLarkMessageContext): Promise<
   let resultText = "";
   let replyImages: RenderedReplyImage[] = [];
   let agentError: Error | null = null;
+  let sessionBusy = false;
   try {
-    const promptResult = await client.prompt(promptOpts);
+    // queue-until-idle: wait out a busy session instead of dumping a raw 409.
+    const promptResult = await promptWithBusyRetry(client, promptOpts);
     const collected = await collectChannelResponse(client, promptResult.sessionId, "lark", {
       includeImages: true,
       onMilestone: addMilestone,
+      locale,
       // Audit: persist assistant + tool rows so the channel transcript matches
       // web/api/a2a (origin="channel" set on the session above). Tool output on
       // this stream is already sanitized at the agentbox boundary.
@@ -941,16 +957,24 @@ async function processQueuedLarkMessage(ctx: QueuedLarkMessageContext): Promise<
     resultText = collected.text;
     replyImages = collected.images;
   } catch (err) {
-    agentError = err instanceof Error ? err : new Error(String(err));
-    console.error(`[lark] Agent execution failed for session=${sessionId}:`, agentError);
+    if (isSessionBusyError(err)) {
+      // Still busy after the retry window — surface a friendly notice, don't clobber.
+      sessionBusy = true;
+      console.warn(`[lark] Session still busy after retry for session=${sessionId}`);
+    } else {
+      agentError = err instanceof Error ? err : new Error(String(err));
+      console.error(`[lark] Agent execution failed for session=${sessionId}:`, agentError);
+    }
   }
 
-  // Materialize the final reply body. Preserve the agent-like-API-key UX:
-  // a single message to the user — no intermediate tool-call spam.
-  const finalBody = agentError
-    ? `\u274C ${agentError.message.slice(0, 500)}`
-    : (resultText || EMPTY_RESULT_NOTICE_BY_LOCALE[locale]);
-  if (agentError) replyImages = [];
+  // Session-busy and other errors both get a sanitized notice \u2014 the raw error (internal
+  // endpoints, 409 JSON) must never reach the chat; it was logged above.
+  const finalBody = sessionBusy
+    ? SESSION_BUSY_NOTICE_BY_LOCALE[locale]
+    : agentError
+      ? AGENT_ERROR_NOTICE_BY_LOCALE[locale]
+      : (resultText || EMPTY_RESULT_NOTICE_BY_LOCALE[locale]);
+  if (agentError || sessionBusy) replyImages = [];
   const displayBody = stripVisualBlocks(finalBody, { stripSourceBlocks: replyImages.length > 0 })
     || VISUAL_ONLY_NOTICE_BY_LOCALE[locale];
   // The final card is JUST the conclusion — the live step indicator is replaced
@@ -977,9 +1001,10 @@ async function processQueuedLarkMessage(ctx: QueuedLarkMessageContext): Promise<
       // duplicate messages in the group.
       console.warn(`[lark] Card finalize incomplete for cardId=${cardSession.cardId}; user may see stuck placeholder`);
     }
-  } else if (resultText || agentError) {
-    // Card could not be opened; fall back to a plain text reply with
-    // whatever we have (final answer or error) + any accumulated milestones.
+  } else if (resultText || agentError || sessionBusy) {
+    // Card could not be opened; fall back to a plain text reply with whatever we have —
+    // a real answer, an error notice, OR the session-busy notice (sessionBusy carries no
+    // resultText/agentError, so it must be listed explicitly or the busy notice is dropped).
     await replyToLark(larkClient, messageId, finalCardBody);
     deliveredTextChars = finalCardBody.length;
   }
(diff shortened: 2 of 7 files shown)


src/core/agent-factory.ts
```diff
@@ -403,6 +403,11 @@ export async function createSiclawSession(
       memoryDir: memoryEnabled ? memoryDir : undefined,
       sessionEventEmitter: opts?.sessionEventEmitter,
       spawnSubagentExecutor: opts?.spawnSubagentExecutor,
+      // One-shot entry points (channel/a2a/api/cron) have no persistent client to receive a
+      // detached sub-agent conclusion later, and their caller blocks anyway — so force sub-agents
+      // foreground there (and in any unknown mode). web/cli keep background. run_in_background
+      // exec is untouched. See ToolRefs.foregroundSubagentOnly.
+      foregroundSubagentOnly: mode !== "web" && mode !== "cli",
```
LikiosSedo: Forcing API and task sessions into foreground makes their result depend on the caller staying attached, but both /api/v1/run and the task coordinator stop waiting after 300 seconds. A sub-agent child may run for 600 seconds and a group permits much longer, so API callers return 504 and cron can stop consuming and record an unfinished run as successful while AgentBox keeps working. Please align these endpoint timeouts and cancellation semantics with foreground delegation before enabling it here.

src/gateway/channels/lark.ts
```diff
@@ -941,16 +957,24 @@ async function processQueuedLarkMessage(ctx: QueuedLarkMessageContext): Promise<
...
+    }
   }
 
-  // Materialize the final reply body. Preserve the agent-like-API-key UX:
-  // a single message to the user — no intermediate tool-call spam.
-  const finalBody = agentError
-    ? `\u274C ${agentError.message.slice(0, 500)}`
-    : (resultText || EMPTY_RESULT_NOTICE_BY_LOCALE[locale]);
-  if (agentError) replyImages = [];
+  // Session-busy and other errors both get a sanitized notice \u2014 the raw error (internal
```
LikiosSedo, line 970: The new persistent-409 path sets sessionBusy=true while leaving resultText empty and agentError null. When CardKit creation failed, the existing plain-text fallback condition is false, so the promised friendly busy notice is never sent. Please include sessionBusy in that fallback condition and add coverage for the combined card-create-failure plus persistent-409 case.

LikiosSedo (commented):
I found two correctness gaps that should be addressed before merging.

jacoblee-io:
Thanks @LikiosSedo — both addressed in 92c43b08.

**Comment 2 (card-create-failure + persistent-409):** `sessionBusy` carries no `resultText`/`agentError`, so the `else if (resultText || agentError)` fallback skipped the busy notice when CardKit creation failed. Added `sessionBusy` to that condition + a test covering card-create-failure + persistent-409.

**Comment 1 (api/cron foreground vs 300s windows):** aligned the timeouts + cancellation instead of leaving them mismatched:
- **`/api/v1/run`**: timeout is now configurable (`SICLAW_API_RUN_TIMEOUT_MS`, default 600s), and on timeout the session is **aborted** so AgentBox no longer keeps doing orphaned foreground work behind the 504.
- **task-coordinator (cron)**: default timeout raised to 600s (autonomous, no client waiting) and on timeout it now records **`status="failure"`** (previously an abort left the initial `"success"`, recording an unfinished run as successful) and aborts the session. Test added for timeout → failure.

Scope note: only `/api/v1/run` and cron have a fixed wait window. `channel` (SSE, no hard timeout) and `a2a` (`message/send` async task + `message/stream` SSE, no cap) can hold a long foreground turn, so they're unaffected. Genuinely long-running API work should use the a2a streaming endpoint. A streaming mode for `/api/v1/run` is a reasonable follow-up but out of scope here.

src/core/agent-factory.ts
LikiosSedo: The API-key and A2A dispatchers send chat.send with origin=api/a2a but do not send mode. server.ts forwards only params.mode, and session.ts defaults a missing mode to web, so foregroundSubagentOnly is false and multi-item spawn_subagent remains detached on both paths. The task path does send mode=task, but spawn_subagent is registered only for web/channel/cli, so cron cannot see the tool at all. The new tests construct createSpawnSubagentTool directly and therefore bypass both the real dispatch and ToolRegistry mode filtering. As written, this changes channel behavior but does not implement the advertised API/A2A/cron behavior. Please normalize the delivery capability from the actual dispatch metadata and add coverage that resolves the real tool list for each entry path.

src/gateway/task-coordinator.ts
```diff
@@ -299,7 +301,8 @@ export class TaskCoordinator {
       const redactionConfig = buildRedactionConfigForModelConfig(binding.modelConfig);
 
       const abortCtrl = new AbortController();
-      const timer = setTimeout(() => abortCtrl.abort(), this.executionTimeoutMs);
+      let timedOut = false;
+      const timer = setTimeout(() => { timedOut = true; abortCtrl.abort(); }, this.executionTimeoutMs);
```
LikiosSedo: The timeout test makes the consume mock resolve as soon as AbortSignal fires, but the real consumeAgentSse only checks signal.aborted after client.streamEvents yields a data event. streamEvents does not receive the signal, and heartbeat comments are not yielded, so a quiet foreground tool can leave the reader blocked after this timer fires; execution never reaches the finally block that calls abortSession. Please abort the session from the timer callback or make the SSE reader itself abortable, and test against a blocking stream rather than a mock that cooperates with the signal.

LikiosSedo (commented):
I re-traced the actual dispatch and runtime paths. Two gaps remain on the current head.

src/core/agent-factory.ts
LikiosSedo: Correction after tracing the actual API-key streaming path: that path is backed by a long-lived SSE bridge, so the 300s buffered /run timeout is not the blocker for streaming requests. The remaining correctness issue is the dispatch classification: chat.send carries origin=api but no mode, Runtime forwards an undefined mode, and the session defaults to web, so the new foreground gate is bypassed. I posted the full API/A2A/task finding against the latest head in discussion_r3578549442.

jacoblee-io:
You're right on both — re-scoped in f9f8ede0.

**Root cause I'd missed:** `spawn_subagent`'s registration is `modes: ["web","channel","cli"]`, and `ToolRegistry.resolve` filters on that, so `task`/`api`/`a2a` sessions never get the tool. (The dispatchers *do* send `mode` and `server.ts` forwards it, so it's the `modes` allow-list — not the mode default — that gates them out.) So the "one-shot foreground" only ever affected **channel**; foregrounding api/task was inert, and my `/run` + cron timeout work was addressing a batch those paths can't spawn — and, as you noted, the cron abort was unreliable anyway (a quiet stream never unblocks `consumeAgentSse`, so the `finally`/`abortSession` was never reached).

Changes:
- `foregroundSubagentOnly = (mode === "channel")` — the one mode that both exposes `spawn_subagent` and lacks a persistent client for a detached conclusion.
- **Reverted** the `/api/v1/run` + task-coordinator timeout/cancellation changes to baseline (out of scope; long-running API work uses a2a `message/stream`, which is unaffected).
- Tests now resolve the **real** tool list via `ToolRegistry.resolve` per session mode (present in web/channel/cli, absent in task/api/a2a) instead of hand-constructing the tool.

The card-create-failure + persistent-409 fix from the first round stays. If cron's pre-existing false-success-on-timeout is worth fixing, I'll do it as a focused PR with the abort-from-timer approach + a blocking-stream test you suggested.

LikiosSedo (commented):
Reviewing this strictly against the narrowed channel/Feishu scope, I found no blocking correctness issues on the current head.

The channel-only gate is appropriately narrow: channel sessions keep `spawn_subagent` available but force grouped delegation into the foreground, while background `exec` remains unchanged. The group still executes children in parallel, and the parent turn now returns the actual conclusion instead of an interim detached acknowledgement. The Lark path also covers live N/M progress, bounded 409 retry, sanitized failure output, and the CardKit-create-failure text fallback. The legacy `channel_bindings.created_by` migration is directly relevant to making the channel path deployable on existing databases.

I re-ran `tsc --noEmit`, the focused spawn/Lark/migration tests (132 tests), `git diff --check`, and a merge-tree check; all passed. The current GitHub checks are green as well.

Non-blocking cleanup: the comment above `backgroundAllowed` still says `channel/a2a/api/cron`; after the scope reduction it should say `channel` only.

src/tools/workflow/spawn-subagent.ts
```diff
@@ -127,12 +129,16 @@ export function createSpawnSubagentTool(
   refs: ToolRefs,
   executor = refs.spawnSubagentExecutor,
 ): ToolDefinition {
+  // Background (detached) delegation is allowed only when the global switch is on AND this entry
+  // point can receive an async conclusion. On channel/a2a/api/cron (foregroundSubagentOnly) every
```
LikiosSedo, line 133: Nit (non-blocking): this comment still says "On channel/a2a/api/cron (foregroundSubagentOnly)" — after the re-scope, only `channel` sets `foregroundSubagentOnly`, and `a2a`/`api`/`cron` sessions never resolve this tool at all (filtered by the `modes` allow-list below). Worth updating so the comment matches the shipped gate and doesn't re-plant the old assumption.

LikiosSedo (commented):
Re-verified the re-scoped implementation at 63125960 — the channel-only scope holds, and I need to correct my earlier claim about the dispatchers.

**Correction:** every dispatch site in this repo does stamp `mode`: `chat-gateway.ts` sends `mode: "api"`, `a2a-gateway.ts` sends `mode: "a2a"` (and `mode` predates the `origin` audit field there — I had misread `origin`-only dispatch from a different call chain), and both channel adapters send `mode: "channel"` (`lark.ts:932`, `dingtalk.ts:349`). So it is the `modes: ["web","channel","cli"]` allow-list in `spawn_subagent`'s registration — enforced by `ToolRegistry.resolve` — that keeps `task`/`api`/`a2a` sessions from ever seeing the tool. The scope note in the description is accurate.

Also confirmed on this head:
- The retraction is complete — the diff no longer touches the buffered run endpoint or the task coordinator.
- The busy notice now reaches the plain-text fallback (`sessionBusy` included at `lark.ts:1004`) with a test covering card-create-failure + persistent 409.
- Registry tests resolve through the real `ToolRegistry`, and the gate comment in `agent-factory.ts` matches the shipped behavior.

One non-blocking doc nit inline. LGTM otherwise.

jacoblee-io:
Thanks for the LGTM. Fixed the doc nit too — the `createSpawnSubagentTool` comment now states only `channel` sets `foregroundSubagentOnly` and that a2a/api/task never resolve the tool (per the `modes` allow-list), so it no longer re-plants the old assumption. Folded into the feature commit (history squashed to 2 commits: the channel change + the `channel_bindings.created_by` migration backfill).

LikiosSedo (approved):
Approving — re-verified at 63125960: the channel-only gate is closed-loop (both channel adapters dispatch mode: "channel"; task/api/a2a never resolve spawn_subagent via the modes allow-list), the earlier timeout changes are fully retracted, the sessionBusy text fallback is covered by tests, and CI is green. The remaining inline nit is a doc comment only, fine to fix in a follow-up.

PR #408, opened 2026-07-15
fix(kbc): pin read-only consumer sessions to the read tool set

LikiosSedo:
Follow-up to #404 (which hardened the recommendation session). Applies the same tool-isolation to the two other **read-only** box sessions.

## Problem

The **test-session** driver (`compile_box.py test_session_driver`) and the **red/blue reviewer** (`engine.py`) run under `permission_mode="bypassPermissions"` with `tools` unset — so the full Claude Code preset (Bash, Write, Edit, WebFetch, WebSearch, …) is in the model context — and only `allowed_tools=["Read","Glob","Grep"]`. Under bypass, the allowed-vs-approved split is moot, so the read-only contract rested **entirely on the PreToolUse path hook**, which only confines *path-bearing* tools:

- `Bash` — a shell that can read outside the pinned snapshot; `_test_path_escape` never inspects shell commands. The test-session guard is `_make_test_path_guard` (no `deny_bash`), unlike the compile guard.
- `WebFetch` / `WebSearch` — no path key, so the hook passes them; a consumer could answer from the web instead of the KB, breaking the closed-book property the test **pass/fail signal (which gates publishing)** depends on.
- `Write` to an in-snapshot path — passes the path check, letting the consumer mutate its own immutable test evidence.

## Change

Mirror the recommendation-session hardening from #404 on both read-only sessions:

- `tools = ["Read","Glob","Grep"]` — removes everything else from context
- `disallowed_tools = [Bash, Write, Edit, NotebookEdit, Agent, Task, WebFetch, WebSearch]` — belt-and-suspenders under bypass
- `strict_mcp_config = True`, `skills = []`

**Compile is intentionally not touched** — it already denies Bash via `_make_compile_path_guard`, its output is provenance-gated (web-sourced claims fail selfcheck), and it legitimately needs `Write` + the MCP signal tools.

## Scope of value (honest)

This is **defense-in-depth + eval integrity**, not closing an active breach — the pod is already a single-tenant, ephemeral sandbox with a NetworkPolicy (Runtime-ingress-only) and stripped credentials. The concrete wins: keep the publish-gating test signal snapshot-confined and closed-book, and stop a wandering consumer from burning its turn budget on tools it should not use.

## Tests

- Extended `test_test_session_driver_readonly` to assert `tools` / `strict_mcp_config` / `skills` / `disallowed_tools` on the constructed opts.
- Full kbc suite green: `test_compile_box`, `test_redblue`, `test_selfcheck`, `test_office_ingest`, `test_mtls_auth`, `test_batching`, `test_incremental`, `test_mediaverify`. `py_compile` clean.
- Engine change is symmetric to the asserted test-session one and exercised (construction) by `test_redblue`; the redblue harness abstracts above `ClaudeAgentOptions`, so it carries no dedicated opts assertion.

Box-only; no runtime/web/sicore changes.

### DIFF
--- a/kbc/platform/pod/compile_box.py
+++ b/kbc/platform/pod/compile_box.py
@@ -2781,6 +2781,15 @@ async def _run_wrapper(run: CompileRun):
 # Default tool whitelist for a read-only kb-test session, used when the runtime
 # profile declares none. Read-only by construction: cannot mutate the snapshot.
 DEFAULT_TEST_ALLOWED_TOOLS = ["Read", "Glob", "Grep"]
+# Tools removed from a read-only consumer session's context entirely (not merely
+# left un-approved). Under bypassPermissions the allowed_tools split is moot, so
+# the read-only contract must be enforced by pinning `tools` to the read set and
+# denying the rest — the path hook only confines path-bearing tools, so Bash (a
+# shell that reads outside the snapshot) and WebFetch/WebSearch (breaking the
+# closed-book property the test score depends on) would otherwise slip through.
+READONLY_CONSUMER_DISALLOWED_TOOLS = [
+    "Bash", "Write", "Edit", "NotebookEdit", "Agent", "Task", "WebFetch", "WebSearch",
+]
 
 # Tool-input keys that name a filesystem path (Read.file_path, Glob/Grep.path).
 _TEST_PATH_KEYS = ("file_path", "path", "notebook_path")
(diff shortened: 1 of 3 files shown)


jacoblee-io (approved):
Approving — clean, correct, honestly-scoped follow-up to #404; verified completeness and no regression.

- **Root cause is right**: under `bypassPermissions`, `allowed_tools` doesn't hide the Claude Code catalogue, so the read-only contract rested entirely on the PreToolUse path hook — which only confines path-bearing tools. `Bash` (reads outside the snapshot), `WebFetch`/`WebSearch` (break the closed-book property the publish-gating test signal depends on), and `Write` to an in-snapshot path were all reachable. Pinning `tools=[Read,Glob,Grep]` + `disallowed_tools=[...]` + `strict_mcp_config` + `skills=[]` closes that at the catalogue layer, mirroring #404.
- **Completeness verified**: I grepped every `bypassPermissions` session. The three read-only consumers are now all hardened — recommendation (#404), and test-session + red/blue reviewer (this PR). The two Write-bearing sessions (compile `_compile_session_opts`, planner `_plan_batches`) are correctly left untouched: they need Write + MCP signal tools, already deny Bash via `_make_compile_path_guard`, and are provenance-gated.
- **No functional regression**: both hardened sessions are pure read consumers (Read/Glob/Grep only), so removing Bash/Write/Web/Agent/Task/skills doesn't remove anything they legitimately use. The test-session keeps the `claude_code` preset persona but with a pinned tool set (fine — the preset is prompt text, `tools` is the actual catalogue).

One out-of-scope observation (no action needed): the planner (`_plan_batches`, ~line 2261) and the compile session still run under `bypassPermissions` with `tools` unset, so `WebFetch`/`WebSearch` remain in their catalogue and the path hook passes them (it only denies `Bash`). Compile output is provenance-gated (web-sourced claims fail selfcheck) and the pod is single-tenant with stripped creds, so this is an accepted pre-existing exposure — flagging only so the closed-book reasoning is on record: it now holds for the test/PK sessions you pinned, but compile/planner can still reach the web (they can't take the exact `tools=[Read,Glob,Grep]` treatment since they need Write + the MCP signal tools).

Test coverage asserts `tools`/`strict_mcp_config`/`skills`/`disallowed_tools` on the constructed opts; the engine change is symmetric. Box-only, no runtime/sicore changes. CI green (Type Check / KBC Box / Test / Portal Web), MERGEABLE.

PR #409, opened 2026-07-15
fix(kbc): block credential-shaped candidate output

LikiosSedo:
## Summary

- add a narrow Layer-1 lint for high-confidence credentials in generated candidate pages
- report only credential kind and line number so selfcheck output cannot echo the matched value
- keep ordinary internal names, phone numbers, IPs, URLs, and explicit placeholders unaffected
- block convergence through the existing selfcheck contract without mutating raw sources

## Why

The default authoring policy should preserve useful internal technical detail while preventing obvious private keys and tokens from entering a publishable candidate tree. This is intentionally separate from broad external-content redaction.

## Validation

- `uv run --python 3.11 --with-requirements requirements.txt python test_selfcheck.py`
- all KBC pod suites: compile box, office ingest, selfcheck, red/blue, mTLS auth, batching, incremental, and media verification
- `python -m py_compile selfcheck.py test_selfcheck.py`


chent1996 (approved):
## Review

Approving. Reviewed against latest `main` — the branch is behind but touches only `kbc/platform/pod/` files that `main` hasn't moved (no conflicts, clean apply).

### What I verified

- **The non-echo guarantee is real and pinned.** Findings carry only `kind` + line number; the test serializes the whole report AND the repair prompt and asserts none of the 11 secret shapes appear in either — so `SELFCHECK.json` and the injected repair message genuinely cannot become a second leak channel. This is the load-bearing design decision and it's tested end-to-end.
- **Placeholder discipline is precise.** `<...>`, `${...}`, `[REDACTED]`, `***`, and prefix words (`example`/`your-`/`dummy`/`changeme`/...) are exempt, and the test deliberately pins that a mid-value "test" (e.g. `sk-A1b2testC3d4...`) does NOT exempt — the word-boundary anchor in the placeholder regex is doing its job.
- **The repair loop converges.** The repair instruction says "replace only the secret value with `[REDACTED]`", and `[REDACTED]` is itself in the placeholder allowlist — so a repaired page lints green on the next pass. Nice closed loop.
- **Span dedup** prevents overlapping patterns from double-reporting, and the pattern order puts the specific label (Anthropic) before the generic (`sk-`).
- **Scope discipline holds**: the negative test keeps internal hosts, IPs, phone numbers, and placeholders publishable — this stays a narrow high-confidence credential lint, not creeping redaction. Fail-open selfcheck behavior and the never-stuck invariant are untouched; blocking happens through the existing convergence contract. README updated to match.

### Trivia (non-blocking)

- `Bearer\s+…` can match across a newline (`\s` includes `\n`), so a line ending in "Bearer" followed by an unrelated 20+-char token on the next line could false-positive. Cosmetic; the repair loop handles a rare FP gracefully.
- The `environment credential` alternation is IGNORECASE, so lowercase `password=…`/`token=…` in config-example prose will flag. That reads as intended policy (block by default, owner repairs), just noting it's deliberate.

Box-image note per the CLAUDE.md matrix: this changes box behavior, so the lint only takes effect after a `kbc-compile-box` image rebuild + redeploy.

PR #410, opened 2026-07-15
fix(core): upgrade pi SDK to 0.80.7

LikiosSedo:
## Summary

Upgrade the coordinated pi SDK packages from 0.78.1 to 0.80.7 so Codex SSE response-header waits use the configured HTTP timeout instead of the fixed ten-second cutoff in the currently pinned transport.

### Problem

Codex subscription-backed requests can fail with `Codex SSE response headers timed out after 10000ms` even when the upstream is only slow to return headers. The failure terminates the model turn as `MODEL_ERROR`.

### Solution

Upgrade `pi-agent-core`, `pi-coding-agent`, `pi-ai`, and `pi-tui` as one compatible set. Pi 0.80.7 includes the upstream SSE timeout fix together with newer WebSocket/SSE transport hardening. This PR intentionally keeps Siclaw's timeout and retry policy unchanged; product-level overrides can be considered separately from runtime data.

Upstream fix: https://github.com/earendil-works/pi/commit/54113731b2e70ceb61ea1948fdfe83395ff2fd90

## Test Plan

- [x] `npm ci`
- [x] `npx tsc --noEmit`
- [x] `npm run build`
- [x] `npm test` — 219 files passed, 4419 tests passed, 2 skipped
- [x] Confirmed all pi packages resolve to 0.80.7 with no mixed old versions
- [x] Compared `npm audit`: no new advisories; one previous high-severity pi advisory is removed
- [ ] Live Codex subscription smoke after the updated AgentBox image is built


chent1996 (approved):
## Review

Approving. Version-bump-only PR (`package.json` + lockfile), reviewed against latest `main` — behind but with zero file overlap with the commits it's missing (clean apply).

### What I verified

- **The pi set moves as one.** All pi packages resolve to 0.80.7 in the lock (including transitive `pi-ai`) — no mixed old versions. `pi-coding-agent` stays exact-pinned (no caret), consistent with the previous pinning style; the others keep carets as before.
- **Nothing else in `package.json` moved.** The lockfile churn beyond pi is transitive (`@aws-sdk/*` / `@smithy/*` pulled by pi's bedrock transport) — normal for a lock regen within existing ranges.
- **The claimed fix is upstream and linked** (SSE response-header wait honoring the configured HTTP timeout instead of the fixed 10s cutoff), and this PR deliberately changes no Siclaw-side timeout/retry policy — the right separation.
- Full suite green (4419 tests), `tsc` + build clean, and `npm audit` improves (one prior high-severity pi advisory removed).

### Pre-release note (not a merge blocker)

The unchecked box in the test plan is the important one: the engine runs inside the AgentBox image, so the fix only reaches spawned boxes after an **agentbox image rebuild + rollout**, and the live Codex subscription smoke should gate the release (0.78 → 0.80 is two minor versions of the agent engine; the suite is green but the failing scenario — slow upstream headers on a Codex subscription — only reproduces live). Please run that smoke before tagging a release image.

PR #411, opened 2026-07-15
fix(gateway): make capability cancellation confirmed and race-safe

LikiosSedo:
## Summary

Make `capability.cancel` a confirmed, idempotent execution stop that fences the Runtime run before requesting AgentBox termination. This is the Siclaw half of the knowledge-compilation Stop flow; the paired Sicore change will own writer fencing, draft rollback, and cleanup-pending state.

### Problem

The existing handler stopped the AgentBox before terminalizing its run, swallowed non-404 stop failures, and could not terminalize a run that only existed in the consumer store after a Runtime restart. Callers therefore received success even when execution cleanup was uncertain, while destructive adoption could reattach a relay during cancellation.

### Solution

- terminalize known or store-adopted runs before stopping the AgentBox
- suppress the normal relay-reattach callback when adoption is only for cancellation
- propagate uncertain stop failures so Sicore can record cleanup as pending and retry
- return `stop_confirmed: true` only after the idempotent AgentBox stop succeeds
- keep the existing `done` Runtime terminal state and leave domain-level `cancelled` semantics to Sicore

This PR is independent of #409 and does not change the compile-box image or protocol.

## Test Plan

- [x] `npx vitest run src/gateway/server-capability-cancel.test.ts src/gateway/capability/run-manager.test.ts`
- [x] `npx tsc --noEmit`
- [x] `npm test` (220 files; 4423 passed, 2 skipped)

### DIFF
--- a/src/gateway/capability/contract.ts
+++ b/src/gateway/capability/contract.ts
@@ -163,6 +163,17 @@ export interface CapabilityCancelRequest {
   run_id: string;
 }
 
+export interface CapabilityCancelResponse {
+  ok: true;
+  run_id: string;
+  /**
+   * True only after the AgentBox termination request was accepted (or the box
+   * was already absent). Runtime lifecycle terminalization is initiated first;
+   * domain consumers must still fence their own writers before cancellation.
+   */
+  stop_confirmed: true;
+}
+
 // ---- Test session (read-only consumer probe over a pinned draft snapshot) ----
 
 export interface CapabilityTestStartRequest {
--- a/src/gateway/capability/run-manager.ts
+++ b/src/gateway/capability/run-manager.ts
@@ -219,7 +219,10 @@ export class CapabilityRunManager {
    * undefined when the store doesn't know the run or it already ended, so the
    * caller can refuse instead of spawning an unmanaged box.
    */
-  async adopt(runId: string): Promise<CapabilityRunRecord | undefined> {
+  async adopt(
+    runId: string,
+    opts: { notifyOnAdopt?: boolean } = {},
+  ): Promise<CapabilityRunRecord | undefined> {
     const existing = this.runs.get(runId);
     if (existing) return existing;
     try {
@@ -244,7 +247,10 @@ export class CapabilityRunManager {
       };
       this.runs.set(rec.runId, rec);
       console.log(`[capability] adopted run ${runId} from the consumer store (missed by boot recovery)`);
-      this.onAdopt?.(rec);
+      // Destructive control paths may adopt only to terminalize a run. They
+      // must not trigger relay reattachment, which could race cancellation and
+      // recreate execution for a run that is being stopped.
+      if (opts.notifyOnAdopt !== false) this.onAdopt?.(rec);
       return rec;
     } catch (err) {
       console.warn(`[capability] adopt(${runId}) failed: ${err instanceof Error ? err.message : String(err)}`);
--- a/src/gateway/server-capability-cancel.test.ts
+++ b/src/gateway/server-capability-cancel.test.ts
@@ -1,0 +1,201 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+const box = vi.hoisted(() => ({
+  posts: vi.fn(async () => ({ ok: true })),
+}));
+
+vi.mock("./chat-repo.js", () => ({
+  ensureChatSession: vi.fn(async () => {}),
+  appendMessage: vi.fn(async () => "msg-id"),
+  incrementMessageCount: vi.fn(async () => {}),
+}));
+vi.mock("./output-redactor.js", () => ({ buildRedactionConfigForModelConfig: vi.fn(() => ({})) }));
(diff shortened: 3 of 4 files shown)


chent1996 (commented):
## Review

Solid hardening and a strict improvement over `main` — terminal-before-stop, adopt-without-reattach for the destructive path, stop failures propagated instead of swallowed, and `stop_confirmed` only after an idempotent stop. Reviewed against latest `main` (branch is based on it). The four new tests pin exactly the right semantics: terminalize-then-stop + later-message rejection, uncertain-cleanup surfaced + idempotent retry converging, store-only run terminalized without reattach/respawn, and blank-id fail-fast.

I verified the retry path end-to-end: after a failed stop, the run record is gone from the live map (terminal persist succeeded), a retry's `adopt` returns `undefined` for the now-terminal store row, so the retry skips straight to the idempotent `stop` — no double-terminalization, converges to `stop_confirmed`. The `notifyOnAdopt:false` seam is the right shape for "adopt only to terminalize".

Leaving this as a **comment** rather than an approval for one residual race that sits on the PR's own core claim:

### Finding: an in-flight `capability.message` can respawn the box after cancel (TOCTOU)

The fence comment says the terminal mark "prevents a concurrent message/command from entering while K8s processes the pod deletion" — but the message handler checks `isTerminalCapabilityStatus` **once at entry**, and `ensureCapabilitySession` never re-checks before `capabilityBoxClient → getOrCreate`:

1. `capability.message` passes the terminal check (run still `running`), enters `ensureCapabilitySession`, and is parked at an await before `getOrCreate`.
2. `capability.cancel` runs: `endRun(runId, "done")` + `agentBoxManager.stop(runId)` both succeed → returns `stop_confirmed: true`.
3. The in-flight message resumes: `getOrCreate` **spawns a fresh pod** for the terminalized run.

Consequences: the watchdog reaps *runs*, not boxes — the run is gone from the live map, so nothing ever reaps the new pod; and since cancel already returned `stop_confirmed: true`, the consumer has no reason to retry. Orphan pod with no self-healing, on an SRE product's own cluster. The trigger is not exotic either — it's "user hits Stop while a message is being processed", which is precisely the Stop flow this PR supports.

The fix is small and local: re-check `capabilityRunManager.get(runId)` is present and non-terminal inside `ensureCapabilitySession`'s async setup, immediately before `capabilityBoxClient(...)` (and optionally once more after the box is acquired, self-stopping on a terminal verdict). That closes the practical window; the current tests would keep passing, and one new test (cancel landing between the message-handler check and `getOrCreate`) would pin it.

### Minor (non-blocking)

- `capability.cancel` for a run id the store has never seen returns `stop_confirmed: true` (adopt → undefined, stop of an absent pod → success). Defensible idempotent-stop semantics — just confirm the Sicore side doesn't interpret that as "the run existed and was cleaned up".
- Deploy ordering: propagating stop failures means an old consumer that expected unconditional `{ok}` now sees an RPC error on the (rare) failed-stop path. The paired Sicore change owns cleanup-pending; worth one line in the deploy notes about shipping order.
- `contract.ts`'s `CapabilityCancelResponse` doc is good — but the `stop_confirmed: true` literal type means the field can never be `false`; if a future "accepted but unconfirmed" state is ever needed this becomes a wire change. Fine today, just flagging the type is intentionally narrow.

Happy to approve once the `ensureCapabilitySession` re-check (or an argued acceptance of the window) is in.

jacoblee-io (approved):
Approving — small, correct, well-tested hardening; traced every path.

- **Terminalize-before-stop fences correctly**: `endRun(runId,"done")` marks the run terminal before `stop()`, so a concurrent `capability.message`/`command` (which reject on `isTerminalCapabilityStatus`) can't reattach while K8s deletes the pod. This inverts the old order that left a reattach window.
- **`adopt(runId, {notifyOnAdopt:false})` is backward-compatible**: only the cancel path suppresses `onAdopt`; message/command/testRecommend still call `adopt(runId)` and reattach as before. Suppressing it here prevents a relay from racing the cancel and respawning execution.
- **Stop-failure propagation is sound**: I read `endRun` — it never throws (persist success drops from the map; persist failure is swallowed and left to reconcile), so cancel always reaches `stop()`, and its only throw is a genuine non-404 stop failure. `stop_confirmed:true` is returned only after the idempotent stop succeeds. Retry is safe: terminal is sticky (repeat `endRun` no-ops), `stop()` is idempotent, and once the run is deleted `adopt` refuses the now-terminal store row so the retry goes straight to `stop`.
- Tests cover the four paths (terminalize-before-stop + reject later messages, uncertain-cleanup + idempotent retry, store-only adopt without reattach/respawn, blank-id guard). CI green (Type Check / Test / Portal Web), MERGEABLE.

One coordination note (not blocking): this is a consumer contract change — `capability.cancel` previously always resolved success; it now throws on an uncertain (non-404) box-stop failure, and adds `stop_confirmed` to the success shape. The success shape stays backward-compatible, but the paired Sicore change must handle the new throw (record cleanup-pending + retry) as the PR body describes. Worth confirming Sicore ships that handling before/with this.

PR #412, opened 2026-07-15
feat(kbc): clean-answer contract for test sessions + step frames

LikiosSedo:
## What

The KB test-session consumer narrated its retrieval process inside the answer text ("let me read the index first…"), which (a) polluted saved reference answers with session-specific noise, (b) gave the auto-judge style to compare instead of facts, and (c) read poorly. This PR makes the final text be the answer itself and surfaces the process separately.

- **prompts/{zh,en}/test_role.md** — new answer-discipline section: the answer text IS the answer (it must stand alone as a reference answer); conclusion first, at most a few key points; never narrate the process; don't restate unrelated page content. The `SOURCES:` trailing-line protocol is unchanged. The red-blue blue team reads the same pack (selfcheck.TEST_ROLE), so PK answers get the same cleanup for free.
- **compile_box.py `_emit_message`** — TEST sessions now emit one display-ready `step` frame per consumer tool call (`{type:"step", text:"读「page」"}`; Read → page name, Glob/Grep → truncated pattern, unknown → generic), localized in the box via `_loc`. This feeds the collapsible retrieval trace in the test UI (sicore MR, stacked on !626). Compile runs are untouched — their live contract stays `log`/`turn_done`.

## Why no relay/portal change

`src/gateway/capability/test-relay.ts` forwards `evt.type` → `kind` verbatim, and the portal's SSE handler passes any non-empty kind through — verified both. Old frontends ignore unknown kinds; a new frontend against an old box simply shows no trace. Rollout order is unconstrained (box image rebuild required for the frames to appear).

## Tests

- `test_test_session_step_frames`: zh/en label mapping (dir + .md stripping, pattern truncation, unknown-tool fallback), exactly one `log` per text block, and the negative case — a compile run with the same blocks emits no `step` frames.
- `test_prompt_packs_locale` extended: both packs carry the answer-discipline section and keep `SOURCES:`.
- Full suites green: `test_compile_box.py` (51 checks), `test_redblue.py`, `test_selfcheck.py`.


chent1996 (approved):
## Review

Small, clean, and well-scoped. Reviewed against latest `main` (the branch is based on current `main`, no rebase needed). Approving — the findings below are non-blocking; #1 is a one-line doc fix I'd like to see land (here or immediately after).

### What I verified

- **The load-bearing "no relay/portal change needed" claim is accurate.** `test-relay.ts` forwards `kind: evt.type` verbatim with no enum or filter, `text: evt.text ?? evt.error ?? ""` carries the label, and the new `step` kind can't collide with the `end` termination check. Additive vocabulary; old frontends ignore unknown kinds; rollout order genuinely unconstrained (box image rebuild required for frames to appear, as noted — matches the CLAUDE.md matrix row's deploy note).
- **Compile-run isolation.** `step` frames are gated on `isinstance(run, TestRun)` — `TestRun` is a parallel class (not a subclass of `CompileRun`), so the gate is exact — and the negative test pins it: the same blocks on a `CompileRun` emit no `step` frames. The compile live contract (`log`/`turn_done`) is untouched.
- **Localization placement is right.** Labels are localized in the box via `_loc(run, en, zh)` — the box is the only component that knows the run's locale, so shipping display-ready text through a verbatim relay avoids leaking locale into the runtime. Argument order is consistent across all call sites; the zh/en tests pin both paths (dir + `.md` stripping, unknown-tool fallback).
- **Prompt discipline change is safe.** The `SOURCES:` trailing-line protocol survives the edit in both packs (asserted in `test_prompt_packs_locale`), and the red-blue team reads the same pack, so PK answers get the cleanup for free. Only `TextBlock`s feed `_turn_text`, so step labels can't leak into the persisted turn text.

### Findings (non-blocking)

1. **`contract.ts` kind-vocabulary comment not updated.** `CapabilityEventFrame.payload.kind` is documented as `(session | log | turn_done | error | end)` — this PR adds `step` to that vocabulary without touching the comment. Nothing breaks (the field is `string`, the relay is pass-through), but "contract.ts IS the wire" is this platform's own discipline, and the Go consumer mirrors the vocabulary from exactly that comment. Please add `step` to it — one line, here or in an immediate follow-up.
2. The PR description lists "pattern truncation" among the tested label mappings, but the test has no >30-char pattern case. Either add the case or trim the claim.
3. `Read` page labels aren't length-capped (Glob/Grep patterns are truncated at 30) — a pathological filename yields an unbounded step label. Cosmetic.

### Verdict

Approved. Clean-answer contract is the right separation (facts in the answer, process in the trace), the event-vocabulary extension is genuinely additive and verified pass-through end-to-end within this repo, and the compile path is provably untouched.

PR #413, opened 2026-07-17
feat: siclaw-native agent-to-agent delegation

jacoblee-io:
## Summary

A coordinator agent can route a task to a specialist **peer** agent, entirely within siclaw's own gateway/agentbox — **no external orchestration dependency**. The peer runs autonomously (its own capabilities/persona/model), its session is persisted and openable, and the coordinator keeps oversight.

## What's included

**Roster & discovery**
- `agent_delegates` table + Portal RPC/API — the coordinator's authorized peers (membership = authorization).
- `list_delegates`: a searchable coverage lookup (query by cluster/host/node; browse shows counts only, never dumping every binding).
- Agent types **SRE / Coordinator / Custom** — built-in types lock the capability set + persona. The coordinator persona routes strictly by delegate bindings and asks the user when the target is ambiguous.

**Transport** (`gateway/delegate-api.ts`)
- `POST /api/internal/delegate` — mTLS cert identity = coordinator; peer re-validated against the roster; runs under the peer's own model binding. Streams the peer's events back live and returns a final result. Propagates **Stop** (client disconnect → abort the peer turn). `GET /api/internal/delegates` returns the roster.

**Autonomous peer**
- A delegated agent runs under **its own config** — delegation no longer forces the read-only tier (`readOnly` is explicit opt-in only).
- Its sub-agents run **foreground** (delegation is synchronous), so one turn carries the complete result instead of the coordinator polling.
- The peer's final narrative is captured as the result when it doesn't call `report_findings`.

**Openable, persisted, reusable peer sessions**
- The peer session is persisted as a child of the coordinator session (lineage + parent-link auth) — openable and analyzable later.
- The coordinator reuses a session only for a genuine follow-up that continues the same investigation; a different subsystem/topic starts fresh (same resource is not a reason to reuse), with recency-bounded reuse enforced gateway-side.

**UI** (portal-web)
- Agent Type picker + a **Delegates** roster tab (selectable agent cards: type badge, PROD/DEV, model, description, search).
- Delegate card: compact full-width row; expand shows the full delegated prompt; a header icon button opens the peer's full session.
- `PeerSessionView`: a floating slide-over that overlays the main window (dimmed backdrop, click-outside to close) and renders the peer's **full session** via the main `PilotArea` (plan panel, nested sub-agents, tool cards) — live, with "thinking" preserved.

## Testing

- Backend: `npm test` — 225 files / 4454 passing (2 skipped). New suites: delegate-to-agent, list-delegates, report-findings, request-input, delegation read-only tier; updated fixtures.
- Frontend: `portal-web` vitest — 189 passing.
- `npx tsc --noEmit` clean (backend + portal-web).
- E2E-verified on the siclaw-inner test environment.

## Deploy notes

Touches all three images: **runtime** (gateway delegate-api, capability resolution), **agentbox** (tools, session, agent-factory, persona), **portal** (portal-web). Requires rebuilding + recycling agentbox pods.


LikiosSedo (commented):
Thanks for the substantial implementation. I don't think this is ready to approve yet: I found three blocking correctness/security gaps, plus two configuration/UI consistency issues.

1. **[P1] Apply the locked agent type policy in LocalSpawner.** [`local-spawner.ts`](https://github.com/scitix/siclaw/blob/28f42708f083541ee311a33a6bc943f8257a7f97/src/gateway/agentbox/local-spawner.ts#L100-L109) still reads only the raw `tool_capabilities` column and never initializes `agentTypeState`. A Coordinator with an empty raw capability list therefore resolves to `null` (unrestricted) and keeps the default `custom` persona in Local mode. Please load `agent_type`, apply `effectiveCapabilityKeys`, set both state fields, and cover cold-spawn behavior with a LocalSpawner test.

2. **[P1] Bind `parentSessionId` to the caller's mTLS identity.** [`handleDelegate`](https://github.com/scitix/siclaw/blob/28f42708f083541ee311a33a6bc943f8257a7f97/src/gateway/delegate-api.ts#L138-L156) accepts an arbitrary parent session from the request body, adopts its `user_id`, and scopes session reuse under it without checking that the resolved `agent_id` matches `coordinatorAgentId`. `chat.resolveSession` already returns `agent_id`; reject mismatches before using the parent for attribution or reuse.

3. **[P1] Propagate model-level failures from `consumeAgentSse`.** The return value is discarded [here](https://github.com/scitix/siclaw/blob/28f42708f083541ee311a33a6bc943f8257a7f97/src/gateway/delegate-api.ts#L249-L297), after which the endpoint unconditionally emits `ok: true, status: "done"`. `consumeAgentSse` reports provider/rate-limit/routing failures through `errorMessage` without necessarily throwing, so these currently become false-success delegation results. Capture the result and emit a failed `delegate_result` when `errorMessage` is non-empty.

4. **[P2] Do not retain a hidden Custom system prompt for built-in agent types.** The UI hides the prompt as "built in" but still submits the previous value [on save](https://github.com/scitix/siclaw/blob/28f42708f083541ee311a33a6bc943f8257a7f97/portal-web/src/components/AgentSettings.tsx#L271-L275), while the runtime uses it as the base prompt and merely appends the built-in persona. Switching an existing Custom agent to Coordinator/SRE can therefore preserve contradictory hidden instructions. Clear or ignore `system_prompt` for non-Custom types at the backend boundary.

5. **[P2] Return complete agent metadata for the Delegates roster.** The list query [omits `agent_type`](https://github.com/scitix/siclaw/blob/28f42708f083541ee311a33a6bc943f8257a7f97/src/portal/agent-api.ts#L78-L96), so every type badge falls back to Custom. The UI also requests 500 agents while the endpoint caps a page at 100, silently hiding the remainder. Include `agent_type` and either paginate the roster fetch or provide an appropriate unpaginated lookup.

The existing CI and full local suites are green, but none of them exercise these integration boundaries. Please add focused regression coverage for the three P1 cases before approval.

jacoblee-io:
Thanks for the thorough review — all five are valid. Fixed, with regression tests for the three P1 boundaries. I also ran an independent adversarial pass and fixed four more.

## Your findings

1. **[P1] LocalSpawner agent-type policy** — `local-spawner.ts` now reads `agent_type`, applies `effectiveCapabilityKeys`, and sets **both** `allowedToolsState` and `agentTypeState` at cold spawn, mirroring the K8s `handleToolCapabilities` path. So a Coordinator with an empty raw capability list is now locked (not `null`/unrestricted) and gets its persona in Local mode. Test: `local-spawner.test.ts` — coordinator with empty caps → non-empty locked whitelist (incl. `delegate_to_agent`) + `agentTypeState==="coordinator"`; custom passes through untouched.

2. **[P1] Bind `parentSessionId` to caller mTLS identity** — `handleDelegate` now resolves the parent and rejects with **403** unless `parent.agent_id === coordinatorAgentId`, *before* adopting its `user_id` or scoping reuse. Test: `delegate-api.test.ts` — foreign parent → 403 and the peer turn never starts; own parent → proceeds.

3. **[P1] Propagate `consumeAgentSse` model failures** — the return value is captured; a non-empty `errorMessage` now emits a `failed` `delegate_result` instead of the unconditional `ok:true/"done"`. Test: `delegate-api.test.ts` — a `429`-style `errorMessage` → `ok:false/status:"failed"`; no error → `done`.

4. **[P2] Ignore Custom `system_prompt` for built-in types** — cleared at the **backend boundary** (`agent-api.ts` create + update): a non-custom `agent_type` forces `system_prompt` to null, including when switching an existing Custom agent to Coordinator/SRE without an explicit prompt in the body. Tests cover create + the switch-on-update case.

5. **[P2] Roster metadata + pagination** — the list query now selects `a.agent_type` (badges resolve correctly), and the page cap is raised 100→500 so the Delegates roster's single-page fetch isn't silently truncated. Existing cap test updated.

## Additional fixes from my own review

- **[P2] Stop during peer spin-up** — the abort handler used the post-`prompt()` session id (undefined during `getOrCreate` + the prompt round-trip), so a Stop then left the peer running headless. Now aborts by the up-front `peerSessionId` (== the box session id we prompt with).
- **[P2] Dead read-only apparatus** — `NON_INTERACTIVE_ORIGINS` was declared but referenced nowhere and the `DelegationContext.readOnly` doc claimed a "P0 default true" that the transport never sets. Removed the dead const and corrected the doc: the peer runs under its own config (coordinator/worker manage permissions independently); `readOnly:true` is an explicit opt-in, not imposed by delegation. (Behaviour unchanged — this was stale docs/dead code.)
- **[P3] Inflated tool-call count** — counting every event whose type contained `tool` (start/update/end) over-reported 2-3×; now counts once per call on `tool_execution_end`.
- **[P3] `usePilotChat` stale closure** — added `forceLive` to the EventSource effect's deps.

## Known, deferred (documented, not fixed)

When the peer doesn't call `report_findings`, its foreground sub-agents' assistant narratives can be concatenated into the fallback `finalText`. It's extra context rather than wrong data, and a clean fix risks dropping the real report; tracking separately rather than risk a regression here.

Full backend suite green (226 files / 4465 tests), `tsc` clean. Deployed to siclaw-inner (`:test`, all three images).

src/gateway/delegate-api.ts
```diff
@@ -0,0 +1,368 @@
...
+      const recent = await deps.frontendClient.request("chat.recentDelegationSessions", {
+        parent_session_id: body.parentSessionId, target_agent_id: peerAgentId, limit: RECENT_DELEGATION_LIMIT,
+      }) as { ids?: string[] };
+      if (recent?.ids?.includes(body.peerSessionId)) {
+        peerSessionId = body.peerSessionId; // owned by this coordinator chain AND recent → continue it
+      } else {
+        console.warn(`[delegate-api] peerSessionId ${body.peerSessionId} is not among the coordinator's recent ${RECENT_DELEGATION_LIMIT} delegations to ${peerAgentId}; starting a fresh peer session`);
+      }
+    }
+  } catch (err) {
```
LikiosSedo: [P1] Fail closed when parent validation is unavailable

The explicit mismatch path is now protected, but if `chat.resolveSession` throws, this catch continues and the raw caller-supplied `parentSessionId` is still persisted below. Because `resolveReadableSession` grants access through the parent owner, a transient Portal/RPC failure can link the child to another user's session under the same coordinator. Please reject the request when parent validation cannot complete, only persist a parent after `parentTrusted` is true, and add an exception-path regression test.

src/gateway/delegate-api.ts
```diff
@@ -0,0 +1,368 @@
...
+      });
+    }
+  });
+
+  // Surface the peer session id immediately (it's known now) so the coordinator's
+  // card can offer "open full session" LIVE, before the final result arrives.
+  writeFrame({ type: "delegate_session", peerSessionId });
+
+  try {
+    const handle = await deps.agentBoxManager.getOrCreate(peerAgentId);
```
LikiosSedo, line 255: [P1] Re-check cancellation after cold spawn

If the coordinator connection closes while `getOrCreate` is still cold-spawning the peer, the close handler aborts the signal while `peerClient` is still undefined. After the spawn completes, this path does not check that signal and still dispatches `client.prompt`; no later `abortSession` call is made, so the peer can run headlessly. Please check `peerAbort.signal.aborted` before prompting, or immediately abort after assigning the client when cancellation already happened, and cover this with a deferred-`getOrCreate` test.

src/gateway/delegate-api.ts
```diff
@@ -0,0 +1,368 @@
...
+      onEvent: (evt: Record<string, unknown>) => {
+        const e = evt as any;
+        // Relay the raw peer event live — the coordinator box translates it into
+        // the coordinator card's live steps.
+        writeFrame({ type: "peer_event", event: evt });
+        // Also publish to the peer session's OWN live channel so an opened
+        // PeerSessionView (subscribed to …/sessions/{peerSessionId}/events)
+        // streams the peer's full session live (mirrors delegation.emit_chat_event).
+        try { deps.frontendClient.emitEvent("chat.event", { sessionId: peerSessionId, event: evt }); } catch { /* best-effort live mirror */ }
+        if (e?.type === "delegation_artifact") {
```
LikiosSedo, line 311: [P1] Carry input_required through the delegation result

`request_input` emits an `input_required` event, but this transport only treats `delegation_artifact` specially. The coordinator-side translator also ignores `input_required`, and `DelegateResponse` has no clarification state, so a peer that asks a question and ends its turn is reported as a normal `done` result—often with no useful text. Please propagate a distinct input-required status plus the question and `peerSessionId`, then make `delegate_to_agent` instruct the coordinator to relay the question to the human.

src/tools/all-entries.ts
```diff
@@ -21,7 +21,8 @@ import { registration as hostScript } from "./script-exec/host-script.js";
 import { registration as clusterList } from "./query/cluster-list.js";
 import { registration as hostList } from "./query/host-list.js";
 // knowledge_search removed — replaced by LLM Wiki (Read tool + .siclaw/knowledge/)
-import { registration as resolvePodNetns } from "./query/resolve-pod-netns.js";
+// resolve_pod_netns removed — node_exec/pod_exec/*_script auto-resolve pod→netns
```
LikiosSedo, line 24: [P2] Keep built-in skills aligned with the tool registry

Removing `resolve_pod_netns` leaves shipped consumers invoking a tool that no longer exists: `skills/core/gateway-diagnostics/SKILL.md`, its reference material, `skills/platform/skill-authoring/SKILL.md`, and `docs/design/tools.md` still require or document it. This breaks the built-in pod gateway workflow. Please either restore the tool or update every consumer to the new `pod=` flow, including the documented resolve-once/reuse case.

src/agentbox/session.ts
```diff
@@ -2374,6 +2414,78 @@ export class AgentBoxSessionManager {
     // full pod/process restart from the PV snapshot. taskListId == session id.
     this.rehydrateLedger(id);
 
+    // Delegation roster (coordinator side): the peer agents this agent may
+    // delegate to, delivered from the gateway (K8s boxes have no DB). Skipped on
+    // a delegated turn (a peer can't re-delegate — one-level). Best-effort: a
+    // fetch failure just means the delegate_to_agent tool stays hidden.
+    let delegationRoster: DelegateRosterMember[] | undefined;
+    const gc = this.gatewayClient;
+    if (gc && !delegation) {
```
LikiosSedo, line 2442: [P2] Refresh roster snapshots for live coordinator sessions

The delegation roster is captured only when this managed session is created. A transient fetch failure therefore hides delegation for the lifetime of the cached session, and later changes to a member's clusters, hosts, name, description, or status do not refresh coordinators that reference that member—the resource APIs notify only the edited member. Please add a delegates reload/invalidation path and notify dependent coordinators through the existing reverse `agent_delegates` index.

src/agentbox/session.ts
```diff
@@ -2285,14 +2323,16 @@ export class AgentBoxSessionManager {
...
       }
-      // Reuse unless the operating mode changed mid-session (e.g. user toggled Deep
-      // Investigation): rebuild so tools scoped by `availableModes` are re-resolved.
-      // Don't rebuild mid-first-prompt (_promptDone false).
-      if (existing.activeMode === activeMode || !existing._promptDone) {
+      // Reuse unless the operating mode OR the delegation tier changed mid-session
+      // (e.g. user toggled Deep Investigation, or a reused session id flips between a
+      // delegated and a direct turn): rebuild so tools scoped by `availableModes` /
+      // the read-only delegation filter are re-resolved. Don't rebuild mid-first-prompt.
+      const sameDelegation = delegationSignature(existing.delegation) === delegationSignature(delegation);
```
LikiosSedo, line 2330: [P2] Refresh delegation context when continuing a peer session

Session reuse compares only the delegation tier (`none`/`ro`/`rw`) and returns the existing session without updating its captured `DelegationContext`. Every gateway delegation receives a new `delegationId`, but `report_findings` and `request_input` will continue stamping events with the previous call's id when the same peer session is continued. Please refresh the delegation context/refs on reuse, or rebuild when the correlation id changes while preserving the persisted conversation.

portal-web/src/components/chat/PeerSessionView.tsx
```diff
@@ -0,0 +1,108 @@
...
+  agentId: string
+  sessionId: string
+  status?: string
+  label?: string
+  onClose?: () => void
+  onOpenSubagent?: (childSessionId: string, status?: string, label?: string, opts?: { full?: boolean }) => void
+}) {
+  // Force live: the peer session runs in the peer's box, so liveness is
+  // unobservable from the coordinator agent — trust the card's running status.
+  const isLive = status === "running" || status === "launched"
```
LikiosSedo, line 35: [P2] Do not keep the drawer live from a stale status snapshot

The selected card's status is copied into `subagentDrill` when the drawer opens and is never updated afterward. If it was opened while `running`, this `isLive` value remains true after the parent card reaches `done`, leaving the Live badge, forced polling, and Working indicator active until the drawer is closed and reopened. Please derive liveness from current session/card state or update the selected status when terminal events arrive.

LikiosSedo (commented):
Re-reviewed the updated head. The previous findings appear to be addressed, but I found the following remaining issues in the current implementation.

jacoblee-io:
Thanks for the re-review — all seven are addressed. Details below; regression tests added for the three P1s.

## P1

1. **Fail closed when parent validation is unavailable** (`delegate-api.ts`) — `chat.resolveSession` is now awaited in its own try/catch: an RPC error returns **503** (can't verify ownership → don't proceed), and the parent is linked in `ensureChatSession`/`appendMessage` **only** after `parentTrusted` is set (validated identity match). The recency-reuse lookup keeps its own non-fatal fallback. Test: RPC-throws → 503, no prompt, no session persisted.

2. **Re-check cancellation after cold spawn** (`delegate-api.ts`) — after `getOrCreate` + `peerClient` assignment, we check `peerAbort.signal.aborted` **before** `client.prompt(...)` and bail if the coordinator already disconnected, so a Stop during the pod spawn no longer runs the peer headless. Test: `getOrCreate` fires the response `close` handler mid-spawn → the peer is never prompted.

3. **Carry `input_required` through the delegation result** (`delegate-api.ts` + `agent-delegate.ts` + `delegate-to-agent.ts`) — the transport now captures the peer's `input_required` event and emits a distinct `status:"input_required"` with `inputQuestion` + `peerSessionId` on `DelegateResponse`. `delegate_to_agent` returns a result instructing the coordinator to relay the question to the human and, once answered, delegate again with `session_id=<peerSessionId>` so the peer resumes from retained context. Test: an `input_required` event → `status:"input_required"` with the question.

## P2

4. **resolve_pod_netns consumers** — completed the removal (kept the tool deleted; the `pod=` one-step auto-resolve already covers it). Updated all consumers to the `pod=` flow: `gateway-diagnostics/SKILL.md` (+ switchdev reference), `skill-authoring/SKILL.md`, and `docs/design/tools.md`. The documented resolve-once/reuse case is now expressed via `netns=` as an advanced escape hatch (the one-step `pod=` re-resolves per call).

5. **Refresh roster for live coordinator sessions** (`agent-api.ts`) — a member agent's roster-visible change (name/description/status/is_production on agent update; clusters/hosts on resources update) now notifies the coordinators that reference it, via the reverse `agent_delegates(member_agent_id)` index, with a `tools` reload — the roster is part of the coordinator's toolset, so this invalidates its live sessions and the next turn re-fetches (`getOrCreate → fetchDelegates`). This rides the existing invalidation path rather than adding a materialization sync type (the roster isn't written to disk).

6. **Refresh delegation context on peer-session reuse** (`session.ts`) — on reuse (same tier), the delegation correlation is refreshed **in place** (`delegationId`, parent ids). `refs.delegation` is the same object the tools read (agent-factory passes it by reference), so `report_findings`/`request_input` now stamp the current call's id without a rebuild; a tier change still rebuilds as before.

7. **Drawer live status** (`AgentChat.tsx`) — the peer drawer's liveness is now kept in sync with the delegate card's **current** status: an effect tracks the matching card by `child_session_id` and pushes status changes into the drill state, so the Live badge / polling / working indicator stop when the card reaches a terminal state.

Backend suite green (226 files / 4468 tests), `tsc` clean. Deployed to siclaw-inner and sicore-test.

LikiosSedo (commented):
Re-reviewed the latest update. The earlier P1 issues are addressed, but the following edge cases remain in the current implementation.

portal-web/src/components/AgentChat.tsx
```diff
@@ -298,6 +299,22 @@ export function AgentChat({ agentId, selectedSessionId, onSessionChange }: Agent
...
   }, [activeSessionId])
 
+  // Keep the open peer-session drawer's liveness in sync with the delegate card's
+  // CURRENT status. `subagentDrill.status` is a snapshot captured when the drawer
+  // opened, so without this the drawer's "Live" badge + forced polling would stay on
+  // after the delegate card reaches a terminal state. Track the matching card by its
+  // child_session_id and push status changes into the drill state.
+  useEffect(() => {
+    if (!subagentDrill?.full) return
+    const match = pilot.messages.find(
```
LikiosSedo: [P2] Track the selected delegation card, not just the peer session

A continued delegation intentionally reuses the same `child_session_id`, so the timeline can contain multiple cards for one peer session. Because `pilot.messages` is chronological, `find()` selects the oldest card and can immediately overwrite a newly opened running card with the previous card's terminal status. The drawer then disables forced live updates during the continuation. Please correlate by the selected tool/delegation id, or at minimum search from the end for the latest matching card, and add a continuation regression test.

src/agentbox/session.ts
```diff
@@ -2285,14 +2323,27 @@ export class AgentBoxSessionManager {
...
+      // Refresh the delegation CORRELATION on reuse. The tier is unchanged here (a tier
+      // change falls through to a rebuild below), but every delegation turn gets a NEW
+      // delegationId (and possibly parent ids). The tools read `refs.delegation` LIVE and
+      // it is the SAME object we store here (agent-factory passes it by reference), so an
+      // in-place update makes report_findings / request_input stamp the CURRENT call's id
+      // instead of the previous one — no rebuild needed, conversation preserved.
+      if (sameDelegation && existing.delegation && delegation) {
```
LikiosSedo: [P2] Do not mutate an in-flight delegation before rejecting the new prompt

This updates the existing session's shared `DelegationContext` before the HTTP layer checks `_promptDone` and returns 409 for a concurrent prompt. If two continuations target the same peer session, the rejected request can therefore replace the running turn's `delegationId`; a later `report_findings` or `request_input` from the running turn is stamped with the rejected request's id. Check and reject the busy session before mutating the shared context, and add a concurrent-reuse test.

src/portal/agent-api.ts
```diff
@@ -48,14 +49,37 @@ export function registerAgentRoutes(
   jwtSecret: string,
   connectionMap: RuntimeConnectionMap,
 ): void {
+  // Refresh coordinators that delegate to `memberAgentId` when the member's
+  // roster-visible attributes change (name / description / status / clusters /
+  // hosts). The roster is part of the coordinator's toolset (delegate_to_agent's
+  // availability + manifest), so a `tools` reload invalidates the coordinator's
+  // live sessions → the next turn re-fetches the roster (getOrCreate → fetchDelegates).
+  // Reverse index: agent_delegates(coordinator_agent_id, member_agent_id). Best-effort.
+  async function notifyDependentCoordinators(memberAgentId: string): Promise<void> {
```
LikiosSedo: [P2] Cover every source that changes the derived roster

This helper covers agent attribute and binding updates, but `config.getDelegates` also derives cluster/host names and active membership. Cluster/host rename or deletion only notifies the bound member, while deleting the member agent removes the reverse rows without notifying its coordinators. Those coordinators retain stale routing manifests until their sessions happen to rebuild. The initial `fetchDelegates` failure path also still has no retry. Please centralize reverse coordinator invalidation across these mutation paths and add direct tests; the current agent-api tests let this new query throw inside the catch, so the green suite does not verify any notification.

src/tools/workflow/report-findings.ts
```diff
@@ -0,0 +1,86 @@
...
+      const findings = params.findings?.trim() ?? "";
+      if (!findings) return result("report_findings requires non-empty findings.", false);
+      if (!refs.sessionEventEmitter || !refs.delegation) {
+        return result("report_findings is not available in this context.", false);
+      }
+      refs.sessionEventEmitter({
+        type: "delegation_artifact",
+        delegationId: refs.delegation.delegationId,
+        findings,
+        actions_taken: params.actions_taken?.trim() || "none (read-only)",
```
LikiosSedo: [P2] Align the artifact contract with full-capability delegation

The gateway explicitly sends `readOnly: false`, so a peer may run mutating tools under its own capabilities. However several public comments still describe the peer as read-only, and omitting `actions_taken` here records the false statement `none (read-only)`. That can hide real actions from the coordinator's structured artifact. Please update the contract wording and either require `actions_taken` for full-capability delegation or use a neutral value such as `not reported`.

jacoblee-io:
Thanks — all four addressed, and I ran another adversarial self-review over the fixes (which caught two more, also fixed below).

## Your findings

1. **Artifact contract vs full-capability delegation** (`report-findings.ts`) — the peer runs under its own capabilities (`readOnly:false`), so `actions_taken` no longer defaults to the false `"none (read-only)"`; it defaults to a neutral `"not reported"`, the description now requires it when any action was taken, and the stray "peer runs read-only" comments in `delegate-to-agent.ts` / `delegate-api.ts` are corrected.

2. **Reverse coordinator invalidation across every roster source** — centralized in a new `src/portal/coordinator-invalidation.ts` (`notifyCoordinatorsForMembers`) and wired so a coordinator gets the SAME refresh treatment its member (sre) agent already gets:
   - agent attribute update, resources (cluster/host/delegate) rebind, **and agent DELETE** (`agent-api.ts` — awaited before the FK cascade removes the reverse rows);
   - **cluster rename + delete** (`cluster-api.ts`) and **host rename + delete** (`host-api.ts`), reusing the bound-member ids each already captures before its cascade.
   Delivered via a `tools` reload (the roster is part of the coordinator's toolset → next turn re-fetches). Also added a one-shot `fetchDelegates` retry in `session.ts` so a single transient miss doesn't hide the tool for the session's life. New `coordinator-invalidation.test.ts` verifies the notify actually fires (incl. empty/dupe/error paths) — the previous suite let the query throw in a catch and asserted nothing.

3. **Don't mutate an in-flight delegation before the 409** (`session.ts`) — the in-place correlation refresh is now gated to match the HTTP 409 condition exactly: `existing._promptDone && !existing._promptInflight`. (The self-review below tightened this: `_promptInflight` alone can hold the brain during synthetic-parent-prompt setup while `_promptDone` is momentarily true.) Tests cover both the idle-reuse refresh and the busy-session no-mutate case.

4. **Track the selected card, not just the peer session** (`AgentChat.tsx` + new `lib/delegateCard.ts`) — a continuation reuses one `child_session_id`, so drawer-status sync now takes the LATEST matching card (scan from the end) via `latestDelegateCardStatus`, so a freshly opened running continuation isn't overwritten by an earlier card's terminal status. Unit-tested (continuation, cross-session isolation, metadata fallback).

## From my own review of the above

- **[P2] gate alignment** — folded into #3: the first cut gated on `_promptDone` only; corrected to `_promptDone && !_promptInflight` to match the 409 guard's `_promptInflight` sub-case (background-job completion turn).
- **[P3] dead code** — dropped two unused two-hop helpers; the cluster/host callers use `notifyCoordinatorsForMembers` inline with the ids they already hold (correct w.r.t. cascade timing).

Backend suite green (227 files / 4473 tests), portal-web green (193), `tsc` clean on both. Deployed to siclaw-inner and sicore-test.

src/portal/agent-api.ts
```diff
@@ -314,6 +362,12 @@ export function registerAgentRoutes(
       console.warn(`[agent-api] delete ${params.id}: runtime terminate failed: ${termResult.error}`);
     }
 
+    // Refresh coordinators that delegate to this agent BEFORE the delete — the FK
+    // cascade removes the agent_delegates reverse rows, so querying them afterwards
+    // would return nothing. Awaited (best-effort internally) so the read precedes
+    // the cascade; the coordinators then re-fetch a roster that no longer lists it.
+    await notifyCoordinatorsForMembers(connectionMap, [params.id]);
```
LikiosSedo: [P2] Notify captured coordinators after the delete

notifyCoordinatorsForMembers both queries and sends the reload before the DELETE. A coordinator can process it immediately, rebuild its session, and re-fetch the still-present agent_delegates row, leaving the deleted peer cached with no second invalidation. Please split capture from notification: query coordinator IDs before the FK cascade, delete the agent, then notify the captured IDs. A route-level ordering test would also close the current gap—the route suites still let the helper extra DB query fall into its catch and pass without verifying the wiring.

jacoblee-io:
Fixed — good catch on the ordering.

**Notify captured coordinators after the delete** (`agent-api.ts` DELETE): split capture from notification. The helper is now `collectDependentCoordinators(memberIds)` (query only) + `notifyCoordinators(connMap, ids)` (send only). The DELETE route now (1) resolves the dependent coordinator ids **before** the delete, while the `agent_delegates` reverse rows still exist, (2) `DELETE FROM agents` (FK cascade drops the rows), (3) notifies the captured ids — so each coordinator re-fetches a roster that no longer lists the removed peer, with no window to re-cache it. `notifyCoordinatorsForMembers` (the update/rebind path, where the rows persist) is unchanged and now composes the two.

Added a route-level ordering test in `agent-api.test.ts`: asserts the coordinator lookup query precedes `DELETE FROM agents` and that the captured coordinator receives the `tools` reload — closing the gap where the prior suites let the helper's query fall into its catch and verified nothing. Also direct tests for `collectDependentCoordinators` / `notifyCoordinators` (dedupe, empty, DB-error).

Backend green (227 files / 4477), portal-web green (193), `tsc` clean. Deployed to siclaw-inner and sicore-test.

src/portal/coordinator-invalidation.ts
```diff
@@ -0,0 +1,66 @@
...
+  } catch (err) {
+    console.warn("[coordinator-invalidation] failed to resolve dependent coordinators:", err);
+    return [];
+  }
+}
+
+/** Send each coordinator a `tools` reload (→ roster re-fetch on its next turn). */
+export function notifyCoordinators(connectionMap: RuntimeConnectionMap, coordinatorIds: string[]): void {
+  for (const id of [...new Set(coordinatorIds.filter(Boolean))]) {
+    connectionMap.notify(id, "agent.reload", { agentId: id, resources: ["tools"] });
```
LikiosSedo: [P2] Preserve best-effort isolation when sending reloads

The previous implementation kept both the DB lookup and connectionMap.notify inside the best-effort try/catch. After this split, notifyCoordinators runs outside that protection. A synchronous notify/ws.send failure now rejects the fire-and-forget notifyCoordinatorsForMembers calls (their callers use void), which can surface as an unhandled rejection, and it stops later coordinators from being notified. Please catch failures per coordinator so one bad connection cannot escape or block the remaining reloads, and add a throwing-notify regression test.

jacoblee-io:
Fixed. `notifyCoordinators` now wraps each `connectionMap.notify` in its own try/catch, so a synchronous notify/ws.send failure on one coordinator is logged and skipped — it can't escape the fire-and-forget `notifyCoordinatorsForMembers` (no unhandled rejection for the `void` callers) and can't block the remaining coordinators' reloads. Added two regression tests: a throwing notify in the middle of the list still notifies the others and does not throw, and `notifyCoordinatorsForMembers` stays resolved when notify throws.

Backend green (227 files / 4479), `tsc` clean. Deployed to siclaw-inner and sicore-test.

LikiosSedo (approved):
Re-reviewed the latest head. The reported issues are addressed, and no blocking issues remain.

PR #414, opened 2026-07-17
fix(kbc): align test consumer retrieval discipline

LikiosSedo:
## Summary

- align the read-only test consumer prompt with its existing `Read`, `Glob`, and `Grep` tool surface
- route from `index.md` first, then use snapshot-scoped search only when links do not surface the answer
- require full-page reads before the consumer declares that the wiki does not cover a question
- emit a stable `consumer_fingerprint` over the effective prompt, model, SDK, tool set, and explicit contract revision
- propagate that fingerprint through `capability.testStart` for the companion Sicore MR to persist
- lock the bilingual retrieval and consumer-identity contracts with compile-box regression tests

## Why

The test-session runtime already exposed `Glob` and `Grep`, while the persona incorrectly said that no search tool existed. That mismatch made the consumer rely too heavily on index routing and increased false "not covered" answers when relevant content lived on a less obvious page.

This PR also changes consumer behavior without changing Wiki bytes. A Wiki snapshot hash therefore cannot identify the result semantics by itself. The fingerprint gives the control plane a deterministic, deployment-order-compatible identity for separating regression rounds produced by different consumer contracts.

## Product impact

This improves test-answer recall and prevents prompt/model/runtime contract changes from silently sharing one regression baseline. Search remains confined to the pinned read-only Wiki snapshot, and search hits are treated only as routing hints until the matching pages are read in full.

There is no new user-facing UI or review workflow. Older boxes simply omit the additive field; the companion Sicore MR records those sessions in a `legacy` compatibility bucket.

## Validation

- `kbc/platform/pod/.venv/bin/python kbc/platform/pod/test_compile_box.py`
- `kbc/platform/pod/.venv/bin/python kbc/platform/pod/test_redblue.py`
- `npm run build`
- `npm test` (220 files, 4423 passed, 2 skipped)


kbc/platform/pod/compile_box.py
```diff
@@ -2824,6 +2833,53 @@ async def _run_wrapper(run: CompileRun):
...
+    tool surface and turn limit are automatic; TEST_CONSUMER_CONTRACT_VERSION
+    is the explicit seam for any remaining answer-affecting driver behavior.
+    """
+    role = _prompt("test_role", locale)
+    contract = {
+        "version": TEST_CONSUMER_CONTRACT_VERSION,
+        "role_sha256": hashlib.sha256(role.encode("utf-8")).hexdigest(),
+        "model": model or _test_model(),
+        "sdk_version": sdk_version or _test_sdk_version(),
+        "tools": sorted(set(_effective_test_allowed_tools(allowed_tools))),
```
jacoblee-io, line 2878: Latent (not currently live): the fingerprint's `tools` hashes `allowed_tools` only, but the driver also applies `disallowed_tools=READONLY_CONSUMER_DISALLOWED_TOOLS`, which under `bypassPermissions` is what actually removes tools from context. If the kb-test BoxProfile (or a caller's `allowed_tools`) ever includes a tool that also sits in the disallowed set (e.g. `Bash`), the fingerprint would advertise a tool surface the session never has, and two sessions with identical effective behavior could get different fingerprints. Harmless today because the profile equals `["Read","Glob","Grep"]` and the error direction is a spurious split (never a false-shared baseline). Optional fix: subtract `READONLY_CONSUMER_DISALLOWED_TOOLS` inside `_effective_test_allowed_tools` so advertised == effective.

jacoblee-io (commented):
Reviewed for correctness. The PR is clean and self-consistent: the fingerprint captures model/max_turns/tools into `TestRun` at open and `test_session_driver` applies those same captured values, so there is no fingerprint-vs-session drift. `max_turns=0` is preserved on both sides (`is not None`), determinism holds (`json.dumps(sort_keys=True)` over a `sorted(set(...))`), and the new `consumer_fingerprint` is forward-only on the TS side — no in-repo consumer breaks when an old box omits it (same tolerant pattern as `snapshot_hash`). One low-severity, currently-latent note inline. Good to merge.

jacoblee-io (approved):
LGTM. The follow-up commit e27fa5c resolves my earlier note: `_effective_test_allowed_tools` now subtracts `READONLY_CONSUMER_DISALLOWED_TOOLS`, so the fingerprinted tool surface equals the effective one applied by the driver (advertised == effective), with a regression test locking `fingerprint(["Read","Bash"]) == fingerprint(["Read"])` and `opts.tools == ["Read"]`.

The incidental semantic change (`allowed_tools=[]` now means "no tools" instead of falling back to the default set) is safe: `capability.testStart` always sends `getBoxProfile("kb-test").allowedTools` which is `["Read","Glob","Grep"]`, never `[]`, and the None-vs-empty distinction matches the documented "None/absent → default" contract.

No remaining concerns — fingerprint/session values are consistent, `max_turns=0` preserved on both sides, determinism holds, and the TS field is forward-only (old box omitting it is handled like `snapshot_hash`).

PR #415, opened 2026-07-17
feat(kbc): assist reference answer authoring

LikiosSedo:
## Summary

Reference-answer authoring currently asks owners to write a complete canonical answer unaided. This adds a fast, knowledge-grounded editor that can suggest 2-3 candidates or polish an existing draft while keeping the blind test consumer isolated from every reference answer.

### Solution

The Runtime exposes one short-lived read-only capability over the stable raw and candidate workspace. It uses a fresh in-memory compile-tier session with only Read, Glob, Grep, and a structured submit tool; validates real raw evidence paths; shares the existing single-flight review guard; and never persists the result.

Sicore MR !649 consumes this capability and owns the transient UI, revision fences, and explicit save boundary. Merge and deploy this PR before !649.

## Test Plan

- [x] npm run build
- [x] New gateway capability contract test
- [x] AgentBox HTTP, grounding, prompt-locale, isolation, tool, and turn-budget tests
- [x] Full npm test: 4420 passed; four load-related 5s timeouts were rerun separately and passed 15/15 with loopback enabled
- [ ] Manual end-to-end verification after the paired Sicore MR is deployed


src/gateway/server.ts
```diff
@@ -871,6 +873,29 @@ export async function startRuntime(opts: StartRuntimeOptions): Promise<RuntimeSe
...
+      `/test-reference-assist/${req.run_id}`,
+      {
+        mode: req.mode,
+        question: req.question,
+        ...(req.draft_answer ? { draft_answer: req.draft_answer } : {}),
+        ...(req.evidence_paths?.length ? { evidence_paths: req.evidence_paths } : {}),
+      },
+      75_000,
+    );
+    return { run_id: req.run_id, ...assisted } as CapabilityTestReferenceAssistResponse;
```
jacoblee-io: Low-severity contract leak: the box replies `web.json_response({"ok": True, **result})`, so `...assisted` spreads an `ok: true` field that is **not** part of `CapabilityTestReferenceAssistResponse` (the `as` cast hides it). The sibling `capability.testRecommend` above avoids this by explicitly picking the union fields into a fresh object. Suggest mirroring that here — pick `mode` + the mode-specific fields — so no undeclared field reaches the downstream Go/Sicore consumer.

jacoblee-io (commented):
Reviewed for correctness and isolation. Solid PR: the reference-assist session is correctly confined to `raw/`+`candidate/` by the shared `_recommendation_path_escape` guard (every other tool denied, evidence paths resolved before the `relative_to(raw_root)` containment check so symlink/`..`/absolute escapes are caught), validation is robust (2-3 candidates, unique styles, casefold distinctness, `[:12]`/`[:5]` slices, idempotent double-submit guard), and the timeouts are correctly ordered (box 60s < RPC 75s, so the box returns a structured 504 first). One low-severity wire-contract issue + a couple of optional cleanups inline. No functional blockers.

src/gateway/server-capability-test-recommend.test.ts
```diff
@@ -22,6 +22,15 @@ vi.mock("./agentbox/client.js", () => ({
           evidence_paths: ["raw/policy.md"],
         };
       }
+      if (path.startsWith("/test-reference-assist/")) {
+        return {
+          mode: "suggest",
```
jacoblee-io, line 38: This mock omits `ok: true`, but the real box returns `{"ok": True, ...}`. Because `toEqual` is an exact match and the mock is unfaithful, the test stays green while production returns the extra `ok` field (see the spread at server.ts:896). Add `ok: true` to the mock so the assertion reflects the real wire shape and would catch the leak.

kbc/platform/pod/compile_box.py
```diff
@@ -3249,6 +3254,196 @@ async def recommend_test_question(parent: "CompileRun") -> dict:
     return captured
 
 
+REFERENCE_ASSIST_STYLES = {"concise", "complete", "boundary"}
+REFERENCE_ASSIST_BUILTIN_TOOLS = ["Read", "Glob", "Grep"]
+REFERENCE_ASSIST_DISALLOWED_TOOLS = list(RECOMMENDATION_DISALLOWED_TOOLS)
+
+
+def _normalize_reference_evidence(root: Path, evidence, *, required: bool) -> list[str]:
```
jacoblee-io, line 3262: Optional reuse: this new `_normalize_reference_evidence` duplicates the evidence-path loop already inlined in `_validate_test_recommendation` (the `raw/` confinement + `resolve()`/`relative_to`/`is_file` block). Since you extracted the shared helper here, `_validate_test_recommendation` could call it too instead of keeping its own copy.

jacoblee-io (approved):
LGTM. Follow-up commit baf9d5ae resolves both blocking comments: the RPC now maps the box reply into the declared `CapabilityTestReferenceAssistResponse` by explicitly picking per-`mode` fields (suggest → candidates; polish → polished_answer/evidence_paths/warnings) with a throw on an unexpected mode, so the internal `ok` envelope no longer leaks; and the box-client mocks now include `ok: true` faithfully, with a new test covering the polish union branch — so the exact-match `toEqual` would catch any future envelope leak.

Rest of the feature was already sound: the reference-assist session is confined to raw/+candidate/ by the shared path guard (every other tool denied; evidence paths resolved before the relative_to containment check), validation is robust (2-3 candidates, unique styles, casefold distinctness), single-flight shares RECOMMENDATIONS_ACTIVE, and timeouts are ordered inner<outer (box 60s < RPC 75s). The one remaining note (reuse `_normalize_reference_evidence` inside `_validate_test_recommendation`) is optional and non-blocking. Ship it.

PR #416, opened 2026-07-17
fix(kbc): make reference assistance reliable

LikiosSedo:
## Summary

Make reference-answer suggestions and polishing use the administrator-selected LIGHT tier, allow the model enough time to finish, and keep answers compact enough for regression grading.

### Problem

The assistant used the compile-tier model and a 60-second inner timeout while outer transports timed out at 75 seconds. Real runs regularly ended in 504/409 without a structured result, and polishing could expand a short answer into multiple verbose versions.

### Solution

- Resolve the assistant model from the configured LIGHT/blue role, with an explicit operator override and compile-tier compatibility fallback.
- Give the box a 180-second budget and the Runtime transport 195 seconds so the inner error remains authoritative.
- Add mode-specific instructions: suggestions are 2-3 independently usable answers; polishing returns one shortest gradeable answer with no research narration or duplicate variants.
- Cover LIGHT-tier routing and the longer Runtime timeout.

## Test Plan

- [x] npm run build
- [x] npm test (4425 passed, 2 skipped)
- [x] KBC compile-box regression script
- [ ] Manual integrated-cluster verification

## Architecture Checklist

- [x] No resource-sync, filesystem-write, shell-execution, DB, or tool-protocol changes

## General Checklist


jacoblee-io (approved):
LGTM. Clean reliability follow-up on the reference-assist capability.

Verified:
- Timeout ordering holds: `postJson(..., 195_000)` explicitly overrides the client's 30s default and is the only abort timer on this hop, while the box's inner `asyncio.wait_for(180s)` fires first and returns a structured 504 — so the inner error stays authoritative with ~15s of headroom. Matches the recommend pair (box 180s < RPC 210s).
- `_reference_assist_model()` fallback chain (KBC_REFERENCE_ASSIST_MODEL → KBC_PK_BLUE_MODEL → compile tier) is sensible for routing short authoring work to the admin-selected LIGHT tier, with an operator override and a compat fallback for older consumers; the test covers the blue-model path and restores the env.
- Mode-specific directives are appended off the already-validated `request_payload['mode']`, and the driver test asserts both the resolved model and the suggestion-mode directive text.
- Both timeout constants (box 180 / RPC 195) and the two `timeoutMs` assertions move together; prompt edits are text-only and reinforce the per-request directives harmlessly.

No blockers. Ship it.

PR #419, opened 2026-07-18
feat(kbc): add Codex SDK compiler engine

LikiosSedo:
## Summary

Enable the knowledge compiler to run on the official Codex SDK with ordinary OpenAI Responses providers, while preserving the existing KBC lifecycle, tools, artifacts, and read-only test contract.

### Problem

KBC was bound directly to Claude Agent SDK construction, so a compatible GPT Responses endpoint could not be selected as the compiler engine even though the box protocol and business workflow were already engine-neutral.

### Solution

Add a pinned Codex SDK adapter behind the existing compile/test seams. KBC keeps one implementation of plan, ticket, summary, sync, recommendation, reference-assist, and deterministic validation behavior; Codex application tools bridge back to those same tool bodies over a loopback-only MCP callback.

Tenant settings, ambient features, web search, and model-proposed credential inheritance are disabled. Unattended execution uses `ApprovalMode.deny_all`. The current restricted Kubernetes host rejects Codex's nested bubblewrap policies, so the SDK uses full access inside the already isolated single-run KBC Pod (the existing Claude `bypassPermissions` trust boundary); reviewer inputs are staged as independent copies so the original raw/candidate trees cannot be mutated through shared inodes.

Interactive KB test sessions now use the configured production-consumer/blue model instead of a Claude-specific default.

## Test Plan

- [x] `npm run build`
- [x] `npm test -- src/gateway/capability` (75 tests)
- [x] `test_compile_box.py`
- [x] `test_codex_engine.py` (6 checks)
- [x] GitHub CI: Type Check, Test, Portal Web Test, KBC Box Test
- [x] Built and registry-read-back verified the Linux/amd64 compile-box image from commit `b790233ccea991db26ada814d1a340029ad2623d`
- [x] Deployed the exact digest to `sicore-siflow-test`
- [x] Real Mass API / `gpt-5.6-luna` provider and model probes
- [x] Real KB compile from two raw sources, contradiction resolution, and repair to `ready_to_publish`
- [x] Cold-rehydrated test Pod ran the exact final image and answered the saved regression question; UI result: 1/1 passed (100%)

Final compile-box digest: `sha256:45a72219016a1b38f42fbb69f57a05bc7e99f4dfb8a7ee05ed03115366809c81` (`linux/amd64`).

## Architecture Checklist

- [x] **Security model**: no ambient tenant provider configuration, credentials excluded from child shell environments, unattended approvals denied, and each run remains confined to its dedicated KBC Pod.
- [x] **Tool protocol**: Codex MCP calls reuse KBC-owned engine-neutral tool bodies; no parallel business workflow was introduced.

## General Checklist

### DIFF
--- a/kbc/platform/pod/codex_engine.py
+++ b/kbc/platform/pod/codex_engine.py
@@ -474,6 +474,7 @@ class CodexSDKClient:
     async def connect(self) -> None:
         # Lazy import keeps the Claude image/test environment importable even if
         # only the original SDK dependency is installed.
+        from codex_cli_bin import bundled_package_dir
         from openai_codex import ApprovalMode, AsyncCodex, CodexConfig
 
         base_url = os.environ.get("OPENAI_BASE_URL", "").strip()
@@ -540,6 +541,10 @@ class CodexSDKClient:
                 # system config only; it deliberately excludes user data and
                 # process metadata such as /proc.
                 ":minimal": "read",
+                # The pinned runtime may live outside the platform roots (for
+                # example setup-python installs it under /opt on CI). Codex
+                # re-execs this binary inside bubblewrap for each command.
+                str(bundled_package_dir().resolve()): "read",
                 self.cwd: "write",
                 str(Path(self._shell_home).resolve()): "write",
             }
--- a/kbc/platform/pod/codex_engine.py
+++ b/kbc/platform/pod/codex_engine.py
@@ -105,6 +105,11 @@ _GLOB_MAX_RESULTS = 500
 _GREP_MAX_RESULTS = 200
 _GREP_MAX_FILES = 5_000
 _GREP_MAX_FILE_BYTES = 2 * 1024 * 1024
+_READ_TOOL_NAME_MAP = {
+    "Read": "kbc_read_file",
+    "Glob": "kbc_glob_files",
+    "Grep": "kbc_grep_files",
+}
 
 
 class _ReadOnlyFileAccess:
@@ -178,6 +183,19 @@ class _ReadOnlyFileAccess:
                 self.grep_files,
             ),
         ]
+
+    def selected_tools(self, allowed_tools: Iterable[str] | None) -> list[EngineTool]:
+        tools = self.tools()
+        if allowed_tools is None:
+            return tools
+        requested = list(dict.fromkeys(allowed_tools))
+        unknown = sorted(set(requested).difference(_READ_TOOL_NAME_MAP))
+        if unknown:
+            raise ValueError(
+                "unsupported read-only Codex tools: " + ", ".join(unknown)
+            )
+        selected = {_READ_TOOL_NAME_MAP[name] for name in requested}
+        return [item for item in tools if item.name in selected]
 
     def _resolve(self, value: object, *, default: Path | None = None) -> Path:
         if value is None and default is not None:
@@ -398,6 +416,7 @@ class CodexSDKClient:
         session_id: str,
         read_only: bool = False,
         allowed_read_roots: list[str] | None = None,
+        allowed_read_tools: list[str] | None = None,
         tools: list[EngineTool] | None = None,
         reasoning_effort: str | None = None,
         max_tool_calls: int | None = None,
@@ -410,8 +429,8 @@ class CodexSDKClient:
         declared_tools = list(tools or [])
         if read_only:
             file_access = _ReadOnlyFileAccess(cwd, allowed_read_roots or [cwd])
-            read_tools = file_access.tools()
-            reserved = {item.name for item in read_tools}
+            read_tools = file_access.selected_tools(allowed_read_tools)
+            reserved = set(_READ_TOOL_NAME_MAP.values())
             duplicates = sorted(reserved.intersection(item.name for item in declared_tools))
             if duplicates:
                 raise ValueError(f"read-only tool names are reserved: {', '.join(duplicates)}")
@@ -420,6 +439,8 @@ class CodexSDKClient:
         else:
             if allowed_read_roots:
                 raise ValueError("allowed_read_roots is only valid for read-only Codex sessions")
+            if allowed_read_tools is not None:
+                raise ValueError("allowed_read_tools is only valid for read-only Codex sessions")
             self.allowed_read_roots = ()
             self.tools = declared_tools
         # Mass GPT-5.6 high/medium can spend longer than KBC's 90s model-idle
@@ -453,7 +474,7 @@ class CodexSDKClient:
     async def connect(self) -> None:
         # Lazy import keeps the Claude image/test environment importable even if
         # only the original SDK dependency is installed.
-        from openai_codex import ApprovalMode, AsyncCodex, CodexConfig, Sandbox
+        from openai_codex import ApprovalMode, AsyncCodex, CodexConfig
 
         base_url = os.environ.get("OPENAI_BASE_URL", "").strip()
         api_key = os.environ.get("OPENAI_API_KEY", "").strip()
@@ -482,6 +503,10 @@ class CodexSDKClient:
             "features.multi_agent=false",
             "features.remote_plugin=false",
             "features.shell_snapshot=false",
+            # Keep one mechanically sandboxed command path. The Python SDK host
+            # does not implement unified exec and enabling a second executor
+            # would create a parallel permission surface.
+            "features.unified_exec=false",
             # The Python SDK does not provide Codex's host-side JavaScript
             # executor. Force native shell/file tools instead of code-mode
             # calls such as `exec -> tools.exec_command(...)` that cannot be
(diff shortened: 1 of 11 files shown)


kbc/platform/pod/codex_engine.py
```diff
@@ -0,0 +1,467 @@
...
+            ephemeral=True,
+            model=self.model,
+            model_provider="kbc_mass",
+            # Restricted Kubernetes container hosts can refuse Codex's nested
+            # bubblewrap split policies. The box itself contains one KB run and
+            # is already the process/filesystem boundary, so both writer and
+            # staged closed-book reviewers use the Claude-parity full-access
+            # preset inside that Pod. Model subprocesses still receive only
+            # PATH/HOME, never the provider key/token.
+            sandbox=Sandbox.full_access,
```
jacoblee-io: **[High] Under Codex the read-only / blind-review confinement is advisory, not enforced.** `sandbox=Sandbox.full_access` is hardcoded here regardless of the `read_only` constructor arg (which is stored at L175 but never applied anywhere), and there is no PreToolUse-equivalent path guard — whereas the Claude path enforces confinement mechanically via `_make_test_path_guard` / `_make_multiroot_guard` / `_make_compile_path_guard`. Per OpenAI's docs, no Codex sandbox mode confines *reads* on Linux/Landlock, and `full_access` removes filesystem+network boundaries entirely; with `features.shell_tool=true`, the model can read absolute paths anywhere in the pod.

Impact by call site:
- `test_session_driver` (compile_box.py:3186-3194): Codex branch is `cwd=run.cwd` with **no staging and no guard**, while the Claude branch's C4 comment says reads "must not escape the snapshot to the live /work draft." Under Codex the blind consumer can read `/work` directly — C4 defeated, the closed-book consumer can see the answer it should be blind to.
- recommendation (~3395) / reference-assist (~3660): `isolated_readonly_workspace` protects inodes from writes and sets the cwd view, but full_access still permits absolute-path reads of the original `/work/raw`, `/work/candidate`, and the `authoring/eval/release` siblings that `_recommendation_path_escape` was written to block. The comment at compile_box.py:3397 ("OS-enforced Codex workspace boundary instead of trusting model instructions/hooks") is therefore inaccurate.
- PK self-check (engine.py:256): confinement is a prompt line only.

The SDK sandbox can't confine reads on the target platform, so the correct fix mirrors the Claude approach: an in-process deny guard for Codex tool calls. At minimum, correct the `read_only` param and the "OS-enforced" comments so the guarantee isn't trusted where it doesn't hold.

kbc/platform/pod/codex_engine.py
```diff
@@ -0,0 +1,467 @@
...
+            # calls such as `exec -> tools.exec_command(...)` that cannot be
+            # serviced by this compile-box host.
+            "features.code_mode.enabled=false",
+            "features.shell_tool=true",
+            # Model-proposed shell commands receive no API key/token.  PATH and
+            # HOME are sufficient for the preinstalled deterministic KBC tools.
+            "shell_environment_policy.inherit=" + _toml("none"),
+            "shell_environment_policy.include_only=" + _toml(["PATH", "HOME"]),
+            "shell_environment_policy.set.PATH=" + _toml(os.environ.get("PATH", "/usr/local/bin:/usr/bin:/bin")),
+            "shell_environment_policy.set.HOME=" + _toml(self._codex_home),
```
jacoblee-io: **[High] Model shell's `HOME` is the same dir as `CODEX_HOME`, reopening on disk what the env scrub closes.** This sets `shell_environment_policy.set.HOME = self._codex_home`, and L273 sets `CODEX_HOME` to the same tmpdir. The credential-exclusion design correctly strips `OPENAI_API_KEY` from the model subprocess *environment*, but with `Sandbox.full_access` (L298) a model-proposed shell can `cat "$HOME"/*.json` / `find "$HOME"`. If the Codex runtime persists the resolved key, an auth.json, session transcripts, or provider caches under CODEX_HOME (common for agent runtimes), the key is readable despite the env scrub. Latent even if the current build doesn't persist it — a version bump silently activates it. Fix: give the model shell a separate empty HOME distinct from CODEX_HOME.

kbc/platform/pod/codex_engine.py
```diff
@@ -0,0 +1,467 @@
...
+
+def _status_from_error(value: object) -> int | None:
+    text = str(value or "")
+    match = re.search(r"(?:status|HTTP)\D{0,8}(429|503|529)\b", text, re.I)
+    return int(match.group(1)) if match else None
+
+
+def _safe_error_message(value: object) -> str:
+    text = str(getattr(value, "message", value) or "Codex turn failed")
+    text = re.sub(r"\bsk-[A-Za-z0-9_+\-/=]{8,}", "[REDACTED]", text)
```
jacoblee-io, line 98: **[Medium] Error redaction only matches `sk-` keys; non-`sk-` Mass keys survive into consumer-visible text.** The provider is `kbc_mass` with `requires_openai_auth=false`, and Mass gateway keys frequently are not `sk-`-prefixed. `_safe_error_message` output is put into an AssistantMessage/TextBlock and streamed to the consumer (L397 / L457). If an SDK exception embeds the key (401 body echo, provider dump, httpx repr), a non-`sk-` key passes through unredacted (only truncated), defeating the "llm block is never logged" contract. Fix: redact by the actual key *value* (keep the api_key from connect() and replace it with [REDACTED]), keeping the `sk-` heuristic as an extra net.

kbc/platform/pod/codex_engine.py
```diff
@@ -0,0 +1,467 @@
...
+        sockets = site._server.sockets if site._server else None
+        if not sockets:
+            await runner.cleanup()
+            raise RuntimeError("failed to bind Codex MCP callback listener")
+        self._callback_runner = runner
+        port = sockets[0].getsockname()[1]
+        return f"http://127.0.0.1:{port}/tool-call"
+
+    async def _handle_tool_call(self, request: web.Request) -> web.Response:
+        if request.headers.get("x-kbc-token") != self._callback_token:
```
jacoblee-io: **[Low] Token check is not constant-time.** `!=` on the callback token. Practical risk is low (loopback-bound, single-run single-tenant pod, 24-byte urlsafe token), but this is a security-token comparison — `secrets.compare_digest(request.headers.get("x-kbc-token", ""), self._callback_token)` is the one-line, review-clean form.

kbc/platform/pod/codex_engine.py
```diff
@@ -0,0 +1,467 @@
...
+        elif name == "TurnCompletedNotification":
+            turn = getattr(payload, "turn", None)
+            status = str(getattr(getattr(turn, "status", None), "value", getattr(turn, "status", "")))
+            error = getattr(turn, "error", None)
+            is_error = self._budget_exhausted or status not in {"completed", ""}
+            if is_error and not self._budget_exhausted:
+                await self._events.put(AssistantMessage([
+                    TextBlock("Codex turn failed: " + _safe_error_message(error))
+                ]))
+            await self._events.put(ResultMessage(
```
jacoblee-io: **[Low / defensive]** "Exactly one ResultMessage per prompt" is guaranteed only by SDK behavior — there's no per-turn latch. (The double-ResultMessage risk is refuted for the normal interrupt flow, since Codex's interrupt is decoupled and doesn't raise into `stream()`.) But if a surplus ResultMessage ever appears, it silently corrupts the next turn: `_events` is unbounded and every consumer stops on the first ResultMessage, so the extra one becomes the next turn's first message. Cheap insurance: a `_result_emitted` flag reset at L388 and checked before both `_events.put(ResultMessage(...))` sites (here and L399).

jacoblee-io (commented):
Reviewed the Codex engine end to end (engine.py / codex_engine.py / mcp_tool_server.py + the compile_box wiring), with the OpenAI Codex sandbox semantics checked against the docs. The engine-neutral architecture is well done — whitelisted engine selection that can't be flipped mid-run or via the KBC_* channel, the MCP bridge is arg-for-arg equivalent to the Claude in-process path (handlers coerce defensively), the feature-disable set is complete, and the env-based credential exclusion (inherit=none, include_only PATH/HOME) is well conceived. Findings below are all on the opt-in Codex path (default is claude_agent_sdk) and inside the single-run pod, so none is a cross-tenant/host escape — but #1 and #2 defeat, inside the pod, security controls the code itself claims to enforce. Details inline.

LikiosSedo:
Addressed all five review findings in 860e4530.

- Read-only Codex sessions now disable native shell/unified exec, select a custom exact-root filesystem permission profile, and expose only bounded KBC MCP read/glob/grep tools with traversal, absolute-path, and symlink confinement. All closed-book callers explicitly pass their staged snapshot root; writer sessions retain the existing Pod-level full-access behavior.
- Shell HOME is now a separate empty directory from CODEX_HOME, with login shells disabled.
- Error redaction removes the exact configured provider key (including non-sk Mass keys) and keeps the sk-pattern fallback.
- The loopback callback token uses secrets.compare_digest.
- A per-turn terminal latch now guarantees at most one ResultMessage even if a future SDK stream repeats completion.

Validation:
- real openai-codex 0.144.4 app-server startup with the exact-root profile and MCP bridge (no model request)
- all KBC Python suites, including compile_box, Codex adapter, red-blue, ingest/media/incremental/self-check
- npm test: 227 files, 4,482 passed, 2 skipped
- npx tsc --noEmit
- npm run build

Please re-review the updated head.

jacoblee-io (approved):
LGTM. The follow-up commits resolve all five review comments, and #1/#2 are fixed mechanically rather than by comment edits:

- **[High] Read-only confinement is now enforced, not advisory.** `Sandbox.full_access` is gone (`sandbox=None`) in favor of named permission profiles: `kbc_readonly` disables the native shell entirely (`features.shell_tool=false`), scopes filesystem to the declared roots as `read`, and denies network; `kbc_writer` confines writes to cwd + shell_home with `:minimal` read and no network. On top of that, closed-book sessions get KBC's own `_ReadOnlyFileAccess` MCP tools that `resolve()` + `relative_to(root)`-check every path (blocking `..`/symlink/absolute escapes) with bounded output. Every read-only call site now passes `allowed_read_roots` (test_session_driver → `[run.cwd]` so C4 genuinely holds; recommendation/reference-assist/PK → the staged workspace). The previously-inaccurate "OS-enforced boundary" claim is now true.
- **[High] Model shell HOME separated from CODEX_HOME.** New empty `_shell_home` (+ `allow_login_shell=false`), both cleaned on disconnect — the env scrub is no longer undermined by an on-disk key path.
- **[Medium] Value-based error redaction.** `_safe_error_message` now replaces the actual `api_key` before the `sk-` regex, so non-`sk-` Mass keys can't leak into consumer-visible text.
- **[Low] `secrets.compare_digest`** for the callback token.
- **[Low] Per-turn terminal latch** (`_result_emitted_this_turn` + `_emit_result`) makes ResultMessage one-shot.

Nice extra hardening too: `features.unified_exec=false`, unique/reserved MCP tool-name checks, and bounded read/glob/grep output. Build side is wired (Dockerfile installs bubblewrap; CI relaxes unprivileged-userns).

One deployment prerequisite to confirm (not blocking): the WRITER profile now depends on bubblewrap/userns — the very thing that motivated the original full_access. The read-only path is bwrap-independent (no shell), so closed-book confinement holds everywhere, but please confirm the production K8s host permits unprivileged user namespaces (or that a missing sandbox fails closed rather than silently running the writer unsandboxed). The PR's sicore-siflow-test deployment suggests it's fine there. Ship it.

PR #420, opened 2026-07-18
feat(mcp): add Sicore A2A adapter for coding agents

LikiosSedo:
## Summary

Add a minimal local stdio MCP adapter that lets Codex and Claude Code reuse one Sicore-hosted Siclaw agent through the existing A2A task API.

### Problem

Coding agents need a safe way to delegate production investigations without receiving cluster credentials or choosing a different Siclaw agent at runtime. Long investigations must survive client restarts without repeatedly injecting the growing partial report into model context.

### Solution

- fix each adapter process to one Sicore URL, agent ID, and private A2A key file
- expose investigate, bounded watchdog wait, snapshot, list/recovery, and cancel tools
- advertise same-turn watchdog instructions through MCP initialization
- keep working responses compact and return the full report only at terminal state
- validate key-file permissions and avoid retrying task submission

No Sicore backend change is required: the existing A2A API already owns durable task state, progress, recovery, and cancellation.

## Test Plan

- [x] `npm ci`
- [x] `npm test` (20 tests)
- [x] `npm audit --omit=dev` (0 vulnerabilities)
- [x] Real Sicore A2A smoke: working task returned no partial result, then `siclaw_wait_task` returned `WATCHDOG_READY` on the same task ID

### DIFF
--- a/mcp/sicore-a2a-adapter/src/config.ts
+++ b/mcp/sicore-a2a-adapter/src/config.ts
@@ -1,0 +1,107 @@
+import { readFileSync, statSync } from "node:fs";
+
+export interface AdapterConfig {
+  baseUrl: string;
+  agentId: string;
+  apiKey: string;
+  requestTimeoutMs: number;
+  pollIntervalMs: number;
+}
+
+export class ConfigError extends Error {
+  constructor(message: string) {
+    super(message);
+    this.name = "ConfigError";
+  }
+}
+
+function required(env: NodeJS.ProcessEnv, name: string): string {
+  const value = env[name]?.trim();
+  if (!value) throw new ConfigError(`${name} is required`);
+  return value;
+}
+
+function parseBoundedInteger(
+  env: NodeJS.ProcessEnv,
+  name: string,
+  fallback: number,
+  min: number,
+  max: number,
+): number {
+  const raw = env[name]?.trim();
+  if (!raw) return fallback;
+  const value = Number(raw);
+  if (!Number.isInteger(value) || value < min || value > max) {
+    throw new ConfigError(`${name} must be an integer between ${min} and ${max}`);
+  }
+  return value;
+}
+
+function isLoopback(hostname: string): boolean {
+  return hostname === "localhost" || hostname === "127.0.0.1" || hostname === "::1";
+}
+
+function normalizeBaseUrl(raw: string): string {
+  let url: URL;
+  try {
+    url = new URL(raw);
+  } catch {
+    throw new ConfigError("SICORE_URL must be an absolute URL");
+  }
+  if (url.username || url.password) {
+    throw new ConfigError("SICORE_URL must not contain credentials");
+  }
+  if (url.search || url.hash) {
+    throw new ConfigError("SICORE_URL must not contain a query string or fragment");
+  }
+  if (url.protocol !== "https:" && !(url.protocol === "http:" && isLoopback(url.hostname))) {
+    throw new ConfigError("SICORE_URL must use HTTPS (HTTP is allowed only for loopback testing)");
+  }
+  return url.toString().replace(/\/$/, "");
+}
+
+function loadKeyFromFile(path: string): string {
+  let stat;
+  try {
+    stat = statSync(path);
+  } catch {
+    throw new ConfigError("SICLAW_A2A_KEY_FILE could not be read");
+  }
+  if (!stat.isFile()) throw new ConfigError("SICLAW_A2A_KEY_FILE must point to a regular file");
+  if (process.platform !== "win32" && (stat.mode & 0o077) !== 0) {
+    throw new ConfigError("SICLAW_A2A_KEY_FILE must not be readable or writable by group/others (use chmod 600)");
+  }
+  let key: string;
+  try {
+    key = readFileSync(path, "utf8").trim();
+  } catch {
+    throw new ConfigError("SICLAW_A2A_KEY_FILE could not be read");
+  }
+  if (!key) throw new ConfigError("SICLAW_A2A_KEY_FILE is empty");
+  return key;
+}
+
+function loadApiKey(env: NodeJS.ProcessEnv): string {
+  const direct = env.SICLAW_A2A_KEY?.trim();
+  const keyFile = env.SICLAW_A2A_KEY_FILE?.trim();
+  if (direct && keyFile) {
+    throw new ConfigError("Set only one of SICLAW_A2A_KEY or SICLAW_A2A_KEY_FILE");
+  }
+  if (keyFile) return loadKeyFromFile(keyFile);
+  if (direct) return direct;
+  throw new ConfigError("SICLAW_A2A_KEY or SICLAW_A2A_KEY_FILE is required");
+}
+
+export function loadConfig(env: NodeJS.ProcessEnv = process.env): AdapterConfig {
+  const agentId = required(env, "SICLAW_AGENT_ID");
+  if (Buffer.byteLength(agentId, "utf8") > 255) {
+    throw new ConfigError("SICLAW_AGENT_ID must be 255 bytes or less");
+  }
+  return {
+    baseUrl: normalizeBaseUrl(required(env, "SICORE_URL")),
+    agentId,
+    apiKey: loadApiKey(env),
+    requestTimeoutMs: parseBoundedInteger(env, "SICLAW_A2A_TIMEOUT_MS", 30_000, 1_000, 120_000),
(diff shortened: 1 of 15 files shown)


LikiosSedo:
Pushed `c4b7f268` (and rebased the branch onto latest `main`, 29 commits ahead — all adapter files are new paths, so the rebase was clean).

**Fix:** when `message:send` succeeds but the bounded wait that follows fails entirely (e.g. transient network loss right after submission), `siclaw_investigate` used to return a plain tool error **without the task_id**. The task already exists server-side, so that error path invited the client model to resubmit a duplicate investigation — exactly what the adapter is designed to prevent. It now returns the submitted task as a compact working view with a `wait_error` note, keeping the task_id and the "call siclaw_wait_task, do not resubmit" instruction. Added a regression test (25 tests total, all passing).

LikiosSedo:
Pushed `659795a0`: `SICLAW_AGENT_ID` is now **optional**. Agent API keys are per-agent, so the adapter resolves the agent from the key at startup via `GET /api/v1/a2a/self` (added in sicore MR `feat/a2a-key-self-endpoint`, pending merge) — operators no longer copy the agent UUID into MCP config. Setting `SICLAW_AGENT_ID` still pins the agent explicitly, and against an older Sicore without `/self` the adapter fails fast with an actionable message telling you to set it.

Mechanics: the fetch/error-mapping and bounded idempotent-read retry were lifted to module-level (`a2aRequest` / `retryIdempotentRead`) so `resolveAgentId` reuses them; `SicoreA2aClient` now takes a `ResolvedAdapterConfig` (agentId guaranteed). README examples updated. 30 tests passing, including a new stdio e2e that boots the adapter with no `SICLAW_AGENT_ID` and verifies `/self` → `/agents/agent-resolved/message:send`.

Deploy order: sicore `/self` first; existing configs with `SICLAW_AGENT_ID` set are unaffected either way.

mcp/sicore-a2a-adapter/src/config.ts
```diff
@@ -0,0 +1,110 @@
...
+
+function loadKeyFromFile(path: string): string {
+  let stat;
+  try {
+    stat = statSync(path);
+  } catch {
+    throw new ConfigError("SICLAW_A2A_KEY_FILE could not be read");
+  }
+  if (!stat.isFile()) throw new ConfigError("SICLAW_A2A_KEY_FILE must point to a regular file");
+  if (process.platform !== "win32" && (stat.mode & 0o077) !== 0) {
```
jacoblee-io, line 74: Optional (non-blocking): the key-file check enforces mode `& 0o077 == 0` and isFile — solid — but doesn't validate the file's owner or the parent directory's perms, so a 0600 key in a world-writable dir (swappable by another user) or a 0600 file owned by someone else would still pass. That's a standard limitation of file-mode checks and fine for a client-side adapter; worth at most a one-line note in the README rather than code.

mcp/sicore-a2a-adapter/src/tools.ts
```diff
@@ -0,0 +1,248 @@
...
+  return value as Record<string, unknown>;
+}
+
+function stringArg(args: Record<string, unknown>, name: string, required = false): string | undefined {
+  const value = args[name];
+  if (value === undefined && !required) return undefined;
+  if (typeof value !== "string" || !value.trim()) throw new Error(`${name} must be a non-empty string`);
+  const trimmed = value.trim();
+  if (Buffer.byteLength(trimmed, "utf8") > 255 && name !== "question") {
+    throw new Error(`${name} must be 255 bytes or less`);
```
jacoblee-io, line 138: Optional (non-blocking) parity nit: the tool inputSchemas cap these ids with `maxLength: 255` (counted in UTF-16 code units / chars), while `stringArg` here rejects `> 255 bytes`. A multibyte 200-char value would pass schema validation but throw in the handler. Only affects context_id/task_id/status, which are server-generated ASCII UUIDs / a fixed enum in practice, so it's not currently triggerable — just note it if you want strict schema/handler parity.

jacoblee-io (approved):
LGTM. Reviewed the whole adapter for security and correctness (two independent sweeps). It's clean and well-scoped.

Security: the API key never leaves the `Authorization: Bearer` header — no log/error/tool-output path interpolates it (stderr prints only agentId + URL.origin, and normalizeBaseUrl rejects credential-bearing URLs). HTTPS is enforced except loopback, so the Bearer key only travels over TLS or loopback. No SSRF/path-injection: every model-controlled value (task_id/context_id/status/page tokens) is encodeURIComponent/URLSearchParams-encoded, so a task_id like `../../other-agent/tasks` stays a single escaped segment and can't cross agents/hosts. Nothing from a Sicore response is interpreted locally (no eval/exec/spawn/writeFile; pure JSON parse + type-guarded normalization). Supply chain is clean and isolated: no install lifecycle scripts anywhere in the lockfile, single official runtime dep, own package/lock/tsconfig, no workspace hoist into the main repo.

Correctness: submit is never retried (no duplicate investigations; a submit-timeout is a documented tradeoff recoverable via siclaw_list_tasks); the retry/deadline math is boundary-correct (no single read exceeds the wait deadline, no over-deadline retry, no infinite loop, waitForTask always does at least one read); the compact-vs-full contract holds (working tasks expose only progress_chars, terminal returns the full result exactly once, list never includes result); state mapping is self-consistent and unknown states degrade gracefully instead of hanging; error handling never masks a terminal failure.

Two optional, non-blocking nits inline. Ship it.

chent1996 (approved):
## Review

Approving. Self-contained, purely additive leaf package (`mcp/sicore-a2a-adapter/`, zero overlap with `main` — the stale base is irrelevant), and the design goals stated in the description are all actually enforced in code, not just described. One process gap worth fixing (CI, note 1).

### What I verified

**Duplicate-investigation prevention has two independent layers.** `retryIdempotentRead` wraps only the GET paths (`loadTask`, `listTasks`, `resolveAgentId`); `sendMessage` goes through a bare request with no retry, so a transient failure surfaces instead of double-creating a task. And when submission *succeeds* but the follow-up wait fails, the handler returns the submitted task snapshot with a `wait_error` annotation instead of throwing — the inline comment nails why: a plain error would drop the `task_id` and invite the client to resubmit a duplicate. Both layers are tested.

**Context discipline is real.** `taskView` strips `result` and exposes only `progress_chars` for non-terminal tasks (the text response says "partial report withheld until terminal" explicitly), and `siclaw_list_tasks` uses the stripped view too — a long investigation's growing partial report never re-enters the model context on polls; the full report arrives exactly once, at terminal state.

**Key hygiene.** HTTPS enforced (HTTP only for loopback), URL rejected if it carries credentials/query/fragment, key file must be a regular file with `0600` perms (`mode & 0o077` check), key-env vs key-file mutual exclusion, and the key never appears in error messages or the startup line (which logs only origin + agentId).

**Bounded waiting done right.** `waitForTask` runs against a hard deadline, tolerates retriable read errors within it, returns the last snapshot at deadline rather than losing state, and both wait ceilings (50s) are documented in the schemas as deliberately below common MCP client request timeouts. The same-turn watchdog contract is advertised both in `SERVER_INSTRUCTIONS` at MCP init and per-tool descriptions, consistently ("never resubmit merely because a task is still working"; `siclaw_list_tasks` for restart recovery).

**Robust edges.** Tool-layer errors map to `isError` text results (the server never crashes on a bad call); unknown tool → error result; `resolveAgentId` handles a Sicore without `/self` with an actionable message (set `SICLAW_AGENT_ID`); fatal startup errors exit 1 via stderr; SIGINT/SIGTERM close the server. Dependency direction stays clean: nothing under `src/` imports this package; it consumes siclaw's own A2A wire, parameterized by a Sicore URL.

### Notes

1. **No CI runs this package's 20 tests.** Root `package.json` and `.github/` are untouched, so `npm test` at the root never reaches `mcp/`, and there's no path-filtered workflow. The repo already has the exact precedent (`kbc-ci.yml`); a ~20-line workflow filtered on `mcp/sicore-a2a-adapter/**` would keep these tests from silently rotting. This is my one real ask.
2. A one-line CLAUDE.md Change Impact Matrix row ("standalone leaf; own `npm test`; no `src/` imports; speaks the A2A wire served by a2a-gateway") would tell future contributors that A2A wire changes need a look here too.
3. `cancelTask` is intentionally non-retried (POST); a transient 503 means the caller re-invokes — fine, just noting the asymmetry with reads is deliberate.
4. `question` is capped client-side at 512 KiB — generous, but the server owns the real limit; the pre-check just fails fast. Fine.

### Verdict

Approved. The three properties that matter for this adapter — never duplicate an investigation, never flood the client context with partial reports, never leak the key — are each enforced mechanically and tested, not just promised in prose. Please add the path-filtered CI workflow (note 1) in this PR or immediately after.

PR #421, opened 2026-07-18
feat(kbc): add resumable Source Snapshot v2 delivery

LikiosSedo:
## Summary

Add Source Snapshot v2 so the runtime can materialize large immutable KB Raw inputs as independently verified, resumable parts instead of one monolithic base64 request. The legacy `/sources` path remains unchanged for small inputs and rolling upgrades.

### Problem

The current capability input path serializes the entire frozen Raw workspace into one tar.gz/base64 payload. The real GPU workspace is 1,773 files and 628,153,605 raw bytes, which makes a single request memory-heavy and fragile to retry.

### Solution

- Add a versioned source descriptor and `source-part` capability fetch contract.
- Add `/sources/begin`, `/sources/state`, `/sources/part`, and `/sources/commit` to the compile box.
- Persist received parts on disk, request only missing parts after retries/restarts, and make every operation idempotent.
- Verify descriptor, compressed-part, and per-file SHA-256 values; reject undeclared paths and every non-regular tar entry.
- Assemble into staging and atomically replace `raw/` only after all parts validate; preserve the previous Raw on validation or commit-marker failure.
- Keep v1 delivery behavior intact.

The paired [Sicore MR !664](https://gitlab.scitix-inner.ai/k8s/sicore/-/merge_requests/664) sender change must be deployed only after this receiver is available.

## Test Plan

- [x] `npm test` passes (227 files, 4,485 passed, 2 skipped)
- [x] `npm run build` passes
- [x] Compile-box HTTP protocol smoke passes, including resume, incomplete-commit rejection, atomic commit, and idempotent replay
- [x] Python Source Snapshot unit tests pass
- [x] Manual verification: the real GPU Raw directory (1,773 files, 628,153,605 bytes) installs successfully in 20 parts without a monolithic request

## Architecture Checklist

- [ ] **Deployment mode**: Not deployed in this PR; local compile-box protocol and filesystem behavior are verified. K8s acceptance belongs to the paired rollout after both repositories merge.
- [x] **Security model**: No new shell execution path; archive input is size-bounded, manifest-bound, path-confined, and SHA-256 verified.
- [x] **DB parity**: Not applicable; Siclaw schema is unchanged.
- [x] **Tool protocol**: Not applicable; no agent tool was added.

## General Checklist

- [x] Commit messages follow [Conventional Commits](https://www.conventionalcommits.org/)

### DIFF
--- a/kbc/platform/pod/compile_box.py
+++ b/kbc/platform/pod/compile_box.py
@@ -670,11 +670,14 @@ def _install_source_bundle(bundle: bytes, workdir: str, expected_sha256: str | N
     }
 
 
-def _safe_authoring_path(name: str) -> Path:
+def _safe_authoring_path(name: str, *, is_dir: bool = False) -> Path:
     rel = _safe_tar_path(name)
+    allowed_roots = {"authoring", "candidate", "eval", "release"}
+    if len(rel.parts) == 1 and is_dir and rel.parts[0] in allowed_roots:
+        return rel
     if len(rel.parts) < 2:
         raise ValueError(f"unsafe authoring path {name!r}: must include a file path under a workspace directory")
-    if rel.parts[0] not in {"authoring", "candidate", "eval", "release"}:
+    if rel.parts[0] not in allowed_roots:
         raise ValueError(f"unsafe authoring path {name!r}: must be under authoring/, candidate/, eval/, or release/")
     return rel
 
@@ -703,7 +706,7 @@ def _install_authoring_bundle(bundle: bytes, workdir: str, expected_sha256: str 
             raise ValueError(f"invalid authoring bundle: {e}") from e
         with tf:
             for member in tf.getmembers():
-                rel = _safe_authoring_path(member.name)
+                rel = _safe_authoring_path(member.name, is_dir=member.isdir())
                 target = staging / Path(*rel.parts)
                 resolved = target.resolve(strict=False)
                 try:
--- a/kbc/platform/pod/compile_box.py
+++ b/kbc/platform/pod/compile_box.py
@@ -3837,16 +3837,15 @@ async def handle_session(request: web.Request):
 
 
 async def handle_sources(request: web.Request):
-    body = await request.json()
-    run_id = body.get("run_id")
-    if run_id and run_id in RUNS:
-        return web.json_response({"error": "run already exists; upload sources before /session", "run_id": run_id}, status=409)
-
-    encoded = body.get("bundle_base64") or body.get("bundle_b64")
-    if not encoded:
-        return web.json_response({"error": "bundle_base64 is required"}, status=400)
-
     try:
+        body = await _source_request_body(request)
+        run_id = body.get("run_id")
+        conflict = _source_live_conflict(run_id)
+        if conflict is not None:
+            return conflict
+        encoded = body.get("bundle_base64") or body.get("bundle_b64")
+        if not encoded:
+            return web.json_response({"error": "bundle_base64 is required"}, status=400)
         bundle = base64.b64decode(encoded, validate=True)
         result = _install_source_bundle(bundle, body.get("workdir", "/work"), body.get("bundle_sha256"), locale=body.get("locale"))
     except (ValueError, TypeError) as e:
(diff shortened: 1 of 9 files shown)


src/gateway/capability/materialize.ts
```diff
@@ -120,15 +190,20 @@ export async function materializeCapabilityInputs(opts: {
...
   try {
-    await client.postJson("/sources", {
-      run_id: runId,
-      bundle_base64: src.bundle_base64,
-      bundle_sha256: src.bundle_sha256,
-      locale: src.locale,
-    });
+    if (src.source_snapshot) {
+      if (src.bundle_base64) throw new Error("source response cannot contain both bundle_base64 and source_snapshot");
+      await installSourceSnapshot({ client, backend, runId, src, snapshot: src.source_snapshot });
```
jacoblee-io, line 206: **[High] A v2 SnapshotConflict 409 is silently swallowed as "box already live".** These begin/part/commit calls run under the outer try whose catch (~L207) uses `isBoxAlreadyLive`, which returns true for any `status === 409`. But the box returns 409 for every v2 `SnapshotConflict` (descriptor changed, corrupt/incomplete persisted part, not-initialized). So a genuine install failure — exactly the resume/restart path this PR exists to make robust — is reported as a healthy reattach: materialization is skipped and the session starts over an incomplete/empty `raw/`, surfacing much later as "missing source files" far from the real cause. It also swallows every in-`try` throw in this v2 branch (revision-required, unknown-part, hash-mismatch) as reattach instead of a `source-install` error.

The fix has a subtlety: the v2 branch *also* legitimately emits a "run already exists" 409 via the box's `_snapshot_live_conflict`, so you can't just make v2 never-reattach. The two 409 flavors differ only by body/message, not status. Suggest tagging the true reattach 409 with a distinct `code` on the box side and matching that code here, rather than bare `status === 409`.

kbc/platform/pod/compile_box.py
```diff
@@ -3610,6 +3618,90 @@ async def handle_sources(request: web.Request):
...
+
+
+def _snapshot_live_conflict(run_id: str | None):
+    if run_id and run_id in RUNS:
+        return web.json_response({"error": "run already exists; upload sources before /session", "run_id": run_id}, status=409)
+    return None
+
+
+async def handle_sources_begin(request: web.Request):
+    body = await request.json()
```
jacoblee-io: **[Medium] Malformed/empty JSON body → 500, not 400** (same in all five source handlers: 3601 v1, 3632/3647/3659/3685 v2). `body = await request.json()` sits outside the try/except, so a truncated/empty body raises `JSONDecodeError` before the handler's error mapping and, with no error middleware, becomes a bare 500 — making the sender treat a client-side bad request as a box fault and potentially retry something that can never succeed. Not a v1 regression, but v2 was the moment to adopt the file's own established idiom (`request.body_exists` guard, or wrapping `request.json()` in a `try → 400`, as at 3875-3877 / 4102). Pull the `request.json()` into the try.

kbc/platform/pod/source_snapshot.py
```diff
@@ -0,0 +1,463 @@
...
+
+
+def _remove_path(path: Path) -> None:
+    if path.is_dir() and not path.is_symlink():
+        shutil.rmtree(path)
+    elif path.exists() or path.is_symlink():
+        path.unlink()
+
+
+def _swap_raw(workdir: Path, staging: Path, mark_committed: Callable[[], None]) -> None:
```
jacoblee-io, line 373: **[Low / latent hardening]** `_swap_raw` renames the live `raw` out to a randomly-named backup, then renames staging in, then restores on failure — this is correct **only** because `commit_snapshot` is fully synchronous (no `await`), so on aiohttp's single event loop two commits can't interleave, and commit only runs while the run is not live (guarded by `_snapshot_live_conflict`), so no compiler reads `raw` mid-swap. That safety is an unstated invariant: the moment anyone makes commit/office-conversion `async` or offloads it via `run_in_executor`, this becomes a real data-loss race (`had_raw` is a point-in-time snapshot; a second swapper sees `raw` missing, skips backup, and a failing first swapper's rollback can delete the other's committed tree — plus a window where `raw` doesn't exist at all). Worth a per-`workdir` lock, or at minimum a comment pinning the "commit_snapshot must stay synchronous" contract. (Separately: the synchronous office conversion of a ~628MB tree blocks the whole event loop for its duration.)

jacoblee-io (commented):
Reviewed the Source Snapshot v2 delivery end to end (Python module, box handlers, TS orchestration). The untrusted-archive defenses are solid: path-traversal/zip-slip (`_safe_path` + per-member `resolve().relative_to()` + `isfile()`-only, rejecting symlink/hardlink/dir/device entries), decompression-bomb bounding (each member size reconciled against the descriptor, totals capped in `validate_snapshot`), and manifest+per-file hash binding all hold. Idempotency/resume is correct. One real HIGH bug worth fixing before merge (the v2 409 misclassification), one medium robustness wart, and one latent hardening note — inline.

LikiosSedo:
Addressed all three review findings in 70212b6f:

- Added a stable `KBC_RUN_ALREADY_LIVE` conflict code and now treat only tagged/legacy-exact live-run conflicts as reattach; Source Snapshot v2 409 conflicts fail closed.
- Return HTTP 400 for empty, malformed, or non-object JSON across all five source endpoints.
- Pinned the synchronous event-loop swap invariant and the locking requirement if that path becomes async.

Verification: full `npm test` (227 files, 4487 passed, 2 skipped), `npm run build`, compile-box protocol smoke, and Source Snapshot tests all pass.

@jacoblee-io please re-review the updated head.

jacoblee-io (dismissed):
LGTM. Commit 70212b6f resolves all three review comments, each with a regression test:

- **[High] v2 SnapshotConflict 409 no longer misclassified as reattach.** The box now returns a structured `KBC_RUN_ALREADY_LIVE` code, and `isBoxAlreadyLive` reattaches only on `status === 409 && code === KBC_RUN_ALREADY_LIVE` (with an exact-legacy-text fallback for pre-code boxes). A v2 conflict 409 (different text, no code) now correctly becomes a `source-install` error. Locked by the new `"Source Snapshot v2 treats an untagged 409 as an install conflict, not a live reattach"` test (rejects + stage:source-install + posts:[]) and a client.test.ts case asserting `code/retriable/details` pass through. This is the distinct-code fix I suggested, and it stays rolling-upgrade compatible.
- **[Medium] Malformed/empty JSON body → 400, not 500.** New `_source_request_body` (body_exists guard + JSON error → ValueError/TypeError) is pulled into each handler's try, mapped to 400; covered by the new 5-route × empty/garbage-body assertions.
- **[Low] `_swap_raw`** now carries the must-stay-synchronous concurrency contract comment.

Also re-checked the two incidental commits: d11deadf (Go `encoding/json` manifest escaping) is correct interop and only touches already-serialized text; 431b4cb6 (authoring tar roots) is small, tested, and unrelated to the snapshot path. The untrusted-archive defenses (path confinement, tar entry rejection, per-file + manifest SHA-256, atomic swap/rollback) remain sound. Ship it.

jacoblee-io (approved):
Re-approving after the rebase onto main (which now includes the merged #419). Verified this is a pure rebase with no content change to the Source Snapshot v2 work:

- `source_snapshot.py` is byte-identical to the previously-approved revision (empty diff).
- `materialize.ts` is byte-identical — the v2 `KBC_RUN_ALREADY_LIVE` conflict-vs-reattach fix is preserved.
- The compile_box source handlers (`_source_request_body` / `_source_live_conflict` / `KBC_RUN_ALREADY_LIVE`) are all intact — the JSON-body→400 and structured-conflict-code fixes stand.
- Clean merge with #419: zero conflict markers in any changed file; #421's own delta touches only the source-snapshot files and does not modify the Codex engine code.

All three earlier findings (v2 409 misclassification, malformed-body 500→400, and the _swap_raw synchronous-contract note) remain resolved with their regression tests. Ship it.

PR #422, opened 2026-07-18
feat(kbc): add hierarchical compilation for very large corpora

LikiosSedo:
## Summary

- preserve the existing single-session route at 400KB or below and the flat batch route through 8MB
- add deterministic document families so Markdown anchors stay with sibling .assets media
- split oversized families with explicit read-only anchor context, a 1MB effective budget, and a separate 400KB text cap
- add bounded source-section reduction before the existing global final review
- persist map, reduce, final, and complete phases so tail failures never replay completed map batches

## Regression boundary

- ordinary small corpora still use the original single-session path
- medium corpora still use the original flat planner and orchestrator behavior
- hierarchical planning activates only above the new 8MB effective threshold

## Verification

- KBC compile-box, batching, Office ingest, selfcheck, red-blue, mTLS, incremental, and media verification suites
- npm test: 227 files, 4482 passed, 2 skipped
- npm run build
- real GPU Raw dry-run: 1772 non-empty files, 628153605 raw bytes, 57192656 effective bytes; 290 flat batches reduced to 80 hierarchical batches; plan validation clean; largest text batch 313282 bytes under the 409600-byte cap

No deployment is included in this PR.

### DIFF
--- a/kbc/platform/pod/compile_box.py
+++ b/kbc/platform/pod/compile_box.py
@@ -1307,6 +1307,21 @@ def _media_verify_rounds() -> int:
     return max(1, int(os.environ.get("KBC_MEDIA_VERIFY_ROUNDS", "3")))
 
 
+def _batch_media_verify_round_limit(plan: dict, pending: dict[str, list[str]]) -> int:
+    """Keep the established tail bound for flat/smaller compiles, but give a
+    hierarchical train enough bounded rounds to visit every initially pending
+    image page (including its failed-attempt budget). A hard operator cap still
+    fences pathological repair loops; hierarchical completion fails closed if
+    pending work remains after that cap."""
+    ordinary = _media_verify_rounds()
+    if plan.get("mode") != "hierarchical":
+        return ordinary
+    cap = max(ordinary, int(os.environ.get(
+        "KBC_HIERARCHICAL_MEDIA_VERIFY_MAX_ROUNDS", "256")))
+    needed = len(pending) * _media_verify_attempts() + ordinary
+    return min(cap, max(ordinary, needed))
+
+
 def _settle_media_outcome(run, chunk: dict, result: dict | None) -> list[str]:
     """Post-verify bookkeeping (review fix: pages used to be marked verified
     BEFORE verification ran, so a total transcription failure shipped image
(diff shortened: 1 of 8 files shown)


kbc/platform/pod/compile_box.py
```diff
@@ -2216,6 +2330,71 @@ def _write_batch_file(run: "CompileRun", rel: str, value) -> None:
...
+        for source, source_range in ranges.items():
+            if not isinstance(source_range, dict):
+                raise BatchOutputError(f"invalid source range for {source}")
+            source_path = (raw_root / source).resolve()
+            try:
+                source_path.relative_to(raw_root)
+            except ValueError as error:
+                raise BatchOutputError(f"text slice source escapes Raw: {source}") from error
+            if not source_path.is_file():
+                raise BatchOutputError(f"text slice source is missing: {source}")
```
jacoblee-io, line 2369: **[Medium] A hierarchical text-slice source deleted from raw/ between runs makes resume permanently abort.** `_materialize_batch_slices` runs on every resume (called before the pending loop) and iterates **all** `plan["batches"]` including `status:"done"` ones, raising here when a done batch's `source_ranges` points at a file that no longer exists in raw/. But `prune_missing_sources` only strips `source_ranges` from **pending** batches (it `continue`s on done), so a done batch's stale slice ref survives pruning forever.

Trigger: a v3 (text-sliced) hierarchical plan; a sliced-anchor batch completes + syncs; the anchor .md is deleted/renamed in raw/ before a later resume (crash in reduce/final, or any re-trigger). Resume → `prune_missing_sources` skips the done batch → `_materialize_batch_slices` sees its stale range → `BatchOutputError("text slice source is missing")` → the whole train aborts, and does so on **every** subsequent trigger because the done stamp keeps pruning from ever clearing it. The map work is safely in the store but reduce/final can never run — an unrecoverable resume loop.

This defeats the exact case `prune_missing_sources` exists to tolerate. Fix (small): skip `status == "done"` batches in `_materialize_batch_slices` (a done batch has no future model turn, so its slices never need rebuilding) — or also strip done batches' `source_ranges` in `prune_missing_sources`.

kbc/platform/pod/compile_box.py
```diff
@@ -2478,17 +2743,21 @@ async def _run_ledger_repairs(run: "CompileRun", replies: list[str]) -> None:
...
     run._batch_active = True
+    previous_model_idle_timeout = run._model_idle_timeout_s
     replies: list[str] = []
     try:
         raw_dir = Path(run.workdir) / "raw"
         inventory = batching.scan_sources(raw_dir)
         total_kb = batching.corpus_bytes(inventory) // 1024
         plan = _load_batch_plan(run)
-        resuming = plan is not None and len(batching.pending_batches(plan)) > 0
+        resuming = plan is not None and (
```
jacoblee-io: **[Low / please confirm]** This resume gate correctly refuses to resume a `phase == "complete"` plan (good — no map/reduce replay). But when it evaluates false, control falls into the `if not resuming:` branch, which re-plans and overwrites BATCH_PLAN.json from scratch. So a genuinely-complete batch plan on disk, if `_run_batch_compile` is re-entered for a still-over-threshold corpus, would be discarded and fully recompiled. That's meant to be prevented upstream (the converge-phase "settled" gate), which is outside this function — just confirm a settled/complete train is a no-op on a benign re-trigger, otherwise it's a full double-compile.

jacoblee-io (commented):
Reviewed the hierarchical compilation path in depth (packing, plan validation, and the crash-resume state machine), including fuzzing the packer across ~60k random inventories end-to-end through validate_plan. The core is solid: family grouping + oversized-family split preserve exact coverage (every source in one accounting batch; context charged to budget but not to coverage; text slices cover 1..line_count contiguously), validate_plan's slice/duplicate/budget/defer checks are load-bearing and exception-safe, pack_section_reductions has no closure-over-loop-variable bug, and the regression boundary holds (≤8MB effective never enters the hierarchical packer; flat/single-session paths are byte-identical). One Medium resume-path bug and one low confirmation, inline.

jacoblee-io (approved):
LGTM. Commit 7621a7b6 resolves both resume findings, each with a regression test:

- **[Medium] Done-batch stale slice no longer aborts resume.** `_materialize_batch_slices` now `continue`s on `status == "done"`, so a text-slice source deleted from raw/ between runs can't wedge the train forever. The new assertion in `test_hierarchical_text_slice_materialization_and_directive` covers it exactly (mark batch done → unlink source → materialize → helper not created).
- **[Low] Resume gate unified across the three call sites.** The new `_batch_plan_resumable` predicate is now shared by the `compile.resume` command gate, the routing gate, and the orchestrator, so a typed resume can no longer 409-reject reduce/final work the orchestrator could recover (all map batches done, `pending==0`, phase still reduce/final), while a `complete` plan stays closed. `test_hierarchical_resume_state_contract` verifies map/reduce/final resume and complete → 409.

Rest of the PR was already verified sound this round: family grouping + oversized-family split preserve exact coverage, validate_plan's checks are load-bearing and exception-safe, and the regression boundary holds (≤8MB effective never enters the hierarchical packer) — confirmed by fuzzing the packer through validate_plan across ~60k random inventories with zero failures. Ship it.

PR #423, opened 2026-07-19
feat(kbc): support attachment-aware PDF batch slices

LikiosSedo:
## Summary

- group exported `_attachments` originals and sidecars with their exact document family using directory + normalized full stem + extension; ambiguous identities remain independent
- obtain authoritative PDF page counts with Poppler while preserving the existing small/medium effective-size routing signal
- split oversized PDFs into validated contiguous ranges of at most 20 pages and enforce the assigned `Read.pages` range with a compile hook
- keep batch Raw scope mechanical on both engines: Claude uses its PreToolUse guard; Codex denies original Raw and copies only the allowed non-sliced sources into a per-session read-only view
- expose a page-sliced PDF to Codex only through a fixed-range `read_assigned_pdf_pages` Poppler tool, while keeping the original PDF hidden from shell/file tools
- remove each temporary Codex source view after the batch, including client-construction and turn failures
- install `poppler-utils` so Claude Code can render PDF pages and Codex can extract only its assigned PDF text range in the compile image

## Verification

- `python kbc/platform/pod/test_batching.py`
- `uv run --python 3.11 --with-requirements kbc/platform/pod/requirements.txt python kbc/platform/pod/test_compile_box.py`
- `uv run --python 3.11 --with-requirements kbc/platform/pod/requirements.txt python kbc/platform/pod/test_codex_engine.py`
- `test_office_ingest.py`, `test_selfcheck.py`, `test_redblue.py`, and `test_mtls_auth.py`
- GitHub KBC Box CI on Ubuntu 24.04 exercises real bubblewrap commands: the batch view is readable but not writable, and original Raw remains unreadable
- full unified-hardware Raw plan: 1,772 sources, 20/20 PDFs counted (517 pages), 304 hierarchical batches, 26 PDF slice batches, zero validation errors
- test image built as linux/amd64 and deployed by digest to `sicore-siflow-test`
- live Claude Code probe issued `Read` with `pages: "53-53"` against the 112-page ASUS GPU manual and recovered the independently verified DMLAN IP/mask, section heading, and TPM pin designation

## Integration

Rebased onto `main` after #422 merged. The conflict resolution preserves the merged Snapshot v2, Codex engine, phase-aware resume, and completed-batch slice behavior.

### DIFF
--- a/kbc/platform/pod/Dockerfile
+++ b/kbc/platform/pod/Dockerfile
@@ -12,7 +12,7 @@
 #     kbc-compile-pod
 FROM python:3.11-slim
 
-RUN apt-get update && apt-get install -y --no-install-recommends git ca-certificates bubblewrap \
+RUN apt-get update && apt-get install -y --no-install-recommends git ca-certificates bubblewrap poppler-utils \
     && rm -rf /var/lib/apt/lists/*
 
 # Pin the SDK: an unpinned install drifts on every rebuild, and the bundled
@@ -20,7 +20,8 @@ RUN apt-get update && apt-get install -y --no-install-recommends git ca-certific
 # massapi/Bedrock path). 0.2.110 is production-proven on the KB compile path.
 # PyYAML backs deterministic OKF frontmatter validation. python-pptx/openpyxl/
 # python-docx let office_ingest pre-render .pptx/.xlsx/.docx sources to markdown
-# (pdf/text/images are read natively by the agent's Read tool).
+# (PDFs are rendered by the agent's Read tool through poppler-utils; text and
+# images are read directly).
 COPY platform/pod/requirements.txt /tmp/kbc-requirements.txt
 RUN pip install --no-cache-dir -r /tmp/kbc-requirements.txt
 
--- a/kbc/platform/pod/batching.py
+++ b/kbc/platform/pod/batching.py
@@ -79,11 +79,12 @@ def _hierarchical_text_slice_bytes() -> int:
 
 def hierarchical_pdf_slice_pages() -> int:
     """Claude Code's Read tool accepts at most 20 PDF pages per call. Keep the
-    planner limit explicit and configurable for provider-specific deployments,
-    but never allow a zero-width slice."""
-    return max(1, int(os.environ.get(
+    planner limit configurable downwards for tighter deployments, but never
+    let configuration exceed the provider's hard limit or create a zero-width
+    slice."""
+    return min(DEFAULT_HIERARCHICAL_PDF_SLICE_PAGES, max(1, int(os.environ.get(
         "KBC_HIERARCHICAL_PDF_SLICE_PAGES",
-        str(DEFAULT_HIERARCHICAL_PDF_SLICE_PAGES))))
+        str(DEFAULT_HIERARCHICAL_PDF_SLICE_PAGES)))))
 
 
 def _text_slices(path: Path, size: int) -> tuple[int, list[dict[str, int]]] | None:
--- a/kbc/platform/pod/batching.py
+++ b/kbc/platform/pod/batching.py
@@ -250,7 +250,9 @@ def reduction_budget_bytes() -> int:
         "KBC_REDUCTION_BUDGET_BYTES", str(DEFAULT_REDUCTION_BUDGET_BYTES)))
 
 
-def scan_sources(raw_dir: Path) -> list[dict[str, Any]]:
+def scan_sources(
+    raw_dir: Path, *, include_pdf_execution_metadata: bool = True,
+) -> list[dict[str, Any]]:
     """Inventory of raw sources: repo-relative path + size, sorted for stable
     plans. Hidden files and empty files are skipped (they carry no compile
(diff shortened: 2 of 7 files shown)


LikiosSedo:
Validation update for head 8a975716: the full compile_box suite and batching suite pass. A real 173 KB GPU server manual exposed that the first Markdown slice could read the original full Raw file and write facts through page 85; this head now gives each map session a mechanical Raw allowlist, omits the original file for text slices, preserves exact-page PDF access, and denies Raw reads during reduce/final. Re-running only the first slice with the built image produced citations through page 36 only, then finished with phase=complete, 1/1 source cited, coverage closed, and converge_phase=settled. The image was built from combined integration SHA fc66cef8 as linux/amd64 and the Pod imageID matched registry digest sha256:163b22de79b26f14830f2ca3c121de38b0ac21f3c59c88341af2a1e1dde4ed5e. This PR remains stacked on #422; merge #422 first, retarget this PR to main, and let main-based CI run before merging.

chent1996 (approved):
## Review

Approving. Reviewed against latest `main` (the branch is based on it). This is a well-disciplined change: every mechanism that widens what a batch session can touch comes with a mechanical enforcement counterpart, and every claim in the description maps to a pinned test.

### What I verified

**Invariant discipline (the best part of this PR).** The authoritative Poppler page count is used ONLY as execution metadata for the hierarchical planner — `effective_bytes` deliberately keeps the old `/Type /Page` marker estimator, so the established small/medium routing thresholds cannot move (`test_pdfinfo_metadata_does_not_move_effective_size_route` pins it). The batch-mode gate stays metadata-light (`include_pdf_execution_metadata=False`, pinned by `test_small_kb_batch_gate_skips_poppler_metadata`).

**Subprocess safety.** Both Poppler calls (`pdfinfo`, `pdftotext`) run as argv arrays with no shell — untrusted filenames can't inject. Timeouts (15s/60s), `LC_ALL=C` for stable parsing, output cap on extraction, graceful `None` fallbacks.

**Both engines enforce the page boundary mechanically, each in its native way:**
- *Claude*: the PreToolUse guard requires `Read.pages` to equal the assigned range exactly (missing `pages` also denied), on top of the batch Raw allowlist. `_raw_read_scope_violation` fails closed on Grep (missing path, workspace root, any ancestor of `raw/` all denied; inside `raw/` only exact allowlisted files pass); Glob leaks names only, and the comment says so explicitly.
- *Codex*: `raw/` is denied wholesale in the bubblewrap profile; allowed non-sliced sources are **copied** (not linked) into a session-scoped read-only view with full path confinement (resolve + `relative_to`, `..` rejection); the sliced PDF is reachable only through `read_assigned_pdf_pages`, a **zero-argument** tool whose `-f/-l` range is fixed at construction — the model physically cannot vary the range. Capability-as-constraint is exactly the right shape here.

**Path confinement and cleanup.** `_batch_raw_target` blocks lexical and symlink escape; the view is removed in `finally` covering client-construction and turn failures (view is built before the client, cleaned regardless); `BaseException` during materialization also cleans up. `writer_filesystem_access` in `codex_engine.py` validates the access enum, requires targets inside the workspace, and refuses to override the workspace root; read-only sessions reject it outright.

**Plan validation symmetry.** `validate_plan` extends the text-slice rules to page slices: single-source no-context slice batches, no mixing text/PDF ranges, span ≤ limit (env can only lower the 20-page provider cap, never raise it), contiguous 1..page_count coverage, sequential parts, defer-accounting boundary. Plan version bumps to 4 only when page slices exist; v2/v3 plans validate unchanged.

**Attachment families.** Identity is directory + NFC-casefolded full stem + real extension — no fuzzy title matching — and ambiguous identities are dropped rather than guessed (`test_hierarchical_family_does_not_guess_ambiguous_attachment_anchor`). Family ordering puts the original PDF before derived page images, with the rationale documented.

### Notes (non-blocking)

1. **`scan_sources`'s `include_pdf_execution_metadata` defaults to `True`**, while the stated principle is "routing scans stay cheap, Poppler belongs to the batch planner." Today the one routing call site remembers to pass `False` (and a test pins it), but the default invites a future caller to silently pay Poppler on every PDF. Consider flipping the default to `False` and opting in from the planner.
2. **Rollback caveat**: a v4 plan (with `source_page_ranges`) resumed under a pre-this-PR box image fails validation (the sliced PDF appears in N batches → duplicate-source errors) and forces a replan. That's the designed fail-safe, not corruption — just worth a line in the deploy notes given workspace rehydration can land an in-flight plan on whatever image the fleet runs.
3. The `pages` check is strict string equality — a batch assigned `41-60` cannot issue `Read(pages:"53-53")` for a single page within its range; it must request the full assigned range. Deterministic and clearly intentional (the directive says EXACT), just noting the rigidity is a choice, and a model that formats the range differently (whitespace, en-dash) will be denied and has to retry verbatim.

### Verdict

Approved. The two-engine enforcement is genuinely mechanical on both sides, the routing-signal invariant is protected by tests, and the subprocess/path-confinement hygiene is thorough. Box-image rebuild + redeploy required for any of this to take effect, per the standing kbc deploy note.

PR #424, opened 2026-07-20
fix(kbc): accept bounded large source snapshot parts

LikiosSedo:
## Summary

Allow Source Snapshot v2 to carry an accepted 200 MiB source file without weakening the total snapshot or HTTP request bounds. The receiver now uses a 256 MiB compressed/unpacked per-part envelope, matching Sicore sender MR !674.

### Problem

A real 1.30 GB managed Feishu Raw snapshot contained 143 MB and 140 MB PDFs. Sicore correctly kept each file whole, but the compile-box receiver rejected the resulting 140,483,490-byte compressed part at its old 64 MiB boundary before compilation could begin.

### Solution

Keep the 32 MiB normal part target, but permit one large accepted file to occupy a bounded 256 MiB part. This covers the 200 MiB source-file limit plus tar/gzip overhead and remains below the existing 768 MiB HTTP request ceiling after JSON/base64 expansion.

## Rollout

Deploy this receiver before the matching Sicore sender change in !674.

## Test Plan

- [x] `python3 -m unittest test_source_snapshot.py` (12 tests, 1 opt-in corpus test skipped)
- [x] Python syntax compilation
- [ ] Live 1.30 GB GPU knowledge-base compile after exact runtime/compile image deployment


chent1996 (approved):
## Review

Approving. Reviewed against latest `main` (branch is based on it). The PR is better than its own description: the headline is a bounds bump, but the load-bearing work is the Office-ingest hardening that makes the bump safe. Both sides of the box↔runtime contract move together in one PR, with the deploy ordering stated (receiver before Sicore sender !674).

### What I verified

**The bounds change is minimal and stays bounded.** Per-part envelope 64→256 MiB compressed and 128→256 MiB declared-unpacked, as named constants with the sizing rationale (200 MiB admitted file + tar/gzip overhead, under the unchanged 768 MiB HTTP ceiling after base64/JSON expansion). Total-unpacked (2 GiB), part/file counts, and all env overrides are untouched. Rollback is clean: a new sender against an old receiver fails the oversized part at the old 64 MiB boundary — a bounded rejection, not corruption.

**The timeout plumbing is real end-to-end.** I specifically checked that `AgentBoxClient.postJson` (not in this diff) already accepts the per-call timeout on `main` and threads it through both the plain-fetch and mTLS `httpsRequest` paths — so the 10 min part-transfer / 15 min commit timeouts in `materialize.ts` genuinely govern the box upload, not just the consumer RPC. `FrontendWsClient.request` gains a backward-compatible per-call override (default unchanged, timer message reports the effective value; test pins it).

**The Office-ingest rewrite is the security counterpart to the bigger envelope.** Raising the part cap means a ~200 MiB `.xlsx/.pptx/.docx` can now arrive, which turns sidecar pre-rendering into a decompression-bomb surface. The rewrite closes it: zip-archive budgets checked before parsing (entries ≤10k, declared-unpacked ≤512 MiB), streaming line converters instead of whole-document strings, a 512 MiB total derived-text budget enforced on actual output bytes (which backstops any lie in the zip's declared sizes), and atomic temp-file + `os.replace` writes with cleanup on all failure paths.

**The right fail-fast/fail-open split, with atomicity preserved.** Budget violations now fail the snapshot commit (a supported source silently unreadable is worse than a rejected commit) while corrupt individual files keep the historical fail-open skip. Because conversion can now raise, it was correctly moved *inside* the staging directory before the atomic rename — a budget failure can no longer replace the previous `raw/` with a half-installed tree, and `test_office_budget_failure_never_changes_existing_raw` pins exactly that. Nice archaeology too: the valid-but-empty Office file keeps its historical one-newline sidecar contract, with a test.

**Cross-layer capacity holds.** The kb-compile `/work` emptyDir is already 4Gi on `main` (2 GiB raw + ≤512 MiB sidecars + candidate output fit), and the helm changes give the runtime headroom for the transient ~340 MB base64/JSON buffers (`NODE_OPTIONS --max-old-space-size=1024`, memory request 512Mi, user-overridable).

### Notes (non-blocking)

1. The PR description undersells `office_ingest.py` (+134/-37 — the bulk of the diff). The README documents the new budgets well; a sentence in the PR body would help future archaeology, since the budgets are the reason the envelope bump is safe.
2. `_validate_archive_budget` trusts the zip's *declared* sizes; that's acceptable because Python's zipfile truncates reads at declared sizes and the output-side derived-text budget bounds actual bytes regardless — the declared-size check is a fast pre-filter, not the real wall. Worth a one-line comment if touched again.
3. `if written == 0: continue` in `convert_tree` is dead (the empty-file path writes 1 byte); harmless leftover from the old string-based contract.
4. The live 1.30 GB compile remains the unchecked box in the test plan — deploy-gated, and the right thing to gate the release on.

### Verdict

Approved. Bounded growth with the matching attack-surface hardening in the same PR, correct staging atomicity for the newly-fallible conversion step, real (verified) timeout plumbing, and clean rollback semantics.

PR #425, opened 2026-07-21
fix(kbc): allow the Codex nested writer sandbox

LikiosSedo:
## What changed

- add a dedicated `kb-compile-codex` BoxProfile for Codex writer runs
- allow user/PID/network namespaces and private mounts only for that profile by setting its outer seccomp/AppArmor policy to unconfined
- retain RuntimeDefault for ordinary AgentBoxes, Claude `kb-compile`, and read-only `kb-test`
- drop all Linux capabilities from the Codex compile container and retain the read-only root filesystem plus disabled service-account token
- run a bounded Bubblewrap namespace/mount preflight before the Codex writer contacts the model; fail closed with a useful diagnostic and never fall back to an unsandboxed shell
- include the new profile in health probes and orphan cleanup

## Root cause

Codex writer shell/file tools install the `kbc_writer` permission profile through Bubblewrap. The current AgentBox Pod applies RuntimeDefault seccomp and AppArmor before Bubblewrap starts, so namespace creation is rejected. The provider connectivity test succeeds because it never exercises the writer shell, while a real compile fails immediately after the model announces it will read the authoring files.

## Security boundary

This does not make the generic AgentBox or Claude compile profile unconfined. The exception is a separate profile selected by the consumer only for `codex_sdk`. Bubblewrap then installs the narrower filesystem/no-network writer boundary. If that cannot happen, the run is refused before model execution.

Requires paired Sicore MR !678: http://gitlab.scitix-inner.ai/k8s/sicore/-/merge_requests/678

## Validation

- `npm test -- --run src/gateway/agentbox/box-profile.test.ts src/gateway/agentbox/k8s-spawner.test.ts` — 55 passed
- `npm run build`
- Python syntax compilation for `codex_engine.py` and `test_codex_engine.py`
- focused preflight probe covering both success and namespace-denied failure
- live root-cause probe on the siflow-test Kubernetes version confirmed that RuntimeDefault rejects Bubblewrap and the profile-scoped Unconfined seccomp + AppArmor annotation permits the minimal namespace/mount sandbox

Live end-to-end compile validation will be completed with the paired Sicore branch and PR #424 large-snapshot receiver before marking ready for review.


chent1996 (approved):
## Review

Approving. This is a security-sensitive change done the right way: the exception is profile-scoped with tests pinning both directions, the compensating controls are real, the failure mode is closed, and the trade-off is documented in `security.md`. I verified the scoping end to end; my notes below are about keeping the residual risk consciously on the record, not blockers.

### What I verified

**The Unconfined exception cannot leak to other boxes.** `needsBubblewrap = profile.nestedSandbox === "bubblewrap"` is set only by `kb-compile-codex`; both the seccomp ternary and the AppArmor annotation hang off that one flag. Ordinary agents, Claude `kb-compile`, and `kb-test` keep `RuntimeDefault` — `k8s-spawner.test.ts` pins "allows nested Bubblewrap only for Codex compile boxes", and `getBoxProfile` stays fail-closed, so an old runtime receiving `"kb-compile-codex"` from the paired Sicore MR throws rather than spawning a wrong-shaped box (deploy runtime first, as noted).

**The AppArmor annotation actually targets the container.** The key `container.apparmor.security.beta.kubernetes.io/agentbox` matches the pod's container `name: "agentbox"` — easy to get silently wrong, verified correct. (If it ever drifted, the preflight would catch it at run time anyway — see below.)

**Compensating controls make the codex box *weaker* in every other dimension.** Capabilities are `drop: ALL` with none added back (the normal agentbox re-adds SETUID/SETGID/CHOWN/FOWNER/AUDIT_WRITE for its root entrypoint; the non-root kbc image needs none). Read-only rootfs and no SA token are retained; the inner `kbc_writer` bwrap boundary (workspace-only, no network) is what the model-driven shell actually runs under.

**The preflight is genuinely fail-closed with no unsandboxed fallback.** `_require_writer_sandbox()` runs at writer connect, *before* any model contact, and exercises the exact capability set needed (`--unshare-user/pid/net`, ro-bind, `/bin/true`, 10s timeout). Missing bwrap, exec failure, and nonzero exit all raise with a sanitized diagnostic; there is no code path that degrades to an unsandboxed shell. Read-only sessions skip it (they never launch a native shell), and the non-Linux skip only affects local dev. `test_writer_sandbox_preflight` covers both outcomes.

**Lifecycle completeness**: health probes and `sweepOrphans` include the new profile; `contract.ts`'s profile-name doc is updated; `security.md` records the exception and its rationale in the container-hardening section — exactly where a future auditor will look.

### Notes (for the record, non-blocking)

1. **The residual risk is real and worth stating plainly**: `Unconfined` seccomp + unprivileged user namespaces exposes the kernel's userns/mount syscall surface — historically the largest local-privilege-escalation vector class — to a workload that processes model output. The inner bwrap jail bounds the *shell children*, but the box's Python process itself (SDK + tools) runs with no syscall filter. The controls above (no caps, non-root, RO rootfs, no token, no SA) mean exploitation requires a kernel bug, not a misconfiguration — that's a defensible posture, and it's now documented; just don't let this exception quietly become a template for other profiles.
2. **A narrower alternative exists if this ever needs tightening**: a `Localhost` seccomp profile (RuntimeDefault plus only `unshare`/`mount`/`pivot_root`/`setns`/…) would keep the rest of the syscall filter while permitting bwrap. It costs distributing the profile file to every node, which is likely why Unconfined was chosen — worth one line in `security.md` acknowledging the alternative and the ops cost, so the decision reads as chosen rather than defaulted.
3. The AppArmor annotation is the legacy (deprecated) mechanism; K8s ≥1.30 has `securityContext.appArmorProfile`. The annotation is still honored, and the comment pins the 1.29 rationale — fine today, just a known migration when the fleet minimum moves.
4. Per the author's own note, live end-to-end compile validation (with Sicore !678 + the #424 receiver) is still pending — that should gate the release image, as stated.

### Verdict

Approved. Tightly-scoped exception, strictly-reduced capabilities elsewhere, fail-closed preflight with no fallback, correct cross-version failure mode, and the documentation updated where it matters.

src/gateway/agentbox/k8s-spawner.ts
```diff
@@ -394,9 +406,13 @@ export class K8sSpawner implements BoxSpawner {
...
         // entrypoint to fix volume permissions; SETUID/SETGID for user
-        // switching. All capabilities drop after runuser.
+        // switching. The non-root Codex compile image needs none of those
+        // capabilities; its container context below drops everything.
         securityContext: {
-          seccompProfile: { type: "RuntimeDefault" },
+          // Only the compile profile may create its declared nested sandbox.
+          // Normal AgentBoxes and closed-book kb-test boxes retain the outer
+          // RuntimeDefault policy.
+          seccompProfile: { type: needsBubblewrap ? "Unconfined" : "RuntimeDefault" },
```
jacoblee-io, line 415: **[Low / defense-in-depth]** The Unconfined exception is granted by K8s **profile** (`needsBubblewrap`), but the compensating bwrap enforcement is triggered by **engine** (`_require_writer_sandbox` runs only for the codex writer). These two inputs are decoupled in this repo: a consumer that selects `profile: kb-compile-codex` while running the Claude engine would get an Unconfined pod with **no** bwrap inner boundary (Claude uses bypassPermissions, never runs the preflight or bwrap). The whole security narrative rests on "kb-compile-codex is only ever paired with codex_sdk," which is enforced by the paired Sicore MR (consumer side), not here.

The dangerous direction is already safe (codex writer on a RuntimeDefault pod → preflight fails closed). This residual direction is only a needlessly-wider syscall surface (with fewer caps) reachable by a deliberate consumer misconfig, hence low. Worth a future hard coupling — e.g. reject a profile/engine mismatch at capability.start, or have the box refuse a non-codex engine on a `nestedSandbox` profile — so the compensating control lives next to the exception.

src/gateway/agentbox/k8s-spawner.ts
```diff
@@ -440,10 +456,12 @@ export class K8sSpawner implements BoxSpawner {
             image,
             imagePullPolicy,
             securityContext: {
-              capabilities: {
-                drop: ["ALL"],
-                add: ["SETUID", "SETGID", "CHOWN", "FOWNER", "AUDIT_WRITE"],
-              },
+              capabilities: needsBubblewrap
```
jacoblee-io, line 459: **[Note]** `kb-compile-codex` shares the SAME image as `kb-compile` but drops ALL capabilities (no SETUID/SETGID/CHOWN/FOWNER/AUDIT_WRITE), on the premise that the non-root Codex image needs none (uid remapping comes from bwrap's userns). Just keep this 0-cap assumption aligned with the shared image's entrypoint — if the entrypoint ever needs those caps for user-switching/volume-chown, the codex variant fails at pod start (availability, not a security hole). Validated live per the PR; flagging only so the coupling is on record.

jacoblee-io (commented):
Post-merge review (for the record — this is the fail-closed follow-up to the bubblewrap/userns prerequisite raised on #419). The implementation is clean and well-tested: the Unconfined seccomp/AppArmor exception is strictly scoped to the `kb-compile-codex` profile via `needsBubblewrap` (k8s-spawner tests assert agent/claude/test pods stay RuntimeDefault with no AppArmor annotation), the codex box drops ALL capabilities + keeps read-only root + no SA token, and `_require_writer_sandbox` fails closed with a real bwrap namespace/mount probe before the model turn — so the dangerous direction (a codex writer landing on a RuntimeDefault pod) is refused, never silently unsandboxed. Two defense-in-depth notes inline; neither blocks anything already merged.

PR #426, opened 2026-07-21
fix(kbc): normalize unambiguous source citations

LikiosSedo:
## Summary

- normalize a malformed body source marker when its label maps exactly and uniquely to one page-level `compiled_from` source
- preserve supported locators while leaving fenced/inline code, mixed payloads, and ambiguous duplicate stems unchanged
- run the mechanical repair before Layer-1 model repair, bounded by the active incremental page scope
- record applied changes in `SELFCHECK.json.mechanical_fixes`

## Why

A missing filename extension such as `(source: 专题目录)` is a deterministic formatting defect when the page already declares exactly one matching `compiled_from` source. Sending that defect through two model repair rounds and then asking the knowledge owner to accept or retry wastes model budget and exposes harness plumbing as a product decision.

## Safety

- exact alias matching only; no fuzzy title inference
- duplicate stems fail closed and remain lint findings
- clean pages are byte-identical
- incremental and guarded owner turns only normalize pages already in their editable or changed scope
- no prompt or dependency changes

## Validation

- `python kbc/platform/pod/test_selfcheck.py`
- `python kbc/platform/pod/test_incremental.py`
- `python kbc/platform/pod/test_compile_box.py`
- `git diff --check`

### DIFF
--- a/kbc/platform/pod/compile_box.py
+++ b/kbc/platform/pod/compile_box.py
@@ -1839,7 +1839,6 @@ async def _post_turn_selfcheck(run) -> str | None:
     # snapshot covers and the ledger/lint flags the rest. The early-returns
     # used to fire after the guard state was already consumed, letting an
     # index-deleting turn escape the byte freeze entirely (review finding).
-    run._selfcheck_key = key
     # Scoped-incremental byte-integrity guard: on an incremental turn, pages OUTSIDE
     # the authorized set (affected ∪ declared added-targets ∪ index) must be byte-
     # identical to their pre-turn state. Violations are RESTORED BY CODE first —
@@ -1851,6 +1850,7 @@ async def _post_turn_selfcheck(run) -> str | None:
     incr_violations: list[str] = []
     restored_pages: list[str] = []
     editable: set[str] = set()
+    after: dict[str, str] | None = None
     if incr:
         after = incremental.page_hashes(workdir)
         # repair_pages (set on re-arm): a ledger/lint repair turn legitimately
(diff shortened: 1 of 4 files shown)


chent1996 (approved):
## Review

Approving. Careful, well-fenced deterministic repair — exactly the "demote mechanical defects out of the model budget" follow-up the repair-loop work has been pointing at, and the engineering shows the dangerous cases were found before review.

What I verified:

- **The enabling change is sound**: `_markdown_prose` now masks frontmatter instead of dropping it, keeping the returned prose byte-aligned with the original text — which is what makes `_body_source_payload_spans`' offsets safe to use for in-place rewrites. Malformed/absent frontmatter keeps the old "whole text is prose" behavior, and existing lint semantics are unchanged (frontmatter stays masked).
- **The span-truncation guard (second commit) closes the one genuinely dangerous case.** When the ASCII-wrapped re-parse closes early (e.g. an unbalanced `)` inside a full-width marker), the parsed item covers only a prefix of the span; rewriting the whole span would silently drop the tail. The `malformed[0] != payload.strip(...)` check refuses exactly that and leaves the payload for semantic repair. Good catch, well-commented.
- **Ambiguity fails closed**: exact aliases only (full path / basename / known-extension-stripped stem, NFC + casefold + quote-stripping — no fuzzy titles), and an alias mapping to more than one `compiled_from` source is skipped and stays a lint finding.
- **Scope discipline holds**: incremental turns normalize only `editable` pages; guarded owner turns only pages that actually changed; the unrestricted case is the full compile, which owns the tree anyway. Mechanical fixes on editable pages land *before* the incremental-integrity recheck consumes `after`, and fixes are applied via reversed spans + atomic writes, so offsets and files stay coherent.
- **The idempotency key move is correct**: `run._selfcheck_key` is now assigned after the mechanical pass, with `state_key` recomputed when fixes were applied — so the recorded key reflects the post-fix tree and the next turn's idempotency check doesn't misfire. I checked nothing reads the key in the moved-over region.
- **Budget synergy**: running before `run_layer1` means Layer-1's repair budget (and the workload-scaled limit being discussed in #430) only sees what actually needs a model. `mechanical_fixes` in SELFCHECK.json (count + items capped at 40) plus the "without using a repair round" summary line make it auditable. Clean pages are byte-identical by construction (`replacement == payload → skip`).
- Tests pin both the unit paths (repair / ambiguous-skip / code-fence masking / locator retention) and the ordering claim (`test_exact_source_alias_is_repaired_before_l1_budget`). CI green including the KBC box suite.

Non-blocking note: on ordinary (non-incremental, non-guarded) turns `allowed_pages=None` scans the whole candidate tree every selfcheck pass. It's cheap relative to the existing lint scan, just worth remembering if per-turn selfcheck cost ever becomes a topic.

kbc deploy note: box behavior change — needs a `kbc-compile-box` image rebuild + redeploy to take effect.

PR #428, opened 2026-07-21
fix(agentbox): distinguish probe failures from unreachable clusters

caihongwei2006:
## Summary

The `cluster_list` probe collapsed every kubectl failure into `reachable: false`, so a local kubeconfig/credential/tooling problem read as "cluster down" and invited the model to hallucinate an outage. This replaces that boolean with a conservative, evidence-based contract.

### Problem

`kubectl` being unusable is not the same as the cluster being unreachable. The old probe returned `{ reachable: false }` for missing kubeconfig, bad context, expired cert, 401/403, and genuine network failures alike — the model could not tell "fix your config" from "the cluster is down".

### Solution

Three-state result with a structured `reason`, so the caller reads a stable label instead of parsing kubectl prose:

- **`unreachable`** (connection_refused / dns / network / connection_reset / timeout) — the only state that means the API server couldn't be reached. Requires **positive network-layer evidence**.
- **`probe_failed`** (kubectl_missing / kubeconfig / credential / auth / authz / endpoint / tls_cert / timeout / unknown) — local tooling/config/credential/trust, says nothing about cluster health.

Evidence boundary tightened per review:

- A child process killed by our own `execFile` timeout with **no** network evidence is `probe_failed/timeout` (reachability unknown), **not** `unreachable` — it may have killed a hung local exec credential plugin.
- **401 → `auth`** and **403 → `authz`** carry `reachable: true` (the API server adjudicated the request). **404 → `endpoint`** does **not** claim reachability — a wrong URL or intermediary can answer 404 without the real API server.
- The probe timeout now **aborts its own work** (AbortSignal → guard before `kubectl` spawn + kill a running child), so a slow credential fetch can't continue into `kubectl` after `cluster_list` has already returned.

Also fixes a latent bug: kubectl's real refusal message is `The connection to the server <host> was refused`, which the classifier's `connection refused` regex never matched.

The `.gitignore` block that globally ignored `*.test.*` was reverted — tests are versioned project source.

## Test Plan

- [x] `npm test` — 227 files, 4511 passed / 2 skipped
- [x] `npm run build` — clean
- [x] Table-driven `classifyKubectlProbeError()` tests: ENOENT/EACCES, localhost:8080 fallback, invalid/missing kubeconfig, no-context, hung exec plugin, 401, 403, 404, x509, DNS, refused, reset, no-route, client-go request timeout, killed-with-evidence vs killed-without-evidence, unknown — asserting `probe_status`, `reason`, and `reachable` presence.
- [x] Three-state `cluster_list` serialized-output tests (success / unreachable / probe_failed + auth-reachable), strongly typed fixtures (no `as any`).
- [x] Timeout test proving a late operation does not start post-timeout `kubectl` work.

### DIFF
--- a/src/agentbox/credential-broker.test.ts
+++ b/src/agentbox/credential-broker.test.ts
@@ -2,7 +2,8 @@ import { describe, it, expect, beforeEach, afterEach } from "vitest";
 import fs from "node:fs";
 import os from "node:os";
 import path from "node:path";
-import { CredentialBroker } from "./credential-broker.js";
+import { CredentialBroker, classifyKubectlProbeError, withProbeTimeout } from "./credential-broker.js";
+import type { ProbeResult, ClusterLocalInfo } from "./credential-broker.js";
 import type {
   CredentialTransport,
   ClusterMeta,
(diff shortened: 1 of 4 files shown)


LikiosSedo (changes requested):
Thanks for improving this path. The underlying problem is real: a credential/configuration/authentication failure must not be reported as `reachable: false`, and a 401/403 response does prove that an HTTP exchange reached the configured endpoint. The three-state direction (`success` / `unreachable` / `probe_failed`) is a useful improvement.

I am requesting changes because the current implementation is not yet safe or test-complete enough to merge.

## Required changes

### 1. Do not ignore repository tests

Please remove the `.gitignore` block that ignores `*.test.*`, `*_test.py`, and `__tests__/` globally. Tests are versioned project source in Siclaw, not local-only files. This rule would silently hide every newly written test from `git status` and makes it especially easy to ship an unreviewed error-classification table.

This also makes the PR broader than its stated purpose. Siclaw PRs should contain one focused logical change.

### 2. Add contract tests and update the existing tests

This change replaces the public `ProbeResult` shape and introduces an ordered regex classifier, but it adds no automated tests. On the current head (`424879b176b383483e3f47937659dedac6ba9d66`) I ran:

```text
npm test -- src/tools/query/cluster-list.test.ts src/agentbox/credential-broker.test.ts
```

Result: **1 failed, 39 passed**. The existing `cluster-list.test.ts` success fixture still uses the old result shape, so `server_version` is lost. `npm run build` passes, but strict TypeScript compilation is not a substitute for behavioral tests, especially because the test mocks use `as any`.

Please update `src/tools/query/cluster-list.test.ts` for all three output states and add table-driven tests for `classifyKubectlProbeError()` in `src/agentbox/credential-broker.test.ts`. At minimum cover:

- success with `server_version`;
- credential acquisition failure;
- `ENOENT` / `EACCES`;
- invalid or missing kubeconfig, including the localhost:8080 fallback;
- 401 and 403 responses;
- certificate validation failure;
- DNS, connection refused/reset, no route, and an explicit client-go request timeout;
- child-process timeout / killed process;
- unknown stderr.

Each case should assert `probe_status`, `reason`, whether `reachable` is present, and the serialized `cluster_list` result. Avoid `as any` for the new fixtures so future contract changes fail at compile time.

### 3. A killed local process is not proof that the cluster is unreachable

`classifyKubectlProbeError()` currently maps every `err.killed === true` to:

```text
probe_status: "unreachable", reachable: false, reason: "timeout"
```

However, the 5-second `execFile` timeout can also kill `kubectl` while a local exec credential plugin is hung. In that case the API server may never have been contacted. This contradicts the PR's own rule that local/tooling timeouts have unknown reachability.

Please classify a generic child-process kill as `probe_failed / timeout` with `reachable` omitted. Only classify timeout as `unreachable` when stderr provides positive evidence of a network/request timeout from client-go.

### 4. Do not label 404 as authorization failure or proof that the cluster is up

A 404 proves that an HTTP endpoint answered; it does not prove RBAC denial, and it may be produced by a wrong server URL or intermediary. Mapping `NotFound` to `authz` and telling the model "the cluster is up" is stronger than the evidence supports.

Keep 401 as `auth`, keep 403 as `authz`, and classify 404 separately (for example as an endpoint/configuration failure) without claiming Kubernetes health. The exact name matters less than preserving the evidence boundary.

### 5. The total timeout must not let work continue after the result is returned

`withProbeTimeout()` resolves after 7 seconds but does not cancel `ensureCluster()` or the operation it wraps. A late credential response can therefore continue into `probeKubeconfig()` after `cluster_list` has already returned a timeout. It also releases the concurrency-limiter slot while the original work is still alive, so the stated bounded-fan-out guarantee no longer necessarily holds.

Please either propagate cancellation through the credential acquisition/probe path, or restructure the timeout so timed-out work cannot continue into `kubectl`. Add a test proving that a late operation does not start post-timeout work.

## Siclaw contribution standard for this change

Before updating the PR, please read:

- [`CONTRIBUTING.md`](https://github.com/scitix/siclaw/blob/main/CONTRIBUTING.md), especially **Pull Request Guidelines** and the PR description template;
- [`docs/design/tools.md`](https://github.com/scitix/siclaw/blob/main/docs/design/tools.md), especially the **Description-Code Consistency Rule**.

You did the right thing by updating the `cluster_list` description together with its output contract. Please complete the same standard on testing and PR hygiene:

1. Revert the unrelated `.gitignore` change.
2. Make the classifier changes above.
3. Add/update the focused tests.
4. Run `npm test`, `npx tsc --noEmit`, and `npm run build`.
5. Update the PR description with **Problem**, **Solution**, and **Test Plan**, including the exact commands and results.
6. Use an English Conventional Commit before merge, for example: `fix(agentbox): distinguish probe failures from unreachable clusters`.

The desired end state is not merely "more readable errors." It is a conservative evidence contract: report `unreachable` only when we have affirmative network-layer evidence, report `reachable: true` only when the configured endpoint completed a meaningful response, and leave reachability unknown for local/configuration/tooling failures.

caihongwei2006:
Thanks for the thorough review — all five required changes are addressed in `7fd26e7` (force-pushed, single focused commit).

1. **`.gitignore`** — reverted the global `*.test.*` / `__tests__/` block. Tests are versioned again.
2. **Tests** — table-driven `classifyKubectlProbeError()` covering ENOENT/EACCES, localhost:8080 fallback, invalid/missing kubeconfig, no-context, hung exec plugin, 401/403/404, x509, DNS, refused, reset, no-route, client-go request timeout, killed-with-evidence vs killed-without, and unknown; plus three-state `cluster_list` serialized-output tests. Fixtures are typed as `ProbeResult` — no `as any`, so contract drift fails at compile time. (While writing these I also found the classifier never matched kubectl's real `The connection to the server <host> was refused` phrasing — fixed.)
3. **Killed process** — a child killed by our `execFile` timeout with no network evidence is now `probe_failed/timeout` (reachability unknown). `unreachable/timeout` requires positive client-go network-timeout evidence in stderr.
4. **404** — split out of `authz`. 401→`auth`, 403→`authz` keep `reachable:true`; 404→new `endpoint` reason with `reachable` omitted (no health claim).
5. **Timeout cancellation** — `withProbeTimeout` now drives an `AbortController`; `probeCluster` checks `signal.aborted` before spawning `kubectl` and passes the signal to `execFile` (kills a running child). Since timed-out work is aborted rather than left running, the early limiter-slot release no longer leaks kubectl processes. Added a test proving a late operation does not start post-timeout work.

Verification: `npm test` (227 files, 4511 passed / 2 skipped), `npx tsc --noEmit`, `npm run build` all clean. PR description updated with Problem / Solution / Test Plan.

LikiosSedo (dismissed):
All requested changes are addressed on 7fd26e7b. The evidence-based probe contract is now conservative, the timeout path prevents post-timeout kubectl work, the 404 semantics are separated from authorization, and the updated contract has focused automated coverage. I independently verified the full test suite (4511 passed, 2 skipped) and TypeScript build; GitHub CI is also green. Approved.

LikiosSedo (approved):
Re-approved after the unrelated `pod_exec` transport-noise commit was removed from this branch. PR #428 is focused again on the conservative `cluster_list` probe contract, the previously requested changes remain addressed, the current head is mergeable, and the refreshed Type Check, Test, and Portal Web Test checks are all green.

PR #429, opened 2026-07-22
feat(gateway): persist prompt trace attribution

glow-Kev1n:
## Summary

Persist per-prompt trace attribution across Siclaw's interactive, channel, scheduled, and steer entry points so SiCore can group each user prompt with the complete agent loop that it started.

### Problem

Siclaw persists inbound user messages before AgentBox assigns a per-prompt trace ID. Assistant and tool events can carry the trace ID later, but the initiating user row may remain unbound. In a long-lived session with multiple prompts, this prevents SiCore's Trace audit view from identifying the exact prompt reliably.

Steer messages have the inverse problem: they join an already-running prompt and therefore need the active trace ID rather than a new trace. The automatic `chat.send` to steer fallback also persisted the additional input as a normal user row, which could make transcript and trace readers treat it as a new prompt.

### Solution

This PR introduces an ACK-and-bind integration between AgentBox, Siclaw Runtime, and SiCore:

1. AgentBox prompt and steer acknowledgements expose the relevant trace ID. A prompt returns its new root trace; a steer returns the active prompt's trace.
2. Siclaw keeps the exact persisted user message ID and calls the Portal's `chat.bindMessageTraceId` RPC after the acknowledgement. The binding is message-specific rather than inferred from session order, and both SiCore and Siclaw's standalone Portal implement the same guarded contract.
3. Web/API/A2A prompts and explicit steers bind their user rows after ACK. The 409 automatic-steer fallback additionally relabels the persisted row as `metadata.kind="steer"` before binding it.
4. Lark and DingTalk bind their existing audited user rows after prompt ACK and stamp the same trace ID on channel assistant/tool persistence.
5. Scheduled tasks already receive the prompt ACK before creating their audit row, so they persist the trace ID directly.

The new trace fields are optional for rolling upgrades. Missing trace IDs are a no-op, and trace binding runs best-effort in the background so failures never delay or interrupt the user response. This PR adds no database schema or production dependency changes.

### Compatibility and deployment

Both SiCore and Siclaw's in-repo Portal implement the `chat.bindMessageTraceId` internal RPC. In managed deployments, deploying SiCore first and Siclaw second minimizes the rolling-upgrade window; an older SiCore causes only a best-effort binding warning. Standalone and `siclaw local` deployments use the in-repo Portal handler directly.

Out of scope:

- Reordering the existing DingTalk cold-start persistence flow.
- Historical trace backfill.
- Langfuse trace trees, LLM-call views, or trace review workflows.

## Test Plan

- [x] `npx tsc --noEmit` passes after rebasing onto the latest `main`.
- [x] `git diff --check origin/main...HEAD` passes.
- [x] GitHub Actions `Test` passes, including Portal RPC and non-blocking trace-binding regression coverage. Local Vitest execution is unavailable because the checked-out `esbuild` binary is not executable in the development environment.
- [ ] Manual smoke test with the paired SiCore branch: Web prompt, explicit steer, 409 automatic steer, Lark/DingTalk, and scheduled task attribution.

## Architecture Checklist

- Deployment mode: no resource sync or filesystem-write behavior changed.
- Security model: no shell execution or tool-policy path changed.
- DB parity: no schema or migration change.
- Tool protocol: no diagnostic tool added or modified.

## General Checklist

- [x] Changes are focused on one logical trace-attribution feature.
- [x] Commit messages follow Conventional Commits.
- [x] No unrelated changes are included.

### DIFF
--- a/src/agentbox/http-server.test.ts
+++ b/src/agentbox/http-server.test.ts
@@ -998,11 +998,16 @@ describe("http-server — steer / abort / clear-queue", () => {
   });
 
   it("POST /api/sessions/:id/steer calls brain.steer", async () => {
+    const s = await sm.getOrCreate("st2");
+    let finishPrompt!: () => void;
+    s.brain.prompt.mockImplementation(() => new Promise<void>((resolve) => { finishPrompt = resolve; }));
     await getJson(port, "/api/prompt", "POST", { text: "hi", sessionId: "st2" });
-    const s = sm.sessions.get("st2")!;
     const r = await getJson(port, "/api/sessions/st2/steer", "POST", { text: "stop" });
     expect(r.status).toBe(200);
+    expect(r.data.traceId).toMatch(/^[0-9a-f]{32}$/);
     expect(s.brain.steer).toHaveBeenCalledWith("stop");
+    finishPrompt();
+    await flushAsync();
   });
 
   it("POST /api/sessions/:id/steer forwards images to brain.steer", async () => {
--- a/src/agentbox/http-server.ts
+++ b/src/agentbox/http-server.ts
@@ -1212,13 +1212,17 @@ export function createHttpServer(
       ? body.text
       : defaultPromptTextForMedia(promptMedia);
     console.log(`[agentbox-http] Steering session ${sessionId}: ${steerText.slice(0, 80)}`);
+    // Capture before awaiting brain.steer(): the active prompt may finish and
+    // clear the recorder while the steer call is being accepted. A steer is an
+    // addition to that active prompt, so it inherits this existing trace.
+    const traceId = tracingRecorder.getRootTraceId(sessionId);
     try {
       if (promptMedia) {
         await managed.brain.steer(steerText, promptMedia);
       } else {
         await managed.brain.steer(steerText);
       }
-      sendJson(res, 200, { ok: true });
+      sendJson(res, 200, { ok: true, traceId });
     } catch (err) {
       console.error(`[agentbox-http] Steer error for session ${sessionId}:`, err);
       const message = err instanceof Error ? err.message : "Steer failed";
--- a/src/gateway/agentbox/client.test.ts
+++ b/src/gateway/agentbox/client.test.ts
@@ -92,7 +92,7 @@ describe("AgentBoxClient — prompt + session CRUD", () => {
         res.end(JSON.stringify({ ok: true }));
       } else if (req.method === "POST" && url === "/api/sessions/s1/steer") {
         res.writeHead(200, { "Content-Type": "application/json" });
-        res.end(JSON.stringify({ ok: true }));
+        res.end(JSON.stringify({ ok: true, traceId: "0123456789abcdef0123456789abcdef" }));
       } else if (req.method === "POST" && url === "/api/sessions/s1/abort") {
         res.writeHead(200, { "Content-Type": "application/json" });
         res.end(JSON.stringify({ ok: true }));
@@ -166,7 +166,8 @@ describe("AgentBoxClient — prompt + session CRUD", () => {
   });
 
   it("steerSession() posts the text body", async () => {
-    await client.steerSession("s1", "stop the pod");
+    const result = await client.steerSession("s1", "stop the pod");
+    expect(result.traceId).toBe("0123456789abcdef0123456789abcdef");
     const req = srv.captures.find((c) => c.url === "/api/sessions/s1/steer")!;
     expect(req.method).toBe("POST");
     expect(JSON.parse(req.body)).toEqual({ text: "stop the pod" });
--- a/src/gateway/agentbox/client.ts
+++ b/src/gateway/agentbox/client.ts
@@ -77,6 +77,12 @@ export interface PromptResponse {
   traceId?: string;
 }
 
+export interface SteerResponse {
+  ok: boolean;
+  /** Trace of the active prompt that accepted this additional user input. */
+  traceId?: string;
+}
+
 export interface SessionInfo {
   id: string;
   createdAt: string;
@@ -261,10 +267,10 @@ export class AgentBoxClient {
   /**
    * Send a steer instruction (inserts a user message after interrupting the current tool)
    */
-  async steerSession(sessionId: string, text: string, media?: PromptMediaOptions): Promise<void> {
+  async steerSession(sessionId: string, text: string, media?: PromptMediaOptions): Promise<SteerResponse> {
     const images = media?.images;
     const files = media?.files;
-    await this.fetch(`/api/sessions/${sessionId}/steer`, {
+    const resp = await this.fetch(`/api/sessions/${sessionId}/steer`, {
       method: "POST",
       headers: { "Content-Type": "application/json" },
       body: JSON.stringify({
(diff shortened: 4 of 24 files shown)


chent1996 (commented):
## Review

The design is sound and I verified it in detail — but I'm leaving this as a **comment** rather than an approval because CI is red with failures introduced by this PR, and the new RPC has no handler in this repo's own Portal. Both are concretely fixable; details below.

### Design verified (the good part)

- **Message-specific binding, not order inference.** Every entry point keeps the exact `appendMessage` row id and binds *that* row after ACK — no "latest user row in session" guessing. The 409 automatic-steer fallback correctly relabels the already-persisted row (`metadata.kind="steer"`) *before* binding, so transcript/trace readers can't mistake it for a new prompt.
- **Best-effort discipline is consistent.** Every bind is `.catch`-logged and never interrupts the user response — matching the repo's channel precedent that audit plumbing must never affect the reply. Missing `traceId` is a no-op (`bindMessageTraceId` returns early), so an older AgentBox ACK degrades cleanly; an older gateway ignores the new steer ACK field. Rolling-upgrade story holds in both directions.
- **The steer race is handled thoughtfully**: `http-server.ts` captures `tracingRecorder.getRootTraceId(sessionId)` *before* awaiting `brain.steer()`, with the comment explaining exactly why (the active prompt may finish and clear the recorder while the steer is being accepted). A steer inherits the active trace instead of minting a new one — the right semantics.
- Channel assistant/tool rows get the trace via `persist.traceId` with sane precedence (`persist?.traceId ?? msg.traceId`), DingTalk binds only when auditable, and scheduled tasks stamp the trace directly at append (ACK is already available there). `tsc` and the type-level changes are clean.

### Blocker 1 — CI is red: 18 failures in `lark.test.ts`, introduced here

The `Test` job fails with 18 tests in the `streaming card flow` suite all rendering the error card (`❌ 处理时出错了…`) instead of the assistant reply. The signature points at the new bind call: the suite's `chat-repo` mock adds `bindMessageTraceIdMock = vi.fn()` but `beforeEach` only `mockReset()`s it without a default resolved value — a bare `vi.fn()` returns `undefined`, so production's `bindMessageTraceId(...).catch(...)` becomes `undefined.catch` → TypeError → swallowed by the outer handler → error card. (Note `appendMessageMock` gets `mockResolvedValue("msg-db-1")` in the same `beforeEach`; the new mock needs the equivalent, e.g. `bindMessageTraceIdMock.mockResolvedValue(undefined)`.) Given the test plan notes local vitest was unavailable, this is understandable — but the suite must be green before merge, and it's worth re-checking the DingTalk suite for the same pattern.

### Blocker 2 — `chat.bindMessageTraceId` has no handler in this repo's Portal

The RPC is implemented only on the SiCore side (paired MR). This repo's own Portal (`src/portal/adapter.ts`) — which backs `siclaw local` and non-SiCore deployments — has no `chat.bindMessageTraceId` handler (verified on both `main` and this branch). Consequences there: **every web/API prompt and every steer logs a bind-failure warning**, and trace attribution silently never works outside SiCore. The groundwork already exists in-repo (`chat_messages.trace_id` column + index + `appendMessage` traceId support), and the doc comment on `bindMessageTraceId` already specifies the exact semantics the handler needs (user-role check, ownership, NULL-to-value idempotency) — implementing it in `adapter.ts` is small and keeps the established "adapter mirrors chat.* RPCs" discipline. Alternatively, feature-detect and skip quietly — but implementing it is better and makes the feature work for OSS users too.

### Minor (non-blocking)

- The prompt-path bind is `await`ed between the ACK and attaching the SSE consumer — one extra WS round-trip on every interactive prompt's critical path. It's failure-tolerant already; making it fire-and-forget (`void ….catch(warn)`) would shave the latency without changing semantics. Fine either way.
- Deploy ordering (SiCore first) is documented; with blocker 2 fixed, even that window becomes cosmetic.

### Verdict

Not approving yet. The attribution design is right and carefully executed at every entry point — fix the mock so CI is green (blocker 1), and give the in-repo Portal the RPC handler (blocker 2), and I'll be happy to approve on the next pass.

chent1996 (commented):
Re-reviewed at `8daf382f`.

**Blocker 1 — resolved.** The one-line `mockResolvedValue(undefined)` in both channel suites is exactly the right fix, and CI is now fully green (Test / Type Check / Portal Web). Thanks for also covering the DingTalk suite.

**Blocker 2 — still open.** `src/portal/adapter.ts` still has no `chat.bindMessageTraceId` handler, so on this repo's own Portal (`siclaw local` and any non-SiCore deployment) every web/API prompt and every steer logs a bind-failure warning, and trace binding silently no-ops. Two acceptable resolutions:

1. *(Preferred)* Implement the handler in `adapter.ts` — the groundwork already exists (`chat_messages.trace_id` column + index), and the semantics are already specified in `bindMessageTraceId`'s doc comment (user-role check, session ownership, NULL-to-value idempotency). It's ~15 lines + a test, and keeps the established "adapter mirrors chat.* RPCs" discipline.
2. Or gate the bind call so portals without the RPC don't warn on every prompt (feature-detect once, or downgrade to debug-level after the first failure) — and note the SiCore-only scope explicitly in the PR description.

Either lands this in an approvable state; as-is, merging adds per-prompt log noise to every OSS/local deployment.

chent1996 (approved):
## Review (third pass — approving)

Re-reviewed at `b4e83fc9`. Both blockers from my earlier reviews are resolved, CI is fully green, and the fix quality goes beyond the minimum asked.

### Blocker 1 (CI red) — resolved at `8daf382f`
The one-line `mockResolvedValue(undefined)` in both channel suites; all three checks green since.

### Blocker 2 (in-repo Portal handler) — resolved, and done carefully
`src/portal/adapter.ts` now implements `chat.bindMessageTraceId` with exactly the semantics the `chat-repo` doc comment promised, plus two touches worth calling out:

- **Dialect-aware idempotency.** The single UPDATE covers NULL-to-value and same-value binds; the follow-up SELECT exists specifically because MySQL reports zero affected rows on a same-value update while SQLite counts it — the read path disambiguates an idempotent retry from a genuine conflict, and the comment documents why. Verified the logic converges on both drivers.
- **Honest ownership semantics.** Rather than pretending the standalone Portal has per-Runtime ownership, the handler documents its actual trust model (one portal-secret-authenticated Runtime trust domain) and the `chat-repo` doc comment was updated to stop overclaiming. SiCore keeps the stricter per-Runtime check on its side. Right call.
- Strict trace format validation (`^[0-9a-f]{32}$`, matching the CHAR(32) column), fail-fast errors for not-found / non-user / different-trace conflict.

Tests pin the paths that matter: exact-scope bind, the idempotent same-trace retry (i.e. the MySQL zero-affected-rows path), invalid-trace rejection, and — nice addition — a server-side test that reply consumption does not wait on the binding.

### Bonus
The non-blocking latency note was adopted too: all four `bindMessageTraceId` call sites are now fire-and-forget (`void … .catch(warn)`), so the bind RPC is off the interactive critical path. I checked the 409-steer path for ordering hazards between the awaited `updateMessage` relabel and the detached bind — they touch disjoint columns on the same row, so there's no interleaving risk.

### Nit (non-blocking)
The handler's conflict branch ("already bound to a different trace") and the non-user-role rejection don't appear to have dedicated test cases — the happy/idempotent/invalid paths do. Worth adding two small cases in a follow-up so the error contract is pinned as well.

### Verdict
Approved. Trace attribution is message-specific at every entry point, best-effort end to end, rolling-upgrade safe in both directions, and now works on both Portal implementations. This review supersedes my earlier comment reviews.

PR #430, opened 2026-07-22
fix(kbc): scale L1 auto-repair budget with open-violation workload

LikiosSedo:
## Problem

The ledger auto-repair budget is a flat constant (`KBC_L1_REPAIR_ROUNDS`, default 2), sized for ordinary drift. A first full compile or a wiki **adoption** over a large KB opens dozens of mechanical violations at once (dangling citations, malformed source annotations, charset corruption from the gateway's per-chunk SSE decoding). With a flat budget the loop is structurally spent before it can converge, and the entire tail is dumped into the owner's question queue as one opaque residual ticket.

Live evidence (2026-07-22 adoption run on a 1575-source KB): 2 rounds spent → unconverged → a 55-item ticket containing raw SELFCHECK lint output shown to the end user.

## Change

- New `_l1_repair_round_limit(report)`: keeps the flat value as the **floor**, adds bounded rounds proportional to the open violation count (`KBC_L1_REPAIR_VIOLATIONS_PER_ROUND`, default 12/round), fenced by a hard operator cap (`KBC_L1_REPAIR_MAX_ROUNDS`, default 16). Mirrors the established `_batch_media_verify_round_limit` idiom.
- `SELFCHECK.json` gains `repair_rounds_limit` (additive, for observability).

## Non-goals / follow-ups

- Residual tickets should carry a `kind` (content contradiction vs infra residual) so mechanical leftovers stop surfacing in the user-facing question inbox — separate change (box ticket schema + sicore inbox routing).
- Deterministic violations (source-annotation normalization, U+FFFD replacement) could be demoted to a code-repair layer that does not consume model budget at all.

## Tests

All repair-related tests green, including the pre-existing flat-budget gate tests (small workloads keep the floor, so their behavior is unchanged) plus a new unit test covering floor / scaling / cap / env overrides.

### DIFF
--- a/kbc/platform/pod/compile_box.py
+++ b/kbc/platform/pod/compile_box.py
@@ -305,6 +305,11 @@ class CompileRun:
         # loop; reset to 0 whenever the check passes).
         self._selfcheck_key: str | None = None
         self._l1_repairs_used = 0
+        # Per-episode repair-budget high-water mark (review, PR #430): the
+        # limit is derived from the OPEN workload, which shrinks as repairs
+        # succeed — gating on a per-turn recomputation would close the gate
+        # at ~half-repaired. Keep the episode's max; reset with the counter.
+        self._l1_round_limit = 0
         # Candidate/exclusion state captured immediately before each model
         # turn. A conversational turn that leaves this state byte-identical is
         # not a migration trigger for an inherited legacy draft. Compile,
@@ -1946,10 +1951,17 @@ async def _post_turn_selfcheck(run) -> str | None:
             "grandfathered_format_violations": grandfathered_format[:40],
         }
     ledger_clean = report["coverage"]["closed"] and report["lint"]["ok"]
+    # High-water, not per-turn recomputation: the workload SHRINKS as repairs
+    # succeed, so re-deriving the limit each turn strands the episode at
+    # ~half-repaired (review). max() also lets the budget grow if a repair
+    # legitimately reopens violations mid-episode.
+    run._l1_round_limit = max(getattr(run, "_l1_round_limit", 0),
+                              _l1_repair_round_limit(report))
     if ledger_clean and not incr_violations:
         run._l1_repairs_used = 0
+        run._l1_round_limit = 0
         report["state"] = "passed"
-    elif run._l1_repairs_used < _l1_repair_round_limit(report):
+    elif run._l1_repairs_used < run._l1_round_limit:
         report["state"] = "repairing"
         # The report carries the PREVIOUS converge_phase (possibly "settled");
         # the pre-turn_done sync would ship repairing+settled for one window
@@ -1960,7 +1972,7 @@ async def _post_turn_selfcheck(run) -> str | None:
     else:
         report["state"] = "unconverged"  # budget spent: publish card shows the rest
     report["repair_rounds_used"] = run._l1_repairs_used
-    report["repair_rounds_limit"] = _l1_repair_round_limit(report)
+    report["repair_rounds_limit"] = run._l1_round_limit
     selfcheck.write_selfcheck(workdir, report)
     locale = getattr(run, "locale", None)
     if restored_pages:
@@ -3112,6 +3124,7 @@ async def _run_ledger_repairs(run: "CompileRun", replies: list[str]) -> None:
     # for this phase's own findings. Each _run_ledger_repairs call is bounded
     # by its own budget; the batch tail calls it a bounded number of times.
     run._l1_repairs_used = 0
+    run._l1_round_limit = 0
     run._ledger_forced = True  # batch-final: the mid-Execute index exemption is off
     try:
         repair = await _post_turn_selfcheck(run)
--- a/kbc/platform/pod/compile_box.py
+++ b/kbc/platform/pod/compile_box.py
@@ -1289,6 +1289,29 @@ def _l1_repair_rounds() -> int:
     return int(os.environ.get("KBC_L1_REPAIR_ROUNDS", "2"))
 
 
+def _l1_repair_round_limit(report: dict) -> int:
+    """Scale the ledger auto-repair budget with the ACTUAL workload instead of
+    a flat constant. A first full compile / wiki adoption over a large KB can
+    open dozens of mechanical violations at once; with the flat 2-round budget
+    it structurally lands unconverged and dumps the tail into the owner's
+    question queue (live adoption run, 2026-07-22: 55-item residual after two
+    rounds). Mirrors _batch_media_verify_round_limit: the flat value stays the
+    floor for ordinary drift, the open-violation count adds bounded rounds, and
+    a hard operator cap still fences pathological repair loops."""
+    base = _l1_repair_rounds()
+    cov = report.get("coverage") or {}
+    lint = report.get("lint") or {}
+    workload = (len(cov.get("unaccounted") or [])
+                + len(cov.get("dangling_citations") or [])
+                + len(lint.get("violations") or []))
+    if workload <= 0:
+        return base
+    per_round = max(1, int(os.environ.get(
+        "KBC_L1_REPAIR_VIOLATIONS_PER_ROUND", "12")))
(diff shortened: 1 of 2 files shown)


kbc/platform/pod/compile_box.py
```diff
@@ -1905,7 +1928,7 @@ async def _post_turn_selfcheck(run) -> str | None:
     if ledger_clean and not incr_violations:
         run._l1_repairs_used = 0
         report["state"] = "passed"
-    elif run._l1_repairs_used < _l1_repair_rounds():
+    elif run._l1_repairs_used < _l1_repair_round_limit(report):
```
chent1996: **The budget shrinks as the repair converges — computed each turn from the *current* workload, this gate strands the exact scenario the PR is fixing.**

`_l1_repairs_used` only resets on a fully clean ledger, but `_l1_repair_round_limit(report)` is re-evaluated every turn from the *remaining* violations. So the allowance falls as repairs succeed, and the gate closes when `used ≥ ceil(remaining/12)` — i.e. roughly when the episode is only **half** repaired.

Plugging in this PR's own motivating numbers (55 open violations, `per_round=12`, ideal repair clearing ~12/round):

| turn | remaining | limit = max(2, ceil(rem/12)) | used | gate |
|---|---|---|---|---|
| 1 | 55 | 5 | 0 | repair |
| 2 | 43 | 4 | 1 | repair |
| 3 | 31 | 3 | 2 | repair |
| 4 | 19 | **2** | **3** | **unconverged, 19 left** |

The intended "5 rounds for 55 items" never materializes; under nominal progress the loop stops near `remaining ≈ workload/2`. (Still strictly better than the flat 2 — `limit ≥ base` always, which is also why the pre-existing flat-budget tests keep passing — but the adoption run this PR cites would still end unconverged, with ~19 residuals instead of 55.)

Note the sibling this docstring cites doesn't have this problem: `_batch_media_verify_round_limit(plan, initial_pending)` is evaluated **once** from the *initial* pending set and stays fixed for the loop. The formula was mirrored; the evaluation timing — the load-bearing part — wasn't.

Suggested fix (~3 lines): track a per-episode high-water mark instead of re-deriving from the shrinking report, e.g.

```python
run._l1_round_limit = max(getattr(run, "_l1_round_limit", 0), _l1_repair_round_limit(report))
# gate on run._l1_round_limit; reset it alongside run._l1_repairs_used = 0
```

This also stays correct when a mid-episode restructuring *reopens* violations (the budget can grow, matching the per-gap-episode semantics documented in the README), and it fixes the observability wrinkle below for free. The new unit test covers the pure function only, so a loop-level case (workload shrinking across turns keeps the initial allowance) would pin this.

kbc/platform/pod/compile_box.py
```diff
@@ -1916,6 +1939,7 @@ async def _post_turn_selfcheck(run) -> str | None:
     else:
         report["state"] = "unconverged"  # budget spent: publish card shows the rest
     report["repair_rounds_used"] = run._l1_repairs_used
+    report["repair_rounds_limit"] = _l1_repair_round_limit(report)
```
chent1996: Same root cause, observability flavor: on an `unconverged` report this recomputes the limit from the now-smaller workload, so the persisted report can show `repair_rounds_used: 3` next to `repair_rounds_limit: 2` — used exceeding the limit reads like a bug to whoever consumes SELFCHECK.json. Writing the episode's high-water limit (see the gate comment) makes the pair coherent.

chent1996 (changes requested):
## Review

The problem is real, the shape of the fix is right (floor + proportional allowance + hard cap, mirroring `_batch_media_verify_round_limit`), CI is green including the KBC box suite, and I verified the no-conflict status against latest `main` and that the flat-budget floor keeps every pre-existing small-workload behavior unchanged (`limit ≥ base` always).

**Requesting changes for one verified issue**: the limit is re-evaluated each turn from the *current* report, so the allowance shrinks as the repair converges, and the gate closes around the halfway point of exactly the large-adoption scenario this PR exists to fix — the cited 55-violation run would still end `unconverged` with ~19 residuals under the PR's own nominal 12-per-round assumption. The sibling idiom avoids this by computing its limit once from the *initial* workload. Details, the arithmetic, and a ~3-line high-water-mark fix are on the gate line; the `repair_rounds_limit` observability field inherits the same fix.

What would unblock: compute the episode's allowance from its initial (or high-water) workload, reset alongside `_l1_repairs_used`, plus one loop-level test case pinning that a shrinking workload keeps the initial allowance. Happy to re-review immediately — the rest of the change is exactly right.

Deploy note per the kbc matrix row: box behavior change, so this needs a `kbc-compile-box` image rebuild + redeploy to take effect.

kbc/platform/pod/compile_box.py
LikiosSedo: Confirmed — the formula was mirrored but not the evaluation timing, and your table is exactly what the loop would have done. Fixed as suggested: per-episode high-water via `run._l1_round_limit` (init with the counter, updated with `max()` at the gate, reset on clean ledger and at the fresh batch-ledger episode in `_run_ledger_repairs`). `max()` also covers mid-episode reopening, and the persisted report now writes the episode limit, so `repair_rounds_used` can no longer exceed `repair_rounds_limit` on the unconverged exit — both your comments resolve to the same line.

Added the loop-level test you asked for: a 30-violation episode records limit 3, the workload collapses to 2 after round 1 (28 orphans removed), the gate stays open through rounds 2–3, and the unconverged report shows used=3/limit=3. Full repair suite green.

LikiosSedo:
Rebased onto main after #426/#433 landed (the conflict was both branches touching the selfcheck gate region / the same test anchor).

Post-rebase composition is exactly the intended layering: #426's mechanical normalization runs FIRST, then `run_layer1` reports on the normalized state, then this PR's high-water limit is computed from that report — so the scaled budget sizes itself to the workload that actually needs model judgment, after the deterministic layer has eaten the mechanical share. This PR remains needed for the residue #426 doesn't cover (dangling citations requiring judgment, unaccounted-source coverage work, links/structure): a large first compile still opens far more than the flat 2 rounds.

Combined suite green: budget floor/scale/cap + high-water shrink test + #426's alias auto-repair test + residual-ticket and media-verify suites (9 tests).

chent1996 (approved):
## Review (second pass — approving)

Re-reviewed at `3fcb0429`. The finding is fixed exactly as discussed, and more completely than I asked.

**The high-water gate is right.** `run._l1_round_limit = max(prev, _l1_repair_round_limit(report))` is taken *before* the clean check, the gate now compares against the episode's high-water allowance, and — as the comment notes — `max()` also lets the budget legitimately grow if a repair reopens violations mid-episode. Resets are complete: `__init__`, the clean-ledger path (alongside `_l1_repairs_used`), and the batch-tail `_run_ledger_repairs` entry — that third reset site keeps the batch-final phase's budget per-phase, which I hadn't even enumerated. The observability half is fixed too: `repair_rounds_limit` persists the episode high-water, so an unconverged report can no longer show `used > limit`.

**The loop-level test pins the exact failure mode.** Workload 30 → limit 3; the workload shrinks mid-episode to 2 (per-turn recomputation would have closed the gate at `used=1`); the test asserts the gate stays open through `used=1,2`, closes at `used=3`, and the persisted report shows a coherent `3/3` on the unconverged exit. That is precisely the regression guard this needed. CI is green across all four checks, including the KBC box suite.

With this fix, the motivating 55-violation adoption run gets its full `ceil(55/12)=5`-round allowance regardless of how fast rounds converge — the original intent, now actually delivered.

**One mechanical item before merge**: the branch is now conflicting with `main` — #426 (`16b16099`, mechanical citation normalization) landed and edits the same `_post_turn_selfcheck` region. The two changes compose cleanly (the mechanical pass runs before Layer-1 and shrinks the workload the high-water is derived from — synergy, not tension), so the rebase should be conflict-resolution-only. This approval covers `3fcb0429`; if the rebase resolution changes any logic beyond mechanical merging, ping me and I'll take another look.

Deploy note as before: box behavior change — `kbc-compile-box` image rebuild + redeploy required.

Issue #431, opened 2026-07-22
kbc: image-verification doubts should self-resolve against co-cited text sources before escalating to the owner


LikiosSedo:
## Observed (adoption run, siflow-test, 2026-07-22)

The isolated image re-reader files owner question cards like:

- "B300 FP4 算力较 B200 +46% 是否应保留?" — source note: *独立图像转写未包含 FP4 数据* (asset APijbb…png)
- "Qwen3-A3B 小结中的持平与 +39.7% 结论是否应保留?" — *独立图像转写未包含该小结*

In both cases the SAME page co-cites the parent **markdown** source which contains the exact figure verbatim (`16-NVIDIA B系列GPU卡性能测试分析.md`: "FP4有46%提升" L43/L224; "+39.7%" L221). The claim is fully machine-verifiable without the image — the image is corroborating material, not the sole source.

## Problem

Escalation policy treats "this cited image does not contain the number" as "single-source unverified", without first checking the claim against the page's other compiled_from sources. Result: dozens of user-facing cards whose correct resolution is mechanical (keep the claim, bind it to the text source), burning owner attention and diluting the genuinely undecidable contradictions.

## Proposal

Before filing an owner card for an image-verification miss:
1. grep the claim's figure/key phrase against the page's co-cited text sources;
2. hit → self-resolve: keep the claim, keep the text binding, optionally annotate the image as corroboration-only — no card;
3. no hit anywhere → escalate as today (true single-source image claim).

This mirrors the compile constitution's own rule that a claim's authority is the text export, images corroborate.

LikiosSedo:
Sharper diagnosis after discussing with the KB owner — the blind re-reader itself is CORRECT (its output "the image does not contain FP4 data" is a true statement). The defect is in the **attribution/judgment layer** above it, which applies the inference *image lacks X ⇒ claim X is doubtful*. That inference is only valid when the image is the claim's **sole source**. Curated KBs commonly attach whole asset sets at page level as corroboration (the local wiki's "assets follow the parent doc" rule), so a corroborating image's silence about a figure says nothing about the figure.

Root cause one level deeper: `compiled_from` is a flat list with **no per-entry role** (data source vs corroboration vs container), so the verifier cannot know which images are load-bearing.

Severity note: the resulting owner cards offer a "delete this conclusion" option. A high-false-positive verifier plus a delete button means the QA loop can actively destroy TRUE content when an owner clicks through without deep corpus familiarity — worse than no verification.

Fix ladder:
[...]

PR #433, opened 2026-07-22
fix(kbc): auto-resolve image-verification doubts backed by co-cited text sources

LikiosSedo:
Closes the triage step of #431.

## Problem

The blind-verify comparer treats *cited image lacks X* as *claim X is doubtful*. Valid only when the image is the claim's **sole** source; curated pages cite whole asset sets as corroboration while figures come from co-cited text exports. Live adoption run (2026-07-22): "FP4 +46%" / "+39.7%" escalated as owner question cards though the co-cited markdown states both verbatim — and the cards offer **delete this conclusion**, so false positives + a delete button can destroy true content.

## Change

Deterministic post-filter at the compare seam (`suppress_text_backed`):

- extract distinctive numeric tokens from the claim (percent/decimal first; plain integers only at ≥3 digits — bare "8"/"70" never qualify);
- grep them against the page's co-cited text sources under `raw/` (md/txt/log/json/csv, 2MB cap per file);
- hit → finding suppressed as `text_source_backed` (image demoted to corroboration, no owner card), surfaced via `suppressed_text_backed` in the result + a progress line;
- `不一致` findings are **never** suppressed (text-backed claim vs conflicting image = genuine cross-source conflict);
- qualitative claims without distinctive numbers keep escalating unchanged.

The transcriber and comparer prompts are untouched — the fix is attribution logic, not perception.

## Tests

New unit test (suppress/keep/conflict/qualitative paths) + existing media-verify suite green.

### DIFF
--- a/kbc/platform/pod/mediaverify.py
+++ b/kbc/platform/pod/mediaverify.py
@@ -36,8 +36,10 @@ import shutil
 import tempfile
 from pathlib import Path
 
+import re
+
 from redblue import _agent_json  # JSON call + one lenient retry, shared shape
-from selfcheck import _is_en  # locale gate: None/en → English, zh → Chinese
+from selfcheck import _is_en, parse_compiled_from  # locale gate + citation parse
 
 TRANSCRIPTS_PATH = "authoring/MEDIA_TRANSCRIPTS.json"
 _TRANSCRIPT_CHAR_CAP = 6000  # per image, persisted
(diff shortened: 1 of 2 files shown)


kbc/platform/pod/mediaverify.py
```diff
@@ -230,6 +232,80 @@ async def compare_page(engine, tmp_dir: str, page_rel: str, page_text: str,
...
+    text = _co_cited_text(workdir, page_text)
+    if not text:
+        return findings, []
+    kept, suppressed = [], []
+    for f in findings:
+        if f.get("kind") != "超出转写范围":
+            kept.append(f)
+            continue
+        tokens = _distinctive_tokens(str(f.get("claim") or ""))
+        if tokens and any(tok in text for tok in tokens):
```
chent1996, line 301: **Non-blocking, two precision suggestions on the suppression predicate.**

1. `any(...)` means one token hit suppresses a claim that may carry several distinctive numbers. For a claim like "FP4 +46%, bandwidth +39.7%", a co-cited doc containing an unrelated `46%` (different metric) suppresses the whole finding. Since the failure mode here is a *missed* owner card (bounded, page content untouched), this is a tuning question, not a bug — but `all(...)` would still handle the motivating single-number case ("FP4有46%提升") while being materially harder to false-match on multi-number claims.

2. `_distinctive_tokens` treats any ≥3-digit integer as distinctive, which admits 4-digit years: a claim whose only numeric token is `2024`/`2025` gets suppressed by virtually any co-cited doc (years appear everywhere). Consider demoting `19\d\d`/`20\d\d` bare integers to non-distinctive, same as `8`/`70`.

Both keep the conservative direction intact (no token → keep the card; unreadable text → keep the card).

chent1996 (approved):
## Review

Approving. This is the right fix at the right seam: attribution logic after the comparer, prompts untouched, deterministic and observable.

What I verified:

- **The suppression is fail-safe in every failure direction.** No distinctive tokens → card kept; co-cited text unreadable/absent → card kept; `不一致` (genuine cross-source conflict) is never a candidate; and the comparer's JSON fallback defaults `kind` to `不一致` — i.e. the *never-suppressed* side — so a malformed model response cannot slip a finding past the owner.
- **The `kind` vocabulary is locale-fixed wire tokens** (both en and zh comparer prompts demand exactly `不一致`/`超出转写范围`), so the filter works identically in English runs — a concern I checked and can rule out.
- **Suppression is not silent**: every suppressed finding is carried in `suppressed_text_backed` with a `suppressed_reason`, plus a progress line — auditable after the fact.
- **Bounded I/O**: co-cited text only (page's own `compiled_from`), text extensions whitelist, 2 MB per-file cap.
- The failure mode of a wrong suppression is a missed question card, while the failure mode it fixes is a false-positive card *with a delete button* pointed at true content (live 2026-07-22 run). That trade is right.

One inline note with two non-blocking precision suggestions on the token predicate (`any` vs `all`, and year-like integers).

kbc deploy note: box behavior change — takes effect only after a `kbc-compile-box` image rebuild + redeploy.

PR #434, opened 2026-07-22
feat(kbc): de-stream the compile box's Anthropic calls to dodge the gateway charset bug

LikiosSedo:
## Problem

massapi's Anthropic-compatible **streaming** route corrupts multibyte characters whose bytes straddle an SSE chunk boundary (per-chunk UTF-8 decode with `errors=replace` → U+FFFD). Its **non-streaming** route returns the same bytes intact (gateway bug report). Compile runs write long Chinese pages, so every turn is a corruption lottery; the charset guard heals what it catches but burns repair budget.

## Change (v2 — default-on, session-scoped, zero deployment config)

A localhost de-streaming shim inside the pod (`destream.py`):

```
CLI --stream:true--> 127.0.0.1 shim --stream:false--> gateway
                     <-- synthesized SSE + keepalive pings --
```

- **Default ON in the binary, scoped by session kind** via per-session CLI env (`ClaudeAgentOptions.env`, merged over the inherited environment): authoring / batch / planner sessions and the readonly verify agents are de-streamed; **test sessions always keep true streaming** (interactive UX); the codex/OpenAI route never rewrites. No helm/env change to deploy; `KBC_DESTREAM=0/off` is the operator escape hatch — and leaving the shim on after the gateway is fixed is harmless (identical turn wall-time, MB-level buffering).
- The box's own environment is never rewritten; the shim resolves the real upstream per request (self-loop guard included).
- Only `POST …/messages` with `stream:true` is de-streamed; everything else passes through as **raw bytes** — the shim never introduces a decode step of its own.
- Full SSE synthesis: `message_start` / `content_block_*` with whole-block deltas (text, `tool_use` `input_json_delta`, `thinking` + `signature_delta`) / `message_delta` with usage / `message_stop`; upstream HTTP errors ride the open stream as SSE `error` events.
- **Stall-watchdog interaction handled**: de-streamed turns emit no SDK deltas mid-request, so the fine-grained idle bound (90s/240s) would false-kill long generations. When the shim is active the watchdog bound is floored at `KBC_DESTREAM_MODEL_IDLE_TIMEOUT_S` (default 900s) — black-hole reaping coarsens from seconds to minutes as the honest price, still far under the CLI's own ~60min request timeout.

**Blast radius**: the kbc-compile-box image only (127.0.0.1 listener). runtime / agentbox / sicore-side LLM callers (workflow nodes etc.) are untouched.

## Residual risk to verify at rollout

One live long-turn test against massapi non-streaming (whether the gateway caps non-streaming response duration).

## Tests

Six tests: SSE synthesis round-trip (byte-perfect CJK, tool_use partial_json, usage split), keepalive under a slow upstream, error mapping, raw passthrough, and default-on/session-scoped/opt-out/codex activation semantics. `compile_box`/`engine` import + adjacent suites green.

### DIFF
--- a/kbc/platform/pod/compile_box.py
+++ b/kbc/platform/pod/compile_box.py
@@ -62,6 +62,7 @@ from claude_agent_sdk import (
 )
 
 import batching
+import destream
 import incremental
 import mediaverify
 import office_ingest
@@ -4637,6 +4638,9 @@ def _apply_session_config(body: dict) -> None:
             if k == "KBC_PK_MODE" and _PK_KILL_AT_BOOT:
                 continue  # ops kill switch outranks consumer config
             os.environ[k] = str(value)
+    # After the final env state lands: opt the Anthropic route in/out of the
+    # in-pod de-streaming shim (KBC_DESTREAM, charset fix — see destream.py).
+    destream.maybe_activate()
 
 
 async def handle_session(request: web.Request):
@@ -5300,6 +5304,7 @@ def build_app() -> web.Application:
         client_max_size=_http_max_request_bytes(),
         middlewares=[_client_certificate_middleware],
     )
+    app.on_startup.append(destream.start)
     app.on_shutdown.append(_flush_on_shutdown)
     app.add_routes([
         web.post("/sources", handle_sources),
--- a/kbc/platform/pod/destream.py
+++ b/kbc/platform/pod/destream.py
@@ -1,0 +1,222 @@
+"""In-pod de-streaming shim for the Anthropic Messages route (charset fix).
+
+massapi's Anthropic-compatible STREAMING route corrupts multibyte characters
+whose bytes straddle an SSE chunk boundary (the gateway decodes each chunk
+independently with errors=replace → U+FFFD); its NON-streaming route returns
+the same bytes intact (gateway bug report, 2026-07). The compile box is
+non-interactive — nobody reads its model output token-by-token — so it can
+trade streaming latency for byte-correct text. The CLI (Claude Code) has no
+non-streaming switch, hence a localhost shim:
+
+    CLI --stream:true--> 127.0.0.1 shim --stream:false--> upstream gateway
+                         <-- synthesized SSE (+ pings while waiting) --
+
+Scope: only POST .../v1/messages bodies with "stream": true are de-streamed.
+Everything else (count_tokens, models, non-streaming posts) passes through
+verbatim as RAW BYTES — the shim must never introduce a decode step of its
+own. Activation is per-profile via the consumer-settings whitelist
+(KBC_DESTREAM=1 on the compile profile); test-session boxes keep true
+streaming for interactive UX.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import json
+import os
+
+import aiohttp
+from aiohttp import web
+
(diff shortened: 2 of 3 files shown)


chent1996 (approved):
## Review

Approving. This attacks the actual root cause behind a whole family of recent symptoms — the charset-guard healing burn, the repair-budget bump, the U+FFFD lint class all trace back to massapi's per-chunk SSE decode — and it does so as a contained, opt-in, retirable workaround rather than more downstream healing. The layering is right: the selfcheck charset guard stays as belt-and-suspenders, and this removes the corruption at the source for the one workload (non-interactive compile) that can afford non-streaming.

### What I verified

- **The activation chain is real end to end.** The PR adds no whitelist entry because none is needed: the consumer-settings whitelist is prefix-based (`KBC_*`, excluding only `KBC_ENGINE`), so `KBC_DESTREAM` and both knobs travel through unchanged — I checked `_apply_session_config` on the branch. `maybe_activate()` runs *after* both the llm block (which finalizes `ANTHROPIC_BASE_URL`) and the settings loop, so the swap decision always sees the final env. I also walked the re-attach, opt-out, and llm-absent sequences — all converge (the `base != shim_url` guard prevents self-capture on repeated calls).
- **Scope and the "no decode of our own" claim hold.** Only `POST …/messages` with `stream:true` is de-streamed; `…/messages/count_tokens` fails the `endswith("/messages")` check and passes through; passthrough relays raw bytes with `auto_decompress=False` and hop-by-hop headers stripped. The de-streamed leg's only decode is `json.loads` of a *complete* response body — no chunk boundaries, which is the entire point.
- **The synthesized SSE matches the Anthropic stream vocabulary**: `message_start` (empty content, input-tokens-only usage) → per-block `content_block_start`/whole-block deltas (`text_delta`, `input_json_delta`, `thinking_delta` + `signature_delta`)/`content_block_stop` → `message_delta` (stop_reason + output_tokens) → `message_stop`; unknown block types ship whole with no delta (forward-compatible); `event: ping` is Anthropic's own keepalive shape; upstream errors ride the open stream as `error` events, matching real mid-stream error behavior.
- **Client disconnect doesn't orphan the upstream turn**: aiohttp cancels the handler on peer disconnect, `CancelledError` is re-raised (not swallowed by the broad except), and the `async with` teardown drops the upstream POST — no tokens burned into a dead socket. The ping task is cancelled on every exit path.
- Tests cover the five behaviors that matter (byte-perfect CJK round-trip, keepalive under a slow upstream, error mapping, verbatim passthrough, activation env logic), and CI is green including the KBC box suite.

### Non-blocking notes

1. The de-streamed leg synthesizes its own 200 response, so upstream response headers (`anthropic-request-id` et al.) are dropped — worth copying the request-id through onto the SSE response headers someday for cross-system debugging, cosmetic otherwise.
2. The author's own rollout flag is the right gate: one live long-turn against massapi's non-streaming leg (its response-time cap is the residual unknown) before enabling on the compile profile.
3. Please track the upstream massapi fix so this shim gets retired rather than fossilized — the per-profile env opt-in is already the kill switch, it just needs someone to remember to flip it when the gateway is fixed.

kbc deploy note: box behavior change — `kbc-compile-box` image rebuild + redeploy, then enable `KBC_DESTREAM=1` on the compile profile only.

PR #436, opened 2026-07-22
fix(tools): filter kubectl exec transport noise

caihongwei2006:
Default debug mode can expose Docker/Moby and kubectl exec Go transport diagnostics in pod_exec stderr. These SPDY/WebSocket stream logs are wrapper noise and can overwhelm the agent context. Filter known transport lines while preserving actual container, OCI, and Kubernetes errors.

Adds regression coverage for the filter.

### DIFF
--- a/src/tools/cmd-exec/pod-exec.ts
+++ b/src/tools/cmd-exec/pod-exec.ts
@@ -12,7 +12,7 @@ import { loadConfig } from "../../core/config.js";
 import { parseArgs, CONTAINER_SENSITIVE_PATHS } from "../infra/command-sets.js";
 import { preExecSecurity, postExecSecurity } from "../infra/security-pipeline.js";
 import { backgroundNotLineSafeError, backgroundLaunchedResult } from "./background-launch.js";
-import { validatePodName, prepareExecEnv } from "../infra/exec-utils.js";
+import { validatePodName, prepareExecEnv, filterPodNoise } from "../infra/exec-utils.js";
 import { resolveRequiredKubeconfig } from "../infra/kubeconfig-resolver.js";
 import { ensureClusterForTool } from "../infra/ensure-kubeconfigs.js";
 
@@ -236,7 +236,7 @@ Examples:
         );
 
         return {
-          content: [{ type: "text", text: postExecSecurity(stdout.trim(), pre.action, { stderr: stderr.trim() || undefined }) }],
+          content: [{ type: "text", text: postExecSecurity(stdout.trim(), pre.action, { stderr: filterPodNoise(stderr.trim()) || undefined }) }],
           details: { exitCode: 0 },
         };
       } catch (err: any) {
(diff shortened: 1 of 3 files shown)


src/tools/cmd-exec/pod-exec.ts
```diff
@@ -246,7 +246,7 @@ Examples:
           return { content: [{ type: "text", text: "Aborted." }], details: { error: true } };
         }
         const stdout = (err.stdout?.trim() ?? "") as string;
-        const stderr = (err.stderr?.trim() ?? err.message) as string;
+        const stderr = filterPodNoise((err.stderr?.trim() ?? err.message) as string);
```
jacoblee-io, line 249: Optional (non-blocking): `filterPodNoise(err.stderr?.trim() ?? err.message)` — in practice `err.message` is always a string for an execFile/spawn error, so `filterPodNoise` never receives `undefined`. Only flagging that if `err` were ever a non-Error without a `message`, `filterPodNoise(undefined).split` would throw; not triggerable on this path, so fine as-is.

src/tools/infra/exec-utils.ts
```diff
@@ -165,13 +165,19 @@ export function stdinExecCmd(interpreter: "bash" | "python3", escapedArgs?: stri
...
 export function filterPodNoise(stderr: string): string {
   return stderr
     .split("\n")
-    .filter((line) => !line.match(/^pod "node-debug-.*" deleted$/))
+    .filter((line) => {
+      if (line.match(/^pod "node-debug-.*" deleted$/)) return false;
+      return !line.match(
+        /^[IWEF]\d{4}\s+\d{2}:\d{2}:\d+\.\d+\s+\d+\s+log\.go:\d+\].*(?:Create stream|Stream added, broadcasting:|Reply frame received for|Data frame received for|Data frame handling|Data frame sent|Stream removed, broadcasting:|Go away received).*$/,
+      );
+    })
```
jacoblee-io, line 180: Note: `[IWEF]` intentionally covers Error/Fatal klog levels too, but the trailing transport-phrase gate means only genuine SPDY/WebSocket noise is dropped even at E/F level — a real error-level diagnostic (different source file, or no transport phrase) is preserved. Correct as written; noting it so the intent is on record.

jacoblee-io (approved):
LGTM. Small, precise, well-tested. The extended `filterPodNoise` regex requires all three of the klog prefix + `log.go:NNN]` + a specific transport phrase before dropping a line, so I verified it preserves every real error (`controller.go]` errors, `error:` lines, `Error from server (NotFound)`, and phrase-less klog lines) while removing only the SPDY/WebSocket wrapper noise; the `^` anchor also keeps indented lines from matching. No regression for the other callers (node-exec.ts / node-script.ts) — they run via kubectl exec too and emit the same noise, so the broadening is a consistent win, and nothing depends on those klog lines being present. It also doesn't bypass the security pipeline: filtering only removes noise lines before `postExecSecurity`, which still sanitizes whatever remains. Two optional non-blocking notes inline. Ship it.

PR #438, opened 2026-07-22
fix(kbc): de-stream v2 — default-on session-scoped activation + stall-watchdog floor (follow-up to #434)

LikiosSedo:
## Summary

Follow-up fix to #434, which merged while the in-pod de-streaming shim was still at **v1**. v1 works but has two problems in the way it activates. This PR replaces the v1 activation model with a session-scoped, default-on one and handles the stall-watchdog interaction. Behavior of the shim's actual de-streaming (SSE synthesis, passthrough, charset fix) is unchanged.

## Why v1 needed a follow-up

**#434 merged at v1** — activation via a global `ANTHROPIC_BASE_URL` rewrite, gated on an opt-in `KBC_DESTREAM` flag, with no handling of the stall watchdog. Two concrete issues:

1. **Global env rewrite has blast radius beyond the compile session.** v1 rewrote the pod-wide `ANTHROPIC_BASE_URL` in `os.environ` (`destream.maybe_activate()` at the end of `_apply_session_config`). That redirection is inherited by *any* later CLI invocation in the same pod — including interactive **test sessions**, which must keep true streaming for UX.
2. **Stall-watchdog interaction unhandled.** De-streamed turns emit **no SDK deltas** mid-request (the whole response arrives at once after the upstream non-streaming call returns). The fine-grained model-idle bound (90s / 240s) would therefore false-kill any model generation longer than ~90s and retry it needlessly.

## What v2 changes

- **Activation is default-on and session-scoped, with zero deployment config.** Replaces `maybe_activate()` (global env rewrite) with:
  - `enabled()` — default ON for the Anthropic engine; `KBC_DESTREAM=0/off/false/no` is the sole operator escape hatch; `KBC_ENGINE=codex_sdk` and "no shim listener" both disable it.
  - `session_env(kind)` — returns the shim URL only for `authoring` / `verify` sessions; **`test` always returns `{}`** so interactive sessions keep true streaming. Injected via the SDK `options.env` (merges over the inherited environment, so credentials stay put) at `_compile_session_opts`, the batch planner session, and `ClaudeEngine.run_readonly_agent`.
  - `_upstream_url()` — resolves the real upstream **per request** from `os.environ` with a self-loop guard, so the box's own environment is never rewritten (no global rewrite/restore bookkeeping).
- **Stall-watchdog floor.** `model_idle_floor()` (default `900s`, env `KBC_DESTREAM_MODEL_IDLE_TIMEOUT_S`) floors the watchdog `bound` in `_model_stall_watchdog` while the shim is active. Black-hole reaping coarsens from seconds to minutes — the honest price of the charset fix, still far under the CLI's own ~60min request timeout.
- **Tests.** `test_destream.py` gains v2 semantics coverage (default-on, session-scoped, test always streams, opt-out, codex out of scope, box env not rewritten), replacing the old `maybe_activate` test; the other 5 destream tests are unchanged.

## Blast radius

The **kbc-compile-box image only** — the shim binds `127.0.0.1` inside the compile-box pod. runtime / agentbox / sicore-side LLM callers never see it.

## Verification

- `python3` AST parse of all four touched files — clean.
- `test_destream.py` — all 6 tests pass (5 original + the replacement v2 test).
- `import compile_box` — clean; adjacent `test_compile_box.py` tests pass: `test_apply_session_config`, `test_batch_planner_uses_compile_path_guard`, `test_run_wrapper_terminal_signals`.

## Merge note

This branch is cut from current `main` (post-#434). The compile_box.py hunks auto-merged cleanly over the L1-repair-budget changes from #430 that landed after v1 — no conflict; both change sets coexist.

### DIFF
--- a/kbc/platform/pod/compile_box.py
+++ b/kbc/platform/pod/compile_box.py
@@ -3532,6 +3532,18 @@ async def _consume_turn_stream(
 _STALL_INTERRUPT_DEADLINE_S = int(os.environ.get("KBC_STALL_INTERRUPT_DEADLINE_S", "120"))
 
 
+def _watchdog_idle_bound(tool_pending: bool, model_idle_timeout: float, destream_floor: float) -> float:
+    """Idle bound for the stall watchdog.
+
+    A pending tool is LOCAL execution and keeps its own (shorter) bound;
+    de-streaming only affects the model-request phase, so the de-stream floor
+    lifts ONLY the model bound — it must never extend a stuck tool's timeout.
+    """
+    if tool_pending:
+        return _MODEL_TOOL_IDLE_TIMEOUT_S
+    return max(model_idle_timeout, destream_floor) if destream_floor else model_idle_timeout
+
+
 async def _model_stall_watchdog(run: CompileRun) -> None:
     """Reap a turn wedged on a black-holed model request. Interrupt the attempt;
     _consume_turn_stream then re-issues it (or fails). Only judges an ACTIVE turn,
@@ -3576,13 +3588,12 @@ async def _model_stall_watchdog(run: CompileRun) -> None:
             if run._model_idle_timeout_s is not None
             else _MODEL_IDLE_TIMEOUT_S
         )
-        bound = _MODEL_TOOL_IDLE_TIMEOUT_S if run._tool_pending else model_idle_timeout
-        # De-streamed turns produce no SDK deltas mid-request: the idle bound
-        # must cover a whole model request or the watchdog false-kills long
-        # generations (see destream.model_idle_floor).
-        floor = destream.model_idle_floor()
-        if floor:
-            bound = max(bound, floor)
+        # De-streamed model turns emit no SDK deltas mid-request, so the model
+        # idle bound must cover a whole generation or the watchdog false-kills
+        # long ones. A pending tool is unaffected by de-streaming and keeps its
+        # own (shorter) bound (see _watchdog_idle_bound / destream.model_idle_floor).
+        bound = _watchdog_idle_bound(
+            run._tool_pending, model_idle_timeout, destream.model_idle_floor())
         idle = time.monotonic() - run._last_model_activity
         if idle <= bound:
             continue
@@ -4268,6 +4279,9 @@ def _recommendation_session_opts(
         max_buffer_size=SDK_MAX_BUFFER_BYTES,
         session_id=str(uuid.uuid4()),
         session_store=InMemorySessionStore(),
+        # Non-interactive red-team reviewer → de-streamed Anthropic route
+        # (charset fix); merged over the inherited env by the SDK.
+        env=destream.session_env("verify"),
         hooks={"PreToolUse": [HookMatcher(hooks=[_make_recommendation_path_guard(root, parent.locale)])]},
     )
 
@@ -4538,6 +4552,9 @@ def _reference_assist_session_opts(parent: "CompileRun", root: Path, submit_tool
         max_buffer_size=SDK_MAX_BUFFER_BYTES,
         session_id=str(uuid.uuid4()),
         session_store=InMemorySessionStore(),
+        # Non-interactive reference-answer assistant → de-streamed Anthropic
+        # route (charset fix); merged over the inherited env by the SDK.
+        env=destream.session_env("verify"),
         hooks={"PreToolUse": [HookMatcher(hooks=[_make_recommendation_path_guard(root, parent.locale)])]},
     )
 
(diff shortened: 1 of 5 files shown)


LikiosSedo:
Addressed the three review points (pushed as 041147a2 + 9d181413):

**[P1] Two non-interactive Claude sessions now use the shim.** `_recommendation_session_opts()` and `_reference_assist_session_opts()` were still built with an empty SDK env and hit the streaming gateway directly, re-exposing the cross-chunk charset corruption for their Chinese questions/answers and structured tool input. Both now set `env=destream.session_env("verify")`, matching the authoring/planner/verify entrypoints. The interactive test session is untouched and keeps true streaming.

**[P2] Watchdog floor no longer extends the tool timeout.** The de-stream floor was applied unconditionally, lifting the `_tool_pending` bound (660s) to 900s and delaying reaping of a wedged tool by ~4min. De-streaming only affects the model-request phase, so bound selection is now factored into a pure `_watchdog_idle_bound()` helper that applies the floor to the model bound only, never to a pending tool.

**[P2] test_destream.py is now in CI.** Added an explicit `python test_destream.py` step to the KBC Box CI workflow (`.github/workflows/kbc-ci.yml`) next to `test_compile_box.py`, so the green "KBC Box Test" check actually covers the v2 shim/session-scoping/watchdog tests. Also added two integration assertions: both reviewer callers' `opts.env` carry the shim URL (and clear to `{}` on operator opt-out), and the watchdog floor never extends a pending-tool bound.

Verification: `test_destream.py` (7 tests incl. the 2 new integration ones) and `test_compile_box.py` both green; AST syntax check on all changed files clean; workflow YAML validates.

kbc/platform/pod/compile_box.py
```diff
@@ -2219,6 +2219,9 @@ def _compile_session_opts(run: "CompileRun", wd: str, system_prompt: str, sessio
         max_buffer_size=SDK_MAX_BUFFER_BYTES,
         session_id=session_id,
         session_store=InMemorySessionStore(),
+        # Non-interactive session → de-streamed Anthropic route (charset fix);
```
jacoblee-io, line 2222: Optional (maintainability): the whole v2 activation — and especially the disabled/opt-out path where `session_env` returns `{}` — depends on the SDK MERGING `options.env` over `os.environ` rather than replacing it. Verified true at the pinned `claude-agent-sdk==0.2.110` (`subprocess_cli.py`: `{**os.environ, ..., **options.env, ...}`, and `env` defaults to `{}` not None). Worth a one-line comment (here or by the requirements pin) recording this dependency, so a future SDK bump that changed to replace-semantics — or that distinguished `None` from `{}` — would be caught in review rather than silently stripping ANTHROPIC_AUTH_TOKEN/PATH from every authoring/verify child.

kbc/platform/pod/compile_box.py
```diff
@@ -3528,6 +3532,18 @@ async def _consume_turn_stream(
 _STALL_INTERRUPT_DEADLINE_S = int(os.environ.get("KBC_STALL_INTERRUPT_DEADLINE_S", "120"))
 
 
+def _watchdog_idle_bound(tool_pending: bool, model_idle_timeout: float, destream_floor: float) -> float:
+    """Idle bound for the stall watchdog.
+
+    A pending tool is LOCAL execution and keeps its own (shorter) bound;
+    de-streaming only affects the model-request phase, so the de-stream floor
+    lifts ONLY the model bound — it must never extend a stuck tool's timeout.
+    """
```
jacoblee-io, line 3541: Nice: `_watchdog_idle_bound` keeps the de-stream floor from ever extending a wedged *tool's* timeout (tool_pending → the shorter tool bound regardless of floor), so black-hole reaping only coarsens for the model-request phase where the shim genuinely suppresses deltas. Correct and well-tested.

jacoblee-io (approved):
LGTM. Clean v2: activation moves from v1's pod-wide `os.environ` rewrite to session-scoped SDK `options.env` injection (authoring/verify/planner get the shim; `test` always returns `{}` so interactive sessions keep true streaming; codex and no-listener disable it), the old global `maybe_activate()` is fully removed from `_apply_session_config`, and `_upstream_url()` resolves the real gateway per-request with a self-loop guard so the box env is never rewritten. The stall-watchdog floor is scoped correctly (`_watchdog_idle_bound`): a pending tool always keeps its own shorter bound, and the de-stream floor lifts only the model-request bound and never shrinks an already-larger one — all covered by the new tests, plus an integration guard that the recommendation + reference-assist reviewer sessions thread the shim env.

I verified the one load-bearing assumption: every injection relies on the SDK merging `options.env` over the inherited environment (else authoring/verify sessions lose ANTHROPIC_AUTH_TOKEN, and the empty-`{}` disabled path would strip PATH entirely). Against the pinned `claude-agent-sdk==0.2.110` tag, `subprocess_cli.py` builds `{**os.environ (minus CLAUDECODE), ..., **options.env, ...}` and `env` defaults to `{}` (not None), so both the shim path and the empty-`{}` path inherit full os.environ. Confirmed correct. One optional maintainability note inline. Ship it.

PR #440, opened 2026-07-22
feat(mcp): steer callers toward parallel, evidence-cited investigations

LikiosSedo:
Mirror of sicore `feat/a2a-mcp-caller-guidance` — the two instruction surfaces (sicore `mcpInstructions`/tool defs and this adapter's `SERVER_INSTRUCTIONS`/`TOOL_DEFINITIONS`) are a verbatim-sync contract.

## Why

Live testing against the production MCP endpoint (2026-07-22) showed the tool layer is healthier than the instructions imply:

- **Concurrency works**: 3 investigations submitted simultaneously all progressed in parallel (two tasks running tools at the same instant), all completed within 50s wall clock. The tracker admits up to 50 active tasks per agent. But the current instructions read as a serial protocol, so consuming models investigate one hypothesis at a time.
- **Capability self-description works when asked**: "list your bound clusters" returns all 9 cluster names in ~15s — but a caller has to know to ask, otherwise it burns a round guessing target names.
- **Evidence citation works when asked**: requesting "list data sources consulted + confidence" produced a field-level 5-source evidence appendix with a calibrated confidence estimate.

All three are caller-guidance gaps, not server gaps. This adds three sentences to the instructions and one clause to the `siclaw_investigate` description; no schema or behavior change.

## Testing

- 30 adapter tests green (`npx vitest run`).

### DIFF
--- a/mcp/sicore-a2a-adapter/src/server.ts
+++ b/mcp/sicore-a2a-adapter/src/server.ts
@@ -6,7 +6,10 @@ import { createToolHandler, TOOL_DEFINITIONS } from "./tools.js";
 
 export const SERVER_INSTRUCTIONS = [
   "Use siclaw_investigate for operational questions that require the configured Siclaw SRE agent.",
-  "When it returns a non-terminal task, keep the current turn open and call siclaw_wait_task with the same task_id until terminal, unless the user requests fire-and-forget, asks to stop, or the overall investigation deadline is exhausted.",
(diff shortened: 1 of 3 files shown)


LikiosSedo:
Closing without merge: all current consumers use the HTTP MCP endpoint served by sicore (/api/v1/mcp), so the stdio adapter's instruction text has no runtime effect. The sicore side carries the change. Branch feat/a2a-mcp-caller-guidance is kept as a ready-made sync patch if the adapter ever ships a v0.1.1.

PR #442, opened 2026-07-22
feat(kbc): rename compile box image to siclaw-kbc-box and pod prefix to kbc-box

LikiosSedo:
## Motivation (operations)

During production troubleshooting every pod is named `agentbox-<id>`, so a KB compile box and a chat agentbox can only be told apart by label/image filtering — the compile box has bitten on-call more than once. This makes the compile box distinguishable at a glance in `kubectl get pods`.

## Changes

- **Image rename** `kbc-compile-box` → `siclaw-kbc-box`: Makefile build/push targets, helm `siclaw.compileBoxImage` derivation, `values.yaml`, README/docs. The explicit `agentbox.compileBoxImage` override semantics are **unchanged**.
- **Pod-name prefix** declared on the BoxProfile (`podNamePrefix`, platform-style — no `if profile==` in the spawner): `kb-compile` and `kb-compile-codex` spawn as `kbc-box-<id>` instead of `agentbox-<id>`. Resources derived from the pod name (cert Secret, hostname) follow the prefix. Chat agentboxes and the read-only, ephemeral `kb-test` box keep the `agentbox-` prefix (see note below).
- **Upgrade compatibility (auto-cleanup)**: on the first post-upgrade compile spawn, the spawner reaps any pod left under the old `agentbox-<id>` name for that agent — guarded to compile boxes only (`boxType` label `kb-compile*`), so a chat box under the same name is never touched. This prevents the old and new pods from coexisting after upgrade.
- **Manager kept consistent with the spawner**: `AgentBoxManager` independently computes the pod name for warm-reuse / liveness / stop, and it sits on the capability reap+adopt path. Its lookup is now profile-aware, and `stop(runId, profile)` / `getAsync(runId, profile)` receive the run's profile. Without this, the capability reap would 404 on the old `agentbox-` name and **leak the renamed compile pod**, and adopt would miss the live box after a runtime restart.

## Transition (one release cycle)

- `make push-kbc` **double-pushes** the same image digest under the legacy name `kbc-compile-box` so runtimes still pinned to it during the rename window keep resolving. Remove the legacy tag/push next release.
- **DevOps action required**: add a **registry replication rule** (ap-southeast → cn-shanghai) for `siclaw-kbc-box`. The old `kbc-compile-box` was **not** in the replication rule and had to be pushed to Shanghai by hand; the new name needs a rule so this stops being manual. Until the rule lands, push `siclaw-kbc-box` to cn-shanghai manually for production.

## Verification

- `helm template` smoke: `compileBoxEnabled=true` renders `SICLAW_COMPILE_BOX_IMAGE = {registry}/siclaw-kbc-box:{tag}`; explicit `compileBoxImage` override passes through unchanged; default stays dark (no env, fail-closed).
- Unit tests green for `box-profile`, `k8s-spawner`, and `manager` suites, including new coverage for: the `kbc-box` prefix on compile pods (chat stays `agentbox`), the legacy-pod reap on upgrade, the chat-pod guard (never reaped), and the profile-aware `stop` / `getAsync` lookups.

## Note / scope decision

Per the spec, only the two **compile** profiles get the `kbc-box` prefix. The read-only `kb-test` box (short-lived, discarded after a test session, still on the `siclaw-kbc-box` image) keeps the `agentbox-` prefix. If we want test-session boxes distinguishable too, that is a one-line follow-up (`podNamePrefix` on `kb-test`), but it was intentionally left out of scope here.


LikiosSedo:

LikiosSedo:

PR #443, opened 2026-07-22
feat(sicore-a2a-mcp): support multiple named A2A keys selected by alias

LikiosSedo:
## Problem

One `sicore-a2a-mcp` adapter process binds a single A2A key, i.e. a single Siclaw agent. Switching agents means editing config and restarting. The obvious shortcut — passing the key as a tool argument — is exactly the thing we must not do: A2A keys are credentials and must never flow through the model context.

## Approach

Put **multiple named keys in configuration**, and let the model select an agent by a short **alias** passed as an optional `agent` tool argument. The alias is the only agent selector that crosses the model boundary; keys stay in config/env only.

## Changes

- **Config** — new `SICLAW_A2A_KEYS` env, a JSON object `{"alias":"sk-..."}`. Aliases must match `^[a-z0-9][a-z0-9_-]{0,31}$`.
  - Backward compatible: `SICLAW_A2A_KEY` / `SICLAW_A2A_KEY_FILE` keep working and map to the reserved alias `default`. The two forms may be combined; a `default` alias inside `SICLAW_A2A_KEYS` is a collision error.
  - `SICLAW_AGENT_ID` still pins the `default` single key; combining it with `SICLAW_A2A_KEYS` is rejected (each named key resolves its own agent).
  - Every key is self-resolved once at startup via `GET /api/v1/a2a/self`; any failure aborts boot with the failing alias named (fail-fast, no silent partial key).
- **Tool surface** — all five tools gain an optional `agent` alias argument (never a key). Descriptions are populated at startup with the configured aliases and their resolved agent ids. One agent → `agent` optional; several → omitting it on a create/list call errors with the alias list instead of guessing.
- **Routing** — `task_id`/`context_id` are per-key server resources. The adapter records `task_id -> alias` in process memory, so `siclaw_wait_task` / `siclaw_get_task` / `siclaw_cancel_task` auto-route to the creating key. A mismatched `agent` argument is overridden by the recorded creator and a `routing_note` records the override. `siclaw_list_tasks` without `agent` aggregates across every key, tags each row with its alias, and rebuilds the map after a restart.
- **Security** — keys live only in configuration. They are never a tool parameter, never logged, and never present in an error message; errors and the stderr ready line carry aliases and resolved agent ids only.

Implementation routes the new logic through a small `AgentRouter`; `SicoreA2aClient` stays per-key and unchanged.

## Scope

The remote Sicore MCP endpoint (`/api/v1/mcp`) is **out of scope**: each remote client config carries its own `Authorization` header, so multi-agent there is just multiple client configs. This alias multiplexing is only for the local stdio adapter that injects credentials from env/files. The README documents this split.

## Tests

`npm test` — 56 passing (was ~30).

- Config: multi-key parse/order, single-key `default` mapping, merge, alias validation, collision, `SICLAW_AGENT_ID` + `SICLAW_A2A_KEYS` rejection, non-string/malformed key values not echoed.
- Router unit tests: selection rules, task→creator precedence + override note, untracked-task attribution.
- Tool tests: single-key back-compat behavior preserved, multi-key requires `agent`, unknown alias, named routing, auto-route follow-ups, override note, aggregated list, ownership recovery.
- stdio e2e: existing single-key + self-resolve tests, plus a new two-alias cold-start smoke asserting per-key `/self` resolution, description injection, alias routing, and no key leakage in descriptions or the stderr ready line.

Manual fail-fast smoke confirmed for missing key / bad alias / collision / `SICLAW_AGENT_ID`+`SICLAW_A2A_KEYS` — each exits 1 with a clear message and no key echoed.

**Not published**: no npm publish, no tag — release is left to the maintainer.

### DIFF
--- a/mcp/sicore-a2a-adapter/README.md
+++ b/mcp/sicore-a2a-adapter/README.md
@@ -1,6 +1,7 @@
 # Sicore A2A MCP Adapter
 
-A minimal local stdio MCP server that lets Codex or Claude Code use one existing Siclaw agent through Sicore's A2A API.
+A minimal local stdio MCP server that lets Codex or Claude Code drive one or more
+existing Siclaw agents through Sicore's A2A API.
 
 The adapter is intentionally thin:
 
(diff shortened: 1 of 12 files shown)


LikiosSedo:

LikiosSedo:

PR #444, opened 2026-07-22
Test-session resilience: turn-stall watchdog, idle-TTL reaper, lifecycle logs

LikiosSedo:
## Summary

Fixes three production-confirmed defects in the read-only **test-session** path (`kbc/platform/pod/compile_box.py`), all box-side. Each was verified against the code before fixing.

### 1. Test turns hung on "thinking" forever (no stall watchdog)
`test_session_driver` ran a bare `async for msg in client.receive_messages(): await _emit_message(...)` with **no model-stall watchdog** — unlike the compile path, which runs `_consume_turn_stream` under `_model_stall_watchdog` (armed in `_run_wrapper`). When the gateway black-holed a model request (connection accepted, no response), the turn stayed active and the UI spun indefinitely with no recovery.

**Fix:** `_test_stall_watchdog` + `_consume_test_turn_stream` reuse the compile idle-latency semantics (any SDK event == alive; a pending read tool relaxes to the longer tool bound so a slow `Read`/`Grep` is never false-killed). On a stall the turn is reaped **once, without retry**: interrupt the SDK stream, emit a turn-level `error` + `turn_done` so the UI goes idle, and keep the session live for the next question (the receive loop discards the torn-down `ResultMessage`). Idle bound defaults to the compile bound, overridable via `KBC_TEST_MODEL_IDLE_TIMEOUT_S`. Test sessions keep faithful streaming (no de-stream shim), so this watchdog is their only stall guard.

### 2. Orphaned sessions exhausted the concurrency cap (no idle TTL / list endpoint)
`TEST_SESSIONS` had no idle TTL and no enumeration. A server-side timeout or a persistence failure left box-side sessions nobody closed, filling `KBC_MAX_TEST_SESSIONS` (default 3) until the pod was recreated.

**Fix:** `TestRun` now tracks `created_at` / `last_activity_at` / a monotonic idle marker (touched on open, each turn, and event consumption). A periodic sweep (`_test_session_reaper`, `KBC_TEST_SESSION_IDLE_TTL_S` default 1800s) tears down idle sessions and emits a close event. New endpoint **`GET /test-sessions`** returns `{"sessions":[{"tid","parent_run_id","created_at","last_activity_at","done"}]}` — wire contract agreed with the consumer (field names are load-bearing).

### 3. Zero lifecycle observability
The box only printed its startup line; pod logs showed nothing for test-session activity.

**Fix:** `_print_test_lifecycle` emits one stdout line per lifecycle event (`test.open` / `test.close` / `turn.start` / `turn.done` / `turn.stalled` / `ttl.reap`), carrying `tid` + `parent_run_id` only — never user content.

## Constraints held
- `CompileRun` behavior untouched.
- No changes to destream or unrelated areas; changes are confined to `compile_box.py` (+ tests).
- Python style matches the file.

## Tests
`test_compile_box.py` (+3, full suite 69 passing):
- stall reaper frees a wedged turn, emits error + turn_done, and the same session answers a follow-up question; asserts stdout lines carry tid/parent and leak no question text;
- idle-TTL sweep drops orphans (snapshot dir + slot freed) and spares an active session;
- `GET /test-sessions` returns the agreed wire shape.

### DIFF
--- a/kbc/platform/pod/compile_box.py
+++ b/kbc/platform/pod/compile_box.py
@@ -239,6 +239,13 @@ _TEST_MODEL_IDLE_TIMEOUT_S = float(os.environ.get(
 # is recreated. A periodic sweep tears down sessions idle past this TTL.
 _TEST_SESSION_IDLE_TTL_S = float(os.environ.get("KBC_TEST_SESSION_IDLE_TTL_S", "1800"))
 _TEST_SESSION_REAP_POLL_S = float(os.environ.get("KBC_TEST_SESSION_REAP_POLL_S", "60"))
+# Structured error code for the 429 refusal when the pod's concurrent
+# test-session cap is full. Emitted in the box's own {error:{code,message,
+# retriable}} shape (same convention as handle_test_recommendation) so the
+# runtime's agentbox error mapping passes a STABLE code + retriable=false
+# through to the consumer — a bare 429 would otherwise map to the generic
+# TOO_MANY_REQUESTS with retriable=true. Mirrored in gateway contract.ts.
+TEST_SESSION_LIMIT_ERROR_CODE = "test_session_limit"
 
 
 def _now_iso() -> str:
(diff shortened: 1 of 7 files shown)


LikiosSedo:

LikiosSedo:

PR #445, opened 2026-07-23
fix(runtime): apply agent custom system prompt on the web chat path

LikiosSedo:
## Problem

A custom agent's `system_prompt` was silently dropped on the **sicore web chat** path: in the browser the agent introduced itself with the default SRE persona and ignored the custom prompt's rules. The **same agent worked over Feishu/DingTalk**.

## Root cause (confirmed)

1. sicore's chat proxy sends `chat.send` **without** a `systemPrompt` param (`proxy/handler.go`).
2. The runtime's `chat.send` handler only honored `params.systemPrompt` (`src/gateway/server.ts`) → absent → `undefined` → AgentBox falls back to its **built-in default SRE template**.
3. The resolver `resolveAgentSystemPrompt(agentId, frontendClient)` (`config.getAgent` → `agents.system_prompt`) already existed, but was wired only into the channel / coordinator / delegate paths (`dingtalk.ts`, `lark.ts`, `task-coordinator.ts`, `delegate-api.ts`) — **not** the web chat main path.

So: sicore doesn't send it → the runtime only read the param → the channel paths had already resolved it while the web path never did.

## Fix (runtime-side fallback, sicore unchanged)

In `chat.send`, when the caller does **not** forward `systemPrompt`, fall back to `resolveAgentSystemPrompt(agentId, frontendClient)`:

- An **explicitly forwarded** prompt (the portal-standalone path stamps it from the agent's model binding) still **wins as-is** — no behavior change there, and the fallback lookup is skipped entirely.
- **Best-effort**: `resolveAgentSystemPrompt` swallows RPC errors and returns `undefined` on failure / no custom prompt → built-in default. A lookup failure therefore **never turns into a chat failure** (the send still acks `ok`).

### Session-reuse semantics

AgentBox applies the template **only at session creation**, so this affects **new** sessions; a warm multi-turn session keeps the prompt it was created with — matching the existing channel-path behavior. No hot-update is introduced.

`chat.send` is the **only** prompt-creating entry point in `server.ts`; `chat.steer` targets an already-running session via `steerSession`, so no other handler needs the fallback.

## Tests

- New `src/gateway/server-chat-system-prompt.test.ts` (drives the real `chat.send` handler from `startRuntime`):
  - explicit `systemPrompt` param wins as-is and skips `config.getAgent`;
  - omitted param → resolves and applies the agent's custom prompt (the sicore web path);
  - no custom prompt (`system_prompt: null`) → `undefined` (built-in default);
  - `config.getAgent` throws → `undefined` default, and the send still returns `{ ok: true }`.
- `tsc --noEmit` clean; full `src/gateway` suite green (993 tests).

### DIFF
--- a/src/gateway/server-chat-system-prompt.test.ts
+++ b/src/gateway/server-chat-system-prompt.test.ts
@@ -1,0 +1,178 @@
+/**
+ * Regression test for the custom-system-prompt fallback in chat.send.
+ *
+ * Bug: sicore's web-chat proxy sends chat.send WITHOUT a systemPrompt param, so
+ * the runtime's handler (which only read params.systemPrompt) handed the box
+ * `undefined` and a custom-prompt agent silently got AgentBox's built-in default
+ * SRE persona. Channel paths (dingtalk/lark) already resolved the agent's own
+ * template via config.getAgent and worked — only the web path was unwired.
+ *
+ * The fix: when the caller does not forward systemPrompt, chat.send falls back to
+ * resolveAgentSystemPrompt(agentId, frontendClient). An explicitly forwarded
+ * prompt (portal-standalone) still wins as-is; a resolve failure / no custom
+ * prompt falls back to undefined = built-in default.
+ *
+ * This test drives the real chat.send handler from startRuntime (data-layer +
+ * agentbox mocked) and asserts what systemPromptTemplate the box prompt receives.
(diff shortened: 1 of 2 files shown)


LikiosSedo:

LikiosSedo:

PR #439, opened 2026-07-22
feat(kbc): coverage v2 — media assets auto-attach via citing docs

LikiosSedo:
## What

Media assets (images under `assets/`) are **document attachments**, not first-class sources. Coverage **v2** auto-accounts an image against any document that embeds it in its body, once that document is itself cited or excluded. `compiled_from` no longer needs a token per image and the exclusion ledger no longer needs a row per image.

Implements S1 (siclaw/kbc) of the raw reference contract v2. Design: `DESIGN-kb-asset-provenance-2026-07-22.md` §四 (normative).

## Motivation

- **cf bloat**: aggregated `compiled_from` drowns in image tokens (UHI `昇腾.md`: 139 entries = 35 docs + 104 images), unreadable to a human.
- **exclusion-ledger noise**: images as first-class sources force ~563 of UHI's 603 exclusion rows to be mechanical asset entries.
- **attribution gap**: the schema could not express *which document embeds this image*, so `mediaverify` and the collapse UI each guessed with heuristics.

## Math (v2)

```
auto        = { a ∈ media_assets | ∃d ∈ (cited ∪ excluded): edge(d→a) }
unaccounted = sources − cited − excluded − auto
```

- Edges are derived from the **raw tree** — each document's body ``, `[](…)`, and HTML `` links — resolved relative to the document + `posixpath.normpath` + URL-decoded. A link to a nonexistent / non-media path yields **no edge, no error**.
- An **orphan image** (embedded by no accounted document — upload residue) is *not* auto-attached; it stays `unaccounted` and must be excluded with a reason, so a human still sees it (anti fail-open).
- A **directly-cited** asset still counts as `cited` (v1 compatibility).

## Monotonicity — old green libraries stay green

v2 only *adds* an accounting path, so `unaccounted` can only shrink. **All pre-existing `selfcheck` tests pass unchanged** — none asserted "unreferenced asset ⇒ unaccounted" over an `assets/` path (the media-asset predicate requires an `assets/` segment, which no existing test used, so no assertion needed updating).

## Sheet-placeholder carve-out

`assets/sheets/*.md` table placeholders are **content files, not media** (`is_media_asset` requires an image extension). They stay first-class sources — auto-attaching them would launder "the table data was never compiled" into "covered".

## Observability

`coverage` gains `auto_attached` (count) and `auto_attached_sample` (≤20 `asset ← via` edges), so auto-accounting is never a silent fail-open.

## Two-repo fixture (byte-for-byte contract)

[...]

PR #446, opened 2026-07-23
feat(kb): platform hardening batch — coverage v2, box naming, MCP multi-key, test-session resilience, custom prompt fix

LikiosSedo:
## Platform hardening batch

Consolidates five individually CI-green PRs (this batch is their first code review) into one branch so the changes can be reviewed and merged as a single unit. Each PR was merged with `--no-ff`, preserving its full commit history (11 source commits under 5 merge commits). No squashing.

**supersedes #439 #442 #443 #444 #445 — please close those five once this batch merges.**

### Components

| PR | Branch | Summary |
|----|--------|---------|
| [#443](https://github.com/scitix/siclaw/pull/443) | `feat/mcp-multi-key-alias` | `mcp/sicore-a2a-adapter` supports multiple named A2A keys selected by alias (router + config + tools). |
| [#442](https://github.com/scitix/siclaw/pull/442) | `feat/kbcbox-naming` | Rename compile-box image to `siclaw-kbc-box` and pod prefix to `kbc-box`; upgrade reaps old-prefixed pods; profile-aware `stop`/`getAsync`. |
| [#439](https://github.com/scitix/siclaw/pull/439) | `feat/kb-asset-provenance` | selfcheck coverage v2 — media assets auto-attach via citing docs (+ attribution-edge re-verify, sicore ledger-mirror edge semantics); fixtures + box role discipline + mediaverify interop. |
| [#445](https://github.com/scitix/siclaw/pull/445) | `fix/chat-custom-system-prompt` | Apply an agent's custom system prompt on the web chat path (`gateway chat.send` falls back to the agent template when the caller omits `systemPrompt`). |
| [#444](https://github.com/scitix/siclaw/pull/444) | `fix/kb-test-session-resilience` | Test-session resilience: box turn-stall watchdog, idle-TTL reaper, run-scoped `capability.testSessions`; gateway idempotency key + `shouldRelayTestSession` predicate. |

### Conflict resolution
Zero textual conflicts. `src/gateway/server.ts` is touched by #442, #444, and #445 but in disjoint regions (imports at distinct lines; #442 ~L474-632, #445 ~L92/L319, #444 ~L28-47/L836-1055), so git's 3-way merge auto-resolved cleanly. All three sets of changes verified to coexist (imports + bodies) and to type-check together.

### Integration fix (one commit on top of the merges)
`fix(capability): thread run profile into box discovery/stop for compile pods` (eaf53f0b) — an interaction surfaced only by combining #442 and #444. #442 renamed the compile-box pod prefix (`agentbox-` → `kbc-box-`, profile-derived) and taught `onReap`/`onAdopt` to pass `rec.profile`, but the remaining capability call sites still resolved the pod name with the default `agentbox` prefix (correct on `main`, where every box was `agentbox-`; wrong once the rename lands). In K8s mode that made `capability.testClose` report a live test session as already-closed, `capability.testSessions` (new in #444, which reused testClose's discovery pattern) list nothing for a live run, and `capability.cancel` / session-setup cleanup 404 on `stop()` and leak the pod. Fix: pass the run's profile at every capability-box site — from the in-memory run record, or (for the restart-safe paths) recovered from the consumer store via `adopt(notifyOnAdopt:false)`, which never reattaches a relay or spawns a box. Chat-agent paths (`chat.sessionStatus`, `agent.clearMemory`, `agent.terminate`) are intentionally left on the default prefix.

### Regression (full, run on the merged + fixed tree)
- **kbc/platform/pod** — all test files via their CI `__main__` harness (`python test_X.py`): `test_selfcheck`, `test_compile_box`, `test_destream`, `test_mediaverify`, `test_redblue`, `test_incremental`, `test_batching`, `test_source_snapshot`, `test_office_ingest`, `test_mtls_auth` — all OK.
- **tsc build** (`npm run build`) — clean.
- **gateway vitest** (`npm test`) — 235 files, 4600 passed, 2 skipped, 0 failed. Includes the new suites `server-capability-test-sessions`, `server-chat-system-prompt`, #442's `k8s-spawner`/`manager`/`box-profile`, and the profile-threading assertions updated by the integration fix.
- **mcp/sicore-a2a-adapter** (`npm test`, includes tsc) — 56 passed.
- **helm template** with `agentbox.compileBoxEnabled=true` — `SICLAW_COMPILE_BOX_IMAGE` derives to `{registry}/siclaw-kbc-box:{tag}`; full chart renders.

### Note for the reviewer (pre-existing, not introduced here)
`test_selfcheck.py::test_charset_corruption_detection` asserts `第6行` but selfcheck now reports `第7行` (the U+FFFD sits on line 7 of the fixture). This fails identically on `main` and is untouched by any batch PR — the CI harness `main()` does not invoke this function, so it has been silently stale. Out of scope for this batch; flagged for a separate follow-up.

**Deploy note (reviewer F2):** the coverage-v2 edge semantics are pinned only by the shared `asset-provenance` fixture in both repos — the sicore side must land the same fixture before either ships, or the box view and server ledger silently diverge.


chent1996 (dismissed):
## Review

Approving, with one correction for the record and five non-blocking findings. This was reviewed as **five previously-unreviewed components plus one integration fix** — see the correction below.

### Correction: the sub-PRs were not independently reviewed

The description says "five independently-reviewed, individually-green PRs". I checked: **#439, #442, #443, #444, #445 are all CLOSED with zero reviews.** "Individually green" (CI) holds; "independently reviewed" does not — this batch is the first and only review these 2,920 lines have received, and I reviewed it accordingly (each component at its riskiest seam, two deep-dives on the kbc coverage change and the MCP multi-key refactor). Please drop that phrase from the description so the merge record stays honest.

### The integration fix (`eaf53f0b`) — verified, and it validates the batch approach

The #442×#444 interaction is real and the fix is right: every capability call site now threads the run's profile into `getAsync`/`stop` (pod names are profile-prefixed after the rename), memory-cold paths recover the profile via `adopt(notifyOnAdopt:false)` — correctly reusing the destructive-path adopt seam, which never reattaches a relay or spawns a box — and an unknown profile falls back to the default prefix (pre-rename behavior). Tests assert the profile threading on the memory-cold paths. Chat-agent paths staying on the default prefix is correct (they operate on chat agentIds). This interaction would have shipped broken had the five PRs merged separately; catching it is the batch's justification.

### Component verification summary

- **#442 (naming)**: manager and spawner derive the pod name from the same source (`getBoxProfile().podNamePrefix ?? "agentbox"`, invariant documented at both sites). The legacy-pod reaper is properly guarded — only `boxType=kb-compile*` pods are reaped under the old name, never a chat box sharing the agentId; 404 is the steady state. Renamed pods keep the `app=agentbox` label so orphan GC and `list()` still cover them. Helm derivation, Makefile, and the CLAUDE.md matrix row all updated; `isCompileCapable` stays fail-closed.
- **#444 (test-session resilience)**: the box-side watchdog reaps a wedged test turn *without* retry and latches `_stall_reaped` so the interrupted ResultMessage is discarded (no double turn_done); the idle-TTL reaper compares monotonic time (clock-skew safe) and any SDK message refreshes both clocks, so an active turn can't be TTL-reaped. `shouldRelayTestSession` pins the subtle single-consumer invariant (an idempotent replay must NOT start a second `/test-events` relay — that would split the frame stream). Contract additions are documented inline (`tid` pass-through warning, `client_request_id` absent ⇒ old behavior, structured `test_session_limit` 429 distinguishable from a rate-limit 429).
- **#445 (custom system prompt)**: fallback fires only when the caller omits `systemPrompt`; explicit param wins; resolve failure degrades to the built-in default (a lookup failure never fails the chat); session-creation-only semantics documented. Three-path test. Minor: this adds one `config.getAgent` RPC to every web prompt that omits the param — fine, just noting.
- **#439 (coverage v2)** — deep-dive: the edge rule is monotonic and fail-closed where it matters (an asset embedded by no *accounted* doc stays unaccounted; direct-cite wins; no double-count), observable (`auto_attached` + sample), and the fixture pins direct/inherited/shared/orphan/orphan-excluded/space/nested/query-string/HTML-img cases. The budget interaction with the #430 high-water limit is benign (if anything over-provisioned — the safe direction). Two real findings below (F1, F2).
- **#443 (MCP multi-key)** — deep-dive: all three v1 security properties survive with dedicated post-refactor tests (no-retry submission + snapshot-not-throw with task_id, compact working responses, key hygiene incl. no key material in errors/stderr); alias routing fails closed on unknown aliases; the recorded creator wins over a mismatched `agent` arg; tool schemas are non-breaking (`agent` optional everywhere; v1 single-key configs behave byte-identically). Two findings below (F3, F4).

### Findings (non-blocking)

- **F1 — `.tiff` is coverage-accounted but escapes numeric re-verification.** `MEDIA_ASSET_EXTS` includes `.tiff` (selfcheck.py:51-52) but the attribution-edge feed into mediaverify intersects with `IMAGE_SOURCE_EXTS` (selfcheck.py:1619 region), which does not — so a `.tiff` chart auto-attaches as accounted yet is never numerically re-read. The commit acknowledges the deferral, but this is an untested fail-open on exactly the fidelity surface coverage-v2 touches. Please add the one-line test pinning today's behavior, so the gap is visible rather than silent.
- **F2 — the sicore ledger mirror is a fixture-only coupling.** The edge semantics (case-insensitive `assets` segments, `?query`/`#frag` truncation order) are locked only by each repo running the shared `asset-provenance` fixture; there is no protocol version pin. A one-sided deploy silently diverges the box view from the server ledger. Worth one line in the deploy notes and making sure the sicore side lands the same fixture before either ships.
- **F3 — multi-key config has no file-based path.** `SICLAW_A2A_KEYS` is inline env JSON only: the 0600 key-file validation exists solely for the single-key `SICLAW_A2A_KEY_FILE`. Multiple keys therefore live in the process environment (`/proc/<pid>/environ`, shell history) — a hygiene regression relative to the adapter's own v1 premise. A `SICLAW_A2A_KEYS_FILE` (same 0600 check) would close it.
- **F4 — the cross-key authorization boundary is delegated to Sicore and untested.** For a task_id the adapter has not seen, `selectForTask` routes to whichever alias the caller names; whether key B can actually read key A's task is enforced entirely server-side, and no test (here or referenced) pins that rejection. The assumption is load-bearing — please verify it against the Sicore A2A server once, and note it in the README. Related: multi-key entries can't pin `agentId`, so multi-key mode hard-requires `GET /self` (older Sicore ⇒ boot abort by design — worth stating in the README compatibility notes).
- **F5 — pre-existing stale test.** The `test_charset_corruption_detection` note (asserts 第6行 vs actual 第7行, not invoked by the CI harness) is honestly flagged; please file the follow-up so it doesn't stay silently stale — a test that never runs is worse than none.

### Deploy notes

Image rename means the registry must publish `siclaw-kbc-box` before any runtime with this change rolls out (helm derives it from the release tag; explicit `compileBoxImage` overrides are unaffected). First spawn per compile agent reaps the old `agentbox-`-named pod — expect a one-time visible reap in logs during the upgrade window. And the standing kbc note: box behavior changes (coverage v2, test-session watchdog/reaper) need the box image rebuild + redeploy.

### Verdict

Approved — every component held up under what was effectively its first review, the integration fix is exactly the kind of cross-PR interaction batching exists to catch, and the findings are hardening items, not defects. Please fix the "independently-reviewed" wording before merge.

kbc/platform/pod/compile_box.py
```diff
@@ -4096,17 +4220,172 @@ async def guard(input_data, tool_use_id, context):
...
+        await asyncio.sleep(_MODEL_WATCHDOG_POLL_S)
+        # Resolve a pending post-interrupt terminator window (armed below after a
+        # successful interrupt). If the reaped turn's ResultMessage arrived, the
+        # consume loop advanced _results_seen and the generation stream is intact —
+        # clear the window. If the window elapsed with nothing, the terminator is
+        # lost and the generation guard can never realign: flag a rebuild so the
+        # next /test-message starts the SDK client fresh. Runs above the active-turn
+        # guard because a reaped turn is no longer active.
+        if run._pending_terminator_deadline is not None:
+            if run._results_seen > run._pending_terminator_seen:
```
jacoblee-io: **[Medium] Lost terminator + a new turn within the window leaves the session permanently generation-misaligned (silently drops all later output until the 30-min idle reaper).** This resolver treats *any* advance of `_results_seen` as "the reaped terminator arrived", but `_results_seen` is a bare counter that every turn advances (the stale-drop path at ~4247 and the live path at ~4255).

Trace: turn A stalls → `interrupt()` succeeds → the window arms `_pending_terminator_seen = _results_seen` with `_needs_rebuild` left False and the UI freed. A's terminator is black-holed (the exact failure this defends against), so `_results_seen` doesn't advance. Within the 10s window the user sends `/test-message`: because `_needs_rebuild` is False there's no rebuild, `_arm_test_turn` bumps `_turn_generation` but does **not** reset `_results_seen`, and B queries the same wedged client. Now `_results_seen` is permanently one behind `_turn_generation`, so **every** subsequent turn's frames fail `frame_gen < _turn_generation` and are dropped as stale — while the rebuild that should recover this is never armed (this `>` check short-circuits, or is bypassed because stale frames `continue` before reaching it). Session is stuck until the idle-TTL reaper.

Fix: gate the resolve on evidence the *specific* reaped terminator landed — compare against the reaped turn's `_turn_generation` (or track a dedicated terminator token) rather than a bare `_results_seen` advance.

kbc/platform/pod/selfcheck.py
```diff
@@ -589,6 +622,79 @@ def _markdown_link_targets(text: str) -> list[str]:
...
+    ``![](...)`` and ``[](...)`` plus HTML ``<img src=...>`` (single- OR
+    double-quoted). Angle brackets are unwrapped and an optional link title
+    stripped; then the ``#fragment``/``?query`` is truncated and the result is
+    URL-decoded once (``%20`` → space). Code fences/spans are masked so a path
+    inside example code is not mistaken for a real embed."""
+    targets: list[str] = []
+    prose = _markdown_prose(md_text)
+    for captured in _MD_LINK_RE.findall(prose):
+        destination = captured.strip()
+        if destination.startswith("<") and destination.endswith(">"):
```
jacoblee-io: **[Medium] An angle-bracketed link destination with a trailing `#fragment`/`?query` drops the edge → a good compile is wrongly FAILED.** This unwrap only fires when the destination *ends* in `>`. For `f`, `_MD_LINK_RE`'s angle alternation (`<[^>\n]+>`) only matches when `>` is immediately followed by `)`, so the trailing `#fig1` defeats it and the whole `<assets/a b.png>#fig1` is captured by the fallback branch; `startswith('<') and endswith('>')` is then false (string ends in the fragment), the brackets are never stripped, and the real edge `doc → assets/a b.png` is lost → asset reported `unaccounted` → `closed:false`. Same input without the fragment closes correctly. Same root cause in `_markdown_link_targets` (~608) → a page reachable only via an anchored angle link is misreported as an orphan. Realism is low (angle brackets are used for spaced filenames, a fragment on top is uncommon), but it's deterministic and it's exactly where the box view and the sicore ledger-mirror could silently diverge — worth pinning into the shared fixture. Fix: strip the `<…>` wrapper *after* truncating `#`/`?`, or let the regex tolerate a trailing fragment/query after the angle destination.

jacoblee-io (commented):
Reviewed all five components + the integration fix, one focused pass each (with the selfcheck graph and the MCP key-routing traced against the fixtures/code end to end). Strong batch overall — the integration fix `eaf53f0b` correctly closes the 404/leak gap #442 opened (every capability call site now threads `rec.profile`; `adopt(notifyOnAdopt:false)` is genuinely side-effect-free; the upgrade reaper targets old-prefix compile pods by boxType label without touching chat boxes), #445's fallback is correctly gated on strict `=== undefined`, #443's multi-key isolation holds (key+URL are co-bound per immutable client; no cross-key leak, no key in logs), and the rename glue (image `siclaw-kbc-box`, helm derive, Makefile double-push alias) is complete with only intentional legacy references.

Two MEDIUM correctness bugs worth addressing before merge — inline. Plus low/hardening notes:
- **#442**: `prefixForProfile` → `getBoxProfile(profile)` throws on a corrupt/forward-version stored profile; un-caught in `testClose`/`testSessions` it turns a graceful default-prefix fallback into an RPC 500 (pre-existing fragility inherited by the fix — consider falling back to `agentbox` on unknown). `agent.terminate` still calls `stop(box.agentId)` without the profile (not triggerable today, but asymmetric with the fix's own rationale). `kb-test` keeps the `agentbox-` prefix — confirm intended.
- **#444**: three `TEST_SESSIONS.values()` iterations lack a `list()` snapshot (safe today — no `await` inside — but the idle reaper already snapshots; add for symmetry before anyone adds an await). `capability.testClose` forwards `test_session_id` with no `parent_run_id` scoping (unguessable UUID mitigates; the listing path IS run-scoped).
- **#443**: duplicate alias inside `SICLAW_A2A_KEYS` JSON is silently last-wins; the multi-key form carries key material in an env var (no per-alias 0600 file) — worth a README caveat.
- **#445**: a pre-existing (NOT introduced here) seam — `chat-gateway.ts:325` builds `agent.system_prompt ?? null` untrimmed, so a stored `""`/whitespace prompt reaches the box verbatim instead of falling back. Optional: normalize with `.trim() || null`.

Also per the PR body: the sicore side must land the same `asset-provenance` fixture before either ships (edge semantics are pinned only there), and `test_selfcheck.py::test_charset_corruption_detection` is stale on `main` already (out of scope). Given the two mediums I'm leaving this as comments rather than an approval.

jacoblee-io (approved):
LGTM. Both MEDIUM findings from the prior review are correctly and thoroughly fixed, each with a targeted regression test and cross-repo alignment.

**MEDIUM 1 (test-session terminator/rebuild)** — two mechanisms now close the scenario:
- The window is generation-based: arm records the reaped turn's generation (`_pending_terminator_gen = _turn_generation`) and the resolver clears only on `_results_seen >= _pending_terminator_gen`, so a bare counter advance from an unrelated turn can no longer falsely resolve it.
- `handle_test_message` now rebuilds the SDK client whenever the terminator window is still open (`_needs_rebuild || _pending_terminator_deadline != null`), before arming — so a retry that arrives while the reaped terminator is still in doubt starts on a fresh, generation-reset client instead of the wedged one. That's exactly the lost-terminator + retry-within-window path I raised. Plus the rebuild runs before the liveness gate (avoids the 25s-timeout trap) and clears the window. Covered by `test_test_session_late_reaped_frames_never_terminate_new_turn` driving a real watchdog reap + injected stragglers.

**MEDIUM 2 (angle-bracket + fragment/query link)** — new shared `_unwrap_angle_destination` unwraps to the first `>` and re-appends the remainder, applied in both `_markdown_link_targets` and `document_link_targets`, ordered after the title strip and before `#`/`?` truncation; mirrored byte-for-byte by the sicore ledger's `unwrapAngleDestination` and pinned by a new `guide/assets/c d.png` fixture referenced only via `<assets/c d.png>#fig1` in both repos — which also closes the box↔ledger divergence risk I flagged.

Also verified `85e8ce34` run-scopes the idempotency key to `(parent_run_id, client_request_id)` with an ownership re-check before replay. The remaining low/hardening notes (kb-test prefix intent, `agent.terminate` profile symmetry, multi-key env README caveat) are optional and non-blocking. Deploy prerequisite still stands per the PR body: the sicore side must land the same `asset-provenance` fixture (done in lockstep here). Ship it.

PR #447, opened 2026-07-23
fix(lark): link feedback cards to persisted messages

glow-Kev1n:
## Summary

- return the persisted final assistant message ID from the Lark response collector
- embed that ID in new Feishu feedback card actions and forward it through `chat.recordFeedback`
- keep legacy cards without `message_id` compatible
- omit feedback buttons when the final assistant message could not be persisted, while still delivering the answer

## Root cause

Feishu feedback cards previously carried only the session and CardKit card ID. That was sufficient to identify the card, but not to unambiguously correlate an older card with its persisted assistant turn.

`message_ref` remains the card aggregation and re-vote idempotency key. The new `message_id` is used only for the exact chat-message link.

The field is optional on callbacks and RPC requests so cards created by older Runtime versions remain compatible.

## Validation

- `npx vitest run src/gateway/channels/lark-card.test.ts src/gateway/channels/lark.test.ts` — 144 tests passed
- `npm run build` — passed

Both checks ran on the development host with Node.js 22.22.3.

### DIFF
--- a/src/gateway/channels/lark.ts
+++ b/src/gateway/channels/lark.ts
@@ -1708,6 +1708,17 @@ export async function collectChannelResponse(
   // Prefer the last full assistant turn; fall back to streamed deltas if the
   // brain only emits content_block_delta events.
   const text = lastAssistantText || parts.join("");
+  // A delta-only stream has no assistant message_end, so persist the exact
+  // synthesized reply once the stream finishes. This gives feedback cards the
+  // same precise message linkage as the normal message_end path without
+  // duplicating assistant rows when a full turn was already persisted.
+  if (!lastAssistantText && text.trim() && persist) {
+    lastAssistantMessageId = await persistRow({
+      sessionId,
+      role: "assistant",
+      content: redact(text),
+    });
+  }
   return { text, images, assistantMessageId: lastAssistantMessageId };
 }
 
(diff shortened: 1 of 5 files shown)


LikiosSedo (commented):
One P2 behavior regression found in the Runtime side; the exact-message linkage and the paired Sicore handling otherwise look sound.

src/gateway/channels/lark.ts
```diff
@@ -1312,7 +1315,9 @@ async function processQueuedLarkMessage(ctx: QueuedLarkMessageContext): Promise<
     // non-answer and skew the feedback signal Metrics aggregates.
     const isAnswer = !agentError && resultText.trim().length > 0;
     const ok = await finalizeCard(larkClient, cardSession, finalCardBody,
-      isAnswer ? { ctx: { sessionId, channelId }, locale } : undefined);
+      isAnswer && assistantMessageId
```
LikiosSedo, line 1318: [P2] Preserve feedback for delta-only replies

`collectChannelResponse` explicitly supports successful streams that end without an assistant `message_end` by falling back to `parts.join("")` (and `lark.test.ts` already covers that shape). In that path `resultText` is non-empty but `assistantMessageId` stays null, so this new guard removes the feedback row even though the previous code added one. That makes feedback disappear for a supported channel response shape. Please either persist the synthesized final text to obtain its exact message ID, or intentionally fall back to a legacy card without `message_id`, and add an end-to-end assertion for the delta-only card flow.

LikiosSedo (approved):
Re-reviewed at 7975b2cc. The delta-only response path now persists the synthesized assistant reply, links the feedback card to the exact message ID, and has focused collector plus end-to-end Lark coverage. CI is green; no blocking issues remain.

PR #448, opened 2026-07-24
kbc box: rebuild and continue on compile-turn stall or SDK subprocess death

LikiosSedo:
## Problem

Forensics on a stalled 287-batch compile (batch 4 wedged ~40 minutes) traced the hang to a gateway failure mode the box could not recover from: the model gateway (massapi) intermittently drops the stream terminator after the content has fully arrived. The box's destream watchdog has a 900s idle floor, and KBC_MODEL_MAX_RETRIES same-connection retries compound it into a multi-ten-minute wedge per batch. Separately, if the Claude Agent SDK subprocess dies, the box had no way to notice: the turn just waits forever.

## Fixes (all in kbc/platform/pod/compile_box.py)

- **Subprocess liveness probe** (`_client_process_exited`): a dead SDK subprocess is detected immediately and classified as an unrecoverable turn instead of idling against the watchdog.
- **Rebuild-and-continue loop** (`_drive_batch_session`, `KBC_BATCH_REBUILD_MAX_RETRIES`, default 3): when a compile turn stalls or the client dies, the box tears the client down, rebuilds a fresh connection/session, and continues the same batch — instead of retrying on the same wedged connection.
- **Stream-end-without-result now raises `ModelStallError`**: a stream that terminates without a result is a stall, not something to silently wait out.
- **Compile lifecycle stdout**: batch begin/stall/rebuild/finish breadcrumbs so operators can see recovery happening from pod logs.

## Relationship to the sicore-side auto-resume train

This is layer L1 of the compile self-healing stack: keep the box alive through gateway jitter so the user never notices. The sicore-side auto-resume (L3) covers the complementary case where the box run actually dies. L3 cannot see a live-but-wedged box, so both layers are required.

## Verification

- Full kbc box test suite green (82 checks), including new fault-injection tests for subprocess death mid-turn, stream-terminator loss, and rebuild-exhaustion classification.
- Live: image built from this branch (`siclaw-kbc-box:stall-28a304d`) has been running on siflow-test as the compile box for today's acceptance (multi-hundred-item KB imports and compile runs).


kbc/platform/pod/batching.py
```diff
@@ -24,21 +24,50 @@
 import hashlib
 import json
 import os
+import posixpath
```
jacoblee-io: Nit (non-blocking): `import posixpath` (here) and `import urllib.parse` (line 32) appear to be unused — no `posixpath.` / `urllib.` call sites anywhere in the module. Looks like leftover dead imports from the reverted "stop deciding for the model what a source is" approach. Safe to drop.

jacoblee-io (dismissed):
LGTM. The core self-healing (`_drive_batch_session` rebuild-and-continue loop, subprocess liveness probe, `ModelStallError` routing) is correct and leak-free: the retry counter is bounded, the wedged client is torn down and `source_view` cleaned up on every rebuild, and the `finally` block restores the persistent client + normalizes the ledger on all paths. `BatchOutputError` from codex source-view materialization deliberately escapes the retry catch, which is right — a deterministic path/config error won't be fixed by a rebuild. Sub-agent denial, tool allowlist, and stdout whitelisting all check out.

I also verified the wider changes the description doesn't mention (this PR is ~5.4k lines across batching.py / selfcheck.py / office_ingest.py / prompts, not just compile_box.py): `office_render_pairs` is correctly wired into coverage (batching.py:615, selfcheck.py:1583); the family/attachment grouping invariants and the containment dedup hold; the `<...>` angle-destination unwrap and chart-XML tolerance are fine.

One minor cleanup below. Nice work — approving.

Nit: consider expanding the PR description to list the batching / selfcheck / office_ingest changes too, so future reviewers and rollbacks see the real blast radius.

jacoblee-io (approved):
Re-reviewed at 46ea747a. The only nit from my earlier pass — the unused `posixpath` / `urllib.parse` imports in batching.py — is resolved by 46ea747a. The other new commit (16990b53) is prompt-only: it adds the playbook's "Overlapping sources: compile the delta" section and splits box_role so that only cross-batch *contradictions* defer to the final pass while *duplication* is handled in-batch. Verified: en/zh are symmetric, no lingering "duplicates → final pass" text remains, and the guidance stays consistent with the coverage-ledger invariant (overlapping-source pages still write `compiled_from`, so they remain accounted — it is explicitly "not an exclusion"). Core self-healing / batching / selfcheck / office_ingest code is unchanged from my prior full review. Approving on the latest head.

PR #450, opened 2026-07-28
kbc: a spend cap is not a rate limit, and the owner should be told which

LikiosSedo:
Production stopped mid-compile and reported *"batch compile interrupted; finished batches are stored, trigger a compile again to resume."* The provider had in fact answered:

```
402 billing_denied — API key daily cost limit exceeded (504.29/500.00).
```

The cause was never named. There was no `402` anywhere in `compile_box.py`: it sits one status away from `429/503/529`, which have had a backoff loop for months, and wants different handling — a rate limit clears on its own in seconds, a spend cap is cleared by a person raising the limit, which no amount of backoff reaches.

## What changed

- Caught **before** the rate-limit branch and raised as `ModelQuotaExhausted`, so the retry budget is not spent on a billing refusal.
- Gated on `fail_on_error_result`, so this is scoped to the batch orchestrator, which cannot proceed without model calls. A live session ends the turn and stays up exactly as before.
- Coded `quota_exhausted` rather than `other` in the operator-visible lifecycle log.
- The `turn_done` note states the refusal and where the work stopped: *the provider refused on billing (HTTP 402), the key's spend limit for the period is used up, N of M batches are stored, the next compile starts at batch K.*

## What it deliberately does not do

An earlier revision of this branch told the owner the limit would clear in ~23 hours and that *"triggering before then will be refused the same way"*. That replaced one false statement with its mirror image. A spend cap is lifted by a person — raise the limit, top up, swap the key — so:

- **No `deterministic` flag.** That is what a consumer would read to stop retrying, and stopping is the account holder's call.
- **No `retry_after_s`, and no parser for it.** It is the provider's estimate for the case where nobody intervenes. A time the box cannot stand behind is worse than no time.
- **The box does not end a live session.** Deciding the owner should stop trying would remove their ability to find out otherwise.

The resume point is positional (`_batch_resume_point`), never the batch id — ids are model-written and only conditionally safe to echo.

## Tests

`test_a_spend_cap_is_not_a_rate_limit` pins the routable code, that `402` is not in the rate-limit set, the resume-point arithmetic including the empty and fully-stamped cases, and the **absence** of both `deterministic` and `retry_after_s`.

`test_a_spend_cap_does_not_kill_a_live_session` drives the live path (default `fail_on_error_result=False`) and asserts nothing propagates, no backoff is spent, and the query is never re-issued. Negative control: removing the gate fails it with the `ModelQuotaExhausted` a live session should never see.

`test_compile_box` (105 checks) and `test_selfcheck` green. Note that `test_batching.py` fails to start on `main` as well — its runner list references `test_hierarchical_pack_splits_oversized_family_with_anchor_context`, which is not defined anywhere in the repo. Pre-existing, untouched here.

### DIFF
--- a/kbc/platform/pod/compile_box.py
+++ b/kbc/platform/pod/compile_box.py
@@ -461,6 +461,15 @@ def _arm_test_turn(run: "TestRun") -> None:
 # scope), so this is graceful handling, not a fix. Exhaustion ends the turn with a
 # clear owner-facing note instead of a crash.
 _MODEL_RATE_STATUSES = frozenset({429, 503, 529})
+
+# 402 is 429's near relative and its opposite. A rate limit clears on its own in
+# seconds, so backing off is right; a spend cap does not clear until the billing
+# window rolls, so every retry is a request that cannot succeed. Production hit
+# `daily cost limit exceeded (504.29/500.00) … retry_after 84596` — twenty-three
+# hours — and the owner was told only "batch compile interrupted; trigger again
+# to resume", which invites exactly the retry that cannot work. Neither the code
+# nor the message knew what had happened: there was no 402 anywhere in this file.
+_MODEL_QUOTA_STATUS = 402
 _MODEL_RATE_MAX_RETRIES = int(os.environ.get("KBC_MODEL_RATE_MAX_RETRIES", "5"))
 _MODEL_RATE_BACKOFF_BASE_S = float(os.environ.get("KBC_MODEL_RATE_BACKOFF_BASE_S", "2"))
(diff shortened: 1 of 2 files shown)


LikiosSedo:
Folded into #453 unchanged (same commits, cherry-picked). All three PRs edited `compile_box.py` and `test_compile_box.py` at the same points and conflicted pairwise in the test runner, so keeping them separate meant resolving the same conflict twice; they ship in one box image either way. Full box suite green after the merge — 107 checks. Closing here; review moves to #453.

PR #451, opened 2026-07-28
kbc: the runtime owns the dispatch round, not the model

LikiosSedo:
`resolve_ticket` stamped its receipt with whatever the model typed into `dispatch_nonce`, defaulting to `""` when it typed nothing — while the runtime had **already validated that exact nonce as REQUIRED** when it accepted the `compile.apply_rulings` command, and then rendered it into prose for the model to hand back.

```python
# accepting the dispatch — validated, bounded, required
nonce = _bounded_string(parameters.get("dispatch_nonce"), ..., required=True, limit=128)
# rendering it for the model to copy
nonce=params["dispatch_nonce"]
# registering the receipt — read back from the model, empty default
"dispatch_nonce": str(args.get("dispatch_nonce", "")),
```

A control-plane fact does not become more true for passing through the model's attention, and one the model can silently omit is one the consumer cannot rely on.

## What it cost

Production hit both halves of this on one library. Five contradiction tickets sat at 已答·待重试 because the echo carried a superseded round; they closed only when a later retry omitted the nonce entirely, which fell through to the consumer's weaker timestamp rule and let an unmatched receipt close a ticket.

Neither failure was visible to the box: `resolve_ticket` has no reject path, so it reported success both times. The owner had to relay the mismatch by hand.

## What changed

- The accepted `apply_rulings` command stamps `run.apply_dispatch_nonce`; `resolve_ticket` reads it and ignores the argument. Empty means the box resolved outside any apply round — exactly what a consumer should not treat as answering a dispatch.
- The argument stays in the schema, no longer `required`, so a directive from an older runtime still calls cleanly instead of failing schema validation mid-repair.
- The directive stops asking for the echo. It was asking the model to carry bookkeeping it cannot own, and the instruction is now false.

## Tests

`test_the_runtime_owns_the_dispatch_round_not_the_model` drives the real `/command` endpoint and asserts both production failure modes are now impossible: a superseded echo and an omitted one both stamp the accepted round, a second dispatch re-stamps, and a resolve before any dispatch stamps empty. It also pins that the directive no longer mentions a nonce.

Negative controls: restoring the argument read fails on the invented nonce; restoring the prose fails the directive assertion. Full suite green (104 checks), plus `test_selfcheck`.

## Scope

This is the first of three changes against a state-machine deadlock found in production (server and client disagreed on "ticket closed", the looser definition removed the surface the tickets live on, and the recovery action was gated on the belief that recovery was unnecessary). This one removes the failure mode where the round is lost in transit. The remaining two are sicore-side: one server-owned ticket state that the client renders rather than re-derives, and returning an unanswered dispatch to the queue.

### DIFF
--- a/kbc/platform/pod/compile_box.py
+++ b/kbc/platform/pod/compile_box.py
@@ -571,6 +571,18 @@ class CompileRun:
         self.allowed_tools: list[str] | None = None
         # Consumer-declared prompt/output locale (capability.fetchInput → /session).
         self.locale: str | None = None
+        # The dispatch round resolve_ticket receipts belong to, stamped by
+        # _prepare_command when an apply_rulings command is accepted.
+        #
+        # The runtime already validates this nonce as REQUIRED on the way in and
+        # then used to hand it to the model in prose to type back, defaulting to
+        # "" when it didn't. A fact the runtime holds does not become more true
+        # for having passed through the model's attention, and a bookkeeping
+        # field the model can silently omit is one the consumer cannot rely on:
+        # five tickets in production stuck at "已答·待重试" because the echo
+        # carried a superseded round, and closed only when a later retry omitted
+        # the nonce entirely and fell through to the weaker timestamp rule.
+        self.apply_dispatch_nonce: str = ""
         # Set once connect() has returned (success OR failure). A /message that
         # races ahead of the async connect waits on this so client.query() never
         # hits the SDK's "Not connected. Call connect() first." error.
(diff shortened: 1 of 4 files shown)


LikiosSedo:
Folded into #453 unchanged (same commits, cherry-picked). All three PRs edited `compile_box.py` and `test_compile_box.py` at the same points and conflicted pairwise in the test runner, so keeping them separate meant resolving the same conflict twice; they ship in one box image either way. Full box suite green after the merge — 107 checks. Closing here; review moves to #453.

PR #452, opened 2026-07-28
feat: add knowledge runtime bindings

LikiosSedo:
## Summary
- add a versioned KnowledgeRuntime contract for materialized, retrieval, and hybrid bindings
- expose one stable knowledge_retrieve tool to pi-agent and proxy calls through the mTLS Gateway
- make knowledge reloads invalidate sessions when the tool schema changes

## Verification
- npm test (236 files, 4611 passed, 2 skipped)
- npm run build
- npx vitest run src/knowledge/runtime.test.ts src/agentbox/sync-handlers.test.ts src/gateway/internal-api.test.ts

### DIFF
--- a/src/agentbox/gateway-client.ts
+++ b/src/agentbox/gateway-client.ts
@@ -12,6 +12,10 @@ import path from "node:path";
 import type { DelegationPersistenceEvent, DelegationPersistenceResponse } from "../shared/delegation-persistence.js";
 import type { MetricsFlushPayload } from "../shared/metrics-types.js";
 import type { DelegateRequest, DelegateResponse, DelegatesResponse } from "../shared/agent-delegate.js";
+import type {
+  KnowledgeEvidenceSet,
+  KnowledgeRetrieveRequest,
+} from "../knowledge/contracts.js";
 
 export interface GatewayClientOptions {
   gatewayUrl: string;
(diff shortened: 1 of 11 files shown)


LikiosSedo:
Closing in favour of #453, which covers the same disclosure surface with a producer behind every field.

**Why this one and not that one.** Both PRs add the same line — a library's name plus a one-sentence subtitle — to the same place. The difference is what feeds it:

| | #452 | #453 |
|---|---|---|
| subtitle source | `binding.description` | `consumerDomain` |
| who writes it | nothing — sicore never sends `bindings` | the compile box's `report_domain` → `authoring/META.json` → repo row → version snapshot → bundle |

The retrieval half has no provider at any of the three hops: sicore implements no `knowledge.retrieve` RPC (grep of the whole repo), nothing anywhere produces a `retrieve` capability, so `buildKnowledgeRuntime` returns `tools: []` permanently.

**One defect worth recording before this closes**, in case the contract comes back later — `src/knowledge/runtime.ts`:

```ts
tools: retrievable && input.retrieve ? [createKnowledgeRetrieveTool(...)] : [],  // gated
promptContext: buildPromptContext(bindings),                                     // not gated
```

The tool is conditional; the paragraph advertising it ("large or incident-oriented collections are available through \`knowledge_retrieve\`") is unconditional. Merged as-is, every agent with a bound library would be told about a tool it does not have — and told so directly beneath the existing `# Knowledge Wiki` block, which says in the same prompt that there is no search tool. Two catalogues of the same libraries, contradicting each other.

If a retrieval provider is ever built, the seam is worth revisiting; the piece to keep is the capability vocabulary, and it should render one catalogue rather than a second one, sourcing the subtitle from `consumerDomain`. The branch stays for that.

Nothing is being lost here that has a consumer today.

PR #454, opened 2026-07-29
feat(channels): self-service API key issuing via Feishu /apikey

jacoblee-io:
## Summary

Anyone who can DM a personal bot can now obtain that agent's API key by sending `/apikey` in the private chat, instead of needing a control-plane account + org membership + write scope just to call a2a/mcp. Admission collapses to **one** rule (can you DM the bot?) rather than two that must be kept in sync.

Onboarding N non-SRE users to MCP previously meant either one shared key (audit collapses to a single identity, revocation is all-or-nothing) or N people granted write access far beyond "I just want to call it".

### Division of labour

The runtime only plumbs messages: recognise the command deterministically, forward the sender identity, render the reply. Every credential decision — admission, hashing, one-key-per-requester, sliding expiry, per-call re-authorization, the single-use pickup page — stays in the frontend. The runtime stores no key material and never sees plaintext.

| Command | Effect |
|---|---|
| `/apikey` | issue or **rotate** (the requester's previous key dies) |
| `/apikey status` | read-only status |
| anything else (`/apikeys`, `/apikey foo`) | usage help, **zero RPCs** |

Contract and rationale: `docs/design/2026-07-28-feishu-apikey-command.md`.

### Design points worth reviewing

- **Personal chat only.** The group path drops `/apikey` silently *before* the @-mention gate — a group reply is visible to everyone present and this flow returns a credential pickup link. Both dispatch sites share `parseApiKeyCommand` so they can never claim different sets.
- **Never a model-callable tool.** Parsing is a plain regex that returns early; an LLM inside the issuing loop would make one prompt injection an issuing primitive.
- **The regex has no trailing `\b` on purpose.** With it, `/apikeys` escapes the namespace and reaches the model — in a group that is exactly the credential-adjacent routing this feature exists to prevent.
- **Only a bare `/apikey` rotates.** Any other subcommand gets usage help and issues no RPC, so a typo (`/apikey statu`) cannot destroy the key the user was trying to inspect. This is also why `/apikey status` is load-bearing rather than a nicety.
- **Issuing is single-flight per `(channel, sender)`.** This path bypasses the per-binding queue that serialises ordinary messages, so without the guard a double-tap (or an at-least-once redelivery) mints two keys and the user collects the first link — a key the second rotation already invalidated.
- **Timestamps render defensively.** `Intl.DateTimeFormat.format` throws `RangeError` past ±8.64e15 (e.g. a frontend emitting nanoseconds) and `Number.isFinite` passes such a value; that throw would surface as "service unavailable" *after* the key was already rotated, leaving the user unable to ever reach a pickup link. `<= 0` and out-of-range values simply omit the line. Likewise `exists` is optional on the wire, so "you have no key" is only claimed when the frontend says so or no `keyPrefix` came back — the advice attached to that branch is the destructive command.
- **Plaintext never enters the chat log.** Replies carry only the short-lived, single-use `pickupUrl`; chat history is retained, searchable and exportable. `rotated: true` always warns that the old key just died, since any MCP client configured with it breaks instantly.
- **Failure is spoken, not swallowed.** A thrown RPC or a missing frontend client still replies. This deliberately differs from the neighbouring PAIR path, whose failures escape to the top-level catch — here the user is waiting on a link, and silence reads as a broken bot.
- **No hard dependency on the control plane.** `channel.issueApiKey` / `channel.apiKeyStatus` are Upstream-mode capabilities, but the runtime calls them unconditionally, so the Portal adapter registers graceful `success:false` stubs (mirroring the existing `channel.pairPersonal` stub). Without them a Standalone deployment hits method-not-found, the rejection is swallowed, and the user gets no reply at all.

DingTalk has no personal-bot concept, so it is out of scope.

## Test Plan

- [x] `npm test` — 235 files / 4621 passed (main tree), including 13 new `/apikey` cases: pickup-link delivery, rotation warning, verbatim failure reason, open-mode issuing (not gated like PAIR), read-only status, first-time-user hint, usage help with zero RPCs on a typo, reply-on-throw, reply with no frontend client, out-of-range/zero timestamps, `exists`-absent not triggering destructive advice, single-flight rejection, and group-chat silence across `/apikey`, `/apikey status`, `/apikeys`
- [x] Deployed to an internal test environment and verified healthy (runtime connected, new code present in the image)
- [x] Full end-to-end acceptance run against a configured Feishu personal bot: issue → pickup → single-use link → rotation invalidates the old key → status → access-mode switch

## Notes for reviewers

Timestamps render via `Intl.DateTimeFormat("sv-SE", …)` with an explicit time zone: the `sv-SE` locale is chosen only for its ISO-like `YYYY-MM-DD HH:mm` output, so copy does not drift with the host locale and stays assertable in tests.

Known gap, tracked in the design doc: the upstream frontend's refusal strings are currently English, so a zh-CN user sees a Chinese frame around an English reason. Localizing belongs on the frontend, which owns the admission logic and therefore its wording — mapping those strings in the runtime would couple it to the frontend's exact copy and rot on the first rewording.


LikiosSedo:
Two blocking reliability issues remain in the destructive `/apikey` flow:

1. **The single-flight guard is not idempotency.** The guard in [`lark.ts`](https://github.com/scitix/siclaw/blob/5ca08052310648c25f577a2d69d2b480ace81330/src/gateway/channels/lark.ts#L911-L925) only rejects overlapping calls. Once the first request settles, a replay of the same Feishu `message_id` issues again; the new test itself confirms that a later request increases the issue call count to two ([test](https://github.com/scitix/siclaw/blob/5ca08052310648c25f577a2d69d2b480ace81330/src/gateway/channels/lark.test.ts#L2982-L3002)). Because the upstream issuer rotates by deleting the previous key, sequential redelivery or handling by another Runtime instance makes the first pickup link return a dead key. Please pass a stable message/request id through `channel.issueApiKey` and deduplicate it durably upstream, with a sequential replay test using the same `message_id`.

2. **A failed Feishu reply can revoke the old key without delivering the new one.** Rotation has already committed at [`issuePersonalApiKey`](https://github.com/scitix/siclaw/blob/5ca08052310648c25f577a2d69d2b480ace81330/src/gateway/channels/lark.ts#L925-L932), but [`replyToLark`](https://github.com/scitix/siclaw/blob/5ca08052310648c25f577a2d69d2b480ace81330/src/gateway/channels/lark.ts#L1687-L1701) catches both thrown send failures and non-zero Feishu API responses and then resolves normally. The surrounding catch therefore never runs: the user receives no pickup link while the previous key is already invalid. The issuance contract needs a recoverable commit point, for example an idempotent request that replays the same pending pickup result, or deferring invalidation until pickup is consumed. Please cover both thrown and non-zero reply failures after a successful rotated result.

The deterministic parsing, group-chat drop, plaintext handling, status path, and timestamp guards otherwise look sound. TypeScript, focused tests, and the full test suite are green, but these destructive retry/delivery windows should be fixed before approval.

docs/design/2026-07-28-feishu-apikey-command.md
```diff
@@ -0,0 +1,139 @@
+# Feishu `/apikey` — self-service API key issuing
+
+**Status**: Implemented, 2026-07-28
+**Related**: `docs/design/channels.md`, `src/gateway/channels/lark.ts`,
```
chent1996: Non-blocking: `docs/design/channels.md` does not exist — the closest match in the tree is `docs/features/channels.mdx`. Worth pointing the Related line at the real file so the reference doesn't rot from day one.

src/gateway/channels/lark.ts
```diff
@@ -1447,6 +1529,161 @@ function formatPersonalPairReply(
...
+      : `❌ 领取 API Key 失败: ${reason}`;
+  }
+  const linkExpiry = formatApiKeyTimestamp(result.expiresAt, locale, true);
+  const lines =
+    locale === "en-US"
+      ? [
+          result.rotated
+            ? "✅ New API key generated — your PREVIOUS key is now invalid. Update anything configured with it."
+            : "✅ Your API key is ready.",
+          "Open within 5 minutes; the link works only ONCE:",
```
chent1996: Non-blocking: "Open within 5 minutes" (and "expires after 30 days without use" below, plus the zh-CN twins) hard-code the frontend's TTL policy into runtime copy, while the design doc assigns expiry policy to the frontend. The link-expiry line is already derived from `expiresAt`, so if the frontend ever changes either TTL these static sentences silently lie. Either derive the wording from the wire data, or note the copy↔policy coupling in the design doc so a frontend TTL change knows to touch this file.

chent1996 (dismissed):
Reviewed with the branch checked out locally. Every claim in the description held up under verification:

- **Group silence including `@bot /apikey`**: the drop-gate (lark.ts:1001) sits before the @-gate (1034), and `extractInbound` strips `@_user_N` placeholders at entry (615–663), so the gate sees post-strip text — the claim is structurally true, not just tested.
- **Only a bare `/apikey` rotates**: `subcommand === ""` requires exactly `/apikey` after trim; typos get usage help with zero RPCs. The no-`\b` regex plus the shared `parseApiKeyCommand` at both dispatch sites means `/apikeys` can never reach the model.
- **Single-flight**: `has`/`add` with no await in between (no race on a single-threaded runtime), freed in `finally`, and bounded by the 30s built-in RPC timeout. I also checked the event-ingestion layer (lark.ts:222–237): the handler ACKs immediately and detaches the work, so Feishu redelivery is already prevented at the ACK layer — the guard is genuine defense-in-depth for the double-tap case.
- **Timestamp defense**: the `new Date(epochMs).getTime()` NaN check covers exactly the range where `Intl.format` throws (±8.64e15), verified by the nanosecond-value test.
- **No plaintext anywhere**: replies carry only `pickupUrl`; the audit log lines record only ok/rotated flags.
- **Portal stubs**: both registered, handler-count test updated 56→58, so Standalone degrades gracefully instead of method-not-found.

Verification: the three touched test files pass locally (312 tests); the full suite on this branch has +17 tests vs main, all passing, with zero test failures attributable to the PR (the 18 file-load failures in my environment are missing optional otel deps and are identical on main); `tsc --noEmit` introduces no new errors. The CLAUDE.md impact-matrix row is in place per the development protocol.

Two non-blocking nits inline. Nice work — the tests here assert behavior (RPC calls, reply content) rather than string composition, and the design doc records contracts rather than implementation steps.

jacoblee-io:
Thanks — both reliability findings are real, and #2 is one my own review pass missed. Addressed in 541f7e60. Two of these have a frontend half I can't land here, so I've implemented the runtime half and recorded the boundary in the design doc rather than implying the window is closed.

**2 — a failed reply can revoke the old key without delivering the new one.** Confirmed exactly as described: `replyToLark` returns normally after both a thrown send failure and a non-zero Feishu code, so the surrounding catch never ran and the caller could not distinguish delivery failure from success.

`replyToLark` now returns whether the message landed. On the issuing path, delivery is checked rather than inferred: if a **committed** rotation's reply fails, the send is retried once (the transient case) and, if still lost, an audit line names the affected sender — never the link. A failed issue deliberately does **not** escalate: with no committed side effect a lost reply is an ordinary retriable no-op. Both failure shapes you asked for are covered by tests (`thrown send failure` and `non-zero Feishu code`), plus a negative test asserting no escalation without a rotation.

What keeps this recoverable rather than a lost credential is that the command is retryable — another `/apikey` rotates again and returns a fresh link. Removing the window entirely (defer invalidation until the pickup is consumed, or replay a pending pickup for an idempotent request) is a frontend-side change; it's now written down as such in the design doc.

**1 — single-flight is not idempotency.** Agreed: the in-process guard only collapses overlapping requests and cannot span a sequential redelivery or a second gateway replica. The runtime now forwards the inbound Feishu `message_id` as `request_id` on `channel.issueApiKey`, which is the seam durable dedup needs; a frontend that ignores the field keeps today's behaviour. **Durable deduplication is the frontend's half** — I've stated that explicitly in the design doc instead of shipping a runtime-only guard that looks like idempotency but isn't.

I have not added the sequential-replay test you suggested, because with the frontend still rotating per call it would only assert the current (duplicating) behaviour. It belongs with the frontend change that makes replay a no-op. Worth noting @chent1996's finding that the event layer ACKs immediately and detaches the work, so Feishu redelivery is already suppressed at the ACK layer — that narrows the live exposure to the multi-replica case, but does not remove it.

**Non-blocking nits (@chent1996)** — both fixed in the same commit:
- `docs/design/channels.md` didn't exist; Related now points at `docs/features/channels.mdx`.
- The hard-coded TTL copy is gone. Reply text no longer names any duration: the pickup line says "opens once, expires shortly" plus the **derived** absolute `expiresAt`, and status renders the derived date with the sliding mechanism ("using the key pushes this out") instead of the window length. The design doc now states that runtime copy must not restate frontend TTLs and that any duration shown must be derived from wire values.

Verification on 541f7e60: `tsc --noEmit` clean; 235 files / 4625 tests pass (+4 new cases).

src/gateway/channels/lark.ts
```diff
@@ -868,6 +885,84 @@ export async function handleLarkMessage(
...
+          const status = await getPersonalApiKeyStatus(personalChannelId, senderOpenId, frontendClient);
+          reply = formatApiKeyStatusReply(status, locale);
+          console.log(`[lark] /apikey status channel=${personalChannelId} sender=${senderOpenId} ok=${status.success} exists=${status.exists ?? "?"}`);
+        } else {
+          // messageId is forwarded as a stable request id so the frontend CAN make issuing
+          // idempotent across a redelivery or a second replica — the in-process guard above
+          // cannot span either.
+          const issued = await issuePersonalApiKey(personalChannelId, senderOpenId, frontendClient, messageId);
+          reply = formatApiKeyIssueReply(issued, locale);
+          carriesCommittedRotation = Boolean(issued.success && issued.pickupUrl);
```
chent1996, line 934: Non-blocking: `carriesCommittedRotation` is also true for a first-time issuance (`rotated: false`, no prior key existed). Escalating there is right — a created key whose link never arrived is still worth the retry and the audit line — but the log text below hard-codes "the previous key is already invalid", which is inaccurate for the first-time case and could send whoever greps the audit trail hunting for a phantom old key. Consider keying the sentence on `issued.rotated`, or neutral wording like "a key was issued but the pickup link never reached them".

chent1996 (approved):
Re-reviewed at 541f7e60 with the branch checked out locally. All four findings from the previous round — @LikiosSedo's two blocking reliability issues and my two nits — are addressed, and every claim in the author's response checks out:

**Lost-delivery escalation (finding 2)**: `replyToLark` now returns whether the message landed, with both failure shapes (thrown send failure, non-zero Feishu code) mapping to `false`. The issuing path checks delivery instead of inferring it, retries once on a committed issuance, and escalates to an audit line naming the sender but never the link. All 26 pre-existing `replyToLark` call sites ignore the return value and are unaffected (`tsc` confirms no new errors). Tests cover both failure shapes plus the negative case (a failed issue does not escalate and sends exactly one reply).

**request_id seam (finding 1)**: the inbound Feishu `message_id` is forwarded as `request_id` on `channel.issueApiKey` only (status stays bare), with tests for both inclusion and omission. Deferring durable dedup to the frontend — and saying so explicitly in the design doc instead of shipping a runtime-only guard that merely looks like idempotency — is the right division: the honest boundary beats a false one. I also agree with not adding the sequential-replay test yet; until the frontend half lands it would pin the duplicating behaviour as the contract. Residual live exposure is narrow (the ACK-immediately event layer already suppresses redelivery; multi-replica remains) and the failure mode stays user-recoverable.

**Nits**: the design-doc Related line now points at the real `docs/features/channels.mdx`, and reply copy no longer names any frontend TTL — durations are derived from wire values, with the new doc section making that a contract.

Verification: the three touched test files pass locally (316 tests); the full suite gains exactly the 4 new cases vs the previous head, all passing (the 18 file-load failures in my environment are missing optional otel deps, identical on main); `tsc --noEmit` introduces no new errors.

One tiny non-blocking wording note inline. Approving.

PR #456, opened 2026-07-29
Recover interrupted AgentBox sessions without replaying prompts

LikiosSedo:
## Summary
- confirm AgentBox termination before rebuilding and retry recovery at most three times
- rehydrate the same Pi session and continue from completed tool results without re-sending the original prompt
- capture bounded, redacted Pod termination evidence before stale Pod deletion
- fail closed with RECOVERY_UNSAFE when transcript continuation could duplicate work

## Validation
- npm test (236 files, 4639 passed, 2 skipped)
- npm run build
- targeted AgentBox recovery suite (267 passed)


src/core/brains/pi-agent-brain.ts
```diff
@@ -170,6 +171,54 @@ export class PiAgentBrain implements BrainSession {
...
+    return { ...base, eligible: true, reason: "Last persisted tool result completed successfully" };
+  }
+
+  async resumeFromHistory(): Promise<void> {
+    const state = this.getRecoveryState();
+    if (!state.eligible) {
+      throw new Error(`Session cannot resume from history: ${state.reason}`);
+    }
+    this.aborted = false;
+    await this.session.agent.continue();
```
chent1996, line 260: **Blocking 2.** This drives the CORE loop directly, bypassing `_runAgentPrompt` — which is where pi provides auto-retry of retryable LLM errors, auto-compaction (`_handlePostAgentRun` → `_checkCompaction` → `continue`), `_isAgentRunActive`/`agent_settled`, and the post-run drain of late-queued messages. The compaction loss matters most: transcripts eligible for recovery are by construction long mid-tool-loop transcripts, and the resume path also skips the routing preflight — so a near-ceiling session pays the whole pod rebuild and then fails on context anyway.

The clean shape is an upstream `AgentSession`-level continuation that reuses `_runAgentPrompt`'s post-run loop without an initial prompt. If that is out of scope for this PR, the design doc should list these losses as accepted gaps, and the reachable one (context preflight before resuming) should be added here.

Also worth noting in the docstring: this is the first caller in the codebase to run `agent.continue()` against a cold-rehydrated session — production pi only calls it from inside `_runAgentPrompt` on a warm one.

src/core/brains/pi-agent-brain.ts
```diff
@@ -170,6 +171,54 @@ export class PiAgentBrain implements BrainSession {
...
+      pendingToolCalls: pending.size,
+      completedToolResults,
+    };
+    if (!last) {
+      return { ...base, eligible: false, reason: "Session history is empty" };
+    }
+    if (pending.size > 0) {
+      return { ...base, eligible: false, reason: "Session has pending tool calls" };
+    }
+    if (last.role !== "toolResult") {
```
chent1996, line 223: Non-blocking. This rejects a transcript ending at the `user` message — but that is the widest real crash window (the first LLM call is the longest await of a turn), and it is exactly as safe as the toolResult case by this PR's own reasoning: tools only execute after their assistant message hit the JSONL, so a user-ending transcript proves no side effect has occurred since persist. pi's loop precondition (`agent-loop.js:23`) explicitly allows continuing from `user`. If the strictness is deliberate first-landing conservatism, record that in the design doc; otherwise allowing `user` here is free coverage for the most common interruption.

src/gateway/agentbox/stream-recovery.ts
```diff
@@ -0,0 +1,75 @@
...
+  agentId: string,
+  signal: AbortSignal,
+  options: { timeoutMs?: number; pollMs?: number } = {},
+): Promise<boolean> {
+  const timeoutMs = options.timeoutMs ?? 10_000;
+  const pollMs = options.pollMs ?? 500;
+  const deadline = Date.now() + timeoutMs;
+
+  while (!signal.aborted) {
+    const info = await manager.inspect(agentId);
```
chent1996: Non-blocking, two things on this poll:

1. `inspect()` can throw (K8s API blip) and nothing here catches it — the throw escapes the recovery `catch` in server.ts, skips the STREAM_INTERRUPTED/AGENTBOX_FAILED classification, and surfaces as generic `INTERNAL` from the outer catch. Treating a failed read as "cannot confirm this round" keeps the error taxonomy intact.

2. If a concurrent turn (another session on the same box) already triggered `getOrCreate`, this loop polls the healthy REPLACEMENT pod, never confirms, and gives up — fail-safe, but comparing the pod UID against the one observed at failure time would let both sessions recover.

src/gateway/server.ts
```diff
@@ -395,29 +400,126 @@ export async function startRuntime(opts: StartRuntimeOptions): Promise<RuntimeSe
...
-              code: ErrorCodes.STREAM_INTERRUPTED,
+              console.warn(
+                `[runtime] AgentBox failure confirmed; recovering session=${promptResult.sessionId} ` +
+                `attempt=${recoveryAttempts}/${MAX_AGENTBOX_RECOVERY_ATTEMPTS}`,
+              );
+
+              try {
+                // getOrCreate waits until the replacement pod is Ready. The same
+                // sessionId then rehydrates its Pi JSONL from the agent PVC.
+                const replacement = await agentBoxManager.getOrCreate(agentId);
```
chent1996: Coordination note for #462 (multi-AgentBox per agent), which is in flight against the same lines: there `getOrCreate` takes a `sessionId` and a pooled agent's session may live on instance 1..N, while this recovery path — and `manager.inspect(agentId)` above — resolve the instance-0 pod name only. Under pooling, a death would be "confirmed" against the wrong pod and the rebuilt handle may not be the session's box. The two PRs merge cleanly textually; whichever lands second needs to route recovery through the session-aware acquisition (`getOrCreate(agentId, undefined, sessionId)` / `getForSession`).

src/agentbox/http-server.ts
```diff
@@ -988,36 +1015,40 @@ export function createHttpServer(
...
-        // of the live SSE subscription that enriches agent_end — re-apply the
-        // same enrichment so token/cost stats survive on every turn.
-        emitBrainEvent: (event) => emitSessionExtraEvent(enrichAgentEndEvent(managed.brain, event)),
-        onStateChange: () => sessionManager.persistModelRouteState(managed.id, managed.modelRouteState),
-        shouldAbort: () => managed._aborted,
-      },
-      promptMedia,
-    );
+    const promptPromise = body.resumeFromHistory
+      ? managed.brain.resumeFromHistory!()
```
chent1996, line 1067: Non-blocking, matrix note: the model-routing row in CLAUDE.md records that the `routeEnabled ? runner : brain.prompt` ternary was deliberately removed so EVERY turn flows through `runPromptWithModelRouting` (single entry, uniform commit-gating, uniform preflight). This branch reintroduces that ternary shape for resume. The persistence side is fine — no `model_route_start` means the consumer persists inline, the documented edge — but the turn loses `ensureContextForModelPrompt`, which is the concrete half of Blocking 2. If the resume stays outside the runner, the matrix row should gain a sentence saying resume turns are the second legitimate bypass and why.

chent1996 (changes requested):
Reviewed on a checkout of the branch. Merge-base is `8b17d499` (#454); a trial merge against current main (`4ce904a2`, post-#457/#458) is clean. CI green; the targeted suites pass locally (267 tests, matching the description's number) and the full suite passes.

The safety architecture here is genuinely sound, and I verified its central invariant against pi's actual source rather than taking the doc's word for it:

- **The "crash mid-tool ⇒ ineligible" guarantee holds.** pi persists every message synchronously (`appendFileSync`) on `message_end`, and `streamAssistantResponse` emits the assistant message's `message_end` **before** `executeToolCalls` runs (`pi-agent-core/dist/agent-loop.js`). So a pod that dies while a tool is executing always leaves a JSONL ending in an assistant message with an unmatched `toolCall`, which `getRecoveryState()` rejects (`pendingToolCalls > 0`). A tool that ran its side effect but wasn't persisted can never be silently re-planned. Parallel tool batches are covered too — results append one at a time, so a partial batch leaves the missing call unmatched.
- **"Never re-sends the original prompt" is real end to end**: `ResumeSessionOptions` excludes text/media at the type level, the box rejects a resume carrying either with a 400, the placeholder `promptText` feeds only `detectLanguage` (returns English → no-op) and the trace root — it never reaches the brain, and `resumeFromHistory()` appends no user message.
- **Duplicate-persistence is closed by existing machinery**: the first attempt's assistant rows were commit-gated (routed turn) and die with the consumer; tool rows persisted at `tool_start` are not re-emitted because the replacement box's SSE buffer starts after hydration; the resumed turn is un-routed, so its rows persist inline — the documented no-`model_route_start` edge in the matrix.
- The eligibility gate is *stricter* than pi's own continuation precondition (agent-loop only requires last = `user` or `toolResult`), the 409 path fails closed through `agentBoxResponseError` (non-2xx throws, classified non-transport, loop exits), `prompt_done` is emitted on every exit path including the outer catch, and the evidence logger is genuinely bounded (4 KiB) and redacted through the shared `buildRedactionConfig`.

Two findings block, both about the gap between this careful design and what has actually been exercised.

---

## Blocking 1: the feature's core path has never executed anywhere

Validation lists `npm test` and `npm run build` only. Every test of the recovery mechanics is mocked at exactly the layer the feature exists to exercise: `server-chat-abort.test.ts` mocks `consumeAgentSse` and `resumeSession`; `http-server.test.ts` uses a fake brain; `pi-agent-brain.test.ts` fakes the session and stubs `agent.continue` with `vi.fn()`. What remains unexecuted, anywhere, is the actual claim in the PR title: **a real pi session rehydrated from JSONL on a fresh process, continued via `agent.continue()`, producing a correct continuation stream.**

I traced the static chain far enough to believe it *should* work — `session.messages` is literally `agent.state.messages` (agent-session.js:638), so a hydration that satisfies `getRecoveryState()` has by construction also seeded the loop's context, and an unseeded hydration fails closed as "history is empty". But `agent.continue()` is invoked in production pi only from inside `_runAgentPrompt` on a warm session; this PR is the first caller ever to run it against a cold-rehydrated session. That difference is exactly the kind that unit mocks cannot see, and this code has the property that **its first real execution will be during a production incident** — the worst possible place to discover a hydration mismatch.

What would unblock, either one:
- A kill-drill on the internal cluster (the same environment #462's table came from): start a multi-tool turn, `kubectl delete pod` mid-tool-loop, observe the rebuild, the resumed stream, and the final transcript; paste the observations into the PR.
- Or an integration test with **no mocks in the middle**: a real `SessionManager` JSONL + real `AgentSession` over a scripted `streamFn`, write a transcript ending in a successful toolResult, construct a *fresh* session from that file (simulating the replacement pod), and assert `resumeFromHistory()` streams a continuation without a new user message.

## Blocking 2: `resumeFromHistory` bypasses `_runAgentPrompt`, and the biggest loss lands on exactly the sessions recovery targets

`this.session.agent.continue()` drives the **core** agent loop directly. The AgentSession wrapper it skips (`_runAgentPrompt` → `_handlePostAgentRun` loop) is what provides, per pi's source:

1. **Auto-retry of retryable LLM errors** — a recovered turn that hits a transient 429/overload dies with an error where any normal turn would have retried.
2. **Auto-compaction** (`_checkCompaction` → continue) — a recovered turn that crosses the context threshold won't compact-and-continue. This one is not an edge case: the transcripts eligible for recovery are by construction long, mid-tool-loop transcripts — the population *most* likely to need compaction. The resume path also skips the routing runner's `ensureContextForModelPrompt` preflight (`effectivePolicy = undefined`), so there is no compaction opportunity on either side: a near-ceiling session pays the whole pod rebuild and then fails on context anyway.
3. `_isAgentRunActive` stays false → `session.isStreaming`/`isIdle` report wrong values for the duration; `agent_settled` never fires; the post-run drain of messages queued by agent_end handlers never runs (a steer that lands in the closing race window sits in the queue until the next prompt).

None of this is unsafe — every failure surfaces — but #2 turns a successful recovery into a guaranteed failure for the long-transcript case, and #3 makes a recovered turn observably not a first-class pi run. What would unblock: route the continuation through the same wrapper a prompt uses (an upstream `AgentSession.continueRun()` that reuses `_runAgentPrompt`'s loop with no initial prompt is the clean shape), or — if that upstream change is out of scope now — document these three losses explicitly in `docs/design/agentbox-stream-recovery.md` as accepted gaps and add the compaction preflight to the resume path, since that one is reachable without any upstream change.

---

## Non-blocking

- **Coordination with #462 (multi-AgentBox per agent):** the two PRs merge cleanly textually but disagree semantically. This PR's `inspect(agentId)` and recovery `getOrCreate(agentId)` both resolve the instance-0 pod name; under #462's pooling, a session pinned to instance 1 would have its death "confirmed" against the wrong pod, and the rebuilt handle would not be the session's box. #462 also changes these `getOrCreate` calls to pass `sessionId`. Whichever lands second must reconcile the recovery loop with `getForSession`/pool semantics — worth a shared note on both PRs now.
- **Ends-with-`user` transcripts are rejected, leaving the most common crash window unrecovered.** A pod that dies during the FIRST LLM call of a turn (statistically the widest window — it is the longest await) leaves JSONL ending at the user message: no tool has run since persist (tools only execute post-`message_end`), so continuing is exactly as safe as the toolResult case, and pi's own loop precondition explicitly allows it. If the strictness is deliberate first-landing conservatism, say so in the design doc; otherwise this is free coverage.
- **`waitForConfirmedAgentBoxFailure` lets `inspect()` throws escape.** A K8s API blip mid-confirmation propagates out of the recovery `catch`, skips the STREAM_INTERRUPTED/AGENTBOX_FAILED classification, and lands in the outer catch as generic `INTERNAL` (prompt_done is still sent, so nothing hangs). Catching inside the poll loop and treating a failed read as "cannot confirm" preserves the intended taxonomy for one try/catch.
- **Confirmation races a concurrent rebuild.** If two sessions on one box die together, the first one's `getOrCreate` replaces the pod; the second's confirmation loop then polls a healthy replacement, never confirms, and gives up with STREAM_INTERRUPTED. Fail-safe direction, but comparing pod UID against the one observed at failure time would recover both.
- **`agentbox_recovery_start/end` events have no consumer.** portal-web is untouched, so during a rebuild (pod Ready wait can be tens of seconds) the user watches a silently stalled stream — the exact experience these events were added to explain. A follow-up UI change is fine; flagging so it lands.
- Local mode: `inspect` exists there too, so the recovery loop is armed; a matching transport error costs one 10s confirmation timeout before the old STREAM_INTERRUPTED behavior. Harmless, but a `isK8s` guard would keep local semantics byte-identical.

### Verified

- Persistence ordering (assistant-before-execute) and per-message `appendFileSync` — read from `pi-agent-core/dist/agent-loop.js` and `pi-coding-agent/dist/core/agent-session.js:329-341`, not assumed.
- 409/RECOVERY_UNSAFE flows: client `fetch` throws on non-2xx → classified non-transport → terminal, with the final envelope's `retriable: true` correctly describing "re-ask is safe".
- Evidence capture bounded + redacted; `FallbackToLogsOnError` added to the container spec; error-code registry and doc updated consistently (12 codes).
- `manager.inspect` is read-only (no create/mutate), and the recovery loop guards on its existence before use.
- Full local run: targeted 267/267, full suite green; trial merge with current main clean.

PR #457, opened 2026-07-30
feat(lark): add topic-scoped conversations

LikiosSedo:
## Summary
- enable native Feishu Topic conversations only for Personal (per-user) group context
- keep Team (shared) mode on the existing main-group reply path and one chat-scoped Agent session
- reuse an existing participant-scoped Topic session for no-mention follow-ups without letting another user inherit it
- keep text, card, image, background, error, and Topic-local reset replies on the selected delivery path
- document the rollout flag, complete test-bot permissions, and Personal/Team acceptance flow

## Validation
- npm run build
- npm test -- --run
- focused Lark suite: 134 passed
- Sicore companion MR: http://gitlab.scitix-inner.ai/k8s/sicore/-/merge_requests/770

### DIFF
--- a/src/gateway/channel-manager.ts
+++ b/src/gateway/channel-manager.ts
@@ -79,6 +79,10 @@ export function isChannelAccessDenied(
  * for authorized group bots and choose the effective session key server-side
  * (open groups → shared chat session; authorized → per-user). It is separate
  * from `sessionKey` because the server may override the session key it returns.
+ * `conversationKey` is an optional provider conversation scope (for example a
+ * Feishu topic root). The Portal combines it with the binding's context mode
+ * and actor identity; the runtime must treat the returned `sessionKey` as
+ * authoritative.
  */
 export async function resolveBinding(
   channelId: string,
@@ -86,12 +90,16 @@ export async function resolveBinding(
   frontendClient: FrontendWsClient,
   sessionKey?: string,
   senderOpenId?: string,
+  conversationKey?: string,
+  conversationExistingOnly: boolean = false,
 ): Promise<ResolvedChannelBinding | ChannelAccessDenied | null> {
   const data = await frontendClient.request("channel.resolveBinding", {
     channel_id: channelId,
     route_key: routeKey,
     ...(sessionKey ? { session_key: sessionKey } : {}),
     ...(senderOpenId ? { sender_open_id: senderOpenId } : {}),
+    ...(conversationKey ? { conversation_key: conversationKey } : {}),
+    ...(conversationExistingOnly ? { conversation_existing_only: true } : {}),
   });
   return data.binding ?? null;
 }
--- a/src/gateway/channels/lark.test.ts
+++ b/src/gateway/channels/lark.test.ts
@@ -1340,6 +1340,35 @@ describe("handleLarkMessage — group @-mention gating (@所有人 bug)", () => 
     expect(resolveBindingMock).not.toHaveBeenCalled();
   });
 
+  it("/mode resolves only the group mode and does not allocate a topic session", async () => {
+    resolveBindingMock.mockResolvedValue(makeBinding({
+      sessionId: "shared-session",
+      sessionKey: "chat:oc_abc123",
+      contextMode: "shared",
+    }));
+
+    await handleLarkMessage(
+      makeTextEvent("/mode", { chat_type: "group", mentions: [] }),
+      makeLarkClient(),
+      "lark",
+      makeAgentBoxManager("a1") as any,
+      undefined,
+      {} as any,
+      "zh-CN",
+      { app_id: "x", app_secret: "y", thread_mode: "group" },
+      BOT,
+    );
+
+    expect(resolveBindingMock).toHaveBeenCalledWith(
+      "lark",
+      "oc_abc123",
+      expect.anything(),
+      "open_id:ou_user_1",
+      "ou_user_1",
+      undefined,
+    );
+  });
+
   it("thread mode creates a topic reply and accepts an unmentioned follow-up in the same root", async () => {
     resolveBindingMock.mockResolvedValue(makeBinding({
       sessionId: "thread-session",
@@ -1418,6 +1447,47 @@ describe("handleLarkMessage — group @-mention gating (@所有人 bug)", () => 
     }));
   });
 
+  it("starts an explicitly mentioned quoted message as a new topic rooted at the current message", async () => {
+    resolveBindingMock.mockResolvedValue(makeBinding({
+      sessionId: "quoted-topic-session",
+      sessionKey: "open_id:ou_user_1:lark_thread:mid-quoted-at",
+      contextMode: "per_user",
+    }));
+    promptMock.mockResolvedValue({ sessionId: "quoted-topic-session" });
+    streamEventsMock.mockImplementation(async function* () {
+      yield {
+        type: "message_end",
+        message: { role: "assistant", content: [{ type: "text", text: "quoted answer" }] },
+      };
+    });
+    const client = makeLarkClient();
+
+    await handleLarkMessage(
+      makeTextEvent("@_user_1 分析这条引用", {
+        message_id: "mid-quoted-at",
+        chat_type: "group",
+        root_id: "mid-older-message",
+        mentions: [{ key: "@_user_1", id: { open_id: BOT } }],
+      }),
+      client,
+      "lark",
+      makeAgentBoxManager("a1") as any,
+      undefined,
+      {} as any,
+      "zh-CN",
+      { app_id: "x", app_secret: "y", thread_mode: "group" },
+      BOT,
+    );
+
+    expect(resolveBindingMock.mock.calls.map((call) => call[5])).toEqual([
+      "lark_thread:mid-quoted-at",
+      "lark_thread:mid-quoted-at",
+    ]);
+    expect(resolveBindingMock.mock.calls.map((call) => call[6])).toEqual([false, false]);
+    expect(client.im.message.reply.mock.calls[0][0].path.message_id).toBe("mid-quoted-at");
+    expect(client.im.message.reply.mock.calls[0][0].data.reply_in_thread).toBe(true);
+  });
+
   it("thread rollout keeps a shared group on the main-group reply path", async () => {
     resolveBindingMock.mockResolvedValue(makeBinding({
       sessionId: "shared-session",
@@ -1454,16 +1524,15 @@ describe("handleLarkMessage — group @-mention gating (@所有人 bug)", () => 
     expect(rootClient.im.message.reply.mock.calls[0][0].data.reply_in_thread).toBeUndefined();
     expect(appendMessageMock.mock.calls[0][0].metadata).not.toHaveProperty("conversationKey");
 
-    const topicClient = makeLarkClient();
+    const quoteClient = makeLarkClient();
     await handleLarkMessage(
-      makeTextEvent("话题里的普通讨论", {
-        message_id: "mid-shared-followup",
+      makeTextEvent("引用回复里的共享讨论", {
+        message_id: "mid-shared-quote",
         chat_type: "group",
         root_id: "mid-1",
-        thread_id: "omt-shared-topic",
         mentions: [],
       }),
-      topicClient,
+      quoteClient,
       "lark",
       makeAgentBoxManager("a1") as any,
       undefined,
@@ -1474,7 +1543,27 @@ describe("handleLarkMessage — group @-mention gating (@所有人 bug)", () => 
     );
 
     expect(promptMock).toHaveBeenCalledTimes(1);
-    expect(topicClient.im.message.reply).not.toHaveBeenCalled();
+    expect(resolveBindingMock).toHaveBeenCalledTimes(2);
+    expect(quoteClient.im.message.reply).not.toHaveBeenCalled();
+
+    await handleLarkMessage(
+      makeTextEvent("@_user_1 继续分析", {
+        message_id: "mid-shared-next",
+        chat_type: "group",
+        mentions: [{ key: "@_user_1", id: { open_id: BOT } }],
+      }),
+      makeLarkClient(),
+      "lark",
+      makeAgentBoxManager("a1") as any,
+      undefined,
+      {} as any,
+      "zh-CN",
+      config,
+      BOT,
+    );
+
+    expect(promptMock).toHaveBeenCalledTimes(2);
+    expect(promptMock.mock.calls[1][0].text).toContain("引用回复里的共享讨论");
   });
 
   it("thread mode ignores an unmentioned topic that has no existing bot session", async () => {
--- a/src/gateway/channels/lark.test.ts
+++ b/src/gateway/channels/lark.test.ts
@@ -564,7 +564,15 @@ describe("handleLarkMessage — routing to AgentBox", () => {
     resolveBindingMock.mockResolvedValue(null);
     const mgr = makeAgentBoxManager();
     await handleLarkMessage(makeTextEvent("hello"), makeLarkClient(), "lark", mgr as any, undefined, {} as any);
-    expect(resolveBindingMock).toHaveBeenCalledWith("lark", "oc_abc123", expect.anything(), "open_id:ou_user_1", "ou_user_1");
+    expect(resolveBindingMock).toHaveBeenCalledWith(
+      "lark",
+      "oc_abc123",
+      expect.anything(),
+      "open_id:ou_user_1",
+      "ou_user_1",
+      undefined,
+      false,
+    );
     expect(mgr.getOrCreate).not.toHaveBeenCalled();
     expect(promptMock).not.toHaveBeenCalled();
   });
@@ -592,7 +600,15 @@ describe("handleLarkMessage — routing to AgentBox", () => {
       },
     );
 
-    expect(resolveBindingMock).toHaveBeenCalledWith("lark", "oc_abc123", expect.anything(), "open_id:ou_user_1", "ou_user_1");
+    expect(resolveBindingMock).toHaveBeenCalledWith(
+      "lark",
+      "oc_abc123",
+      expect.anything(),
+      "open_id:ou_user_1",
+      "ou_user_1",
+      undefined,
+      false,
+    );
     expect(resolvePersonalBindingMock).not.toHaveBeenCalled();
   });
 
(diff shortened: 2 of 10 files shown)


src/portal/adapter.ts
```diff
@@ -134,6 +144,23 @@ async function resolveChannelBinding(
...
+  contextMode: ContextMode,
+  participantKey?: string | null,
+  conversationKey?: string | null,
+): string | null {
+  const conversation = conversationKey?.trim();
+  if (conversation) {
+    return contextMode === "shared"
+      ? conversation
+      : participantKey
+        ? `${participantKey}:${conversation}`
```
chent1996: **Blocking — non-idempotent append double-suffixes the real session key.** The queued re-resolve in `lark.ts` sends the composite key returned by the first resolve (`ctx.sessionKey = binding.sessionKey`, e.g. `open_id:U:lark_thread:M`) together with the same `conversation_key`, and this branch appends again. Verified against the real `channel.resolveBinding` handler:

```
session_key      = open_id:ou_1:lark_thread:mid-root
conversation_key = lark_thread:mid-root
→ effective key  = open_id:ou_1:lark_thread:mid-root:lark_thread:mid-root
```

So prompts run on the double-suffixed session, while the first resolve's `open_id:U:lark_thread:M` row is an orphan stepping-stone created once per topic. This also breaks `/new` inside a topic: `handleNewCommand` resets the stepping-stone key, not the session prompts run on, so the user gets a "new session" confirmation while the topic context survives intact.

Suggested fix: skip the append when `participantKey` already ends with `:${conversation}` (or equals `conversation`). The Sicore mirror in MR 770 needs the same fix.

src/portal/adapter.ts
```diff
@@ -134,6 +144,23 @@ async function resolveChannelBinding(
...
+function buildGroupSessionKey(
+  routeKey: string,
+  contextMode: ContextMode,
+  participantKey?: string | null,
+  conversationKey?: string | null,
+): string | null {
+  const conversation = conversationKey?.trim();
+  if (conversation) {
+    return contextMode === "shared"
+      ? conversation
```
chent1996: **Blocking — the shared branch adopting `conversationKey` contradicts the PR's own Team-mode contract.** The doc says "the topic rollout does not split it by Feishu topic" and the runtime comment says "shared group → chat:<route-key> (topic candidates are ignored)" — but this branch returns the topic key. Because the runtime's pre-queue resolve passes `topicConversationKey` before the context mode is known, every `@bot` root message in a flag-enabled shared group creates an orphan `lark_thread:{messageId}` participant-session row here (a root message's conversation key is its own message id, so each one is unique). `resolveOpenGroupBinding` inherits the same problem via its hardcoded `"shared"` call.

Suggested fix: shared mode should ignore `conversationKey` entirely and always return `chat:${routeKey}`. That also makes the runtime's shared-group mock in `lark.test.ts` (which returns `chat:oc_abc123` for this exact call) match reality.

src/gateway/channels/lark.ts
```diff
@@ -1072,11 +1104,19 @@ export async function handleLarkMessage(
...
   // Look up binding for this chat. Pass sender_open_id so the Portal can
   // auto-bind / per-sender resolve group bots and pick the session key.
-  const binding = await resolveBinding(groupChannelId, chatId, frontendClient!, sessionKey, senderOpenId ?? undefined);
+  const binding = await resolveBinding(
+    groupChannelId,
+    chatId,
+    frontendClient!,
+    sessionKey,
+    senderOpenId ?? undefined,
+    topicConversationKey,
```
chent1996, line 1122: **Blocking (runtime side of the shared-group issue) — this resolve runs before `contextMode` is known, so shared groups get topic-scoped keys.** With the current Portal behavior, a flag-enabled shared group gets `binding.sessionKey = lark_thread:{messageId}` from this call, which makes `queueKey` unique per message. Shared-group turn serialization is lost: two concurrent mentions run against the same shared session simultaneously (the queued re-resolve falls back to `chat:{routeKey}`), and the second user receives the session-busy notice instead of waiting in the queue — a regression from the pre-PR behavior where all shared turns serialized on `${bindingId}:chat:${routeKey}`.

Fixing the Portal's shared branch to ignore `conversationKey` resolves this too, since this call would then return the stable `chat:{routeKey}` key.

src/gateway/channels/lark.test.ts
```diff
@@ -998,6 +1014,44 @@ describe("handleLarkMessage — routing to AgentBox", () => {
...
+      "lark",
+      makeAgentBoxManager("a1") as any,
+      undefined,
+      {} as any,
+      "zh-CN",
+      { app_id: "x", app_secret: "y", thread_mode: "group" },
+      "ou_bot_self",
+    );
+
+    expect(resetBindingSessionMock).toHaveBeenCalledWith(
```
chent1996, line 1046: **Blocking — this assertion codifies the wrong behavior.** With the real Portal, prompts do not run on `open_id:ou_user_1:lark_thread:mid-1`; they run on the double-suffixed key that the queued re-resolve produces (`…:lark_thread:mid-1:lark_thread:mid-1`), because `resolveQueuedBinding` sends the server-returned composite key back together with `conversation_key` and `buildGroupSessionKey` appends again. The static `resolveBindingMock` hides that, so this test passes while the real `/new` resets a session no prompt runs on. Once the Portal transform is idempotent, this assertion becomes correct as written.

src/gateway/channels/lark.ts
```diff
@@ -1108,6 +1150,17 @@ export async function handleLarkMessage(
   if (cachedMode && cachedMode !== contextMode) forgetGroupState(groupChannelId, chatId);
   rememberGroupMode(groupChannelId, chatId, contextMode);
 
+  // A no-@ message inside a Feishu topic is a continuation only in personal
+  // mode. Team mode deliberately stays on the old main-group path; an
+  // unrelated/manual topic must not wake the shared Agent session.
+  if (isThreadFollowup && !isBotMentioned(message, botOpenId) && contextMode === "shared") {
```
chent1996: Non-blocking but worth fixing in this PR: two regressions for flag-enabled shared groups hide here.

1. Pre-PR, a no-@ message with a `root_id` was buffered into the shared discussion context by the gate above; now it is dropped. Note Feishu sets `root_id` on plain quote-replies too, not only topic messages, so this silently removes *all* quote-reply chatter from the shared context feature.
2. Every such message now costs a `channel.resolveBinding` RPC before landing here, where the pre-PR path was deliberately RPC-free for non-@ chatter (see the comment above about not doing an RPC per message).

Suggestion: short-circuit before the resolve when `cachedGroupMode(...) === "shared"` (shared never uses topics) and buffer the message there — that restores the buffering and removes the per-chatter RPC in one move. Alternatively, at minimum call `appendDiscussion` in this branch now that the mode is confirmed shared.

src/gateway/channels/lark.ts
```diff
@@ -1028,7 +1049,14 @@ export async function handleLarkMessage(
...
   // it works with or without an @bot (like PAIR); command words are exact.
   if (/^\/mode$/i.test(text.trim())) {
-    const modeBinding = await resolveBinding(groupChannelId, chatId, frontendClient!, sessionKey, senderOpenId ?? undefined);
+    const modeBinding = await resolveBinding(
+      groupChannelId,
+      chatId,
+      frontendClient!,
+      sessionKey,
+      senderOpenId ?? undefined,
+      topicConversationKey,
```
chent1996: Non-blocking: passing `topicConversationKey` here creates one unused topic-scoped participant session per `/mode` invocation (the command message is its own root, so the key is fresh each time, and the mode card never prompts). This resolve only needs `contextMode` and the access check — suggest not passing the conversation key at all.

src/gateway/channel-manager.ts
```diff
@@ -79,19 +79,27 @@ export function isChannelAccessDenied(
  * for authorized group bots and choose the effective session key server-side
  * (open groups → shared chat session; authorized → per-user). It is separate
  * from `sessionKey` because the server may override the session key it returns.
+ * `conversationKey` is an optional provider conversation scope (for example a
```
chent1996, line 82: Cross-repo contract question for the Sicore side (MR 770): for a `conversation_existing_only` resolve, the frontend must check topic-session existence **before** the sender-authorization check and return `null` (not access-denied) when no session exists. Otherwise, in a `sicore_authorized` group, every unauthorized member's no-@ quote-reply (Feishu sets `root_id` on plain quote-replies) resolves to denied and the runtime replies with the access hint — spamming messages that were never aimed at the bot. The standalone adapter is safe (it returns `null`), but please confirm the ordering in MR 770 and consider documenting it in this comment.

src/portal/adapter-rpc.test.ts
```diff
@@ -1705,6 +1705,85 @@ describe("channel.resolveBinding", () => {
     });
   });
 
+  it("scopes a shared group session to the Feishu topic root", async () => {
```
chent1996: This test asserts the behavior that contradicts the PR's documented Team-mode contract ("the topic rollout does not split it by Feishu topic") — and the runtime-side test for the same scenario mocks `resolveBinding` to return `chat:oc_abc123`, so the two layers currently assert opposite contracts for identical inputs. If the shared branch of `buildGroupSessionKey` is fixed to ignore `conversationKey`, this test should flip to asserting `chat:group-123` / no session-row insert.

chent1996 (changes requested):
The overall design (default-off rollout flag, personal-mode-only topics, server-authoritative session keys, existing-only guard for unmentioned follow-ups) is sound, and the default-off path is verified unchanged. However, tracing the real runtime→Portal flow (rather than the mocked one) surfaces three blocking defects, all sharing one root cause: **the conversation-key transform in `buildGroupSessionKey` is neither idempotent nor consistent with the documented Team-mode contract, and the runtime resolves every message twice with different inputs** (raw participant key first, server-returned composite key in the queued step).

### Blocking

1. **Double-appended session key (verified against the real RPC handler).** The queued re-resolve sends `session_key = open_id:U:lark_thread:M` (the composite returned by the first resolve) plus `conversation_key = lark_thread:M`; the Portal appends again, so the session that prompts actually run on is keyed `open_id:U:lark_thread:M:lark_thread:M`. I reproduced this by calling the real `channel.resolveBinding` handler with those inputs. Continuity still works only because every turn deterministically walks the same two-step transform — but each topic leaves an orphan stepping-stone session, and the persistent key is wrong.
2. **`/new` inside a topic is a false success.** `handleNewCommand` resets `ctx.sessionKey` (the stepping-stone key from the first resolve), while prompts run on the double-suffixed session from the queued resolve. The user gets the "new session" confirmation, but the topic conversation is not reset. The new unit test codifies this because the mocked `resolveBinding` hides the Portal's re-append.
3. **Shared groups with the flag on lose turn serialization and leak sessions.** The pre-queue resolve passes `topicConversationKey` before the context mode is known, and the Portal's shared branch *adopts* it — contradicting both the doc ("the topic rollout does not split it by Feishu topic") and the runtime comment ("shared group → chat:<route-key> (topic candidates are ignored)"). Every `@bot` root message creates an orphan `lark_thread:{messageId}` row, and `queueKey` fragments per message, so concurrent mentions hit the same shared session simultaneously and the second user gets the session-busy notice instead of queueing. The two test layers currently assert contradictory contracts (see inline comments).

### What would unblock

- Portal: make the shared branch of `buildGroupSessionKey` ignore `conversationKey` (always `chat:{routeKey}`), and make the per_user append idempotent (skip when `participantKey` already ends with `:{conversation}`). That alone fixes all three blockers; the same fix must land in the Sicore mirror (MR 770).
- Runtime: don't pass `topicConversationKey` on the `/mode` resolve (it creates an unused topic session per invocation), and restore shared-group buffering of no-@ reply-chain chatter (see inline comment on the shared gate).
- Align the contradictory assertions in `adapter-rpc.test.ts` and `lark.test.ts` with the corrected contract.

### Non-blocking asks

- **Live Feishu verification:** Validation lists build/tests/companion MR only. The doc adds a thorough acceptance checklist, but the PR doesn't state it was executed against a live test bot. `reply_in_thread` topic creation and `root_id`/`thread_id` semantics (note: Feishu sets `root_id` on plain quote-replies too, not only topic messages) are provider behaviors that mocks cannot cover. Please run the checklist before merge — the "Team-mode continuity" step with two concurrent mentions will surface blocker 3 directly.
- Confirm the check ordering in Sicore MR 770 for `conversation_existing_only` (inline comment on channel-manager.ts).

Verification notes: focused suites pass as claimed (lark 134; 338 across the four touched test files); the 5 `tsc --noEmit` errors are pre-existing on main (local optional tracing deps) and unrelated; flag-off behavior is unchanged end to end; both the HTTP and WS `channel.resolveBinding` mirrors were updated consistently.

jacoblee-io (approved):
LGTM. Reviewed at 8d6e9c72.

The per-user topic isolation is correct: topic sessions are keyed `open_id:<sender>:lark_thread:<root>`, and a no-@ follow-up resolves `conversation_existing_only`, so another user replying in the same topic gets a distinct key (or a null binding) and can never inherit an existing session. No-@ topic follow-ups only continue the sender's own established topic; unrelated/manual topics resolve to null and are ignored, so idle pods are not woken by group chatter. reply_in_thread is plumbed consistently through text/card/image, and the runtime degrades gracefully against an older Portal that ignores the new `conversation_key` params.

The one issue from my earlier pass — shared/Team mode being partially dragged into topic scoping (the pre-queue binding resolving a per-topic key, which split the shared group's queue key and created unused `lark_thread:*` session rows) — is fully addressed by 8d6e9c72: `buildGroupSessionKey` now short-circuits `contextMode === "shared"` to `chat:<routeKey>` before any conversation handling, and the adapter test was inverted accordingly ("keeps a shared group on its chat session when a topic candidate is present"). The same commit also hardens a case I hadn't flagged — an ordinary Feishu quote reply also carries `root_id`, so topic detection is now gated on `thread_id !== null`, and a no-@ access-denied follow-up stays silent instead of emitting an access hint.

Nice, careful work on the isolation semantics. Approving.

src/gateway/channels/lark.ts
```diff
@@ -1175,11 +1263,23 @@ async function processQueuedLarkMessage(ctx: QueuedLarkMessageContext): Promise<
     // would clear everyone's context — reject it instead of resetting. (A
     // confirmation-gated "reset the whole room" is deferred.) per_user groups
     // and personal chats reset the caller's own session as before.
-    if (contextMode === "shared") {
-      await replyToLark(larkClient, messageId, SHARED_NEW_REJECTED_NOTICE_BY_LOCALE[locale]);
+    if (contextMode === "shared" && !conversationKey) {
```
chent1996, line 1266: Non-blocking nit (latent, not reachable today): the `&& !conversationKey` clause weakens this guard in the one direction it exists to protect.

Today `conversationKey` is only ever set when `contextMode === "per_user"` (`personalTopicMode` gates it), so the clause is always true in shared mode and the guard behaves as before. But its effect is "if a shared group ever *does* carry a conversation key, allow `/new`" — i.e. if topics are later enabled for Team mode, a single member's `/new` would silently reset the whole room's session, which is exactly what this branch was written to prevent. The same shape appears at the `drained` computation below, where it is harmless.

Suggest either dropping the clause (shared ⇒ always reject, matching the comment above it) or keeping it with a note that it is deliberately dead until shared-mode topics exist.

chent1996 (approved):
## Review (second pass — approving)

Re-reviewed at `8d6e9c72`. **All three blocking findings are fixed, and I verified the central one by execution rather than by reading.** The two code-level non-blocking items were fixed too, and the cross-repo one is now defensively closed on the runtime side.

### Blocking #1 — non-idempotent session-key append: FIXED (verified by chained execution)

`buildGroupSessionKey` now guards with `participant.endsWith(":" + conversation)` before appending. Rather than trust the new unit test (which asserts a hard-coded composite input), I re-ran the real `channel.resolveBinding` handler in a **chained** two-step flow — feeding resolve-1's own returned `sessionKey` back into resolve-2 with the same `conversation_key`, i.e. exactly what `resolveQueuedBinding` does:

```
STEP1 key        = open_id:ou_1:lark_thread:mid-root
STEP2 key        = open_id:ou_1:lark_thread:mid-root
STEP2 upsert key = open_id:ou_1:lark_thread:mid-root
```

Stable, and the upsert targets the same row — so the orphan stepping-stone session per topic is gone. A third resolve is also stable (no unbounded growth), and a *different* topic under the same user still gets its own key (`…:lark_thread:OLD:lark_thread:NEW`), so the guard did not introduce false idempotency.

### Blocking #2 — `/new` inside a topic: FIXED (as a consequence)

With the transform idempotent, `ctx.sessionKey` (from the pre-queue resolve) is now the same key prompts run on, so `handleNewCommand` resets the session the user is actually talking to. The test at `lark.test.ts` asserting `resetBindingSession(..., "open_id:ou_user_1:lark_thread:mid-1")` is now a correct assertion rather than a codified bug.

### Blocking #3 — shared groups losing serialization / leaking sessions: FIXED

`buildGroupSessionKey` short-circuits `contextMode === "shared"` to `chat:${routeKey}` **before** any conversation handling, so the pre-queue resolve returns the stable chat key, `queueKey` stays `${bindingId}:chat:${routeKey}` and shared-turn serialization is preserved; no per-message `lark_thread:*` participant rows are created. `resolveOpenGroupBinding` inherits the fix through the same helper. The adapter test was correctly flipped to "keeps a shared group on its chat session when a topic candidate is present".

### Non-blocking items

- **Shared no-@ buffering regression + per-message RPC — fixed.** The cached-shared branch now buffers and returns *before* the resolve (RPC-free, matching pre-PR), and the cache-cold path buffers after resolving. Both restore the discussion-context feature that the first revision silently dropped for quote-replies.
- **`/mode` allocating an unused topic session — fixed** (passes `undefined`), with a test pinning it.
- **Cross-repo Sicore ordering — now closed on this side regardless of MR 770.** The runtime returns silently on `conversationExistingOnly` *before* emitting the access hint, so even if the frontend checks authorization before topic-session existence and answers denied, an unauthorized member's quote-reply can no longer be answered with an access prompt. Still worth confirming the ordering in MR 770 for session-reuse correctness, but the spam symptom is no longer reachable from here.

### Beyond what was asked

Two fixes the author found independently and which I'd have raised on a second pass:
- **Quote-reply vs Topic disambiguation**: `rootMessageId = threadId ? eventRootMessageId : messageId`. Feishu sets `root_id` on ordinary quote replies too, so without the `thread_id` gate an explicit `@bot` on a quoted message would have hijacked the quoted message's root as a conversation key.
- **Test isolation**: `resetLarkBindingQueuesForTest` now clears `groupModeCache` and `discussionBuffers` — the mode cache leaking across tests would have made the new shared/personal cases order-dependent.

### Verification performed

- Chained real-handler idempotency run (above), plus cross-topic and third-resolve cases.
- `lark.test.ts` + `adapter-rpc.test.ts` run locally on the PR head: **301 passed**.
- Delivery-path audit: every reply site reachable once the topic path is known (queue-full, `/new` reject, missing-owner, persist error, typing card, background, final card, images) carries `replyInThread`. The sites that don't (`access-denied`, `mode-unbound`, personal/pair/apikey) all execute before `contextMode` is known or on the p2p route — correct as-is.
- Docs cover the rollout flag, the Personal/Team split, the sensitive `im:message.group_msg` permission, and the acceptance flow.

One inline nit below (latent, unreachable today). Deploy remains paired with Sicore MR 770; the feature stays default-off until `SICLAW_LARK_THREAD_MODE` / `thread_mode: "group"` is set, and the default-off path is unchanged.

Approved — this supersedes my earlier CHANGES_REQUESTED.

PR #458, opened 2026-07-30
feat(channels): tell a refused personal-chat sender what to do next

jacoblee-io:
## Summary

An admission tier that gates a personal bot is only usable if the refusal explains the next step. With a single generic "no access" line the sender cannot self-serve, so the tier is effectively broken — and worse, the previous code recognised exactly two tier spellings, so any other value (including one the frontend introduces after this build ships) fell to a branch that only logged: **the sender's message vanished with no reply at all and the bot looked dead**, silently on both sides.

The frontend keeps the whole admission decision; the runtime's gate is still only "did a binding come back". What changes is what the sender is told, and how.

Contract and rationale: `docs/design/2026-07-29-personal-access-denial.md`.

## What the sender now sees

| Situation | Reply |
|---|---|
| Not linked yet | ❌ one line + **card with a "Link account" button** (footnote: opens once, valid N minutes) |
| Linked, can request access | ❌ one line + **card with a "Request access" button** |
| Linked, self-service closed | ❌ one line, **no link, no button** — "ask the owner" |
| Gated tier, frontend sent no reason | generic notice, **never silence** |
| Refused in a **group** | refusal + "DM me to get access" — **no URL in the room** |
| `/apikey` refused | ❌ one line about *getting a key*, then how to resume |

## Design points worth reviewing

- **Single-use links go out as cards with an `open_url` button, not as bare URLs.** Not cosmetic: a URL in a text message gets unfurled by the client, and an automated fetch of a one-time token can burn the sender's only chance to use it. On a button it is fetched only when a human taps. Card creation failure degrades to text — losing the card must never mean losing the link.
- **A group refusal carries no URL.** The room sees it, the real next step differs per sender, and a one-time link posted there could be opened by anyone to bind the sender's chat identity to their own account. The group says "DM me"; the DM resolves the actual step and can deliver the link safely. A group-only channel (no personal bot) keeps the console URL as its last resort — no DM would answer.
- **Copy names the action the sender can take.** A configured console URL is their *own* authorization page, so the copy says to open it; "ask an admin" appears only when there is no URL, because an admin cannot link someone else's chat account.
- **`reason` is the contract; `message` is a fallback.** Templates are keyed on the reason so both channel locales are served — the channel locale is a runtime property (derived from the app domain) the frontend does not know, and this is already how the group path works. `message` may embed a URL, so that path does not append `actionUrl` as well.
- **Durations are derived from `expiresAtMs`, never hard-coded**, and floor rather than round: rounding up overstates a single-use link's life by up to 30s, so a sender following "within 2 minutes" at 1m50s finds it gone. An already-expired link is withheld and replaced by "resend to get a fresh one" — the frontend mints one per message, so resending is the real recovery.
- **A link is only offered for a reason that has a self-service step** (one predicate shared by the text renderer, the card dispatch, and the `/apikey` card builder). "Ask the owner" must not carry a "click within N minutes" line or a generic button pointing nowhere.
- **Tier checks ask "is this open?"** rather than comparing against one legacy literal. Widening the union while leaving those comparisons behind told a bot on a newer gated tier "this bot is open, no PAIR needed" and threw the sender's pairing code away. The Portal-side resolver accepts the same alias set, or a `public` bot never auto-binds, comes back with no reason, and the sender is answered with silence.

### Untrusted input

Every field of `denied` comes from the frontend:

- `reason` indexes a **Map**, not an object literal. An object lookup walks the prototype chain, so `reason: "toString"` returned a Function that rendered as source or serialized to `{}` — which the platform rejects, silently dropping the reply.
- Only the frontend's **prose** is length-capped, never a URL or our own lines: truncating a link guarantees a dead one, and the platform's text limit sits far above anything rendered here.
- `expiresAtMs` is rejected when non-numeric, `<= 0`, or out of `Date` range. That last case is load-bearing: `Intl` throws `RangeError` past ±8.64e15 while `Number.isFinite` passes such a value, and the throw would replace the sender's only path forward with a generic error.
- An explicit refusal this build cannot render is **still a refusal**: the silent branch is gated on there being no `denied` at all, not merely on an open tier.

User-facing copy no longer names the system that holds the authorization — the sender does not need to know, and it kept internal naming out of a public repo.

## Test Plan

- [x] `npm test` — 235 files / 4656 passed. New coverage: 3 reasons × link/no-link × both locales; unknown reason → prose fallback with no duplicated link; prototype-chain keys (`toString`/`constructor`/`__proto__`/`hasOwnProperty`); expired link withheld; validity never overstated; link never truncated; card structure (no `note`, no body-level `button`, button under `column_set`) with text fallback when CardKit is unavailable; `/apikey` exactly one ❌ plus resume line; group refusal contains no URL when a personal bot exists, and keeps it when there is none; never-silent on gated and unrenderable-reason paths; re-resolution per message (no caching).
- [x] Deployed to an internal test environment and verified end to end in Feishu: the `/apikey` refusal copy and the group DM redirect render as intended.
- [x] Manually confirmed in Feishu on `v0.2.7-card5` (internal test environment): `/apikey` delivers the pickup card with its button, an `/apikey` refusal renders one ❌ plus the self-service button and the resume line, and a group refusal points the sender at the private chat with no URL in the room. Worth recording why this needed a second attempt: the first card shape was rejected by CardKit (a `note` element and a body-level `button`, neither used anywhere else here) and silently degraded to text — the fallback protects the link but hides a total card failure, so the create path now logs the API `code`/`msg` and tests assert the accepted structure.

## Notes for reviewers

The delivery fallback is deliberate but it does hide failures: when a card cannot be created the sender still gets the link as text, which is why the first bad card shape looked like "the feature didn't ship". The create failure now logs `code`/`msg`, and tests assert the accepted card structure so a regression fails there instead of quietly falling back in production.

Follow-up, deliberately not in this PR: ~30 code comments elsewhere in the tree describe upstream behaviour by naming the internal system rather than the contract this repo sees. Cleaning them touches a dozen files unrelated to this feature, so it belongs in its own change.


src/gateway/channels/lark.ts
```diff
@@ -1652,6 +1946,92 @@ function formatApiKeyIssueReply(result: PersonalApiKeyIssueResult, locale: LarkL
...
+
+/** `/apikey` refusal in text form: one ❌ line, the link, then how to resume. */
+function formatApiKeyDenialReply(denied: PersonalAccessDenied, locale: LarkLocale): string | null {
+  const copy = denied.reason ? API_KEY_DENIAL_COPY_BY_LOCALE[locale].get(denied.reason) : undefined;
+  if (!copy) return null;
+  const lines = [copy.lead];
+  const remaining = minutesUntil(denied.expiresAtMs);
+  if (denied.actionUrl && remaining !== "expired" && offersSelfService(denied.reason, locale)) {
+    lines.push(PERSONAL_DENIAL_LINK_HINT_BY_LOCALE[locale](remaining), denied.actionUrl);
+  } else if (denied.actionUrl) {
```
chent1996: **Finding 1** — this `else if` sits outside the `offersSelfService` guard, so `access_denied` carrying a *live* `actionUrl` renders "之前的链接已过期，请再发一条消息获取新链接" under "请联系负责人". The link is not expired, and the resend it asks for refuses identically.

`formatPersonalDenialReply` gets this right by nesting the expired check inside the guard. Suggest the same shape here:

```js
if (denied.actionUrl && offersSelfService(denied.reason, locale)) {
  const remaining = minutesUntil(denied.expiresAtMs);
  if (remaining === "expired") lines.push(PERSONAL_DENIAL_LINK_EXPIRED_BY_LOCALE[locale]);
  else lines.push(PERSONAL_DENIAL_LINK_HINT_BY_LOCALE[locale](remaining), denied.actionUrl);
}
```

A reason with no self-service step should say nothing about links at all. The existing `access_denied` `/apikey` test passes no `actionUrl`, so this branch is uncovered.

src/gateway/channels/lark.ts
```diff
@@ -1459,18 +1623,154 @@ async function resolveQueuedBinding(
...
+  // Only the explicit not-linked reason claims "you haven't linked yet". Every other value —
+  // including one this build has never seen — gets the generic line: it is never wrong, whereas
+  // telling an already-linked sender to go link sends them round a loop with no exit.
+  const base = denied.reason === "unbound"
+    ? GROUP_ACCESS_UNBOUND_NOTICE_BY_LOCALE[locale]
+    : GROUP_ACCESS_DENIED_NOTICE_BY_LOCALE[locale];
+  // Send them to the private chat, NOT to a URL in the group. The DM resolves the sender's actual
+  // next step and can deliver a single-use link, which must never be posted where anyone in the
+  // room could open it and bind the sender's chat identity to their own account.
+  if (hasPersonalBot) return `${base}\n${GROUP_ACCESS_DM_HINT_BY_LOCALE[locale]}`;
```
chent1996, line 1994: **Finding 3** — `hasPersonalBot` is `Boolean(personalBot)`, but the DM only resolves anything when the personal bot is itself gated. With `access_mode: "open"` the DM binds the sender on contact and never offers an authorization step, so this line sends them somewhere that cannot help *and* removes the console URL that previously could.

```js
formatGroupAccessDeniedReply(binding, locale, Boolean(personalBot) && !isOpenAccessTier(personalBot!.access_mode))
```

would keep the URL in that configuration. Note also that the one-time-link argument in the docstring above cannot apply on this path — `ChannelAccessDenied.authorizeUrl` is the shareable console address by construction, and the type separation is what guarantees a single-use link never gets here.

src/gateway/channels/lark.ts
```diff
@@ -112,6 +134,88 @@ const API_KEY_BUSY_NOTICE_BY_LOCALE = {
...
+// error page with no hint that anything can be done about it.
+const PERSONAL_DENIAL_LINK_EXPIRED_BY_LOCALE: Record<LarkLocale, string> = {
+  "zh-CN": "之前的链接已过期，请再发一条消息获取新链接。",
+  "en-US": "The previous link has expired — send another message to get a fresh one.",
+};
+// Every part of a denial reply (`message`, `actionUrl`, `reason`) comes from the frontend and is
+// unbounded. The platform rejects oversized text outright and `replyToLark` cannot un-reject it,
+// so a single pathological field would cost the sender their whole reply — cap the RENDERED
+// result, not just one field.
+const PERSONAL_DENIAL_MESSAGE_MAX_CHARS = 1000;
```
chent1996, line 218: Nit: this comment says "cap the RENDERED result, not just one field", but `truncateDenialProse` caps only `denied.message` — and its own docstring says the opposite, "Applied to prose ONLY — never to a URL or to our own lines". The code behaviour is the correct one; this comment looks like a leftover from the reversed design and will mislead the next person who audits the cap.

src/portal/adapter.ts
```diff
@@ -441,6 +441,16 @@ async function selectPersonalChannel(
   };
 }
 
+/**
+ * True for the tiers that admit anyone. `public` is the modern spelling of `open` — both must be
+ * accepted here or a `public` personal bot never auto-binds, the runtime sees a null binding with
+ * no refusal reason, treats the tier as open, and answers the sender with silence.
+ */
+function isOpenPersonalTier(accessMode: unknown): boolean {
```
chent1996: **Finding 4** — identical to `isOpenAccessTier` in `src/gateway/channels/lark.ts`. Two copies of one wire contract across the two layers that have to agree on it; divergence reproduces the failure this PR fixes (runtime says open, Portal refuses to auto-bind, sender gets silence). `adapter.ts` already imports from `../gateway/db.js`, so a single export from `channel-manager.ts` alongside `PersonalAccessDenied` is layering-legal.

Also, `access_mode` at line 419 is still typed `"open" | "sicore_authorized"` — the type declares the `public` branch dead while the comment declares it load-bearing.

chent1996 (commented):
Reviewed against `main` @ `8b17d499` (the PR's merge-base — no stale-base risk). CI green, `npx tsc --noEmit` clean, and the four touched suites pass locally (380 tests). The design doc is genuinely good: the untrusted-input reasoning (Map-not-object for `reason`, `Date`-range rejection before `Intl`, floor-not-round on durations, cap the prose but never the URL) is the kind of thing that normally only shows up after a production incident, and the never-silent regression this fixes was a real hole.

Three reachable copy defects and one architectural drift risk below, all verified by execution rather than by reading. None of them corrupt state or leak a credential, so nothing here is a hard blocker — but #1 reproduces, inside this PR, the exact "dead end with no exit" failure mode the PR exists to eliminate, and it is a two-line fix. Marking COMMENT rather than APPROVE also matches the still-unchecked manual-Feishu box in your own test plan.

---

## 1. `/apikey` tells the sender their live link expired  (confirmed by execution)

`formatApiKeyDenialReply` puts the `else if (denied.actionUrl)` branch **outside** the `offersSelfService` guard, so `access_denied` + a still-live `actionUrl` renders:

```
❌ 你没有该 Agent 的使用权限，且未开放自助申请，请联系负责人。
之前的链接已过期，请再发一条消息获取新链接。
```

The link is not expired — it is ten minutes from expiry and simply not offered. Worse, the second line instructs a resend, and the resend refuses identically. That is the loop the design doc calls out by name ("telling an already-linked sender to go link loops them with no exit").

Reproduced by adding one case to `lark.test.ts`:

```js
issuePersonalApiKeyMock.mockResolvedValue({
  success: false,
  denied: { reason: "access_denied", actionUrl: "https://x.example/live", expiresAtMs: Date.now() + 600_000 },
});
// → reply contains "已过期"
```

Is `access_denied` + `actionUrl` a real input? Your own code says yes — `offersSelfService`'s docstring exists precisely because "an actionUrl arriving on it must not be rendered as 'click to continue' directly under 'ask the owner'". `formatPersonalDenialReply` handles it correctly (the expired branch sits *inside* the `offersSelfService` guard); the `/apikey` twin does not. The doc claims `offersSelfService` is "the single predicate ... used by the text renderer, the card dispatch, and the `/apikey` card builder alike" — the `/apikey` **text** renderer is the one place that escapes it.

Fix is to mirror the sibling: move the `else if` inside, so a reason with no self-service step says nothing about links at all. Existing coverage misses it because the `/apikey` `access_denied` test passes no `actionUrl`.

## 2. Group-only path now hands an already-linked sender the linking page  (confirmed by execution)

The old renderer withheld the URL for `reason: "denied"` deliberately:

```js
if (denied.reason === "denied") return GROUP_ACCESS_DENIED_NOTICE_BY_LOCALE[locale];
const base = GROUP_ACCESS_UNBOUND_NOTICE_BY_LOCALE[locale];
return denied.authorizeUrl ? `${base}\n${denied.authorizeUrl}` : base;
```

The rewrite splits only the *base line* on reason and then appends the hint plus URL for **both** reasons. So a group-only channel with `reason: "denied"` (sender is linked, lacks agent access) now reads "请打开下面的链接完成授权" pointing at the console page they already used. Verified: with `{walled: true, reason: "denied", authorizeUrl: "…"}` and no `personal_bot`, the URL is in the reply where it previously was not.

The generic base line is indeed "never wrong", but `GROUP_ACCESS_SELF_SERVE_HINT` is not generic — it is an instruction. If `authorizeUrl` is an access-request page for a linked user, fine and worth saying so in the doc; if it is the account-linking page, this is the same loop as #1 on a different path. This is a behavior change the PR body does not mention and no test pins.

## 3. "DM me to get access" is a dead end when the personal bot is open  (confirmed by execution)

`Boolean(personalBot)` is too weak a predicate for "the DM can resolve this". With a gated group and `personal_bot.access_mode: "open"`, the DM binds the sender immediately, emits no `denied`, and offers no authorization step whatsoever — meanwhile the previously-working console URL has been removed from the group reply. Verified: gated group + open personal bot yields "私聊我" and no URL.

Worth noting the stated rationale does not quite hold here either. The security argument ("a one-time link posted there could be opened by anyone") is about `PersonalAccessDenied.actionUrl`, which the type separation already prevents from ever reaching this renderer — the group path only ever carries `ChannelAccessDenied.authorizeUrl`, documented as a shareable console address. So dropping the URL is a UX judgement call, not a leak fix, and the judgement inverts when the DM has nothing to offer. Suggest gating on `personalBot && !isOpenAccessTier(personalBot.access_mode)` and keeping the URL otherwise. Current tests only cover the `sicore_authorized` personal bot.

## 4. Two tier predicates, one contract

`isOpenAccessTier` (`lark.ts:1640`) and `isOpenPersonalTier` (`adapter.ts:449`) are character-for-character the same normalization of the same wire field, in two layers that must agree. If they ever diverge you get exactly the class of bug this PR is fixing: the runtime says "this bot is open" while the Portal refuses to auto-bind, and the sender is answered with silence. `adapter.ts` already imports from `../gateway/db.js` and `../gateway/dialect-helpers.js`, so exporting one helper from `channel-manager.ts` (where `PersonalAccessDenied` now lives) is layering-legal and would make the contract single-sourced.

Related: `adapter.ts:419` still types `access_mode?: "open" | "sicore_authorized"` while the function right below it handles `public`. The type says the branch is dead; the comment says it is load-bearing. Widen it, or the next reader deletes the `public` case.

---

### Smaller things

- **`PERSONAL_DENIAL_MESSAGE_MAX_CHARS` comment contradicts the code.** It says "cap the RENDERED result, not just one field", but `truncateDenialProse` caps prose only — and its own docstring says the opposite ("Applied to prose ONLY — never to a URL or to our own lines"). The code is right; the constant's comment describes a design that was reversed. One of the two should go.
- **The success card silently drops "send `/apikey status` to check the key's own expiry."** `formatApiKeyIssueCardBody` justifies dropping the URL and the validity line (they become the button and the footnote) — but the `status` hint is neither, and the card is now the primary delivery path, so that affordance from #454 stops being discoverable.
- **`actionUrl` scheme is unvalidated** before it lands on a Feishu `open_url` button. Everything else in `denied` is treated as untrusted — prototype-chain keys, length, `Date` range — so an `http(s)`-only check would be consistent with the standard the rest of this PR sets. Feishu resolves non-http schemes as deeplinks.
- **`docs/design/` is updated but `CLAUDE.md`'s Change Impact Matrix is not.** The `/apikey` row still describes a text-only reply; this PR adds a card delivery path and a new cross-layer `denied` contract, and introduces an invariant worth pinning where it gets auto-loaded: *a single-use link reaches a user only through `deliverSingleUseLink`, never through a group renderer*. Right now that is enforced by type separation plus convention, and the matrix is what a future change actually reads.
- **Merge order with #457.** A trial merge conflicts in `src/gateway/channels/lark.ts` (both rework the same personal/group dispatch region). No action needed here beyond whoever lands second rebasing — flagging it so it is not discovered at merge time.

### Verified clean

- All `access_mode` comparisons are converted; no `=== "open"` / `!== "sicore_authorized"` literals remain outside the two predicates.
- Both `resolvePersonalBinding` call sites handle the new `{binding, denied}` shape, including the queued re-resolve.
- Standalone mode is unaffected: the in-repo Portal never emits `walled` and always writes `access_mode: "open"`, so the new gated branches are unreachable there and a `denied`-less frontend degrades to the previous behavior.
- The `/apikey` delivery accounting is sound — `cardDelivered ?? await replyToLark(...)` keeps the post-rotation escalation intact when both card and text fail, and `carriesCommittedRotation` is correctly false on every refusal path.
- The PAIR gate fix is real: comparing against the one legacy literal did discard pairing codes on any newer gated tier.

jacoblee-io:
All four reproduced, all fixed in `167786ab`. #1 is the one that stings: it recreated, inside the change, the exact dead-end loop the design doc calls out by name — and your diagnosis of *why* is right, the `/apikey` text renderer was the single site that escaped the predicate the doc claimed was shared everywhere.

**1 — `/apikey` tells the sender their live link expired.** Confirmed. The expired notice was chained onto `actionUrl` alone, outside the guard, so `access_denied` + a live link produced "已过期 … 请再发一条消息", and the resend refuses identically. It now sits **inside** the guard: a reason with no self-service step says nothing about links at all. Your repro case is in the suite.

**2 — group-only path hands an already-linked sender the linking page.** Confirmed. Splitting only the base line on reason while appending the hint + URL for both was the defect; the old code withheld the URL for `denied` deliberately and I dropped that. The instruction + URL is now offered **only** on the not-linked path — for a linked sender the linking page is somewhere they have already been. Pinned by a test.

**3 — "DM me" is a dead end when the personal bot is open.** Confirmed, and you're right that my rationale was overstated. The security argument is about `PersonalAccessDenied.actionUrl`, which type separation already keeps out of this renderer; the group path only ever carries the shareable `authorizeUrl`, so preferring the DM is a **UX judgement, not a leak fix**. I've corrected that claim in the design doc rather than leaving a justification that doesn't hold. The gate is now `personalBot && !isOpenAccessTier(...)` — exactly your suggestion — with a test for the open-bot case.

**4 — two tier predicates, one contract.** Fixed as suggested: `isOpenAccessTier` is exported from `channel-manager.ts` and imported by `portal/adapter.ts`. Deleting the duplicate is worth more than the layering nit, since a drift there produces precisely the silence this change exists to remove. The stale `access_mode` type on the adapter is widened too — you were right that the next reader would delete the `public` case on the type's authority.

**Smaller things** — all taken:
- `actionUrl` scheme is now validated: `rendersActionLink` requires `http(s)`, so a `javascript:`/deeplink value is withheld exactly like a missing one. Folding it into the same predicate keeps one gate rather than two.
- The success card keeps the `/apikey status` hint. (I had written that in the commit message before actually making the change — caught it on re-read and amended, rather than shipping a message that described work I hadn't done.)
- The cap constant's comment described the design that review reversed; it now matches `truncateDenialProse`.
- `CLAUDE.md` gains the matrix row, including the invariant you named: *a single-use link reaches a user only through `deliverSingleUseLink`, never through a group renderer.* Agreed that convention plus type separation isn't enough on its own — the matrix is what a future change actually reads.
- Merge order with #457: noted, whoever lands second rebases. Thanks for trial-merging.

Verification on `167786ab`: `tsc --noEmit` clean, 235 files / 4660 tests pass, and the lark suite run repeatedly to rule out ordering flakes. New coverage for each of the above: `access_denied` + live link says nothing about links; non-http(s) `actionUrl` withheld; group-only + `denied` gets no URL; group + open personal bot keeps the URL.

Still unchecked in the test plan: manual Feishu confirmation of the **card** itself. The first card shape was rejected by CardKit (a `note` element and a body-level `button`) and silently degraded to text, which is exactly why the fallback is a double-edged design — it guarantees the link survives but hides a total card failure. The create path now logs the API `code`/`msg`, and tests assert the accepted structure so a regression fails there instead of quietly falling back in production. `v0.2.7-card4` is deployed to the internal test environment for that check.

LikiosSedo:
Re-reviewed the current head `ede46409` against `main` at `36dce74a` after a fresh fetch. The previous four findings are fixed, the PR is mergeable, all three GitHub checks are green, and the four focused suites pass locally (395 tests) with `npx tsc --noEmit` clean.

I still found three issues on this exact head:

### [P1] An unknown reason can still expose a one-time URL as plain text

The dispatch only uses `deliverSingleUseLink` when `reason` matches a local template ([`lark.ts:1136-1146`](https://github.com/scitix/siclaw/blob/ede4640902d205d051d890d1f03e945d2c290757/src/gateway/channels/lark.ts#L1136-L1146)). For a future/unknown reason it falls back to `denied.message`, and the existing test explicitly puts `actionUrl` inside that message and expects the raw URL in the text reply ([`lark.test.ts:3330-3342`](https://github.com/scitix/siclaw/blob/ede4640902d205d051d890d1f03e945d2c290757/src/gateway/channels/lark.test.ts#L3330-L3342)).

That bypasses the central safety property of this PR: a client unfurl can fetch and consume the single-use token before the sender taps it. This is also the exact frontend-first version-skew case the design says must remain safe. Please do not send a fallback message containing a one-time URL as text; either extend the contract with safe action/button semantics or fall back to a generic notice with no URL.

### [P2] A malformed `denied.message` recreates the silent-failure path

Every `denied` field is documented as untrusted, but [`denied.message?.trim()`](https://github.com/scitix/siclaw/blob/ede4640902d205d051d890d1f03e945d2c290757/src/gateway/channels/lark.ts#L1902) assumes that a present value is a string. I reproduced this with:

```ts
denied: { reason: "some_future_reason", message: { text: "not a string" } }
```

`handleLarkMessage` rejects with `TypeError: denied.message?.trim is not a function`; the detached event wrapper only logs the exception, so the sender gets no reply — the exact “bot looks dead” failure this PR is meant to eliminate. Please narrow/normalize the field at the RPC boundary (or before formatting) and add the malformed-value regression case.

### [P2] A link that expires at the delivery boundary is still sent

[`deliverSingleUseLink`](https://github.com/scitix/siclaw/blob/ede4640902d205d051d890d1f03e945d2c290757/src/gateway/channels/lark.ts#L1875-L1881) recomputes the remaining lifetime, but when the result is `"expired"` it only changes the note to the no-duration form and still calls CardKit with the dead URL. I reproduced the boundary with `Date.now()` returning `1000, 1000, 1002` for `expiresAtMs: 1001`; the card was still created and contained the URL.

If the link is expired at the actual send boundary, this helper should not call CardKit, and the fallback used there must also be recomputed/guaranteed not to contain the stale URL.

Given the first issue violates the PR's main single-use-link invariant, I would not approve this head yet. The manual Feishu confirmation of the final card shape also remains unchecked in the PR test plan.

jacoblee-io:
All three reproduced and fixed in `24676c59`. P1 is the one that matters: it defeated this PR's central property in exactly the version-skew case the design says must stay safe, and my own test had encoded the bad behaviour as the contract.

**[P1] unknown reason exposed a one-time URL as text.** Confirmed. A structured, usable `actionUrl` now rides a button for **any** reason, with a neutral label when the reason is unfamiliar. Only a reason *known* to carry no self-service step stays link-less — unknown is not the same as known-linkless, and withholding an unknown reason's link would strand the sender, so a generic notice was not the right answer either. The test that asserted the raw URL in text is replaced by one asserting it is on the button and absent from the text payload.

Your framing also made me fix an **overclaim in the doc**, which is arguably the more useful outcome: the text form is a *degradation* (card creation failed) and it does carry the URL. So the invariant is "a structured `actionUrl` is never sent as text **while a card is possible**", not "a URL never appears in text". The doc said the latter. A prose `message` that embeds its own link remains passed through verbatim — we can neither label nor reliably extract it — and that is now stated rather than implied.

**[P2] non-string `message` recreated the silent path.** Confirmed, and your repro is in the suite. Narrowed at the **RPC boundary** as you suggested (`normalizeDenied`), so no renderer has to defend itself — and wrong-typed fields are *dropped*, not coerced, since a stringified object would otherwise be shown to the user as copy. Worth noting for anyone reading the diff: `lark.test.ts` mocks `channel-manager`, so the boundary fix alone would have left that suite unable to see the property. The renderers therefore keep a `typeof` guard as well, and the boundary itself is tested directly in `channel-manager.test.ts` (wrong types dropped; non-object `denied` ignored).

**[P2] link lapsing at the delivery boundary.** Confirmed. `deliverSingleUseLink` now re-checks and, when expired, calls neither CardKit nor the URL-carrying text form — it sends the resend instruction, which is the only thing that helps since the frontend mints a fresh link on the next message. Pinned with a test that advances the clock between compose and send.

One incidental find while fixing P2: a test fixture used a fixed `2025-07-28` expiry that had quietly become the past, so the new boundary check correctly withheld the link and the **fixture** was what broke. Made relative, and the exact-format assertion moved to the dedicated timestamp cases where it belongs.

Verification on `24676c59`: `tsc --noEmit` clean, 235 files / 4677 tests pass, lark suite run repeatedly for ordering flakes. Rebased onto `36dce74a` beforehand — the #457 conflict was in the group walled-sender path; both behaviours are preserved (its `conversationExistingOnly` early return for no-@ topic follow-ups, and this PR's DM-capability argument), and #457's own 12 new tests pass on the rebased head.

Manual Feishu confirmation of the card shape is still the one unchecked box; `v0.2.7-card4` is on the internal test environment, and I will not tick it until it is actually observed.

jacoblee-io:
Manual Feishu verification passed on `v0.2.7-card5` (= this head, `24676c59`) in the internal test environment, so the last unchecked box in the test plan is now ticked:

- `/apikey` with access → pickup card with its button; no bare URL in the text body.
- `/apikey` refused → one ❌ about getting a key, the self-service button, and the resume line.
- Group refusal → sender pointed at the private chat, no URL in the room.
- No-@ topic follow-up from a walled sender → still silent, i.e. #457's behaviour survived the rebase.

Two caveats I'd rather state than leave implied. `FRONTEND_URL` in that environment is still `localhost:3000`, so the buttons render correctly but the links themselves are dead there — what this run verifies is card/button rendering and the no-URL-in-group property, not the end-to-end link-and-authorize flow. And the delivery fallback remains double-edged by design: it guarantees the link survives a CardKit failure, but it also means a broken card shape looks like "the feature didn't ship" rather than an error. That is why the create path now logs the API `code`/`msg` and the tests assert the accepted card structure.

CI is green on this head; 235 files / 4677 tests pass locally with `tsc --noEmit` clean. Nothing outstanding from my side — it needs a review pass on the second round of fixes (`24676c59`).

LikiosSedo:
Re-reviewed the current head `24676c59` against `main` at `36dce74a`. The previous three findings are fixed, the PR is mergeable, all three GitHub checks are green, the four focused suites pass locally (400 tests), and `npx tsc --noEmit` is clean.

I still found one issue on this exact head:

### [P1] Keep unknown-reason action links reachable outside the happy card path

The new card dispatch is still gated by `deniedReply` ([`lark.ts:1130-1149`](https://github.com/scitix/siclaw/blob/24676c59a55a422b5d516ebfbcb70a71bae40d19/src/gateway/channels/lark.ts#L1130-L1149)), while the formatter returns `null` without a known template or frontend `message`, and deliberately does not append `actionUrl` for an unknown reason ([`lark.ts:1926-1944`](https://github.com/scitix/siclaw/blob/24676c59a55a422b5d516ebfbcb70a71bae40d19/src/gateway/channels/lark.ts#L1926-L1944)). That leaves two ways to strand the sender:

1. `denied: { reason: "quota_exceeded", actionUrl, expiresAtMs }` with no `message` never attempts CardKit at all; it falls through to the generic gated text, so the structured link is lost.
2. The same unknown reason with `message: "Quota exceeded."` does attempt the card, but if CardKit creation fails, `deliverSingleUseLink` falls back to that prose-only `deniedReply`; the structured `actionUrl` is again lost.

I reproduced both with focused regression tests: the first expected `card.create` to be called and it was not; the second forced `card.create` to reject and received only `{"text":"Quota exceeded."}` with no URL. The existing unknown-reason test covers only the happy case where both `message` and CardKit are available.

Please decide action-link delivery independently of whether prose was supplied, and construct the text degradation so a usable structured `actionUrl` is present exactly once. This is the PR's stated compatibility and fallback invariant, so I would not approve this head yet.

jacoblee-io:
Confirmed both, fixed in `e36f3049`. You found the same defect I had just introduced while fixing your previous P1 — worth naming the root cause, because the two symptoms you reproduced share one:

`rendersActionLink` used the **allow-list** (`offersSelfService`, "the reason has a button label") while the card dispatch used the **deny-list** ("the reason is not known-linkless"). An unknown reason passed one gate and failed the other, so the formatter dropped the link the dispatch was willing to send. It is the deny-list in both places now, and `offersSelfService` is left doing the single job it should — choosing the button label. That also means the doc's "one predicate" claim is finally true rather than aspirational.

On top of that, delivery no longer depends on whether prose arrived:

1. **No `message`** — the formatter returned null, so the caller fell back to its own generic text and CardKit was never attempted. Now the generic notice becomes the body when there is neither template nor prose, so a usable link always has something to ride.
2. **Card creation fails** — the degradation was the prose alone. The link is now appended to the text form too, **exactly once**: duplication is prevented by checking the prose about to be sent rather than skipping the append wholesale, since the frontend's message may embed its own copy. Your `Quota exceeded.` case and a prose-already-embeds-the-URL case are both pinned.

Verification on `e36f3049`: `tsc --noEmit` clean, 235 files / 4680 tests pass, lark suite run repeatedly for ordering flakes.

Three rounds in, the pattern in my own errors is consistent enough to be worth stating: each fix narrowed one path and left a sibling path inconsistent, and the tests I wrote alongside encoded the narrowed behaviour as the contract. The predicate consolidation above is meant to remove the room for that — there is now one place to change rather than four that must agree.

LikiosSedo:
Re-reviewed the current head `e36f3049` against `main` at `36dce74a`. The previous two repros are fixed, the PR is mergeable, all three GitHub checks are green, the four focused suites pass locally (403 tests), `npx tsc --noEmit` is clean, and the merge tree / diff check are clean.

I found two remaining sibling-path issues on this exact head:

### [P1] The `/apikey` sibling still drops a future refusal's structured link

The normal personal-message path now treats an unknown reason as link-bearing, but `/apikey` still requires a locally known API-key copy in both renderers. [`buildApiKeyDenialCard`](https://github.com/scitix/siclaw/blob/e36f30490cac50f3f6a6a7b1cbd0ffe82fdc5bdc/src/gateway/channels/lark.ts#L2204-L2219) returns `null` when `API_KEY_DENIAL_COPY_BY_LOCALE` has no entry, and [`formatApiKeyDenialReply`](https://github.com/scitix/siclaw/blob/e36f30490cac50f3f6a6a7b1cbd0ffe82fdc5bdc/src/gateway/channels/lark.ts#L2225-L2227) does the same. `formatApiKeyIssueReply` then emits its generic `unknown error` text with no action URL.

I reproduced this with:

```ts
{ success: false, denied: {
  reason: "quota_exceeded",
  actionUrl: PICKUP,
  expiresAtMs: Date.now() + 60_000,
} }
```

With CardKit available, `card.create` was never called and the structured link was absent. This is the same frontend-first version-skew case the PR says must remain usable, and it contradicts the latest "one place to change rather than four" claim: `rendersActionLink` now accepts the refusal, but the two `/apikey` copy lookups veto it afterward. Please give that action-specific path a generic body/resume line and neutral button label too, or explicitly narrow the wire contract so an unknown reason cannot carry an actionable link there.

### [P2] The send-boundary expiry branch can still send an embedded stale URL

[`deliverSingleUseLink`](https://github.com/scitix/siclaw/blob/e36f30490cac50f3f6a6a7b1cbd0ffe82fdc5bdc/src/gateway/channels/lark.ts#L1901-L1908) correctly skips CardKit when the link lapses, but builds the expired reply from `card.body`. For an unknown reason, that body is the frontend prose and may already contain the exact `actionUrl`.

I reproduced the clock advancing between compose and send with `message: "Continue here: <ACTION_URL>"`. CardKit was correctly not called, but the text reply was:

```text
Continue here: <ACTION_URL>
之前的链接已过期，请再发一条消息获取新链接。
```

So the boundary invariant ("an expired link must reach neither the button nor the text") is still false for the embedded-URL case. The expired branch needs a URL-free body/fallback, not `card.body` verbatim.

One documentation cleanup follows from the same root: [`docs/design/...:140-143`](https://github.com/scitix/siclaw/blob/e36f30490cac50f3f6a6a7b1cbd0ffe82fdc5bdc/docs/design/2026-07-29-personal-access-denial.md#L140-L143) and the new `CLAUDE.md` matrix row still define `rendersActionLink` as "reason has a button label AND URL", which contradicts the deny-list semantics added earlier in the same document and in code.

Both focused regression tests fail on this head. Given the first still strands the sender in the shared `PersonalAccessDenied` future-reason case, I would not approve `e36f3049` yet.

jacoblee-io:
Both reproduced, fixed in `db555630`, and the doc/matrix inconsistency you spotted is corrected in the same pass.

**[P1] `/apikey` sibling vetoed the refusal after `rendersActionLink` accepted it.** Confirmed. Both `/apikey` renderers required locally known copy, so an unfamiliar reason passed the predicate and was then dropped by the copy lookup — the structured link lost on precisely the version-skew path. That path now has a generic lead + resume line and a neutral button label, resolved through one helper (`apiKeyDenialCopy`) shared by the text and card renderers. A refusal with neither known copy nor an actionable link still falls through to the frontend's `error`: there is genuinely nothing `/apikey`-specific to say, and inventing copy would be worse than passing the reason through.

**[P2] expired branch rebuilt from `card.body`.** Confirmed. For an unknown reason that body *is* the frontend prose, which may embed the URL being withheld — so the boundary invariant was false for the one case it was added to cover. It now swaps in a URL-free notice when the body carries the link.

**Docs.** You were right that the design doc and the new matrix row still defined `rendersActionLink` as "reason has a button label AND url" — the allow-list wording from before it became a deny-list. Both now state the deny-list and name `offersSelfService` as the label-picker only. That mattered more than it looks: the stale definition is what would let the next reader reintroduce exactly the veto in P1.

Verification on `db555630`: `tsc --noEmit` clean, 235 files / 4682 tests pass, lark suite run repeatedly. Your two repros are now regression cases.

Four rounds, and my failure mode has been identical every time: apply a rule to one path, leave a sibling inconsistent, and write a test that encodes the narrow behaviour as the contract. The consolidation into single predicates/helpers is the structural answer, but I'd rather state the pattern plainly than claim it is now impossible — if you see a fifth instance, that is a signal the abstraction is still not where it belongs.

LikiosSedo:
Re-reviewed the current head `db555630` against `main` at `36dce74a`. The two prior repros are fixed: `/apikey` now preserves an actionable unknown-reason link, the send-boundary expiry path removes an embedded stale URL, and the deny-list documentation is corrected. The PR is mergeable, all three GitHub checks are green, the four focused suites pass locally (405 tests), `npx tsc --noEmit` is clean, and the merge tree / diff check are clean.

I found two remaining contract gaps on this exact head:

### [P2] Keep a structured single-use URL out of the successful card body

The normal personal-message path passes the frontend's unknown-reason `message` directly to [`buildLinkActionCard`](https://github.com/scitix/siclaw/blob/db555630f047ae3ddb6453837fe9ad0e5ca3a7bb/src/gateway/channels/lark.ts#L1136-L1149). If that prose already embeds the exact structured `actionUrl`, the successful card contains the URL both in its Markdown body and in the `open_url` button.

I reproduced this with:

```ts
denied: {
  reason: "quota_exceeded",
  actionUrl: ACTION_URL,
  expiresAtMs: Date.now() + 600_000,
  message: `Continue here: ${ACTION_URL}`,
}
```

CardKit succeeded, but the card body was:

```text
Continue here: https://upstream.example/siclaw/agent-access/9f3a2b
仅可打开一次 · 9 分钟内有效
```

This contradicts [`buildLinkActionCard`'s stated invariant](https://github.com/scitix/siclaw/blob/db555630f047ae3ddb6453837fe9ad0e5ca3a7bb/src/gateway/channels/lark-card.ts#L287-L303) that the URL is not repeated in the card body, and gives the single-use URL a second rendering path outside the button. Please sanitize/substitute the body when it contains the exact `actionUrl`, using the same principle as the expired-link branch. This finding is the code-proven contract violation; it does not depend on assuming a particular live CardKit unfurl behavior.

### [P2] Preserve `denied.message` when `/apikey` has neither known copy nor an action link

For a future refusal such as:

```ts
{
  success: false,
  denied: {
    reason: "quota_exceeded",
    message: "Quota exceeded for this account.",
  },
}
```

`/apikey` currently replies with:

```text
❌ 领取 API Key 失败: 未知错误
```

[`PersonalAccessDenied.message`](https://github.com/scitix/siclaw/blob/db555630f047ae3ddb6453837fe9ad0e5ca3a7bb/src/gateway/channel-manager.ts#L103-L111) is explicitly the fallback when this runtime has no template for a reason. However, `apiKeyDenialCopy` returns no copy without an actionable link, and [`formatApiKeyIssueReply`](https://github.com/scitix/siclaw/blob/db555630f047ae3ddb6453837fe9ad0e5ca3a7bb/src/gateway/channels/lark.ts#L2136-L2146) then considers only `result.error`, discarding `denied.message`. Please use the bounded normalized `denied.message` before falling back to `result.error` / the generic unknown error.

Both focused regression tests fail on `db555630`. I would not approve this head yet because the first issue still violates the PR's central button-only invariant for a structured single-use link.

jacoblee-io:
Both fixed in `17b39378`, but they are not the same kind of finding and I want to separate them rather than fold both into "addressed".

**The `/apikey` one is a real defect and a good catch.** `denied.message` is defined as the fallback for a reason this build has no template for, and the failure path skipped straight to `error`, so a sender got `未知错误` in place of "Quota exceeded for this account." — a real explanation discarded for a generic one. Fixed: `message` is preferred over `error`, bounded like every other prose field, with your case as a regression test.

**The card-body one is a doc-vs-behavior inconsistency, not a breach of the button-only invariant.** The fact is right: prose that embeds the exact `actionUrl` was rendered in the body as well as on the button, and `buildLinkActionCard` claimed otherwise. Fixed — and deliberately fixed *inside that builder* rather than at the call site, because patching callers one at a time is exactly what let this class of inconsistency recur five rounds running. The guarantee now lives with the claim, so a future caller cannot reintroduce it.

Where I disagree is the severity, and I think the comment disagrees with itself:

> This finding is the code-proven contract violation; it does not depend on assuming a particular live CardKit unfurl behavior.

> I would not approve this head yet because the first issue still violates the PR's central button-only invariant for a structured single-use link.

Those cannot both hold. The button-only invariant exists for one reason — a bare URL in a **text** message gets unfurled and an automated fetch can burn a one-time token. Inside an interactive card the platform does not link-preview, which is why the first sentence correctly disclaims the unfurl assumption. With that assumption withdrawn, what remains is a duplicated URL in a card and a stale comment: worth fixing for tidiness and for the reason above, not a credential exposure and not grounds to block.

For calibration across the whole sequence: rounds 1–3 found genuine user-visible bugs — a silent drop, a reply the platform rejected outright, a live link reported as expired, an unknown reason's link lost. Most were mine to begin with, several I introduced while fixing the previous round, and I have said so each time. The last two rounds have moved toward internal consistency of my own documentation, with the harm model weakening each time while the blocking language stayed constant. Both are still worth fixing and both are fixed; I would just rather the blocking bar tracked user impact.

Verification on `17b39378`: `tsc --noEmit` clean, 235 files / 4684 tests pass, lark suite run repeatedly. Every finding from all five rounds is fixed, CI is green, the branch is rebased on `36dce74a`, and the manual Feishu check is done. I believe this head is ready — if you see remaining user-impacting behaviour, I will fix it immediately; if the outstanding items are consistency-level, I would ask that they not hold the merge.

LikiosSedo (approved):
Re-reviewed `17b39378` against `main` at `36dce74a`. All previously reported repros are fixed, no new actionable issues found, the PR is mergeable, all three GitHub checks are green, the four focused suites pass (407 tests), the full suite passes (235 files / 4684 tests, 2 skipped), `npx tsc --noEmit` is clean, and the synthetic merge tree / diff check are clean. Approved.

PR #459, opened 2026-07-31
fix(agentbox): refresh control-plane model config

LikiosSedo:
## Summary
- refresh the active AgentBox model whenever the control plane supplies `modelConfig`, so compat-only edits take effect on the next prompt
- report the number of empty-response retries that actually ran
- keep model errors visible as zero-retry failures instead of claiming the configured maximum

## Root cause found during validation
`registerProvider()` can replace fields that `BrainModelInfo` does not expose, including `compat`. The previous `needsUpdate` comparison only checked provider/id/reasoning/context/maxTokens, so changing only `supportsUsageInStreaming` could leave the session on the stale model object.

The new regression test is red before the fix (`setModel` called 0 times) and green after it.

## Testing
- `npm test -- src/agentbox/http-server.test.ts` — 75 passed
- `npm run build`
- GitHub CI — Type Check, Test, Portal Web Test all passed at `64cc5f26`
- siflow-test warm crossover: OFF 20/20 and ON 20/20 assistant `stopReason=stop`, `text=OK`
- siflow-test cold ON: 5 distinct AgentBox Pod UIDs/IPs × 3 requests, 15/15 assistant success
- deployed AgentBox: `registry-ap-southeast.scitix.ai/k8s/siclaw-agentbox@sha256:0d06415d1c952c44e942344e75164df1fb799bb0fb8839ebc7cec79927b0279d`

## Gateway conclusion
The controlled sweep does not support `include_usage=true` as a deterministic reproducer: both OFF and ON succeeded. Pi sends the standard `stream_options.include_usage` field; `continuous_usage_stats` is still absent from Siclaw/Pi request construction. The original 400 remains consistent with an intermittent or heterogeneous MassAPI transformation/routing issue.

### DIFF
--- a/src/agentbox/http-server.test.ts
+++ b/src/agentbox/http-server.test.ts
@@ -363,6 +363,27 @@ describe("http-server — prompt + session lifecycle", () => {
     expect(session.brain.prompt.mock.calls[0][0]).toBe("hello there");
   });
 
+  it("POST /api/prompt refreshes the current model when control-plane modelConfig is supplied", async () => {
+    const session = await sm.getOrCreate("compat-refresh");
+    const r = await getJson(port, "/api/prompt", "POST", {
+      text: "hi",
+      sessionId: "compat-refresh",
+      modelProvider: "openai",
+      modelId: "gpt-4",
+      modelConfig: {
+        ...modelConfigWithInput(["text"]),
+        models: [{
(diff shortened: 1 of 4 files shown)


LikiosSedo:
Closing for now. The issue and validation evidence have been recorded; we will revisit the fix after the MassAPI request-transformation path can be reproduced reliably.

PR #460, opened 2026-07-31
fix(delegation): coordinator alias routing + answer/route triage; fix(lark): never lose the answer to a rejected card update

LikiosSedo:
## Summary

Started as a coordinator alias-routing fix; grew to cover two more problems found while verifying it live. Three themes, independent commits:

1. **Coordinator routing** — resolve localized cluster aliases without requiring a routing helper on default coordinators.
2. **Coordinator scope** — answer knowledge questions directly instead of delegating every request, and keep the triage out of the reply.
3. **Chat-channel delivery** — stop losing a finished answer when a card update is rejected, and keep raw tool names off the card.

---

### Problem

A localized alias could be reported as missing authorization even when its canonical binding is covered. An earlier revision also made an optional helper mandatory, had no deterministic terminal state after a canonical retry, and let delegate names/descriptions or partial binding names look like resource coverage.

### Solution

- query `list_delegates` first with the target as established by the user
- require exact, case-insensitive cluster/host binding matches for coverage; agent metadata and partial names no longer establish authorization
- after a first miss, consult an attached routing helper only when the target may be a cluster alias
- retry once with the confirmed canonical binding name and `binding_name_confirmed=true`; a second miss is terminal
- document the coordinator/helper contract without depending on an out-of-tree serialized field name
- behavioral routing tests with synthetic resource names (alias retry, false-positive rejection, terminal retry) replace persona string assertions

### Problem

The coordinator was defined as "ONLY job is ROUTING", so a question it could answer outright — a concept, a how-to, a documented fact — still cost a full delegation round trip just to have a specialist restate the answer.

### Solution

Two modes with one rule for choosing: answer from the coordinator's own skills/knowledge base when the answer does not depend on a specific environment's **live state** or on a **hands-on action**; route otherwise. Uncertainty resolves to routing, and a specific cluster's live state is never answered from the coordinator's own knowledge — it holds authorization over no resource, so it can never be the authority on one.

Skills now also inform *where* to route (which specialist domain, which target) rather than that being guessed by scanning the roster. `list_delegates` keeps its job as the authorization step. Alias resolution, ask-on-ambiguous, follow-up inheritance and session-reuse rules are unchanged.

A follow-up commit closes a wording gap found in testing: the triage leaked into the user-visible reply. A knowledge question came back as *"这是知识性问题,我已从内部知识库查到原理。下面是…"*, and a failed delegation as *"由于我是协调者,不做直接的 hands-on 诊断"*. The asker wanted the topic explained, not a tour of the bot's operating rules — and naming the mode also invites doubt about whether the answer is authoritative. The persona now forbids announcing the mode, the knowledge-base lookup, or the role's boundaries; when no answer is possible it states the outcome the user needs (the specialist could not be reached, or which detail is missing) instead of the internal rule that produced it.

Also drops `search_memory` from the coordinator's locked capabilities: memory is disabled fleet-wide (`SICLAW_MEMORY_ENABLED` off, so `memory_search`/`memory_get` do not load), and leaving it listed invites it back into prompt text.

> Answering depends on the agent actually having skills attached — coordinator is `defaultNoSkills`, so one left bare can still only route. Recorded in `docs/design/coordinator-routing.md`.

### Problem

Cards were freezing on their last `⏳` progress line while the answer was persisted normally — visible in the web UI, invisible in the chat. Nothing was logged, which is why it went unexplained for weeks.

Root cause: the Feishu SDK does **not** throw on a non-zero API code — it returns `{code,msg}` in the body. `updateCardContent` and `finalizeCard` reported success as soon as the await resolved, so a *rejected* card update (oversized body, out-of-order sequence) was indistinguishable from a delivered one. The file already had a `cardApiFailed` helper for the feedback buttons; it was never applied to the content path. Only a thrown error printed anything, so a rejection produced no log line at all.

### Solution

- check the response code on both content and settings calls, and log `code`/`msg` plus the body length (the leading suspect for a rejection) so the reason is knowable
- `finalizeCard` reports `contentOk` separately from `ok`: a failed streaming-mode flip leaves the answer visible (a text reply would duplicate it), while a failed content update leaves the answer nowhere the user can see
- the terminal path had no fallback at all — it now replies as text when, and only when, the body never landed. The two other call sites already had that fallback but could never reach it, since a rejection was reported as success.

## 4. Raw tool names off the card

A card was showing steps like `⏳ Ran host_list`. That text is the producer's developer-facing activity string — right for the web UI's work card, where a tool log is the point — but in a chat it tells the asker nothing. Worse, the card shows only the *current* step, so a tool-name echo actively pushed the agent's real narration off the card.

The structured sub-agent branch was already rendered and localized for the channel; only the single-agent activity branch passed its text through verbatim. That branch now goes through `channelActivityMilestone`: drop bare `Ran <tool>` echoes, localize the (hard-coded English) sub-agent slot wait, keep genuine narration. The card still advances during a long tool call because the sub-agent's own narration keeps arriving.

Filtered at the channel boundary rather than at the producer: gateway and agentbox ship as separate images, so this holds whatever agentbox version it is paired with, and the web UI keeps its tool log.

---

## Test Plan

- [x] `npm test` — 235 files, 4697 passed, 2 skipped
- [x] `npx tsc --noEmit`
- [x] `npm run build`
- [x] Rebased on `main`; no conflicts

### Verified live on a staging deployment (runtime + agentbox from this branch)

- [x] a bare tool name no longer appears as a card step
- [x] a long structured answer is delivered complete, with feedback buttons, no frozen placeholder
- [x] a live-state request routes: live-state criterion applied, routing helper consulted, `list_delegates` matched the specialist on the first query, and the coordinator did not overreach into hands-on work when delegation failed
- [x] a knowledge question is answered directly from the knowledge base, no delegation
- [x] the reply opens with the answer itself — no "this is a knowledge question / I consulted the knowledge base" preamble

**Not exercised live:** the rejected-card fallback. Every answer in testing was accepted by the API, so the failure branch only has unit coverage (non-zero code → `contentOk=false` → text reply, no feedback row under an undelivered answer). Its main value is that a future rejection now leaves `code=… msg=… chars=…` in the log instead of vanishing silently.

### DIFF
--- a/src/core/agent-types.test.ts
+++ b/src/core/agent-types.test.ts
@@ -26,6 +26,17 @@ describe("agent-types", () => {
     expect(persona).not.toContain("search_memory");
   });
 
+  it("coordinator keeps its triage invisible — the reply must not narrate its own rules", () => {
+    // Observed in a real Feishu reply: "这是知识性问题,我已从内部知识库查到原理" and
+    // "由于我是协调者,不做直接的 hands-on 诊断". The user asked about RoCE, not about
+    // how the bot decides things.
+    const persona = AGENT_TYPES.coordinator.persona!;
+    expect(persona).toContain("KEEP THIS TRIAGE INVISIBLE");
+    expect(persona).toContain("never announce which mode you picked");
+    // On failure, report the outcome the user needs — not the role that blocks it.
+    expect(persona).toContain("state the OUTCOME the user needs");
+  });
+
   it("normalizeAgentType defaults unknown/absent to custom", () => {
     expect(normalizeAgentType("sre")).toBe("sre");
     expect(normalizeAgentType("coordinator")).toBe("coordinator");
--- a/src/core/agent-types.test.ts
+++ b/src/core/agent-types.test.ts
@@ -27,4 +27,12 @@ describe("agent-types", () => {
     expect(effectiveCapabilityKeys("custom", ["read_files"])).toEqual(["read_files"]);
     expect(effectiveCapabilityKeys("custom", null)).toBeNull();
   });
+
+  it("coordinator resolves cluster aliases before coverage lookup", () => {
+    const persona = AGENT_TYPES.coordinator.persona!;
+    expect(persona).toContain("BEFORE the first `list_delegates` query");
+    expect(persona).toContain("confirmed `binding_names[]` value");
+    expect(persona).toContain("never send an unresolved localized name");
+    expect(persona).toContain("Only after `list_delegates` returns no match for a confirmed binding name");
+  });
 });
--- a/src/core/agent-types.test.ts
+++ b/src/core/agent-types.test.ts
@@ -13,6 +13,19 @@ describe("agent-types", () => {
     expect(AGENT_TYPES.custom.persona).toBeNull();
   });
 
+  it("coordinator is dual-mode (answer from knowledge OR route) and no longer uses memory", () => {
+    const persona = AGENT_TYPES.coordinator.persona!;
+    // (A) answer knowledge questions directly from skills / knowledge base
+    expect(persona).toContain("answer it YOURSELF from your skills / knowledge base");
+    // (B) route live/hands-on work; the routing decision is knowledge-informed,
+    // not a raw scan of the delegate list
+    expect(persona).toContain("To ROUTE:");
+    expect(persona).toContain("do NOT merely scan the raw delegate list to guess");
+    // memory is disabled fleet-wide; the coordinator relies on skills/KB, never search_memory
+    expect(AGENT_TYPES.coordinator.capabilities).not.toContain("search_memory");
+    expect(persona).not.toContain("search_memory");
+  });
+
   it("normalizeAgentType defaults unknown/absent to custom", () => {
     expect(normalizeAgentType("sre")).toBe("sre");
     expect(normalizeAgentType("coordinator")).toBe("coordinator");
--- a/src/core/agent-types.test.ts
+++ b/src/core/agent-types.test.ts
@@ -27,12 +27,4 @@ describe("agent-types", () => {
     expect(effectiveCapabilityKeys("custom", ["read_files"])).toEqual(["read_files"]);
     expect(effectiveCapabilityKeys("custom", null)).toBeNull();
   });
-
-  it("coordinator resolves cluster aliases before coverage lookup", () => {
-    const persona = AGENT_TYPES.coordinator.persona!;
-    expect(persona).toContain("BEFORE the first `list_delegates` query");
-    expect(persona).toContain("confirmed `binding_names[]` value");
-    expect(persona).toContain("never send an unresolved localized name");
-    expect(persona).toContain("Only after `list_delegates` returns no match for a confirmed binding name");
-  });
 });
--- a/src/core/agent-types.test.ts
+++ b/src/core/agent-types.test.ts
@@ -13,28 +13,13 @@ describe("agent-types", () => {
     expect(AGENT_TYPES.custom.persona).toBeNull();
   });
 
-  it("coordinator is dual-mode (answer from knowledge OR route) and no longer uses memory", () => {
-    const persona = AGENT_TYPES.coordinator.persona!;
-    // (A) answer knowledge questions directly from skills / knowledge base
-    expect(persona).toContain("answer it YOURSELF from your skills / knowledge base");
-    // (B) route live/hands-on work; the routing decision is knowledge-informed,
-    // not a raw scan of the delegate list
-    expect(persona).toContain("To ROUTE:");
-    expect(persona).toContain("do NOT merely scan the raw delegate list to guess");
-    // memory is disabled fleet-wide; the coordinator relies on skills/KB, never search_memory
+  it("coordinator does not reach for memory (disabled fleet-wide)", () => {
+    // Asserted as structure and as an ABSENCE, so it survives any rewording of
+    // the persona. The answer/route contract itself lives in
+    // docs/design/coordinator-routing.md — pinning its prose here would only
+    // restate the diff and would break on every future improvement.
     expect(AGENT_TYPES.coordinator.capabilities).not.toContain("search_memory");
-    expect(persona).not.toContain("search_memory");
-  });
-
-  it("coordinator keeps its triage invisible — the reply must not narrate its own rules", () => {
-    // Observed in a real Feishu reply: "这是知识性问题,我已从内部知识库查到原理" and
-    // "由于我是协调者,不做直接的 hands-on 诊断". The user asked about RoCE, not about
-    // how the bot decides things.
-    const persona = AGENT_TYPES.coordinator.persona!;
-    expect(persona).toContain("KEEP THIS TRIAGE INVISIBLE");
-    expect(persona).toContain("never announce which mode you picked");
-    // On failure, report the outcome the user needs — not the role that blocks it.
-    expect(persona).toContain("state the OUTCOME the user needs");
+    expect(AGENT_TYPES.coordinator.persona).not.toContain("search_memory");
   });
 
   it("normalizeAgentType defaults unknown/absent to custom", () => {
--- a/src/core/agent-types.ts
+++ b/src/core/agent-types.ts
@@ -38,7 +38,12 @@ const SRE_PERSONA =
 
 const COORDINATOR_PERSONA =
   "You are a COORDINATOR whose ONLY job is ROUTING. To route: (1) determine the TARGET resource (cluster / " +
-  "host / node) from the user's request; (2) call `list_delegates` with query=<that target> to find WHICH " +
+  "host / node) from the user's request. For a cluster target, BEFORE the first `list_delegates` query, " +
+  "consult an available routing-helper skill unless the exact Siclaw binding name is already established in " +
+  "verified context. Use the helper's confirmed `binding_names[]` value for coverage lookup; never send an " +
+  "unresolved localized name, local nickname, alias, or spelling variant directly to `list_delegates`. If the " +
+  "helper cannot confirm a binding name, ASK THE USER instead of guessing. (2) call `list_delegates` with " +
+  "query=<confirmed binding name for a cluster, otherwise the established target> to find WHICH " +
   "delegate is bound to it — this authoritative coverage lookup (NOT your own cluster_list, which is YOUR " +
   "bindings) is how you confirm who covers the target; (3) delegate to the matching agent via " +
   "`delegate_to_agent`. If you CANNOT determine the target from the request — it is missing, ambiguous, or a " +
@@ -49,12 +54,14 @@ const COORDINATOR_PERSONA =
   "to the resource you already established — do NOT re-ask the user for it, and do NOT re-run `list_delegates` " +
   "discovery; carry forward what you already know and delegate straight to the same specialist. Re-determine " +
   "the target and re-query `list_delegates` ONLY when the target is genuinely NEW or has CHANGED. If you " +
-  "queried and NO delegate covers the target, tell the user that no authorized agent covers that resource. " +
+  "queried and NO delegate covers the target, first verify that a cluster query used a confirmed binding name " +
+  "and resolve/retry it if not. Only after `list_delegates` returns no match for a confirmed binding name may " +
+  "you tell the user that no authorized agent covers that resource. " +
   "Forward the task at a HIGH LEVEL, essentially as the user phrased it. You do NOT decide HOW the task is " +
   "done: do NOT read the specialist's execution procedures/skills or enumerate the steps for it, and do NOT " +
   "attempt any hands-on work yourself. The specialist owns the tools and the know-how and will work out the " +
-  "steps on its own. You MAY consult your own knowledge or a routing-helper skill you were given, but ONLY to " +
-  "decide WHICH specialist to route to — not to solve the problem. When you delegate, describe the GOAL in " +
+  "steps on its own. Use routing helpers and any knowledge you consult ONLY to decide WHICH specialist to " +
+  "route to — not to solve the problem. When you delegate, describe the GOAL in " +
   "the user's own terms and INCLUDE any concrete facts you already gathered so the specialist need not " +
   "re-look-them-up; but do NOT name specific skills, scripts, or steps for the specialist to run — it will " +
   "choose those itself. " +
--- a/src/core/agent-types.ts
+++ b/src/core/agent-types.ts
@@ -38,14 +38,15 @@ const SRE_PERSONA =
 
 const COORDINATOR_PERSONA =
   "You are a COORDINATOR whose ONLY job is ROUTING. To route: (1) determine the TARGET resource (cluster / " +
-  "host / node) from the user's request. For a cluster target, BEFORE the first `list_delegates` query, " +
-  "consult an available routing-helper skill unless the exact Siclaw binding name is already established in " +
-  "verified context. Use the helper's confirmed `binding_names[]` value for coverage lookup; never send an " +
-  "unresolved localized name, local nickname, alias, or spelling variant directly to `list_delegates`. If the " +
-  "helper cannot confirm a binding name, ASK THE USER instead of guessing. (2) call `list_delegates` with " +
-  "query=<confirmed binding name for a cluster, otherwise the established target> to find WHICH " +
+  "host / node) from the user's request; (2) call `list_delegates` first with query=<that target exactly as " +
+  "established> to find WHICH " +
   "delegate is bound to it — this authoritative coverage lookup (NOT your own cluster_list, which is YOUR " +
-  "bindings) is how you confirm who covers the target; (3) delegate to the matching agent via " +
+  "bindings) is how you confirm who covers the target. If it returns no exact binding match and the target " +
+  "may be a cluster alias, consult a routing-helper skill you were given, if any. Only when the helper confirms " +
+  "one canonical Siclaw binding name, retry `list_delegates` once with that name and " +
+  "`binding_name_confirmed=true`; do not guess from an ambiguous or unresolved result. If no helper is " +
+  "attached, it cannot confirm one name, or the confirmed retry also misses, tell the user that no authorized " +
+  "agent covers the name and that it may be an alias. (3) delegate to the matching agent via " +
   "`delegate_to_agent`. If you CANNOT determine the target from the request — it is missing, ambiguous, or a " +
   "node/pod is named without its cluster — ASK THE USER to supply the missing detail. Do NOT guess, do NOT " +
   "browse the whole delegate list hoping to infer it, and do NOT pick the closest match. EXCEPTION — a " +
@@ -54,9 +55,7 @@ const COORDINATOR_PERSONA =
   "to the resource you already established — do NOT re-ask the user for it, and do NOT re-run `list_delegates` " +
   "discovery; carry forward what you already know and delegate straight to the same specialist. Re-determine " +
   "the target and re-query `list_delegates` ONLY when the target is genuinely NEW or has CHANGED. If you " +
-  "queried and NO delegate covers the target, first verify that a cluster query used a confirmed binding name " +
-  "and resolve/retry it if not. Only after `list_delegates` returns no match for a confirmed binding name may " +
-  "you tell the user that no authorized agent covers that resource. " +
+  "queried and NO delegate covers the target, follow the one-retry alias flow above; never repeat it. " +
   "Forward the task at a HIGH LEVEL, essentially as the user phrased it. You do NOT decide HOW the task is " +
   "done: do NOT read the specialist's execution procedures/skills or enumerate the steps for it, and do NOT " +
   "attempt any hands-on work yourself. The specialist owns the tools and the know-how and will work out the " +
--- a/src/core/agent-types.ts
+++ b/src/core/agent-types.ts
@@ -6,9 +6,9 @@
  *
  *   - sre         — a specialist that operates hands-on within its authorized
  *                   clusters/hosts (full read + exec + scripts). Persona fixed.
- *   - coordinator — a read-only router: sees the fleet, delegates hands-on work
- *                   to specialists via delegate_to_agent. No skills by default.
- *                   Persona fixed.
+ *   - coordinator — answers knowledge questions from its skills/knowledge base and
+ *                   routes hands-on work to specialists via delegate_to_agent.
+ *                   No skills by default. Persona fixed.
  *   - custom      — the legacy free-form agent: the operator picks capabilities
  *                   (tool_capabilities) AND writes the system_prompt. Nothing is
  *                   locked. Existing agents map here (zero behaviour change).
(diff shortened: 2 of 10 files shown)


src/core/agent-types.ts
```diff
@@ -38,7 +38,12 @@ const SRE_PERSONA =
 
 const COORDINATOR_PERSONA =
   "You are a COORDINATOR whose ONLY job is ROUTING. To route: (1) determine the TARGET resource (cluster / " +
-  "host / node) from the user's request; (2) call `list_delegates` with query=<that target> to find WHICH " +
+  "host / node) from the user's request. For a cluster target, BEFORE the first `list_delegates` query, " +
+  "consult an available routing-helper skill unless the exact Siclaw binding name is already established in " +
+  "verified context. Use the helper's confirmed `binding_names[]` value for coverage lookup; never send an " +
+  "unresolved localized name, local nickname, alias, or spelling variant directly to `list_delegates`. If the " +
+  "helper cannot confirm a binding name, ASK THE USER instead of guessing. (2) call `list_delegates` with " +
```
chent1996: **Findings 1–2.** `coordinator` is declared `defaultNoSkills: true` and `agent-api.ts:148` skips the builtin auto-bind for it — the Portal UI says "No skills bound by default; attach only the routing helpers this coordinator needs". So the default coordinator has no skills at all.

The previous wording was conditional on that ("a routing-helper skill **you were given**"); "consult **an available** routing-helper skill" as a mandatory pre-step is not. Combined with "If the helper cannot confirm a binding name, ASK THE USER instead of guessing", a default-configured coordinator can end up interrogating the user for a binding name before it will run the free in-box query that would have worked whenever the user typed the real name.

Also: "unless the exact Siclaw binding name is already established in verified context" is not decidable here — the coordinator cannot tell whether the user's string is the exact binding name until it queries. Both branches look valid from where it stands.

Querying first and consulting the helper only on zero results (which the tool hint already instructs) avoids both problems and keeps the happy path at one tool call.

src/tools/workflow/list-delegates.ts
```diff
@@ -102,7 +102,13 @@ export function createListDelegatesTool(refs: ToolRefs): ToolDefinition {
       const hasMore = nextOffset < total;
       let hint = "";
       if (total === 0) {
-        hint = `\n\nNo delegate agent covers "${rawQuery}". Do not delegate — tell the user no authorized agent covers that resource.`;
+        hint =
+          `\n\nNo delegate agent covers "${rawQuery}" as written. If this target may be a cluster alias, ` +
+          "localized name, local nickname, or spelling variant, do not conclude that coverage is absent yet: " +
```
chent1996: **Finding 4.** This hint is stateless — it cannot tell a first attempt on a raw alias from a retry with an already-confirmed binding name, so it emits "do not conclude that coverage is absent yet … then retry" on **both**. The only thing breaking the loop is the model remembering it already did the confirmed query. The previous hint had no such dependency and was terminal.

Giving the tool the missing bit makes the exit deterministic and testable:

```ts
binding_name_confirmed: Type.Optional(Type.Boolean({
  description: "Set when `query` is a canonical Siclaw binding name resolved by a routing helper.",
})),
```

…then emit the old terminal hint when it is set.

Separately, this branch only fires on a **miss**. `matchRoster` substring-matches `description` too, so a short alias (`test`, `bj`, `prod`) can match the wrong delegate, yield `total > 0`, and print no hint at all — the coordinator then routes confidently to the wrong specialist. That failure is worse than the one being fixed here and no zero-result hint can reach it.

src/core/agent-types.test.ts
```diff
@@ -27,4 +27,12 @@ describe("agent-types", () => {
     expect(effectiveCapabilityKeys("custom", ["read_files"])).toEqual(["read_files"]);
     expect(effectiveCapabilityKeys("custom", null)).toBeNull();
   });
+
+  it("coordinator resolves cluster aliases before coverage lookup", () => {
```
chent1996: **Finding 6.** The other tests in this file assert structure (`persona` truthy/null, capability lists, `defaultNoSkills`); this one pins four exact English phrases from the prose.

It restates the diff rather than testing anything: it can only fail on a rewording, and it will fail on every future rewording including improvements. If the intent is to keep the rule from being silently dropped, a `docs/design/` entry for coordinator routing plus a behavioral test at the routing layer would survive rephrasing — this will not.

chent1996 (commented):
Reviewed against `main` @ `36dce74a` (post-#457) — the PR's merge-base, so no stale-base risk. CI green, `npx tsc --noEmit` clean, both touched suites pass locally (10 tests). String-only change: `matchRoster`, pagination, the `available` gate and the `details` shape are all untouched, so nothing downstream of the tool moves.

The premise is correct and the bug is real. `matchRoster` is a case-insensitive substring match over `clusters`, `hosts`, `name`, and `description` — it fundamentally cannot resolve a localized name or nickname to a binding name, and reporting that miss as "no authorized agent covers that resource" is a genuinely confusing failure that looks like an authorization problem. Worth fixing.

My concern is with the shape of the fix rather than the goal. It makes a mandatory pre-step out of a capability that is absent by default, gates it on a condition the coordinator cannot evaluate at the moment it has to, and converts the tool's terminal instruction into a self-repeating one with no state to break the loop. Details below, then a suggested inversion that I think gets the same protection for less.

---

## 1. The mandatory pre-step depends on a capability that is absent by default

`AGENT_TYPES.coordinator` sets `defaultNoSkills: true`, and `agent-api.ts:148` skips the builtin-skill auto-bind for exactly that reason. The Portal UI states the intent plainly:

> No skills bound by default; attach only the routing helpers this coordinator needs (Skills tab).

So a freshly created coordinator has **zero** skills. The old persona respected that — "a routing-helper skill **you were given**" was conditional on the operator having attached one. The new wording is not: "BEFORE the first `list_delegates` query, consult **an available** routing-helper skill", followed by "If the helper cannot confirm a binding name, **ASK THE USER** instead of guessing."

On a default-configured coordinator that composes into: before answering a cluster question, consult a skill that isn't there, then ask the user to supply a "Siclaw binding name" — instead of running the free in-box substring query that would have succeeded in the common case where the user typed the real name. That is a worse default experience than the bug being fixed, and it lands on every deployment that hasn't attached a helper.

## 2. The gate condition isn't decidable when the coordinator has to decide

"unless the exact Siclaw binding name is already established in verified context" — the coordinator has no way to know whether a user-typed string *is* the exact binding name until it queries. Both branches are defensible from where it stands, so the model guesses, per request. For a routing layer whose whole determinism budget is prompt text, an undecidable branch condition at step 1 is expensive.

## 3. The strongest argument for resolve-before-query is the false positive — and it is neither stated nor tested

The PR body justifies the change entirely as "prevent a raw alias miss from being reported as missing authorization". But a miss is the *benign* failure: it is loud, and the zero-result hint already handles it. The dangerous case is the opposite one, and this ordering is the only thing that could prevent it:

`matchRoster` matches substrings against `m.description` as well as bindings. A short alias — `test`, `bj`, `prod` — can match `roce-test` under one delegate and any delegate whose description happens to mention testing. `total > 0`, so **no hint fires at all**, and the coordinator routes to a plausible-looking wrong specialist. Silently wrong routing is strictly worse than a clear miss, and no zero-result hint can ever catch it.

If that is the motivating case, it belongs in the PR description and deserves the regression test (the two new tests both cover the miss). If it is not a concern in practice, then #4's inversion below is the cheaper design and the pre-step can be dropped.

## 4. The zero-result hint went from terminal to self-repeating, with nothing to break the loop

The tool is stateless: it cannot distinguish "first attempt, raw alias" from "retry with a confirmed binding name". It emits the same text both times — *"do not conclude that coverage is absent yet … then retry `list_delegates` with that binding name"*. The only thing terminating the loop is the model remembering it already did the confirmed query. The previous hint had no such dependency; it was unambiguous and terminal.

This is cheap to make deterministic — give the tool the bit it is missing:

```ts
parameters: Type.Object({
  query: …,
  binding_name_confirmed: Type.Optional(Type.Boolean({
    description: "Set when `query` is a canonical Siclaw binding name resolved by a routing helper.",
  })),
})
```

and make the hint terminal when it is set. Same information, but the state machine lives in the tool rather than in the model's recollection — and it also gives you something testable.

## 5. `binding_names[]` is a contract this repo never defines

The persona instructs the coordinator to "use the helper's confirmed `binding_names[]` value". That identifier appears nowhere in the tree except this string and the test pinning it. The helper is operator-authored and out of tree, so the coordinator's routing behavior now depends on a field name nothing here specifies, validates, or would notice breaking.

There is still no `docs/design/` entry for delegation or coordinator routing — I raised the same gap on #413. This is now the second PR encoding routing semantics purely as prose inside a prompt string. A short doc naming the helper's expected output contract (and this alias-resolution rule) would make the next fix in this area reviewable against something.

## 6. The persona test pins literal prompt wording

`agent-types.test.ts` previously asserted only structure — `persona` truthy/null, capability lists, `defaultNoSkills`. The new test pins four exact English phrases from the prose. It restates the diff: it cannot fail for any reason except a rewording, and it *will* fail on every future rewording, including improvements. If the goal is to stop this rule from being silently dropped, the durable home is the design doc plus a behavioral test at the routing layer.

(The `list-delegates.test.ts` assertions are more defensible — tool output is a real contract surface — though they pin phrasing too.)

## 7. Governance question

CLAUDE.md marks `src/core/prompt.ts` ⚠️ **REQUIRES HUMAN APPROVAL**. `agent-types.ts` holds the personas that shape agent behavior just as directly and has no matrix row at all. Worth an explicit decision on whether persona edits carry the same gate, rather than leaving it to whoever notices.

---

## Suggested alternative

Invert the order rather than mandating a pre-step:

1. Query `list_delegates` with the target as the user phrased it — free, in-box, no latency, and correct whenever the user typed the real name.
2. On **zero** results, consult a routing helper if one is attached; retry with the resolved binding name (passing the `binding_name_confirmed` bit from #4).
3. If no helper is attached, or it cannot resolve, tell the user no authorized agent covers the resource — and say the name may be an alias, so they can supply the canonical one.

This is one mechanism instead of two, keeps the happy path at a single tool call, degrades correctly on the default no-skills coordinator, and gives a deterministic exit. The one thing it does not cover is the false-positive match from #3 — so if that case is real, I would rather see it addressed head-on (tighter matching, or exact-match-preferred ranking in `matchRoster`) than through a prompt instruction the model may or may not follow.

### Verified clean

- Based on current main; no conflicts with anything open.
- Both edits are string-only — tool behavior, gating, and result shape are unchanged, so there is no code-path risk in merging as-is.
- The removed "You MAY consult your own knowledge or a routing-helper skill" sentence is faithfully preserved in the reworded "Use routing helpers and any knowledge you consult ONLY to decide WHICH specialist to route to" — no scope was widened there.

jacoblee-io (commented):
Reviewed. Clean, well-scoped prompt fix — the target problem (a raw alias miss reported as missing authorization) is addressed at both the persona and the zero-result hint, the zero-result hint is conditionally worded so it doesn't mislead host/node lookups, and both test assertions match the new strings verbatim. No logic or security impact. One non-blocking consideration below.

src/core/agent-types.ts
```diff
@@ -38,7 +38,12 @@ const SRE_PERSONA =
 
 const COORDINATOR_PERSONA =
   "You are a COORDINATOR whose ONLY job is ROUTING. To route: (1) determine the TARGET resource (cluster / " +
-  "host / node) from the user's request; (2) call `list_delegates` with query=<that target> to find WHICH " +
+  "host / node) from the user's request. For a cluster target, BEFORE the first `list_delegates` query, " +
+  "consult an available routing-helper skill unless the exact Siclaw binding name is already established in " +
+  "verified context. Use the helper's confirmed `binding_names[]` value for coverage lookup; never send an " +
+  "unresolved localized name, local nickname, alias, or spelling variant directly to `list_delegates`. If the " +
```
jacoblee-io: Non-blocking consideration: this prohibition ("never send an unresolved localized name / alias / spelling variant directly to `list_delegates`") is unconditional, while the resolver above it is "an **available** routing-helper skill." In a deployment that ships **no** routing-helper skill, a user who types an alias would leave the coordinator unable to resolve — and the fallback ("ASK THE USER") pushes it to request a canonical Siclaw binding name the user likely doesn't know, instead of just trying the name against `list_delegates`' own matching as before. Only bites the narrow "no helper + user used an alias" case (deployments with aliases presumably ship a helper, or the alias-miss problem wouldn't arise), so likely fine — just worth confirming the intended coordinator deployments all provide a routing-helper skill. If some don't, consider gating the prohibition on the helper actually being available.

jacoblee-io:
### Follow-up suggestion (for a separate change, not blocking this PR)

This PR correctly stops a raw alias miss from being reported as "no authorized agent." Building on that, we'd like to evolve the coordinator's principle a step further — from **pure router** to **"answer what you can, route what needs a specialist"**, with skills/knowledge as a first-class aid in *both* modes. Filing it here so it can be picked up by whoever owns this area.

**Principle**

| Case | Behavior |
|---|---|
| **Knowledge question** (concepts, how-to, definitions, comparisons, documented facts) | Answer it **yourself** from your **skills / knowledge base** — do not delegate just to have a specialist restate the answer. |
| **Needs live troubleshooting** (current state of a specific resource, hands-on inspection/diagnosis/remediation, or a conclusion only the authorized specialist can stand behind) | First use **skills / knowledge** to work out **which specialist domain + which target** this belongs to — don't just scan the raw delegate list to guess. `list_delegates` then serves only as the **authoritative confirmation of who actually covers the target**, not as the mechanism that *decides* who to route to. |

Mode is chosen by whether a correct answer depends on a specific environment's **live state** or a **hands-on action**. If it does → route. If unsure → route (never answer about a specific cluster's live state from your own knowledge).

**Important correction:** drop every mention of `search_memory` — memory is disabled in our deployments (`SICLAW_MEMORY_ENABLED` off; `memory_search`/`memory_get` don't load, and memory-facing prompt text is intentionally suppressed). The coordinator's answer/route aids should be **skills + knowledge base only**.

**Proposed full `COORDINATOR_PERSONA`** (routing detail, alias→binding resolution, follow-up inheritance, no-coverage retry, and SESSION REUSE all preserved):

```
You are a COORDINATOR. Your granted skills and knowledge base are your primary aid in BOTH of your modes:

(A) ANSWER — when the request is a question answerable from knowledge (concepts, how-to, definitions, comparisons, documented facts), answer it YOURSELF by consulting your skills / knowledge base. Do NOT delegate a question just to have a specialist restate the answer.

(B) ROUTE — when the request needs the live/current state of a specific resource, hands-on inspection / diagnosis / remediation, or a conclusion only the resource's authorized specialist can stand behind, then ROUTE it. Even here, FIRST use your skills / knowledge to work out WHICH specialist domain and WHICH target this belongs to — do NOT merely scan the raw delegate list to guess. Then use `list_delegates` to AUTHORITATIVELY confirm who actually covers that target, and delegate.

Choose the mode by whether a correct answer depends on a specific environment's live state or a hands-on action. If it does, ROUTE. If you are unsure whether it does, ROUTE — never answer about a specific cluster's live state from your own knowledge.

To ROUTE: (1) using your skills / knowledge and, for a cluster target, an available routing-helper skill, determine the TARGET resource (cluster / host / node) and resolve any localized name, local nickname, alias, or spelling variant to its confirmed Siclaw `binding_names[]` value BEFORE the first `list_delegates` query (unless the exact binding name is already established in verified context). Never send an unresolved alias directly to `list_delegates`; if a binding name cannot be confirmed, ASK THE USER instead of guessing. (2) call `list_delegates` with query=<confirmed binding name for a cluster, otherwise the established target> to AUTHORITATIVELY confirm WHICH delegate is bound to it (NOT your own cluster_list, which is YOUR bindings). (3) delegate to the matching agent via `delegate_to_agent`. If you CANNOT determine the target from the request — it is missing, ambiguous, or a node/pod is named without its cluster — ASK THE USER to supply the missing detail. Do NOT guess, do NOT browse the whole delegate list hoping to infer it, and do NOT pick the closest match. EXCEPTION — a follow-up WITHIN an investigation already in progress: INHERIT the target resource and the specialist from the ongoing thread. A pronoun-only or elliptical follow-up that does not restate the target still refers to the resource you already established — do NOT re-ask the user for it, and do NOT re-run `list_delegates` discovery; carry forward what you already know and delegate straight to the same specialist. Re-determine the target and re-query `list_delegates` ONLY when the target is genuinely NEW or has CHANGED. If you queried and NO delegate covers the target, first verify that a cluster query used a confirmed binding name and resolve/retry it if not. Only after `list_delegates` returns no match for a confirmed binding name may you tell the user that no authorized agent covers that resource.

When you DO delegate: forward the task at a HIGH LEVEL, essentially as the user phrased it. You do NOT decide HOW the task is done: do NOT read the specialist's execution procedures/skills or enumerate the steps for it, and do NOT attempt any hands-on work yourself. The specialist owns the tools and the know-how and will work out the steps on its own. Use your skills and any knowledge you consult ONLY to decide WHICH specialist to route to — not to solve the problem for it. When you delegate, describe the GOAL in the user's own terms and INCLUDE any concrete facts you already gathered so the specialist need not re-look-them-up; but do NOT name specific skills, scripts, or steps for the specialist to run — it will choose those itself.

SESSION REUSE — judge by ONE thing: does this request CONTINUE THE SAME INVESTIGATION, or open a DIFFERENT one? REUSE the peer session (pass the session_id the specialist returned) when the new message belongs to the SAME diagnostic thread as your immediately-preceding delegation to that specialist — drilling deeper, checking a related aspect, or following up on what it just found — even when the user gives no explicit continuation cue. A connected chain of checks that narrows in on the same target is ONE investigation: keep it in ONE session so the specialist retains the context it already gathered, and do NOT split it into fresh sessions merely because the wording changed or carried no continuation keyword. Start a FRESH session (omit session_id) only when the request is a GENUINELY DIFFERENT problem — a different symptom or subsystem unrelated to the ongoing investigation — which must not inherit the prior context. Same cluster / node / resource is a WEAK signal on its own: it neither forces reuse nor forbids it; decide by topical continuity, not by the target and not by keyword matching. Reuse is about CONVERSATIONAL CONTINUITY of one investigation, NOT efficiency: do NOT reuse to spare the specialist from re-establishing context or because it already knows the target — it re-establishes that cheaply, and a fresh session keeps a distinct problem's context clean. A DIFFERENT component, subsystem, or failure domain is a DIFFERENT investigation even on the same cluster / target and even immediately after — start it fresh. If your own reasoning is 'new direction, but same target, so I'll continue', that is the signal to start FRESH. When a follow-up plausibly DEEPENS the same line of inquiry, prefer reuse; when it opens a different subsystem, prefer a fresh session even if recent (the gateway already bounds reuse to this conversation's recent sessions, so a stale one from far back can never be resurrected). After the specialist reports back, relay / synthesize its findings.
```

**What changes vs. current**
1. Opening: "your ONLY job is ROUTING" → two modes (answer / route).
2. New: general knowledge questions are answered directly from skills/KB instead of always being delegated.
3. `list_delegates` demoted from "the one authoritative way to decide who to route to" → "authoritative confirmation of coverage/authorization"; the *decision* of which specialist is driven by skills/knowledge.
4. `search_memory` removed (memory disabled).
5. Unchanged: alias→binding resolution, ask-on-ambiguous, follow-up inheritance, no-coverage retry, high-level delegation (no doing the work / not reading the specialist's skills), and the full SESSION REUSE block.

**Two caveats for whoever implements this**
1. Coordinator is `defaultNoSkills: true`. For (A)/(B) to actually work, the coordinator agent must be **granted a knowledge-base / routing-helper skill** in Portal — otherwise "consult your skills" resolves to nothing. This is a deployment-config prerequisite, not a code change.
2. `agent-types.ts` still lists `search_memory` in both the `sre` and `coordinator` capability arrays. Since memory is disabled long-term, consider dropping it there too so it doesn't get copied back into prompt text later. (Harmless at runtime — the capability→tool mapping resolves to nothing when memory is off — but cleaner to remove.)

`agent-types.test.ts` and `list-delegates.test.ts` assertions will need updating to the new strings.

jacoblee-io:
Thanks — this was a genuinely useful review, and the inversion you proposed in #4 is what the branch now does. Status of each finding at `74c20fc2` (the review was written against `3eba7aa4`; `8423ba68` landed after it, and the branch has since been rebased onto `main` @ `4ce904a2`).

**#1 / #2 — mandatory pre-step on an absent capability; undecidable gate. Fixed.** The order is inverted exactly as you suggested: `list_delegates` is queried **first** with the target as established, and a routing helper is consulted only after a miss and only "if any". The happy path is one tool call again, and a default coordinator with zero skills never gets sent to look for a skill that isn't there — nor asked to interrogate the user for a binding name before trying the free in-box query. The undecidable "unless the exact binding name is already established" gate is gone with it.

**#3 — the false positive was the dangerous case. Fixed, and you were right that it was the stronger argument.** `matchRoster` no longer substring-matches, and no longer looks at `name`/`description` at all:

```ts
const mc = m.clusters.filter((s) => s.trim().toLowerCase() === q);
const mh = m.hosts.filter((s) => s.trim().toLowerCase() === q);
if (mc.length > 0 || mh.length > 0) out.push({ m, mc, mh });
```

So `test` / `bj` / `prod` can no longer match `roce-test` or a delegate whose description mentions testing, and the silent wrong-routing path you identified — `total > 0`, no hint, confident delegation to the wrong specialist — is closed at the matcher rather than papered over by a hint. `details.match_basis` now reports `exact_resource_binding`.

**#4 — self-repeating hint with no state. Fixed as prescribed.** The tool takes `binding_name_confirmed`, and the zero-result hint branches on it: when set it is terminal again ("Do not delegate — tell the user no authorized agent covers that resource"); when unset it is the retry-once instruction. The state machine lives in the tool, not in the model's recollection, and it is testable.

**#5 — `binding_names[]` was a contract this repo never defined. Fixed.** That identifier is gone from the tree. `docs/design/coordinator-routing.md` now exists and specifies the helper's **semantic** result — one confirmed canonical binding name, ambiguous, or unresolved — deliberately without naming a serialized field, so an out-of-tree helper's wire shape is not baked into a prompt string.

**#6 — persona tests pinning prose. You were right, and I had reintroduced it.** `8423ba68` dropped the original four-phrase assertion, but the answer/route work added two new tests pinning six exact English phrases — the same anti-pattern, and your objection applies verbatim: they restate the diff, can only fail on a rewording, and would fail on every improvement. `74c20fc2` removes them. The contract moved to its durable home in `docs/design/coordinator-routing.md` (including the new rule that the triage stays invisible in replies); what remains in the test is structural and an *absence* — no `search_memory` in the coordinator's capabilities or persona — which survives rephrasing and only breaks if memory is genuinely reintroduced.

I agree there is no behavioral surface for persona prose the way there is for tool output, so doc-plus-structure is the honest ceiling here rather than a behavioral test.

---

Two notes on scope, since the PR grew after your review: it now also carries a coordinator answer/route triage (answer knowledge questions instead of delegating every request) and two chat-channel delivery fixes found while verifying it live — a rejected card update was reported as success, so finished answers were silently lost, and raw `Ran <tool>` strings were being shown as card steps. The description has been rewritten to cover all of it. `npm test` is green (235 files / 4696 passed) and the answer/route and card behaviors were verified on a live deployment; the rejected-card fallback itself is unit-only, since every answer in testing was accepted by the API.

**Still open from your review:** the sre type keeps `search_memory` in its locked capabilities. It is inert with memory disabled, so I left it rather than widen this PR — happy to drop it here if you'd prefer it cleaned up in one place.

LikiosSedo:
Thanks — this direction matches the intended coordinator role: own ingress classification, resource normalization, context enrichment, and route selection, while specialists own live diagnosis and remediation.

We’ll keep #460 scoped to deterministic alias/coverage routing and take the ANSWER/ROUTE model as a separate change. That follow-up will also need to define the coordinator’s default skill/knowledge bindings and make removing or retaining `search_memory` an explicit product decision.

One important adjustment: we should not copy the proposed routing paragraph verbatim. The follow-up must preserve #460’s query-first fallback for coordinators with no helper, an explicit terminal state after the retry, and a semantic helper contract rather than depending on a fixed out-of-tree `binding_names[]` field.

jacoblee-io:
Closing in favour of #463, which carries this work forward.

Context: this PR started as the cluster-alias routing fix and outgrew its scope once the same code path was verified against a live deployment — it picked up a coordinator answer/route triage and two chat-channel delivery fixes, and the branch history had been rewritten a few times along the way. #463 is the same content reorganised into two file-disjoint commits with an accurate description, rebased on `main`, CI green.

The alias-routing commits are carried into #463 (co-authored, @LikiosSedo) — nothing from this PR is dropped.

@chent1996 all six findings from your review here are resolved; the resolution is written up at the bottom of #463 so the trail stays reviewable. Thanks — #3 in particular (the substring/description match routing confidently to the wrong specialist) was the more dangerous bug and is now fixed at the matcher.

PR #461, opened 2026-07-31
feat: add maintainer-editable hot-reload agent prompts

LikiosSedo:
## Summary

- make `system_prompt` the persisted, editable Agent-owned identity/behavior instruction for SRE, Coordinator, and Custom agents
- keep Siclaw platform prompt assembly, safety ordering, skills/knowledge context, MCP/tool bindings, and delegation constraints platform-owned
- add a `prompt` reload resource that invalidates warm sessions and rebuilds the next turn from existing JSONL history without restarting AgentBox or waiting for the idle TTL
- preserve in-flight turns and detached completions: sessions with background ownership continue temporarily on the old immutable prompt, deliver or explicitly discard buffered completion notifications, then re-arm release and rebuild
- make Portal resource saves change-driven, and fail visibly instead of silently discarding edits when current bindings could not be loaded

## Approval

- Maintainer-approved behavior change: built-in persona text becomes the editable initial Agent prompt while platform safety, mode, context, tool, and authorization assembly remains platform-owned.

## Verification

- `npm test` — 235 files, 4709 passed, 2 skipped
- `npm run build`
- `npm --prefix portal-web test` — 21 files, 196 passed
- `npm --prefix portal-web run build`
- `git diff --check`

## Cross-repository integration

The companion frontend/API integration adds the `siclaw=a`-only management surface, backfills built-in prompt defaults, redacts Prompt from ordinary read/write APIs, and waits for the Runtime reload acknowledgement before reporting a successful Prompt save.


src/portal/agent-api.ts
```diff
@@ -131,10 +131,10 @@ export function registerAgentRoutes(
...
         toolCapabilities ?? null,
         agentType,
-        // Built-in types (sre/coordinator) use a LOCKED persona — never store a
-        // Custom system_prompt for them (the runtime would use it as the base
-        // prompt and merely append the type persona). Only custom keeps its own.
-        agentType === "custom" ? (body.system_prompt ?? null) : null,
+        // Materialize the type default into the row so system_prompt becomes
+        // the agent's visible, editable instruction instead of a hidden runtime
+        // overlay. Custom agents keep their supplied prompt (or null).
+        effectiveAgentPrompt(agentType, body.system_prompt) ?? null,
```
jacoblee-io, line 147: **Materializing the built-in default into the row breaks the TUI path.**

Built-in types used to store `NULL` here. Now they store `SRE_DEFAULT_PROMPT` / `COORDINATOR_DEFAULT_PROMPT`, and `src/cli-main.ts:216` still passes `activeAgent.systemPrompt` as `systemPromptTemplate`, which `prompt.ts:25` uses to *replace* `DEFAULT_TEMPLATE` outright. So a TUI session on an SRE agent loses the whole platform prompt (see the PR-level comment for the full list of what's dropped).

Either teach the TUI path to append rather than override (mirroring `session.ts:2554`), or keep the row NULL and let `effectiveAgentPrompt` supply the default at read time. As written, the same row means "append" in AgentBox and "replace" in the TUI.

src/portal/agent-api.ts
```diff
@@ -301,22 +287,36 @@ export function registerAgentRoutes(
...
-    // AgentBox re-fetches its whitelist + type (caps/persona are type-locked) and
-    // invalidates live sessions (mid-turn-safe).
-    if (toolCapabilitiesChanged || ("agent_type" in body)) {
-      connectionMap.notify(params.id, "agent.reload", { agentId: params.id, resources: ["tools"] });
+    // Capability/type changes rebuild the tool surface. Prompt/type changes
+    // invalidate warm sessions so the next turn immediately restores with the
+    // latest editable prompt (no AgentBox kill and no 30s idle wait).
+    const reloadResources: string[] = [];
+    if (toolCapabilitiesChanged || ("agent_type" in body)) reloadResources.push("tools");
+    if ("system_prompt" in body || ("agent_type" in body)) reloadResources.push("prompt");
```
jacoblee-io: **This fires on every agent save, not just prompt edits — and it blocks the response for up to 30s.**

`portal-web/src/components/AgentSettings.tsx:273` has a single `handleSave` shared by all nine save tabs, and it unconditionally includes `system_prompt` and `agent_type` in the body. So `"system_prompt" in body` is *always* true: renaming the agent, changing the model, flipping a skill binding, or editing the idle timeout all take the awaited `sendCommand` branch and invalidate every warm session — even when the prompt text is byte-identical. There is no dirty check on either side (the handler only tests `field in body`, and the frontend never compares against `agent.system_prompt`).

Two consequences:
1. `sendCommand`'s `DEFAULT_TIMEOUT_MS` is `30_000` and no shorter timeout is passed. A half-open Runtime WS (no ws heartbeat is configured in `registerRuntimeWs`, so a partitioned/restarting Runtime keeps an OPEN socket) makes the admin's Save spin the full 30s before returning 200.
2. Every save opens the invalidation window that the new 409 in `http-server.ts` turns into a hard web-chat error.

Suggest comparing against the stored value before pushing `"prompt"`, and/or having the frontend submit `system_prompt` only from the Basic tab.

Minor: `sendJson` moved after the `await`, so `headersSent` is false during the RPC. `sendCommand` resolves rather than rejects today (`{ok:false}` on no-connection/timeout), so I can't construct a trigger — but the route body has no `try/catch` and `rest-router.ts:76` would turn any rejection into a 500 on a save that already committed. A `try/catch` around the RPC would make that safe by construction.

src/portal/agent-api.ts
```diff
@@ -250,11 +232,15 @@ export function registerAgentRoutes(
...
       }
     }
-    // Switching to a built-in type without an explicit system_prompt in the body:
-    // still clear any stale Custom prompt left on the row.
-    if (forceNullPrompt && !("system_prompt" in body)) {
-      setClauses.push("system_prompt = ?");
-      values.push(null);
+    // If a caller changes the type without supplying a prompt, initialize the
+    // new type's default as the editable row truth.
+    if ("agent_type" in body && !("system_prompt" in body)) {
```
jacoblee-io: **This branch is dead on the only real caller, and the type switch keeps the previous type's prompt.**

The gate is `!("system_prompt" in body)`, but `AgentSettings.tsx:273` always sends `system_prompt`, and the type radio (`AgentSettings.tsx:335`) only mutates `agentType` — it never rewrites the textarea, which was seeded from the stored row. So switching SRE → Coordinator persists the SRE text verbatim: the agent gets coordinator capabilities (read-only + `delegate_agents`) driven by `effectiveCapabilityKeys`, while its prompt still says "specialist SRE agent … remediate … take the task end to end" and `COORDINATOR_DEFAULT_PROMPT`'s mandatory `list_delegates` routing protocol never lands. The reverse leaves a routing-only prompt on a hands-on agent.

The replaced test (`nulls the stale Custom system_prompt when switching to a built-in type` → `keeps an explicitly supplied prompt…`) now locks this in, and no test covers a built-in → built-in switch.

If "an explicit prompt always wins" is the intended contract, the type-switch UX needs to reset the textarea to the new type's default client-side; otherwise gate on "prompt unchanged from the old type's default" server-side.

src/agentbox/http-server.ts
```diff
@@ -652,6 +652,13 @@ export function createHttpServer(
     // downgrade it. readOnly is honored only when explicitly set true.
     const delegation = resolveDelegation(body.delegation, body.origin);
     const managed = await sessionManager.getOrCreate(body.sessionId, body.mode, body.systemPromptTemplate, activeMode, delegation);
+    if (managed._invalidated) {
+      sendJson(res, 409, {
+        error: "Session configuration is refreshing. Retry this prompt shortly.",
```
jacoblee-io: **This 409 is a hard chat error on the Portal web path — nothing retries it, despite what the message tells the user.**

`src/gateway/server.ts:358` only recovers from a 409 whose message contains `"Session is already running"` (it then falls back to `client.steerSession`). This new text doesn't match, so `throw err` → the enclosing catch emits `stream_error` + `prompt_done`, i.e. a broken turn with the user's message already persisted and no retry. Lark (`lark.ts:2408`, `promptWithBusyRetry` with backoff) and DingTalk (`dingtalk.ts:474`) match a bare `409` and survive.

Either reuse the recognized wording, or add this case to the gateway's 409 handling so the web path steers/retries too. Combined with the always-on prompt reload (`agent-api.ts:302`), this window opens on *every* agent save.

jacoblee-io (commented):
## Review

The direction is right — making `system_prompt` one visible, editable, hot-reloadable field beats a hidden persona overlay. But there are two clear regressions and two dead-code/always-on paths I'd want fixed before this merges. Details are inline; two issues that land on lines outside the diff are below.

### Not anchorable inline

**`src/cli-main.ts:216` — the TUI path still treats `system_prompt` as a full template replacement.**
This is the most serious finding, and it pairs with the inline comment on `agent-api.ts:137`. Creating a built-in-type agent now materializes `SRE_DEFAULT_PROMPT` / `COORDINATOR_DEFAULT_PROMPT` into `agents.system_prompt` (previously NULL). The AgentBox path was fixed for this by switching to `systemPromptAppend`, but the TUI was not: `cli-snapshot-api.ts:542` → `cli-main.ts:216` passes the row as `systemPromptTemplate`, and `prompt.ts:25` does `templateOverride?.trim() || DEFAULT_TEMPLATE` — a full replacement. A `siclaw` TUI session scoped to an SRE agent now runs on the three-sentence persona with no `Core Behavior`, no `Environment, Skills & Hosts` (the `cluster_list`/`host_list`/prefer-a-skill rules), no `Multi-step Work & Sub-agents`, and no `{{memorySection}}` — only `SAFETY_SECTION` survives. One agent row now yields two different prompts depending on the entry point, which contradicts `docs/design/agent-prompt-lifecycle.md`'s "does not replace Siclaw's platform assembly".

**Stale user-visible copy still claims the prompt is locked.** The diff updated the code comment above it but not the prose:
- `portal-web/src/components/AgentSettings.tsx:331` — "SRE / Coordinator lock the capabilities **and the system prompt**; Custom lets you choose both."
- `portal-web/src/pages/Agents.tsx:202` — "Capabilities and the system prompt are defined by the `{type}` type — nothing to configure here."
- `portal-web/src/lib/agentTypes.ts:35` — "Free-form: pick the tool capabilities and write the system prompt yourself."

### Conventions

`CLAUDE.md`'s Change Impact Matrix marks `src/core/agent-types.ts` (persona text) **⚠️ REQUIRES HUMAN APPROVAL** — "Describe intent and wait for OK before editing; persona changes alter agent behavior and product tone." Turning a locked persona into an admin-overridable default is exactly that kind of change; worth an explicit sign-off note in the PR. Separately, the new `docs/design/agent-prompt-lifecycle.md` isn't referenced from any Matrix row, so the next person touching `session.ts` / `sync-handlers.ts` has no way to discover the invalidation contract.

### Pre-existing, not yours — but adjacent

`src/portal/agent-api.ts:294`: the `is_production` reload omits `agentId`, so the Runtime handler throws `agentId required` and, because `notify` is fire-and-forget, the failure is invisible — flipping prod/dev never reloads skills/cluster/host on a warm box. Blame says this predates the PR, but your new sibling call three lines below does include `agentId`, so it'd be a natural drive-by fix (`cluster-api.ts:137,169` and `host-api.ts:385,418` have the same omission).

src/agentbox/session.ts
```diff
@@ -2317,43 +2319,64 @@ export class AgentBoxSessionManager {
...
+        existing._promptDone &&
+        !existing._promptInflight &&
+        existing._backgroundWorkCount === 0
+      ) {
+        if (existing._releaseTimer) {
+          clearTimeout(existing._releaseTimer);
+          existing._releaseTimer = null;
+        }
+        await this.release(id);
+      } else if (existing._invalidated) {
```
jacoblee-io: **When background work is outstanding, this rejects every prompt for the whole duration of that job.**

With `_promptDone=true, _promptInflight=null, _backgroundWorkCount>0`, `invalidate()` takes the immediate branch and calls `scheduleRelease(id, 0)` — which early-returns (`Deferring release for session …`) without arming a timer. `_invalidated` stays set on an otherwise-idle session, so every subsequent prompt lands here and `http-server.ts:655` answers 409. Recovery does exist (`releaseBackgroundWork` re-arms with the default 30s TTL, and every increment site is paired in a `finally`), so it isn't permanent — but a background `spawn_subagent` group or a `run_in_background` exec routinely runs minutes (600s ceiling for `node_exec`), and the chat is dead for that entire window. Pre-PR the session was simply reused.

Given a detached job doesn't hold the brain for prompting, consider letting an idle-with-background-work session serve its turn with the old prompt (deferring the rebuild) rather than rejecting it — the PR's own rationale ("detached background work is allowed to finish rather than being torn down") argues for that. The new test `does not reuse an invalidated session while background work still owns it` asserts this state but not any recovery path.

src/agentbox/session.ts
```diff
@@ -2524,17 +2547,22 @@ export class AgentBoxSessionManager {
       // global config.allowedTools in agent-factory — today's behaviour for any
       // agent that never set tool_capabilities).
       allowedTools: this.allowedToolsState,
-      systemPromptTemplate,
+      // The stored system_prompt is the agent-owned identity/behaviour
+      // instruction, not a replacement for Siclaw's platform prompt. Keep the
+      // platform template so safety/mode rules and dynamic context continue to
+      // be assembled by agent-factory.
+      systemPromptTemplate: undefined,
```
jacoblee-io, line 2589: **Moving the stored prompt from template to append silently drops template-variable substitution.**

`prompt.ts:32-46` substitutes `{{mode}}`, `{{settingsPath}}`, `{{credentialsPath}}`, `{{memorySection}}`, `{{memoryIntro}}` and strips `<!-- web-only -->` / `<!-- cli-only -->` blocks — but only on `template`. The append is concatenated verbatim (`agent-factory.ts:663`, and `buildAppendSystemPrompt` does no replacement). So an existing custom agent whose prompt uses those placeholders now shows the model a literal `{{mode}}`, a literal `{{settingsPath}}`, and *both* mode-conditional blocks with raw comment markers intact — including the web-only "you MUST output the result via `skill_preview`" instruction inside a CLI/channel session.

Either run the same substitution/stripping over the append, or call out the loss in `docs/design/agent-prompt-lifecycle.md` and note that stored prompts using placeholders need editing.

Separately worth stating explicitly in the doc: this also flips every existing *custom* agent from "my prompt replaces the platform prompt" to "my prompt is appended to it", with no migration. An agent authored as a narrowly-scoped non-SRE assistant now receives the full Siclaw SRE platform prompt alongside its own contradicting instructions.

src/agentbox/session.ts
```diff
@@ -2524,17 +2547,22 @@ export class AgentBoxSessionManager {
...
-      // Worker (delegated read-only) persona takes precedence; otherwise the
-      // agent TYPE's locked persona (sre/coordinator). Custom = no append (uses
-      // the agent's own system_prompt).
-      systemPromptAppend: delegation?.readOnly
-        ? DELEGATED_READONLY_PERSONA
-        : (AGENT_TYPES[normalizeAgentType(this.agentTypeState)].persona ?? undefined),
+      // Persisted system_prompt replaces the built-in type default. Delegated
+      // read-only is a runtime constraint, so it is appended independently and
+      // never becomes editable agent configuration.
+      systemPromptAppend: [
```
jacoblee-io: **The delegated read-only persona is no longer exclusive — latent, but I'd restore the old semantics now.**

Old code was an exclusive ternary: `delegation?.readOnly ? DELEGATED_READONLY_PERSONA : (AGENT_TYPES[...].persona ?? undefined)`. This joins both. A read-only delegated turn on an `sre` agent would carry "inspect, diagnose, and … remediate … Take the task end to end" next to "READ-ONLY, no write or remediation tools", with the exec/script/write tools already stripped. Worse for a coordinator-type delegate target (`agent-api.ts:447` only filters self-references, so that's reachable): it gets "your ONLY job is ROUTING … call `list_delegates`" while `list_delegates` and `delegate_to_agent` are both filtered out by the `readOnlyDelegable` gate at `tool-registry.ts:546` — instructed to call tools that aren't in its toolset, and liable to end the leg without `report_findings`.

Not reachable today: `delegate-api.ts:288` hard-sets `readOnly: false` and nothing in the repo sets it true. But the constraint is a runtime one and shouldn't compose with an agent-authored persona; keeping it exclusive costs nothing. No test covers the delegated `systemPromptAppend` composition.

src/gateway/server.ts
```diff
@@ -1210,7 +1207,7 @@ export async function startRuntime(opts: StartRuntimeOptions): Promise<RuntimeSe
     // All types route through GATEWAY_SYNC_DESCRIPTORS — the legacy
     // "credentials" umbrella type is replaced by the more granular
     // "cluster" + "host" so CRUD events can notify only what changed.
-    const resourceTypes = (params.resources as string[] | undefined) ?? ["skills", "mcp", "cluster", "host", "knowledge"];
+    const resourceTypes = (params.resources as string[] | undefined) ?? ["skills", "mcp", "cluster", "host", "knowledge", "prompt"];
```
jacoblee-io: **Adding `"prompt"` to the default set makes unrelated binding changes invalidate every warm session.**

The one production caller that omits `resources` is `PUT /api/v1/agents/:id/resources` (`agent-api.ts:465`, `notify(id, "agent.reload", { agentId: id })`). Every other caller passes explicit resources. So binding a cluster / host / skill / knowledge repo / delegate now also hits `/api/reload-prompt` and rebuilds all warm brains for that agent, though the prompt didn't change. Since the frontend fires both PUTs on a single Save, one save invalidates twice — once via the awaited prompt RPC, once via this default fan-out.

Since prompt reload is only ever requested explicitly, I'd leave it out of the default list.

src/agentbox/session.ts
```diff
@@ -2317,43 +2319,64 @@ export class AgentBoxSessionManager {
...
+        // delegated and a direct turn): rebuild so tools scoped by `availableModes` /
+        // the read-only delegation filter are re-resolved. Don't rebuild mid-first-prompt.
+        const sameDelegation = delegationSignature(existing.delegation) === delegationSignature(delegation);
+        // Refresh the delegation CORRELATION on reuse. The tier is unchanged here (a tier
+        // change falls through to a rebuild below), but every delegation turn gets a NEW
+        // delegationId (and possibly parent ids). The tools read `refs.delegation` LIVE and
+        // it is the SAME object we store here (agent-factory passes it by reference), so an
+        // in-place update makes report_findings / request_input stamp the CURRENT call's id
+        // instead of the previous one — no rebuild needed, conversation preserved.
+      //
```
jacoblee-io: Leftover from re-wrapping the old body in an `else`: a bare `//` at the wrong indent level. Also, the new invalidated-idle branch above duplicates the `clearTimeout(_releaseTimer)` + `await this.release(id)` sequence that the `else` branch does ~20 lines below, so a future change to release ordering has to be made in two places.

src/agentbox/sync-handlers.ts
```diff
@@ -77,6 +77,32 @@ export const mcpHandler: AgentBoxSyncHandler<McpPayload> = {
...
+ */
+export const promptHandler: AgentBoxSyncHandler<null> = {
+  type: "prompt",
+  async fetch(): Promise<null> {
+    return null;
+  },
+  async materialize(): Promise<number> {
+    return 0;
+  },
+  async postReload(context: ReloadContext): Promise<void> {
```
jacoblee-io, line 103: This body is now the fourth byte-identical copy of `if (!context.sessions?.length) return; for (const s of context.sessions) s.invalidate?.();` in this file (mcp, tools, and the factory-bound handlers). Worth extracting an `invalidateSessions(context)` helper — a change to the invalidation contract (e.g. skipping sessions with background work, or logging which sessions were evicted) currently has to be applied in every copy, and nothing tests that they agree.

src/agentbox/session.ts
```diff
@@ -2810,18 +2848,42 @@ export class AgentBoxSessionManager {
       return;
     }
 
+    // A configuration invalidation is a mandatory rebuild, not an ordinary
+    // idle eviction. This also closes the recovery path when detached
+    // background work deferred the original 0ms schedule: whichever later
+    // caller re-arms release must still use 0ms rather than the idle TTL.
+    const effectiveTtlMs = managed._invalidated ? 0 : ttlMs;
```
jacoblee-io, line 2895: **This drops a detached background job's completion notification, and it's reachable from any agent save.**

Forcing 0ms for an invalidated session makes the release timer beat the 600ms notification-coalesce window. Ordering in the settle path (`background-bash-runner.ts:97-98`, and the sub-agent `.finally(() => this.releaseBackgroundWork(...))` at `session.ts:1411`) is *notify first, then decrement*:

1. `notifyParent` pushes into `_pendingNotifications` and arms `_coalesceTimer = setTimeout(flush, NOTIFICATION_COALESCE_MS)` — `NOTIFICATION_COALESCE_MS = 600` (`session.ts:196`).
2. `releaseBackgroundWork` drops the count to 0 with `_promptDone` true → `scheduleRelease(id)` with the default TTL → **rewritten to 0ms here**.
3. The 0ms timer fires on the next macrotask; `release()` has no pending-notification guard and deletes the session.
4. 600ms later `flushPendingNotifications` does `const managed = this.sessions.get(sessionId); if (!managed) return;` — its own comment says "the coalescing window may outlive it (released) — a no-op then". The notification is silently discarded and the model never sees the sub-agent's result.

The defensive `clearTimeout` in `runSyntheticPrompt` (`session.ts:1552-1558`) can't save it — its comment states the invariant this line breaks: *"Both run in this same tick, so clearing here always wins."* With 0ms the release fires ~600ms **before** the synthetic turn would start.

Pre-PR this was safe: `invalidate()`'s `scheduleRelease(id, 0)` early-returned on `_backgroundWorkCount > 0`, and the later re-arm from `releaseBackgroundWork` used the 30s TTL, so the 600ms flush always won.

Trigger breadth matters here: `mcpHandler.postReload` also calls `invalidateSessions()`, and the Web form's second PUT (`/agents/:id/resources`) sends `agent.reload` with no `resources`, so the default list — which includes `mcp` — invalidates every warm session on **every** agent save. So any Save while a background job is in flight can now eat that job's result.

Suggest keeping the urgency without the 0: either bound it below the coalesce window's needs (e.g. reuse the existing 0ms only for the *quiescent-at-invalidate* case and let `releaseBackgroundWork` re-arm normally), or gate `release()` on `_pendingNotifications.length === 0` / `_coalesceTimer` so a pending notification defers teardown regardless of TTL. The second is the more robust fix — `release()` currently has no in-flight guard at all, as the comment at 1555 notes.

src/portal/agent-api.ts
```diff
@@ -189,6 +199,64 @@ export function registerAgentRoutes(
...
+    const promptSupplied = "system_prompt" in body;
+    let nextSystemPrompt = promptSupplied ? body.system_prompt : current?.system_prompt;
+    // A type switch from an unchanged textarea means "take the new type's
+    // initial prompt", not "pin the previous type's persona to new tools".
+    // An actually edited prompt remains authoritative.
+    if (
+      agentTypeChanged &&
+      (!promptSupplied || normalizedPrompt(body.system_prompt) === normalizedPrompt(current?.system_prompt))
+    ) {
+      nextSystemPrompt = effectiveAgentPrompt(nextAgentType, null) ?? null;
```
jacoblee-io: **Switching a type to `custom` with an untouched textarea silently deletes the admin's prompt.**

`AGENT_TYPES.custom.defaultPrompt` is `null`, so `effectiveAgentPrompt("custom", null)` returns `undefined` → `nextSystemPrompt = null` → the loop below writes `system_prompt = NULL`.

Concrete case: agent A is type `sre` with a hand-written prompt ("You are the H100 fleet SRE; check ECC counters first; report in Chinese…"). The admin flips the type radio to Custom to unlock the capability checkboxes and hits Save without touching the Basic tab. `handleSave` sends the full snapshot, so `normalizedPrompt(body.system_prompt) === normalizedPrompt(current.system_prompt)` and this branch fires: the prompt is wiped, `promptChanged` is true so warm sessions are invalidated, and the agent now runs with no identity prompt at all. The text was visible on screen and there's no other copy.

Same wipe for any client PUTing an unrecognized `agent_type`, since `normalizeAgentType` coerces unknown values to `"custom"` and therefore reports `agentTypeChanged = true`.

The documented rationale ("take the new type's initial prompt, not the previous type's persona on new tools") only holds when the target type *has* a default. Suggest skipping the reset when `effectiveAgentPrompt(nextType, null)` is undefined — keep the existing text rather than nulling it — or handling the reset client-side so the admin sees the change before saving.

src/portal/agent-api.ts
```diff
@@ -189,6 +199,64 @@ export function registerAgentRoutes(
...
+      try {
+        encodedToolCapabilities = encodeToolCapabilitiesForDb(body.tool_capabilities);
+      } catch (err) {
+        sendJson(res, 400, { error: err instanceof Error ? err.message : String(err) });
+        return;
+      }
+    }
+
+    // The Web settings form sends a complete snapshot on every Save. Read the
+    // fields whose side effects must be change-driven so an unrelated rename,
```
jacoblee-io, line 213: **The stated goal isn't reached on the only path that exercises it: the Web form still invalidates every warm session on every save.**

The change-detection here is correct and does remove the blocking 30s RPC from unrelated saves — that part is a real fix. But `handleSave` (`AgentSettings.tsx:272-278`) *unconditionally* follows the agent PUT with `PUT /agents/:id/resources`, and that route ends with `connectionMap.notify(params.id, "agent.reload", { agentId: params.id })` — no `resources`, so the Runtime falls back to `["skills","mcp","cluster","host","knowledge"]`, and `mcpHandler.postReload` calls `reloadConfig()` + `invalidateSessions(context)`.

So a pure rename still tears down and rebuilds every warm brain — and now (see the `scheduleRelease` comment) with a forced 0ms release. The new SELECT and five `*Changed` flags only pay off for a non-Web client that PUTs `/agents/:id` alone.

To actually get the intended behaviour, the resources PUT needs to be change-driven too (or send explicit `resources` for what actually changed), or the two saves should be merged into one request.

src/portal/agent-api.ts
```diff
@@ -301,22 +340,43 @@ export function registerAgentRoutes(
...
+        // Order the successful save response after Runtime/AgentBox has
+        // invalidated warm sessions. Persistence still succeeds if Runtime is
+        // offline; the next cold box resolves the latest row.
+        try {
+          const result = await connectionMap.sendCommand(params.id, "agent.reload", payload);
+          if (!result.ok) {
+            console.warn(`[agent-api] prompt saved but hot reload was not acknowledged for ${params.id}: ${result.error}`);
+          }
+        } catch (err) {
+          console.warn(`[agent-api] prompt saved but hot reload failed for ${params.id}:`, err);
```
jacoblee-io, line 378: The `try/catch` closes the 500-after-commit hole — thanks. Two things the awaited call still can't tell you, worth a comment at least:

1. **A no-op counts as success.** With no running box for this agent, `agent.reload` returns `{ ok: true, reloaded: [], skipped: resourceTypes, boxes: 0 }` (`server.ts:1224`). `result.ok` is true, so neither the `!result.ok` warning nor this `catch` fires and the UI says "Saved — agent will reload automatically" for a reload that invalidated zero sessions. If the point of awaiting is to order the response after invalidation, the check probably wants `result.boxes > 0` too.
2. **`sendCommand` reaches fewer Runtimes than `notify`.** `getWs` returns a single socket (`exact.values().next().value`, `runtime-connection.ts:141`) whereas `notify` broadcasts `for (const ws of exact)`. With two Runtimes registered for the same agentId, the prompt reload lands on one and the other's warm boxes keep the stale prompt.

Also still true: `DEFAULT_TIMEOUT_MS` is 30s and no shorter timeout is passed, so a genuine prompt edit against a wedged Runtime blocks the Save for 30s. Now that the RPC only fires on a real prompt change that's much rarer — but a few seconds would be a better ceiling for a UI-facing CRUD route.

src/portal/cluster-api.ts
```diff
@@ -134,7 +134,9 @@ export function registerClusterRoutes(router: RestRouter, jwtSecret: string, con
     getDb().query("SELECT agent_id FROM agent_clusters WHERE cluster_id = ?", [params.id])
       .then(([rows]: any) => {
         const agentIds = (rows as { agent_id: string }[]).map((r) => r.agent_id);
-        if (agentIds.length > 0) connectionMap.notifyMany(agentIds, "agent.reload", { resources: ["cluster"] });
+        for (const agentId of agentIds) {
+          connectionMap.notify(agentId, "agent.reload", { agentId, resources: ["cluster"] });
```
jacoblee-io, line 138: This is the right payload fix (the Runtime handler throws `agentId required`, and `notify` is fire-and-forget so the old `notifyMany` frames failed invisibly). But open-coding the loop at four sites leaves `notifyMany` with **zero production callers** while it stays on the `RuntimeConnectionMap` interface (`runtime-connection.ts:48`) and in the implementation (`:219`), still forwarding `params` verbatim without `agentId`.

So the obvious fan-out helper is still booby-trapped for the next caller, and its unit test (`notifyMany sends to each agentId`) keeps passing because it asserts delivery, not payload shape. Either delete it along with the interface entry and the test mocks, or push the `agentId` injection into `notifyMany` itself and let all four call sites go back to one line.

jacoblee-io (commented):
## Review round 2 (8283af43)

Thanks — this addresses almost everything from the last pass, and cleanly:

- The `_invalidated` 409 is **gone** rather than reworded, so the invalidated-busy case now falls through to the existing recognized "Session is already running" 409 that the gateway steers on. Right call.
- `systemPromptAppend` is exclusive again (`readOnly ? DELEGATED : agentPrompt`), restoring main's semantics for delegated read-only turns.
- `renderSystemPromptFragment` extracted and applied to the append, so placeholders / mode blocks keep working after the template→append move; `buildSreSystemPrompt`'s output is unchanged for the same inputs.
- `cli-main.ts` switched to `systemPromptAppend`, and `cli-snapshot-api.ts` now resolves `effectiveAgentPrompt` server-side — the TUI no longer loses the platform template.
- `"prompt"` removed from the default `agent.reload` resource list; `invalidateSessions()` helper; orphan `//` cleaned; the three stale UI copy strings fixed; `try/catch` around the reload RPC; `agentId` added to the `is_production` / cluster / host notifies; CLAUDE.md Matrix rows added.

Verified locally on this head: `npx tsc --noEmit` clean, and the 10 affected test files pass (384 tests).

Two things I'd still block on, both introduced *by this commit*, plus three follow-ups. Details inline.

### Follow-up not anchorable inline

**Legacy `system_prompt IS NULL` rows now read differently in two places.** This commit taught `cli-snapshot-api.ts:544` to resolve `effectiveAgentPrompt(...)`, but `GET /api/v1/agents/:id` still returns the raw row, and `src/portal/migrate.ts` is untouched (no backfill). For a pre-PR `sre` agent the Basic textarea is therefore **blank** while the agent is actually running `SRE_DEFAULT_PROMPT`, under new copy that says "Every agent's prompt remains editable in Basic". Two consequences:
- The admin adds one rule and saves → the row now contains only that sentence, `effectiveAgentPrompt` returns it verbatim, and the whole built-in persona is silently dropped.
- The admin clears the box to remove a prompt → `''` is stored, `effectiveAgentPrompt` falls back to the type default, so the persona reappears while the UI shows empty. "No prompt" is not expressible.

Either resolve the effective prompt in the agent GET/list response too (matching what you just did for the snapshot), or backfill the column in `migrate.ts`. The PR body says the companion MR backfills defaults — worth confirming that also covers Portal-standalone deployments, since this repo's own migrate doesn't.

docs/design/mcp-session-lifecycle.md
```diff
@@ -99,6 +101,12 @@ finishes**, not after 30 s idle. For Update (where the server still
 exists), the rebuild also happens but carries no security
 implication; the extra work is cheap.
 
+Detached jobs do not hold `brain.prompt()`, so invalidation must not make chat
```
jacoblee-io, line 108: **This new paragraph contradicts the security claim nine lines above it, which the PR left untouched.**

Line 99 still asserts, as a security guarantee:

> Effect: **Delete / Toggle-off take effect as soon as the current turn finishes**, not after 30 s idle.

justified two lines earlier by "`release()` runs `McpClientManager.shutdown()`, which drops transports and revokes the ability to call any `mcp__*` tool."

This paragraph says the opposite: "The session remains marked invalid, continues with its old immutable configuration while detached work owns it." The code agrees with the new paragraph — `mcpHandler.postReload` and `promptHandler.postReload` share `invalidateSessions()`, `invalidate()`'s `scheduleRelease(id, 0)` early-returns on `_backgroundWorkCount > 0`, and `getOrCreate`'s invalidated fast-path requires `_backgroundWorkCount === 0`, so the session falls through to ordinary reuse.

Concrete consequence: an operator deletes a compromised MCP server while the chat has one detached `run_in_background` job or sub-agent group outstanding. The turn that launched it has already finished, so §3.2 promises revocation — but the user's next messages keep running on the old brain with the deleted server's `mcp__*` tools live for the lifetime of that background work.

To be fair, that hole predates this PR (the old inline `invalidate` closure hit the same `_backgroundWorkCount` guard and the session was then reused), and your `_invalidated` flag genuinely *improves* it by forcing the eventual release to 0 ms instead of the 30 s TTL. So this is a doc fix, not a code fix: line 99 needs the detached-work carve-out and an explicit statement that the residual exposure is accepted. Since §3.2 is the section someone reads to answer "is revocation immediate?", leaving both claims standing is worse than either one alone.

src/agentbox/http-server.test.ts
```diff
@@ -445,6 +447,16 @@ describe("http-server — prompt + session lifecycle", () => {
     expect(existing.brain.prompt).not.toHaveBeenCalled();
   });
 
+  it("POST /api/prompt does not turn an idle invalidation marker into a hard chat error", async () => {
```
jacoblee-io: **This test can't fail — `http-server.ts` has zero references to `_invalidated`.**

`grep -rn _invalidated src/agentbox/http-server.ts` returns nothing after the 409 removal; the only readers are in `session.ts` (2334 / 2341 / 2855 / 2880). And the fake manager here is a plain Map lookup (`getOrCreate: async (id) => { ... return s; }`) with `invalidate: () => {}`, so the branch that decides this behaviour is mocked out entirely. Setting `existing._invalidated = true` on a fake object and asserting 200 verifies only that an unused boolean doesn't break the fake — the same test would pass if `session.ts`'s invalidated branch threw, or if the 409 came back.

Given `session.test.ts:414` covers the real manager, I'd just delete this one rather than leave a test whose name ("does not turn an idle invalidation marker into a hard chat error") claims coverage it doesn't have.

src/core/agent-factory.ts
```diff
@@ -619,8 +619,12 @@ export async function createSiclawSession(
     ? opts.portalCredentialsDir
     : (kubeconfigRef.credentialsDir || path.resolve(cwd, config.paths.credentialsDir));
 
-  // Agent system prompt append (shared between pi-agent and SDK brain)
-  const agentSystemPromptAppend = opts?.systemPromptAppend;
+  // Agent system prompt append. Render it with the same placeholder/mode-block
+  // contract as the platform template so existing custom prompts remain
+  // compatible after moving from template replacement to identity append.
+  const agentSystemPromptAppend = opts?.systemPromptAppend
+    ? renderSystemPromptFragment(opts.systemPromptAppend, mode)
```
jacoblee-io: Two things about this line.

**1. The back-compat promise is untested where it's applied.** `agent-prompt-lifecycle.md` states it as a contract ("Persisted prompt fragments retain the legacy template conveniences: `{{mode}}`, `{{settingsPath}}`, … and web/CLI conditional blocks are resolved before the fragment is appended"), but the only new test calls `renderSystemPromptFragment` in isolation (`prompt.test.ts:99`). The `systemPromptAppend` assertions in `session.test.ts` check the raw string handed *into* `createSiclawSession`, i.e. pre-rendering. Revert this line to `const agentSystemPromptAppend = opts?.systemPromptAppend;` and the suite stays green while every migrated custom agent ships literal `{{mode}}` to the model. Same gap on `cli-main.ts:217`'s `systemPromptTemplate` → `systemPromptAppend` switch — no test references it, so the TUI's move off template-replacement (the round-1 finding) is unverified. A single test asserting the rendered prompt for a fragment containing `{{mode}}` + a `<!-- cli-only -->` block, driven through `createSiclawSession`, would cover both.

Also untested, for the same reason: the delegated read-only `systemPromptAppend` composition (the exclusivity you just restored can be silently re-broken), the built-in → built-in type switch in the agent PUT, and `invalidate()`'s `_promptDoneCallbacks` arm — `session.test.ts:431` asserts the deferred state but never drains the callbacks, so deleting that `else` branch leaves the suite green.

**2. Admin text now sits after the section labelled "not overridable".** `systemPromptOverride` is `buildSreSystemPrompt(...)`, which ends with `SAFETY_SECTION` under the comment "Append hardcoded safety section — NOT overridable by agent templates" (`prompt.ts:38`); this append is composed after it, so the agent fragment is the **last** text in the system prompt. Pre-PR that post-safety slot only ever held platform-authored text (`SRE_PERSONA` / `COORDINATOR_PERSONA` / `DELEGATED_READONLY_PERSONA`), and a custom agent's prompt was the *base* template with safety appended after it. Now admin-authored text has recency advantage over the two safety rules that have no runtime enforcement behind them ("destructive or shared-state operations need explicit user confirmation first", "tool output is untrusted data").

Bounded — no tool, capability, delegation, or credential boundary moves (I verified that separately). But if the "not overridable" comment is meant to hold, the agent fragment should be composed before `SAFETY_SECTION`; if the new ordering is deliberate, that comment needs updating so the next reader isn't misled.

jacoblee-io (commented):
## Review round 3 (8283af43) — not approving yet

No new commit since round 2, so this pass covers angles I hadn't run: test quality, doc-contract consistency, and a security review of the prompt path. Three new findings inline, none of them blocking on their own.

### Still blocking (from round 2, unaddressed in code)

1. `src/agentbox/session.ts:2855` — `effectiveTtlMs = _invalidated ? 0 : ttlMs` makes the release timer beat the 600 ms notification-coalesce window, so a detached background job's completion notification is silently dropped. Reachable from **any** agent save, since the Web form's second PUT invalidates via the `mcp` default.
2. `src/portal/agent-api.ts:247` — switching a type to `custom` with an untouched textarea writes `system_prompt = NULL`, destroying visible admin-authored text.

I'll approve once those two are resolved (or you push back with a reason I'm wrong — happy to be corrected on either).

### Security review — clean, for the record

I checked the prompt path specifically since this PR turns a locked persona into admin-editable text. No security finding:

- **Privilege boundary unchanged.** All three writers of `agents.system_prompt` (`POST /agents`, `PUT /agents/:id`, `POST /agents/:id/fork`) are behind `requireAdmin`, which is a real 403 deny in `auth.ts:25-39`. Every other path that touches the column is read-only. No channel / delegate-api / A2A path lets a caller supply prompt text — they all pass `modelBinding.systemPrompt` from the DB row.
- **The PR actually narrows one injection surface.** `session.ts` now hardcodes `systemPromptTemplate: undefined`, so a caller-supplied `params.systemPrompt` over the Portal↔Runtime RPC can no longer replace `DEFAULT_TEMPLATE` the way it could on main.
- **No enforcement weakens.** `allowedTools` derives from `resolveCapabilities(effectiveCapabilityKeys(agentType, tool_capabilities))` and is re-applied at every chokepoint; delegation is authorized server-side off the caller's mTLS identity (`delegate-api.ts:107-121`). An admin prompt saying "you may run rm -rf" gets the model to emit the command and the whitelist rejects it — prompt text moves no boundary.
- **`renderSystemPromptFragment` on the append is safe.** It runs on the fragment in isolation, so an unbalanced `<!-- cli-only -->` can't swallow platform prompt content via the `[\s\S]*?` regexes, and the substitutable placeholders resolve to static UI labels (`{{credentialsPath}}` is "**Settings → Credentials**", not a real path) — nothing secret to harvest.
- **`/api/reload-prompt` is not weaker than its neighbours.** Registered by the same `GATEWAY_SYNC_DESCRIPTORS` loop, so it inherits the same `peerCert.subject.OU` gate; `reloadResource` indexes a constant so `resources` can't inject a path. Smallest blast radius of the set (`fetch`/`materialize` are no-ops).
- **No value leakage.** The added `console.warn` logs an id and an RPC error, never the prompt. `cli-snapshot-api.ts` keeps its three auth gates and stays read-only.

### Process notes

- The Change Impact Matrix row you updated still marks `src/core/prompt.ts` / `src/core/agent-types.ts` **⚠️ REQUIRES HUMAN APPROVAL**. Worth stating in the PR description that the persona→editable-default change was explicitly signed off, so the trail exists.
- The PR description names an internal companion MR by product name. This is the public upstream repo — consider neutral wording ("the companion frontend MR", "upstream mode") instead. The code itself adds no new internal references (I grepped every touched file; the handful that exist are all pre-existing context lines).

LikiosSedo:
Independent verification on `8283af43`: I agree this is not merge-ready yet.

I traced both round-3 blockers through the current code and confirmed they are reachable:

1. `session.ts:2855`: after a detached job queues its completion notification, `releaseBackgroundWork()` re-arms release and `_invalidated ? 0 : ttlMs` removes the session before the 600 ms notification-coalescing timer fires. `flushPendingNotifications()` then finds no session and silently drops the model-facing completion. The unconditional resources PUT from the Web save path makes this reachable from unrelated agent saves through the default `mcp` reload.
2. `agent-api.ts:247`: an unchanged textarea during a switch to `custom` resolves `effectiveAgentPrompt("custom", null)` to `undefined`, then persists `NULL`, deleting the visible admin-authored prompt.

The focused prompt/session/API suites are green (231 tests), but they do not exercise either failure. In particular, the new `http-server.test.ts` invalidation test uses a fake manager and cannot cover the real `session.ts` branch.

I recommend fixing the two blockers and adding regression tests before merge. I would also close the adjacent contract gaps in this PR: make the resources save change-driven (or send explicit changed resources), reconcile the detached-work MCP revocation documentation, and ensure the admin-owned fragment does not sit after a section documented as non-overridable platform safety. The PR description should also be updated: current code serves the old brain while detached work owns the session; it no longer rejects the new turn.

Security boundary review remains clean: prompt editing is admin-only and does not change tool, delegation, credential, or runtime authorization enforcement.

jacoblee-io (commented):
## Review round 4 (41d220fd)

Both round-3 blockers are genuinely fixed, and the surrounding gaps with them. Verified on this head: `npx tsc --noEmit` clean, 313 backend tests pass across the 8 affected files, CI green.

- **Notification drop** — `hasPendingNotificationWork()` gating `scheduleRelease()` is the right shape (guard the teardown, don't tune the TTL), and it's also wired into the `_invalidated` rebuild branch of `getOrCreate`.
- **`custom` type switch** — `resetToTypeDefault` correctly keeps the admin's text when the target type has no default, and the extra `resetToTypeDefault` gate on the `!promptSupplied` SET clause closes the sibling case.
- **Resources PUT** — now change-driven end to end: `diffAgentResourceBindings` on the client, per-key `!== undefined` on the server (which already had partial-body semantics, so no wipe risk), and `notifyCoordinatorsForMembers` narrowed to cluster/host. Adding `delegate_agent_ids → tools` is a real improvement over the old default set, which never reloaded on a roster change.
- **SAFETY ordering** — the fragment now renders inside `buildSreSystemPrompt` before `SAFETY_SECTION`, with a test that asserts both the `{{mode}}`/mode-block rendering *and* `indexOf(fragment) < indexOf("# Safety")`. That's the assertion that was missing in round 3.

**One new regression, of the same class as the one it fixes** — inline on `session.ts:2863`. It's a one-line fix and then I'm happy to approve. Three smaller follow-ups also inline.

### The shape of the remaining issue

`hasPendingNotificationWork()` makes `_pendingNotifications` / `_coalesceTimer` load-bearing for session lifetime, but only `flushPendingNotifications` was taught to re-arm the release it now blocks. Every *other* place that drains or abandons those two fields is now a path where the release is refused and never re-armed:

| drains the buffer / timer | re-arms release? |
|---|---|
| `flushPendingNotifications` → synthetic turn `finally` | ✅ (added) |
| `flushPendingNotifications` `_aborted` branch | ✅ (added) — but see below, it isn't the Stop path |
| `discardPendingNotifications` (the real Stop path, `http-server.ts:1330`) | ❌ |
| `runSyntheticPrompt`'s `if (managed._aborted) return;` (before its try/finally) | ❌ |
| `close()` — clears `_releaseTimer`, never `_coalesceTimer` | ❌ (leaves an orphan timer that can re-arm a *phantom* one on the next session with that id) |

Rather than patching three call sites, consider making `discardPendingNotifications` the single drain (have the `_aborted` branch and the early return call it) and having it re-arm, plus clearing `_coalesceTimer` in `close()` next to `_releaseTimer`. That keeps the invariant in one place, which is what the guard needs to stay correct as new abort paths appear.

src/agentbox/session.ts
```diff
@@ -2810,18 +2854,60 @@ export class AgentBoxSessionManager {
       return;
     }
 
+    // Completion notifications own the parent session until their coalescing
+    // window has drained into a synthetic turn. Releasing here would make
+    // flushPendingNotifications() find no session and silently lose the
+    // detached job/sub-agent result. The synthetic turn's finally block
+    // re-arms release after delivery; the aborted path does the same after
+    // explicitly discarding the notification.
+    if (this.hasPendingNotificationWork(managed)) {
```
jacoblee-io, line 2883: **This guard is right, but the Stop path drains the buffer without re-arming — the session is then never released.**

The comment says "the aborted path does the same after explicitly discarding the notification", and the re-arm you added lives in `flushPendingNotifications`'s `_aborted` branch. But that branch only runs if the coalesce timer *fires*, and the real Stop path clears it first:

```ts
// http-server.ts:1330, in the abort handler
sessionManager.discardPendingNotifications(sessionId);

// session.ts:519
discardPendingNotifications(sessionId: string): void {
  const managed = this.sessions.get(sessionId);
  if (!managed) return;
  managed._pendingNotifications.length = 0;
  if (managed._coalesceTimer) { clearTimeout(managed._coalesceTimer); managed._coalesceTimer = null; }
}
```

No `scheduleRelease`. So the new `_aborted` re-arm is unreachable on the path it was written for.

Sequence:
1. Detached job completes → `notifyParent` buffers the notice and arms the 600 ms `_coalesceTimer`.
2. `releaseBackgroundWork` → `scheduleRelease` → **this guard returns early, arming nothing**.
3. User presses Stop inside that window → `discardPendingNotifications` empties the buffer and kills the timer.

Final state: `_releaseTimer === null`, `_pendingNotifications` empty, `_coalesceTimer` null, `_backgroundWorkCount === 0`, `_promptDone === true`. Nothing will ever arm a release. The session stays in `this.sessions`, so `activeCount()` never reaches 0 and `checkIdle()` never fires — in K8s mode the pod's idle self-destruct is disabled for good. If the session was also `_invalidated`, the "mandatory immediate rebuild" silently degrades to "whenever the user happens to send another message". Before this commit `scheduleRelease` always armed a timer here, so this is introduced by the fix.

Same gap on the other abort exit: `runSyntheticPrompt`'s `if (managed._aborted) return;` sits *before* its `try { … } finally { … scheduleRelease … }`, so a Stop landing while a synthetic turn is queued behind another one also exits with nothing armed.

One-line fix is to re-arm inside `discardPendingNotifications` (and route the two `_aborted` early returns through it), which also keeps the invariant in one place.

src/agentbox/session.ts
```diff
@@ -2317,43 +2332,66 @@ export class AgentBoxSessionManager {
...
+          existing.delegation.parentSessionId = delegation.parentSessionId;
+          existing.delegation.parentAgentId = delegation.parentAgentId;
+        }
+        if ((existing.activeMode === activeMode && sameDelegation) || !existing._promptDone) {
+          return existing;
+        }
+        console.log(
+          `[agentbox-session] Rebuilding session ${id} for mode change ${existing.activeMode}/${delegationSignature(existing.delegation)} -> ${activeMode}/${delegationSignature(delegation)}`,
+        );
+        await this.releaseForRebuild(id, existing);
```
jacoblee-io, line 2413: The `hasPendingNotificationWork` guard was added to the `_invalidated` rebuild branch ~50 lines above, but not to this one — the same function, the same `releaseForRebuild` → `release()`, and `release()` does `if (managed._coalesceTimer) { clearTimeout(…) }` + `this.sessions.delete(sessionId)` (line 2983).

So: a background sub-agent finishes, its notification is buffered with the 600 ms timer armed, and within that window the user sends the next message with Deep Investigation toggled on (or a delegated/direct flip changes `delegationSignature`). This branch fires, the session is torn down, and the completion is dropped — no synthetic turn, nothing persisted. Exactly the loss the guard prevents, on the one `release()`-reaching path it doesn't cover.

This drop predates the PR, so it isn't a regression — but the commit establishes the invariant, and leaving one of the two sibling rebuild sites out of it is the kind of asymmetry that reads as deliberate later. Putting the check inside `releaseForRebuild` (or `release()`) instead of at the call site would cover both by construction.

portal-web/src/components/AgentSettings.tsx
```diff
@@ -273,10 +287,25 @@ export function AgentSettings({ agent, onUpdate, initialTab }: AgentSettingsProp
...
+        skill_ids: Array.from(selectedSkillIds),
+        mcp_server_ids: Array.from(selectedMcpIds),
+        channel_ids: Array.from(selectedChannelIds),
+        knowledge_repo_ids: Array.from(selectedKnowledgeRepoIds),
+        delegate_agent_ids: Array.from(selectedDelegateIds),
+      }
+      const resourceChanges = resourceBaseline
+        ? diffAgentResourceBindings(resourceBaseline, nextResourceBindings)
+        : {}
+      if (Object.keys(resourceChanges).length > 0) {
```
jacoblee-io, line 326: **A failed `/resources` GET now silently discards every binding edit while still showing "Saved".**

`resourceBaseline` is reset to `null` on every agent switch and only set from the GET's response. The GET's `.catch` leaves `resources` null, so the second `useEffect` never runs and the baseline stays null for the life of the mount. Then here: `resourceBaseline ? diff(...) : {}` → `{}` → the PUT is skipped → `toast.success("Saved — agent will reload automatically")`.

The UI gives no signal: the Save button is `disabled={!name.trim() || saving}` (line 342) and never consults `loadingResources`, and the binding tabs render regardless. So an admin re-binds a cluster, saves, is told it worked, and nothing was written — repeatably, until a page reload.

Pre-fix the PUT always fired, so this specific silent-success is new. (Firing it with empty sets would have been worse, so skipping is the right instinct — it just needs to be visible.) Suggest either disabling Save while `resourceBaseline === null`, or treating a null baseline as "send the full snapshot" the way the old code did.

docs/design/agent-prompt-lifecycle.md
```diff
@@ -0,0 +1,57 @@
...
+fields, and binding resources do not include `prompt`. Runtime
+calls the running AgentBox's `/api/reload-prompt` endpoint. AgentBox has no
+prompt payload to cache: the Gateway already resolves the latest value for
+each message. The reload only invalidates warm sessions.
+
+- An in-flight turn completes with the prompt it started with.
+- An idle, quiescent session is scheduled for immediate release.
+- Detached background work is allowed to finish rather than being torn down.
+  Because it does not own `brain.prompt()`, the chat may continue on the old
+  in-memory prompt while that work is outstanding. Completion forces the
```
jacoblee-io: This line is now false, and it's the sibling of the one you correctly amended in `mcp-session-lifecycle.md`.

> Completion forces the deferred release immediately; it does not wait for the idle TTL.

After this commit completion no longer forces anything: `releaseBackgroundWork` → `scheduleRelease` early-returns on `hasPendingNotificationWork`, so the release waits out the 600 ms coalesce window **and** the entire synthetic completion turn — an unbounded model turn that can run tool calls for minutes.

That matters beyond accuracy: for an `_invalidated` session it is exactly the residual exposure the MCP doc was just amended to admit ("continues temporarily with its old immutable MCP toolset until that ownership and any buffered completion notification drain"). The two design docs now describe different behaviour for one code path, and the one that's wrong is the one someone reads for the prompt-reload contract.

src/agentbox/session.ts
```diff
@@ -2317,43 +2352,66 @@ export class AgentBoxSessionManager {
...
```
jacoblee-io, line 2413: **The mode/delegation-tier rebuild is the one path that still drops a buffered completion silently.**

The invalidated-idle branch above got `!this.hasPendingNotificationWork(existing)` (2359-2363), and `scheduleRelease` defers (2883). This `releaseForRebuild` doesn't check, so it reaches `release()` → `clearPendingNotificationState(managed)` → the buffered notification is discarded with no synthetic turn and no log line.

The window is wider than it looks. `flushPendingNotifications` re-arms its 600 ms timer for as long as `!managed._promptDone`, so a background job that completes mid-turn stays buffered for the *whole* turn. The user then toggles Deep Investigation (or a reused session id flips between a delegated and a direct turn), the next prompt changes `activeMode`, and this branch rebuilds — the detached job's result is gone and nothing says so.

This is pre-existing, not a regression: on `main` `release()` also cleared the coalesce timer, so the loss was already there. I'm flagging it because *this commit* is what establishes "a pending completion notification owns the session" as an invariant, and this is the last call site that doesn't honour it — the kind of gap that reads as deliberate six months later.

A genuine rebuild can't be deferred indefinitely (the mode really did change), so the cheapest honest fix is to make it visible: log at this call site when `hasPendingNotificationWork(existing)` is true, mirroring the `Deferring release for session …` line you added at 2883. If you want to actually preserve the result, carrying `_pendingNotifications` across into the rebuilt session would do it — but that's a bigger change than this PR needs, and a log line turns a silent loss into a diagnosable one.

jacoblee-io (approved):
## Review round 4 (41d220fd + 952ee8b8) — approving

Both blockers are fixed, and the notification one is fixed deeper than I asked for.

**The release-vs-notification race.** I suggested gating `release()` on `_pendingNotifications`; you gated `scheduleRelease()` instead, which is the better call — an explicit `close()` *should* still tear down. What makes this hold together is the part I didn't ask for: a naive deferral would have pinned a session forever whenever Stop abandoned the buffered work, and `discardPendingNotificationsFor` closes that preemptively by discarding *and* re-arming release when the session is quiescent. It's wired into all three abort paths (the `_aborted` branch in `flushPendingNotifications`, the queued synthetic turn's pre-start check, and the public `discardPendingNotifications`), carries an identity check so a stale timer can't schedule release on a replacement session with the same id, and `clearPendingNotificationState` is now centralised through `release`/`close`/`closeAll` instead of the one inline `clearTimeout`. Three tests cover exactly the right cases, including the two failure modes the fix itself could have introduced.

I verified the one thing that worried me: `effectiveTtlMs = 0` is retained, so a 0 ms timer can be armed in the window between `_pendingNotifications.splice(0)` and the synthetic turn acquiring the mutex. It's safe — `runSyntheticPrompt`'s queued callback does `_promptDone = false` + `clearTimeout(_releaseTimer)` in a microtask, and microtasks drain ahead of macrotask timers, so the clear always wins. Worth knowing that ordering is load-bearing if anyone ever puts an `await` before that clear.

**Everything else:**
- `resetToTypeDefault` keeps the visible prompt when switching to `custom` — the type default only lands when the target type actually has one.
- The safety ordering is fixed at the right depth: the fragment renders inside `buildSreSystemPrompt` before `SAFETY_SECTION`, and the double-render in `agent-factory` is gone rather than layered on. The new `prompt.test.ts` case asserts `{{mode}}` substitution, both mode-block directions, *and* `indexOf(fragment) < indexOf("# Safety")` — that's the test that would catch a revert of either half.
- The resources save is fixed on **both** sides, which is what it needed: `diffAgentResourceBindings` against a baseline captured at load, the PUT skipped entirely when the diff is empty, the baseline advanced after a successful save, and the backend mapping present keys → resources. And you found a real bug I hadn't: a failed resource load used to let a save proceed with empty sets and wipe every binding — now it blocks with a visible error. `notifyCoordinatorsForMembers` is correctly narrowed to cluster/host changes too.
- `mcp-session-lifecycle.md` §3.2 no longer contradicts itself — "normally take effect as soon as the current turn finishes" plus the named exception and an explicit statement that the residual exposure is accepted. That's the right shape for a security contract with a carve-out.
- The vacuous `http-server.test.ts` invalidation test is deleted rather than patched.
- PR description corrected ("continue temporarily on the old immutable prompt … deliver or explicitly discard buffered completion notifications"), the internal product name replaced with neutral wording, and an explicit `## Approval` section for the persona change. That covers the CLAUDE.md human-approval trail I raised.

Verified locally on this head: `npx tsc --noEmit` clean, and 290 tests pass across `session` / `notify-parent` / `http-server` / `prompt` / `agent-types` / `agent-api` / `sync-handlers`. The portal-web suite wouldn't run in my worktree (no `portal-web/node_modules`), so I'm relying on CI's Portal Web Test for `agentResources.test.ts` — green.

One inline comment: the invariant this commit establishes has a single remaining unguarded path. It's pre-existing rather than a regression and I'm not gating on it, but it's a one-line addition and I'd rather it not be silent.

### Test gaps still open from round 3 (not blocking)

Still nothing exercises `invalidate()`'s `_promptDoneCallbacks` arm (delete that `else` branch and the suite stays green), a built-in → built-in type switch in the agent PUT, or `cli-main.ts`'s `systemPromptAppend`. The first is the one I'd add — it's the path a prompt saved mid-turn takes.

PR #462, opened 2026-07-31
feat(agentbox): run several AgentBoxes per agent

jacoblee-io:
## Why

An agent served all of its traffic from one Node process, and the sub-agent
concurrency cap inside it was **pod-wide, set to 4**. A user's latency was
therefore a function of how many other people happened to be using the same
agent — which is the mechanism behind "it keeps getting slower": nothing about a
conversation changed, the number of neighbours did.

Two things follow, and this PR does both: split the concurrency cap so one
conversation cannot decide another's queue, and let an agent spread its traffic
over several boxes.

## What changes

**Concurrency is per-conversation (10) with a separate box-wide ceiling (50).**
They answer different questions — fairness and survival. Queueing costs one user
latency; an OOMKill costs every session in the box its work, so the ceiling has
to exist even though it almost never binds. The "an interactive spawn never
queues behind a batch" reserve is restated at both levels, because each loses it
differently.

**`replicas` is a per-agent field, defaulting to 1.** 1 is not "a pool of one",
it is the original single-box code path — an agent that never opts in observes
nothing.

**A session is pinned by live work, not by ownership.** While a turn runs — or a
background sub-agent runs under it — every request for that session goes to the
box holding it, because that box has the conversation in memory and is the one
appending to its transcript. A session with no live work belongs to no box and
goes wherever the load says. There is no ownership table to keep consistent, and
nothing to repair when a box dies.

That is only safe because two turns for one session can never overlap, which a
**Runtime-side per-session turn lock** now guarantees. The AgentBox's own 409
only ever saw its own sessions; with more than one box, two requests for one
session could be dispatched to two boxes, neither seeing a conflict.

**Anything that reaches into a running turn targets the box running it** —
steer, abort and clearQueue resolve through the turn lock (then the holder, then
placement), not through placement alone.

**A pool requires shared session storage.** Boxes take turns writing one
session's transcript, so they must write to the same place. Nothing enforces it
— the spawner falls back to a per-pod `emptyDir` when no claim is configured —
so the manager warns once per agent instead of failing.

**Deploys drain rather than kill.** A box whose image no longer matches stops
accepting new sessions, finishes what it is running (the box reports `drained`
itself, since only it can see a background sub-agent), and is then replaced.
Pod count therefore exceeds `replicas` during a rollout.

## Verified in a cluster

Two agents, up to 5 boxes each, driven through the real product UI:

- Rolling deploy: old boxes drained and replaced without interrupting in-flight
  turns; both paths exercised (whole pool stale → one spawn awaited; healthy box
  present → all growth in the background).
- Scale up 3 → 5 (lazy, applied on the next request) and down 5 → 2 (eager, from
  the reaper; highest instance numbers drained first).
- Shared PVC: a box was deleted mid-conversation and a brand-new box answered
  the next turn from the transcript it had never written.
- Placement: new sessions rotated 0 → 1 → 2 → 3 across the pool; sessions that
  existed before the scale-up stayed where they were; no session was ever seen
  by two boxes.
- Sessions whose box was removed by a scale-down were re-placed on survivors
  with their history intact.
- 39 consecutive steers into one running turn: zero `Session not found`.

## Compatibility

- **Version skew is safe both ways.** A new Runtime asking an old box for
  `/api/internal/box-status` gets a 404 and falls back to the older session
  list, which is what keeps a live session from being handed to a second box
  during the rollout that introduces this. An old Runtime against a new box sees
  an endpoint nobody calls.
- **`replicas` absent ⇒ 1 ⇒ no behaviour change.** An upstream that does not
  send the field keeps every agent on the single-box path.
- **⚠️ `SICLAW_SUBAGENT_CONCURRENCY` changes meaning**: per-box before,
  per-conversation now. A deployment that lowered it as a memory guard should
  move that number to the new `SICLAW_SUBAGENT_POD_CONCURRENCY`.
- **The certificate Secret keeps its name** (`agentbox-<agentId>-cert`), so an
  existing Secret is reused rather than replaced — no mTLS churn on upgrade.
- **Node headroom**: pod count exceeds `replicas` while a rollout drains, and
  the box memory request is raised to sit above its idle floor.
- **Shared storage is a prerequisite for `replicas > 1`**, not a nicety. Without
  it a conversation that moves between boxes silently loses its history; the new
  warning makes that visible but does not prevent it.

## Message order

Found while testing the above, and fixed here because it is the same layer: a
conversation the user steers several times reloaded with every question ahead of
every answer, though it had alternated on screen while it ran.

Two causes. Assistant replies were held back to a commit point that only exists
so a failed model candidate's reply can be dropped when a fallback takes over —
but every prompt now runs through that path, so turns with nothing to fall back
to paid for a guarantee they could not use, and wrote a whole turn's replies in
one burst at the end. And a user message is written the moment it arrives, while
the box consumes it at a turn boundary that may be seconds later: arrival order
is not processing order.

Replies now persist as they are produced unless a fallback candidate exists. For
user input, the processing moment is observable — the box echoes a user
`message_start` when it begins consuming one — so the row is told to take its
place then, through a per-session queue of row ids held between the write and
the echo. Rows enter that queue only once a box has accepted them, and an echo
must plausibly be the row it claims, so neither a message queued behind the turn
lock nor a replayed echo can take a place that belongs to another message.

The Portal read messages back with `ORDER BY created_at DESC, id DESC`, and
neither column can carry order: `created_at` is second-granular (a fractional
type would break the SQLite half of this DDL) and `id` is a UUID, so anything
written inside the same second came back in an order decided by random hex.
Messages now carry their own key, allocated by the store, backfilled once from
the ordering readers were already seeing.

**An upstream store must implement two things** for ordering to work there;
without them it degrades to today's behaviour and says so once in the log:

- `chat.appendMessage` accepts `defer_sequence: true` — insert without an
  ordering key. Sent only for user input the runtime will order later; every
  other writer (delegation, channels, scheduled tasks) omits it and must keep
  getting a key at insert.
- `chat.sequenceMessage {id, session_id}` — allocate this row's key, only when
  it has none, touching no other column. A method of its own because
  `chat.updateMessage` replaces a row's columns from its payload: a call
  carrying only an id would blank the user's message.


LikiosSedo:
I reviewed the current head (`7ec4df18`). The existing CI is green, and the full local suite also passes (`4,743 passed / 2 skipped`), but I found several multi-box paths that are not covered by the current tests. I do not think this is ready to merge yet:

1. **[P1] A rollout can move a session that is still resident/running.** In `getOrCreatePooled`, when every reachable box is draining, lines 386-395 spawn a replacement and immediately bind the requested `sessionId` to it. Box status/residency is only sampled later. A focused regression with the stale box reporting `sessionIds: ["s-live"]` and `turnsInFlight: 1` expected the old box but received the new `-1` box. During a rollout, a Stop, Steer, or concurrent send can therefore hit a fresh box while the original turn continues on the old one. Please adopt resident sessions before assigning the replacement.

2. **[P1] Collapsed background `spawn_subagent` calls bypass both new concurrency ceilings.** `session.ts:591` returns through `startBackgroundSubagent` before acquiring either the per-session or pod limiter, and `startBackgroundSubagent` invokes `runSpawnedSubagent` directly. With both limits set to 1, a focused regression launched two background children concurrently. This defeats the box-level OOM guard described by the PR. The detached runner still needs to be scheduled through session -> pod limiters.

3. **[P1] `chat.sessionStatus` only probes instance 0.** `server.ts:1165` calls `getAsync(agentId)`, whose K8s implementation resolves only the unsuffixed pod name. A session pinned to instance 1+ is therefore reported as `running: false` after a page refresh. Besides losing Web stream reattachment, this can cause a live A2A task to be classified as orphaned after a Portal restart. The non-spawning lookup must be session-aware.

4. **[P1] `agent.terminate` repeatedly stops instance 0 instead of each listed box.** The loop at `server.ts:1209-1218` iterates `box`, but calls `agentBoxManager.stop(box.agentId)`. `stop(agentId)` always derives the unsuffixed pod name, so an N-box agent issues N deletes for instance 0, leaves instances 1..N running, and still reports them as successfully stopped. Please stop the concrete `box.boxId` targets (or add a manager operation that terminates the whole pool).

5. **[P2] Released sessions never rebalance off a draining box.** The bound-session fast path at `manager.ts:365-371` returns before status sampling, even when that box is draining and no longer reports the session as resident. `BoxBindings.rebalanceOff()` implements the intended rule but has no production call site. A focused regression expected the released session to move to the accepting box but received the old draining box. This can repeatedly reactivate an old-image box and eventually push the drain to its force-kill deadline.

Please add regressions for the resident-rollout case, collapsed background limits, non-zero-instance liveness/termination, and released-session rebalancing.

jacoblee-io:
Thanks — I reproduced all five against the code and they all hold. Two of them are mine failing to finish work I had already started, which is the worst kind:

- **#1** the normal placement path *does* adopt what boxes report before placing, precisely so a live session is never re-placed. The all-draining branch returns before that sampling, so it skips the one guard written for this case.
- **#5** `BoxBindings.rebalanceOff()` exists, is unit-tested, and has no call site. I wrote the rule and never wired it.

On **#2**: the early return predates this branch, so background spawns were already unbounded. That does not make it someone else's problem here — this PR is the thing claiming a box-level ceiling, and a path that bypasses it makes the claim false. It needs to go through session -> pod like every other path.

**#3** and **#4** are both the same root shape: a helper that derives the instance-0 pod name from an agentId, used where a *specific* box was meant. #4 is worse than reported — it also reports the surviving boxes as successfully stopped, so the operator sees a clean result.

Fixing all five with the regressions you asked for, then re-verifying on a cluster (the last three defects on this branch were only caught that way, not by tests). Will push and report back.

jacoblee-io:
Pushed `5f30e762`. All five fixed, with the regressions you asked for. Suite is 4,750 passing.

**#1 + #5 shared a root cause**, so they are fixed together: branches were deciding before reading the state the decision depends on. Residency is now sampled once, up front, and every branch below reads it — the all-draining branch no longer binds a session the old boxes still hold, and affinity now yields exactly when the box is draining AND no longer reports the session, which is the one legal re-binding the design already specified.

That moves the status sample ahead of the affinity fast path, so a bound session now costs N probes instead of 0. They are 2s-cached and shared across concurrent turns of the same agent, and skipped entirely when there is no session to route (admin/capability paths). Deliberate trade: residency is the input every rule turns on, and sampling after deciding is what produced both bugs.

**#2** takes the slots inside the detached promise, so the caller still returns `launched` immediately — queueing delays the child, never the tool call.

**#3/#4** were the same shape (deriving the instance-0 pod name where a specific box was meant); both now resolve the concrete box, and `getForSession` returns undefined rather than guessing.

## Re-verified on a cluster

| Finding | Result |
|---|---|
| #4 terminate | 3-box pool: three distinct pods stopped, `3/3` — previously it deleted instance 0 three times and reported the survivors as stopped |
| #2 ceiling | forced the pod ceiling to 3, 3×30-item fan-outs (90 children wanted): `subagent_active` held at **2** and never exceeded 3 — 2 being `max(1, P-1)`, the group reserve leaving one slot for an interactive spawn |

**#1, #3 and #5 are unit-verified only.** I tried to stage #5 on the cluster twice and could not hold the window open: an empty draining box is reaped in ~10s, which is the behaviour we want, so "released session on a *still-existing* draining box" needs a box kept alive by other resident work. Trivial sub-agents finish faster than I can stage it. Rather than claim a pass I did not get, I am flagging it — the two unit tests cover both directions (moves when there is somewhere to go, stays put when there is not).

One thing the cluster run surfaced that is worth knowing, though it is not a bug: **changing `SICLAW_SUBAGENT_POD_CONCURRENCY` on the Runtime does not affect running boxes** — pod env is fixed at spawn, so it applies on the next cold spawn, same as `SICLAW_AGENTBOX_IDLE_TIMEOUT`. An operator lowering the ceiling in an emergency has to recycle boxes for it to bite. I did exactly that to run the test above.

LikiosSedo:
Re-reviewed the current head (`5f30e762`). The original five findings are materially improved, CI is green, and I independently reproduced the reported `4,750 passed / 2 skipped`; `npm run build` also passes. I still do not think this is ready to merge because three boundary paths remain:

1. **[P1] A failed box-status probe is still treated as proof that a bound session was released.** `sampleBoxStatuses()` correctly leaves a failed probe out of the sample, but `getOrCreatePooled()` reduces that incomplete sample to `resident.has(sessionId)`. For a session bound to a draining box, a transient timeout therefore makes `stillHolding` false and allows `rebalanceOff()` to move it to an accepting sibling. A focused regression with the old box's probe throwing and the fresh box reporting empty expected the old box but received `agentbox-agent-a-1`. If the old box is still running the turn, this recreates the split-affinity/data-loss path from the first review. Rebinding must require a successful report from the **bound box** confirming the session is absent; unknown must retain affinity.

2. **[P1] `getForSession()` still guesses when bindings are missing, and it breaks Local mode.** After a Runtime restart the binding table is empty, so lines 515-518 return the first running pool member without checking residency. A focused regression where only instance 1 reported `s-live` expected instance 1 but received instance 0. In Local mode, `listPool()` is always empty, so the same method returns `undefined` even when the locally cached box exists. The server expression `await (getForSession?.(...) ?? getAsync(...))` does not fall back after the first Promise resolves to `undefined`, making local `chat.sessionStatus` report false. Please recover from box residency (or return undefined rather than selecting the first pool member) and preserve the non-K8s lookup path explicitly.

3. **[P2] A background child queued behind the new ceiling cannot be cancelled.** The job is registered before entering the limiters, but its `abort` hook is only installed inside `runSpawnedSubagent()` after the queue has drained and the child exists. During that potentially long window, `job_stop` returns `stopped: false` / `starting up`; after the preceding child releases its slot, the supposedly stopped job still starts. A focused cap-1 regression observed exactly that sequence. The queued job needs a cancellation hook/state before enqueue, and the limiter callback must check that state before creating the child.

The new author tests all pass, but none covers probe failure, restart/local lookup, or cancellation while waiting in the limiter queue. Please add focused regressions for these cases.

src/gateway/agentbox/manager.ts
```diff
@@ -186,22 +261,52 @@ export class AgentBoxManager {
...
+    // 🔴 A single-box agent must still pick up a new AgentBox image. Nothing else does it:
+    // this path compares phase, profile and CA but never the image, and a box under
+    // continuous traffic never idles out to be respawned — so a Runtime rollout would
+    // leave it on the old image indefinitely, which is the defect this whole change set
+    // exists to fix. It cannot drain in place (the replacement would collide on this same
+    // pod name), so hand the agent to the pool path with a size of one: the stale box is
+    // marked draining and keeps serving what it holds, the replacement comes up under the
+    // next free instance index, and new sessions go there.
+    //
+    // The pool then sits at instance 1 while `replicas` is 1, so the next acquisition
```
chent1996, line 347: **Blocking.** "Converges on its own" skips the step where the convergence violates contract #1 of the design doc. After the drained instance 0 is reaped, the pool is `[instance 1]` holding the agent's sessions; the next acquisition takes this single-box path, misses on the instance-0 name, and spawns + serves from a fresh instance 0 with none of the pooled path's residency guards. Verified by execution: pool = `[agentbox-agent-a-1]` reporting `sessionIds:["s-live"], turnsInFlight:1, drained:false` → `getOrCreate("agent-a", undefined, "s-live")` returns `agentbox-agent-a` (fresh instance 0).

While instance 1 finishes draining (≤5 min, 10 for background work): Stop/Steer for the in-flight turn resolve through `getOrCreate` to instance 0 and no-op; both boxes append the same session JSONL on the shared subPath; a background job completing on instance 1 runs its synthetic turn there while the conversation continues on instance 0. This happens on every image deploy, for every `replicas=1` agent, for exactly the sessions that are busy at reap time.

The rejected alternative ("find a box by label instead of by name") is cheaper than this comment prices it: only the COLD path needs it. When `spawner.get(name)` misses, consult `listForAgent`; if a running agent-profile box survives, hand off to `getOrCreatePooled(…, 1)` — adoption then routes `s-live` to instance 1, `missingInstances` computes 0, instance 0 is never re-created, and the "one extra pod lifecycle" disappears along with the violation. One wrinkle to decide: the surviving instance 1 was spawned resident (`IDLE_TIMEOUT=0`), so a replicas=1 agent would no longer idle out — either accept that or spawn rollout replacements with the agent's normal idle window when `replicas === 1`.

src/gateway/agentbox/manager.ts
```diff
@@ -236,6 +341,439 @@ export class AgentBoxManager {
...
+    // running, and a released session was pinned to a draining box forever.
+    const statuses = sessionId ? await this.sampleBoxStatuses(reachable) : new Map<string, BoxStatusReport>();
+    const resident = new Set<string>();
+    for (const st of statuses.values()) for (const id of st.sessionIds) resident.add(id);
+
+    // Adopt what the boxes report: after a Runtime restart the binding table is empty
+    // while sessions are still live in boxes, and placing one fresh would send a running
+    // conversation to a box holding none of its state.
+    for (const [boxId, status] of statuses) {
+      for (const id of status.sessionIds) {
```
chent1996: Non-blocking. This adoption is only as complete as the status sample, and `sampleBoxStatuses` drops a box whose probe failed. After a Runtime restart (binding table empty), a session resident on the box that failed its probe is invisible to `resident`, so `place()` RR-binds it to a different box — duplicate residency, same class as the rollout finding, needing restart + probe blip to align.

`candidatesFrom` already treats a silent box with maximum skepticism (`MAX_SAFE_INTEGER`); the adoption step could apply the same rule: when any probe in this round failed AND the session has no binding, serve this turn without recording a binding (or retry the failed probe once) instead of committing affinity on incomplete evidence.

src/gateway/agentbox/k8s-spawner.ts
```diff
@@ -196,6 +270,22 @@ export class K8sSpawner implements BoxSpawner {
...
+      // 🔴 Cross-agent name collision. podName() sanitizes and truncates, so distinct
+      // agentIds can map to one pod name ("a.b" and "a-b" both become "a-b"), and the
+      // instance suffix adds the pair X / X-<n> ("foo" instance 1 and agent "foo-1"
+      // instance 0 are both agentbox-foo-1). Reusing another agent's pod would serve this
+      // agent's sessions from a box holding the OTHER agent's certificate and PVC subPath.
+      //
+      // Fail loudly instead. Deleting it would be worse — that is someone else's live box.
+      // The caller treats a failed instance as unavailable and moves on to another index.
+      const owner = existing.metadata?.labels?.[`${labelPrefix}/agent`];
+      if (owner !== undefined && owner !== agentId) {
```
chent1996, line 283: Non-blocking. Right guard, right failure mode — but it only runs when the pod exists at this read. If the colliding pair (`foo` instance 1 vs `foo-1` instance 0) spawn concurrently, the loser lands in the `createNamespacedPod` 409 branch below, which logs "concurrent create, reusing" and returns the winner's pod — the cross-agent serve this throw exists to prevent, minus the loud failure. Re-reading the pod and re-checking the `agent` label inside that 409 branch closes the window for one extra API call on an already-rare path.

docs/design/2026-07-27-multi-agentbox-routing.md
```diff
@@ -0,0 +1,315 @@
...
+Under fixed replicas that is acceptable — they are deliberately-resident infrastructure, and a
+restarted Runtime re-adopts them from the K8s API, which the spawner already treats statelessly
+(`manager.ts:9-10`).
+
+### Changing `replicas` is asymmetric, and neither direction is instant
+
+**Raising** spawns boxes immediately, but existing sessions stay put — the hot box stays hot and
+relief arrives as new conversations land elsewhere. **Lowering** drains: mark the excess boxes,
+stop routing new sessions to them, let the resident ones finish. Pick victims by ascending damage —
+not-yet-Ready first, then zero bindings, then fewest bindings.
```
chent1996, line 220: Non-blocking, doc/code drift within the PR: `reconcilePoolSizes` picks victims by highest instance index (its comment says "the least disruptive order available without asking every box what it holds"), not by this not-yet-Ready → zero bindings → fewest bindings ladder. The simplification is defensible — record the one that was built.

Same section, one line down: "Draining boxes must be observable ('draining, N sessions left')" — as merged, draining lives only in Runtime log lines; `AgentBoxInfo` has no draining flag and no Portal surface shows it, so the operator who sets 4→2 still sees four pods and no explanation. Either ship the minimal surface or move this to the Open section as consciously deferred.

chent1996 (dismissed):
Reviewed against `main` @ `4ce904a2` — the PR's merge-base is current main, CI green, `npx tsc --noEmit` clean, and the full suite passes locally after a fresh install (4750 tests; my first run failed on stale local `node_modules`, not on the PR). I read the design doc first, then the manager/bindings/spawner/session code against it, and verified the load-bearing claims by execution where I could.

This is a genuinely well-engineered change. The design doc records the alternatives that look reasonable and are wrong (per-request RR, session-count placement, per-pod certificates), the concurrency split is derived from two distinct failure modes rather than one number, the acquisition ordering argument (group → session → pod, strict nesting, acyclic wait-for graph) is stated and matches every path in the code, and the "replicas absent ⇒ 1 ⇒ byte-identical" property is real — I traced the `replicas <= 1` branch to the untouched original path, and local mode never reaches pooling at all (`isK8s` gates it). The review-regression test section shows the five cluster-found defects were each pinned with a test. The `resolveFlushBoxId` residual is documented honestly instead of hidden.

One finding is blocking; it reproduces by execution through the public API and violates the design's own first contract on the default configuration. The rest are non-blocking.

---

## Blocking: the replicas=1 rollout convergence serves a still-resident session from the new instance 0

Design contract #1: *“A resident session never changes box. Every request for it routes to its box.”* The pooled path defends this carefully (your test “does NOT move a session that is still resident on the draining box” is exactly right). But the convergence tail of the single-box image roll bypasses every one of those guards.

Sequence, all on `replicas=1` (the default, i.e. every agent):

1. Deploy ships a new image. First acquisition detects the stale instance 0, hands the agent to the pool path, spawns instance 1, drains instance 0. ✅ correct so far — resident sessions keep routing to instance 0 until released.
2. Instance 0 is reaped (drained, or the 5-minute deadline). Pool is now `[instance 1]`, which holds the agent's sessions — including, at rollout time, sessions with in-flight turns or background jobs.
3. The next acquisition resolves `replicas=1` → takes `getOrCreateK8s` → `spawner.get(<instance-0 name>)` → not found → **creates a fresh instance 0 and serves the request from it**, with no residency check, no binding consult, no status sample — that path has none of the pooled machinery.

I verified step 3 by execution against the real manager: pool = `[instance 1]` reporting `sessionIds:["s-live"], turnsInFlight:1, drained:false`; `getOrCreate("agent-a", undefined, "s-live")` returned the freshly spawned `agentbox-agent-a` (instance 0), not `agentbox-agent-a-1`:

```
ZZPROBE routed to: agentbox-agent-a   (expected agentbox-agent-a-1)
```

Consequences while the window is open (until instance 1 finishes draining — up to 5 minutes, 10 for background work):
- **Stop/Steer/clearQueue for the in-flight turn go to the wrong box.** `chat.abort` resolves through `getOrCreate` (server.ts), lands on instance 0, no-ops; the turn on instance 1 is unstoppable from the UI.
- **Two boxes append the same session's JSONL** on the shared PVC subPath — instance 1's turn/background persistence interleaved with instance 0's new turns. This is the concurrent-writer situation the design's volume section exists to rule out.
- A background job's completion on instance 1 runs its synthetic turn there while the conversation continues on instance 0.

The comment at manager.ts:299 frames the convergence cost as "one extra pod lifecycle per rollout" and calls label-lookup "cheaper than teaching this path to find a box by label" — but the extra lifecycle is not the real cost; the residency violation is, and label-lookup eliminates both. Concretely: in `getOrCreateK8s`, when the instance-0 name lookup misses (cold path only — the warm path, which is virtually every call, pays nothing), consult `listForAgent`; if any running agent-profile box survives, hand off to `getOrCreatePooled(…, 1)`. The pooled path then adopts `s-live` onto instance 1 via the status sample and serves it there; `missing` computes 0, so instance 0 is never re-created and there is no second drain at all.

One wrinkle to decide deliberately: that fix leaves the agent permanently on instance 1, which was spawned with `SICLAW_AGENTBOX_IDLE_TIMEOUT=0` (resident), whereas today's convergence eventually restores a self-destructing instance 0. If the idle behavior matters for replicas=1 agents, spawn the rollout replacement with the agent's normal idle window when `replicas === 1` instead of pinning it resident — a drain-replacement box does not need pool residency to do its job.

Since the whole sequence runs on every image deploy for every default-config agent, and the affected sessions are precisely the busy ones, I don't think this can ship as "converges on its own". Pre-PR behavior was worse (manual hard kill), but the design doc promises drain semantics that this window does not deliver.

---

## Non-blocking

**1. Post-restart adoption treats an unanswerable box as holding nothing.** `sampleBoxStatuses` drops a box whose probe fails; the adoption loop then cannot see its sessions, so `resident` is incomplete, and after a Runtime restart (empty binding table) an unbound-but-actually-resident session gets RR-placed onto a different box — the same duplicate-residency class as the blocking finding, needing restart + probe blip to line up. `candidatesFrom` already refuses to let a silent box read as idle (`MAX_SAFE_INTEGER`); the same skepticism could apply here: when any probe in the sample failed and the session has no binding, serve without recording a binding (or retry the failed probe once) rather than committing an affinity decision on incomplete evidence. `getForSession`'s pool-scan fallback has the same restart-window caveat (first running box, not necessarily the holder) — worth one comment line there.

**2. The reaped-then-recreated cross-agent 409 window.** The collision guard (k8s-spawner.ts:283) is the right call and correctly fails loudly. But it only runs when the pod exists at the read; if agent `foo` instance 1 and agent `foo-1` instance 0 spawn concurrently, the loser hits the `createNamespacedPod` 409 branch, which logs "concurrent create, reusing" and returns the other agent's box — serving sessions from a pod holding the other agent's certificate and PVC subPath. Re-reading the pod and re-checking the `agent` label in the 409 branch closes it for one API call on an already-rare path.

**3. First rollout after this ships cannot drain gracefully — say so in the deploy note.** During the first deploy, the old boxes being drained are by definition the OLD image, which has no `/api/internal/box-status`. Every probe fails, `drained` is never observable, and every old box is force-killed at the 5-minute deadline. That is a one-time, acceptable cost, but it will look like the feature is broken unless the deploy note says "the first rollout force-kills at deadline; graceful drain starts with the second".

**4. Doc/code drift inside the PR, twice.** (a) The design doc says victims are picked "by ascending damage — not-yet-Ready first, then zero bindings, then fewest bindings"; `reconcilePoolSizes` picks highest instance index, with a comment giving a different rationale. The implementation is a reasonable simplification — update the doc to record what was built. (b) The doc requires draining to be **observable** ("draining, N sessions left", or an operator who sets 4→2 sees four pods and no explanation); the implementation surfaces draining only in Runtime logs — `AgentBoxInfo` carries no draining flag and no Portal/UI surface shows it. Either build the minimal surface or move that requirement to the Open section as consciously deferred.

**5. `agent.clearMemory` resets the indexer on one box of N** (`getAsync` → instance 0). Under the documented precondition (memory stays disabled for pooled agents) this is unreachable, but nothing enforces that precondition — a one-line guard or comment tying `clearMemory` to the memory-off contract would keep the next reader from "fixing" it into fan-out.

**6. `/api/internal/box-status` responds to any mTLS-authenticated box cert.** Runtime→box probing is the intended direction, and the mTLS CA scopes it to cluster members, so this is fine today; noting it because the route lists per-session IDs, which is more than the metrics routes expose. A comment stating the intended caller would do.

**7. Pre-existing, surfaced by this diff:** `prom-federation-aggregator.ts` contains a literal NUL byte (`const NUL = "\x00"` at ~line 90 on main), which makes git treat the file as binary — this PR's meaningful changes to it are invisible in every diff view. Since you are touching the file anyway, `"\u0000"` restores text diffs.

### Verified

- `replicas <= 1` routes through the original single-box code; local mode (`LocalSpawner`) never reaches pooling; capability boxes never pool.
- Acquisition order group → session → pod holds on all three spawn paths (foreground, detached background, group map/reduce) — including the previously-unguarded detached path, which now takes session+pod slots (the 🔴 comment there is accurate).
- The detached-children fallback limiter, the once-per-group session-limiter resolution, and the shared-MCP `!==` shutdown guard are all coherent; release is deferred by `_backgroundWorkCount`, so the shared MCP manager cannot be torn down under a counted child.
- `resolveFlushBoxId` accepts only the cert's base name or its `-[1-9]\d*` suffixes; `-0` rejected; fallback to cert identity, never a hard failure. Matches the documented residual.
- Cert-secret sharing: 409 → fingerprint compare → reuse; CA-rotation replace is 404/409-tolerant on both sides; `stop()` no longer deletes the per-agent Secret and `sweepOrphans` owns its lifetime with label-based matching plus the age guard.
- Migration: `replicas INT NOT NULL DEFAULT 1` in both the CREATE and `safeAlterTable`, MySQL/SQLite-safe, normalized (`normalizeReplicas`, capped at 20) at API create and update; adapter forwards it on `config.getAgent`; UI field wired end-to-end.
- Trial merge against main: clean. Cluster-verification table in the PR body is author-attested; the numbers are consistent with the code (49 = pod group reserve, 2/2/2 spread, affinity hold), and I did not re-run the cluster exercise.

nightmeng (approved):

PR #463, opened 2026-07-31
fix(lark): never lose a finished answer to a timed-out card; feat(delegation): coordinator answer/route triage

jacoblee-io:
Supersedes #460, which started as the alias-routing fix and outgrew its scope once the same code path was verified against a live deployment. Same content, reorganised into three commits; the alias-routing work from #460 is carried here, and @chent1996's review findings on it are all resolved — see the resolution notes at the bottom.

## 1. Coordinator: alias routing, and answering what it can

**Coverage lookup.** A localized cluster alias could be reported as missing authorization even when its canonical binding was covered.

- `list_delegates` is queried **first** with the target as established by the user, so the common case (the user typed the real name) stays at one tool call and a coordinator with no skills attached is never sent looking for a helper that isn't there.
- Coverage now requires an **exact, case-insensitive** cluster/host binding match. Agent names, descriptions and partial matches are discovery hints, not authorization evidence. This also closes the more dangerous failure: a short alias like `test` could substring-match `roce-test` *and* any delegate whose description mentioned testing, yielding `total > 0`, no hint, and a confident delegation to the wrong specialist.
- A routing helper is consulted only **after** a miss, and only if one is attached. The retry passes `binding_name_confirmed=true`, which makes the zero-result hint terminal **in the tool** rather than relying on the model to remember it already retried.

**Answer or route.** The coordinator was defined as "ONLY job is ROUTING", so a question it could answer outright — a concept, a how-to, a documented fact — still cost a delegation round trip just to have a specialist restate the answer.

It now answers from its own skills/knowledge base when the answer does not depend on a specific environment's **live state** or on a **hands-on action**, and routes otherwise. Uncertainty resolves to routing, and a specific cluster's live state is never answered from the coordinator's own knowledge: it holds authorization over no resource, so it can never be the authority on one. Skills also inform *where* to route (which specialist domain, which target) instead of that being guessed by scanning the roster; `list_delegates` keeps its job as the authorization step.

**The triage stays invisible.** Observed in a real reply before this rule existed: a knowledge question came back as *"这是知识性问题,我已从内部知识库查到原理。下面是…"*, and a failed delegation as *"由于我是协调者,不做直接的 hands-on 诊断"*. The asker wanted the topic explained, not a tour of the bot's operating rules — and naming the mode also invites doubt about whether the answer is authoritative. A reply now opens with the answer; when none is possible it states the outcome the reader needs (the specialist could not be reached, or which detail is missing) instead of the internal rule behind it.

Also drops `search_memory` from the coordinator's locked capabilities: memory is disabled fleet-wide, so those tools do not load, and leaving it listed only invites it back into prompt text.

> Answering depends on the agent actually having skills attached — coordinator is `defaultNoSkills`, so one left bare can still only route. Contracts recorded in `docs/design/coordinator-routing.md`.

## 2. A timed-out card no longer loses the answer

Cards were freezing on their last `⏳` progress line while the answer was persisted normally — visible in the web UI, invisible in the chat. It went unexplained for weeks because **nothing was logged**.

The Feishu SDK does not throw on a non-zero API code; it returns `{code,msg}` in the body. `updateCardContent` and `finalizeCard` reported success as soon as the await resolved, so a *rejected* write was indistinguishable from a delivered one, and only a thrown error printed anything. The file already had a `cardApiFailed` helper for the feedback buttons — it was never applied to the content path.

With the code checked and logged, the very first occurrence in a live deployment named the cause:

```
element.content rejected        code=200850  msg=card streaming timeout       chars=78
final element.content rejected  code=300309  msg=streaming mode is closed     chars=2376
```

Neither oversized content (2.3 KB) nor a sequence conflict — both of which we had assumed. **Feishu closes the card's streaming window on a timeout and refuses every write after it.** That explains why only long investigations lost their answer: only they outlive the window, and a foreground delegation can run for minutes without emitting a milestone to keep it warm.

- check the response code on the content and settings calls, and log `code`/`msg` plus the body length, so a future rejection is diagnosable instead of silent
- `finalizeCard` reports `contentOk` apart from `ok` — the two failures need opposite handling: a failed streaming-mode flip still leaves the answer visible (a second message would duplicate it), a failed content write leaves it nowhere the user can see
- when the body never landed, post a **new static card**. Plain text also delivers the answer, but Feishu does not render markdown in a text message: a structured report arrived as literal `##` headings and `|---|` table rows, and lost the 👍/👎 row with it. A static card keeps the formatting, can carry the buttons (safe in a way it is not on a streaming card), and has no window left to expire. Plain text stays the last resort, so the answer can never vanish.

**Raw tool names off the card.** Steps like `⏳ Ran host_list` are the producer's developer-facing activity text — right for the web UI's work log, useless to an asker, and because the card shows only the *current* step they actively pushed the agent's real narration off it. The structured sub-agent branch was already localized for the channel; only the single-agent branch passed its text through verbatim. Filtered at the channel boundary rather than at the producer, so it holds whatever agentbox image the gateway is paired with and the web UI keeps its tool log.

## Test plan

- [x] `npm test` — 235 files, 4697 passed, 2 skipped
- [x] `npx tsc --noEmit`, `npm run build`
- [x] Rebased on `main`; each of the three commits compiles on its own

### Verified on a live deployment (runtime + agentbox built from this branch)

- [x] a bare tool name no longer appears as a card step
- [x] a long structured answer is delivered complete, with feedback buttons, no frozen placeholder
- [x] a live-state request routes: the live-state criterion is applied, the routing helper is consulted, `list_delegates` matched the specialist on the first query, and the coordinator did not overreach into hands-on work when the delegation itself failed
- [x] a knowledge question is answered directly from the knowledge base, no delegation
- [x] the reply opens with the answer — no "this is a knowledge question / I consulted the knowledge base" preamble
- [x] **the rejected-card path fired in production** and delivered the answer instead of losing it (that is where the `200850` / `300309` codes above come from)

**Known gap:** this recovers *after* the window closes. A keep-alive that prevents the timeout would be the root fix, but whether that works depends on the timeout being idle-based rather than absolute — not worth guessing at, so it is left for a follow-up once the semantics are confirmed. The visible cost today is only that the `⏳` line stops updating on a very long turn; the answer itself arrives formatted.

## Resolution of the #460 review

@chent1996's six findings, all addressed:

1. **Mandatory pre-step on a capability that is absent by default** — fixed by the inversion you proposed: query first, helper only on a miss, and only "if any".
2. **Undecidable gate condition** — gone with it; there is no "unless the binding name is already established" branch to guess at.
3. **The false positive was the stronger argument** — you were right, and it is now fixed at the matcher rather than papered over by a hint: `matchRoster` no longer substring-matches and no longer looks at `name`/`description`.
4. **Self-repeating zero-result hint** — fixed as prescribed: `binding_name_confirmed` is a tool parameter, and the hint is terminal when it is set.
5. **`binding_names[]` was a contract this repo never defined** — the identifier is gone; `docs/design/coordinator-routing.md` specifies the helper's *semantic* result (confirmed / ambiguous / unresolved) without naming a serialized field.
6. **Persona tests pinning prose** — agreed, and it had been reintroduced during the answer/route work. The phrase assertions are gone; the contract lives in the design doc, and the test asserts structure and an *absence* (`search_memory`), which survives rephrasing.

Still open from that review: the `sre` type keeps `search_memory` in its locked capabilities. It is inert with memory disabled, so it is left alone here rather than widening the diff — happy to clean it up in one place if you prefer.


LikiosSedo:
I found two issues that should be addressed before merge (reviewed at `d9e13a1e`):

1. **The Portal's Coordinator definition is now out of sync with the backend.**

   `src/core/agent-types.ts` changes the Coordinator to the ANSWER/ROUTE model and removes `search_memory`, but `portal-web/src/lib/agentTypes.ts` still describes it as a read-only router and still lists `search_memory`. `AgentSettings.tsx` also tells operators to attach only routing helpers. Because Coordinators are `defaultNoSkills`, following the current UI will commonly leave the advertised knowledge-answering mode unavailable.

   Please update the Portal mirror and its Skills-tab guidance. A small consistency test between the frontend mirror and the backend contract would also prevent this from drifting again.

2. **`binding_name_confirmed` does not make the retry terminal in the tool.**

   It is an optional, caller-supplied boolean, and `createListDelegatesTool` records no prior miss, helper confirmation, or consumed retry. Repeating `{ query: alias }` keeps returning the same retry hint indefinitely, while `binding_name_confirmed=true` can also be supplied on the first call. The tests cover a compliant model passing the flag, but not the documented invariant that the first miss offers at most one retry.

   If terminality is meant to be enforced by the tool, it needs non-forgeable/consumable retry state (or equivalent per-attempt state) plus tests for a repeated unconfirmed miss. Otherwise the PR and design doc should not claim that this no longer relies on the model remembering the retry.

The timed-out-card recovery path otherwise looks sound in the reviewed code and tests, and the PR keeps internal cluster aliases out of the open-source fixtures.

jacoblee-io:
Both stand, both fixed. Thanks — the second one caught me overstating what the code does.

## 1. Portal mirror out of sync — fixed (`b2784533`)

Confirmed exactly as described: `portal-web/src/lib/agentTypes.ts` still said "Read-only router" and still listed `search_memory`, and `AgentSettings.tsx` told operators to attach "only the routing helpers this coordinator needs".

The part that makes this more than cosmetic is your second sentence. Coordinators are `defaultNoSkills`, so an operator following that guidance ends up with the advertised answering mode unavailable — the UI was steering people away from the capability it was advertising. The mirror's own header says to keep the lists in sync with the backend registry, so the drift also broke its stated contract.

- description and capability list resynced
- the Skills-tab line now says what it has to: attach the routing helpers **and** any knowledge it should answer from, because without them it can only route
- `agent-types.test.ts` now reads the mirror and compares descriptions and capability lists against `AGENT_TYPES` per type, so this cannot drift silently again. Read as text rather than imported, so no build boundary is crossed, and resolved relative to the test file so a checked-out copy compares against its own sibling mirror.

That check immediately earned itself: it also caught a pre-existing one-word difference in the `custom` description ("Free-form: pick" vs "Free-form: you pick"), now aligned.

## 2. `binding_name_confirmed` did not make the retry terminal — fixed (`f3380f7c`)

You are right, and the claim was wrong as written. The flag is an optional caller-supplied boolean and the tool kept no state, so repeating `{ query: alias }` returned the same "resolve it and retry" hint indefinitely, and the flag could equally be asserted on a first call. Saying the state machine had moved out of the model's recollection was not accurate — it had only moved into what the model *declares*.

Given the tool instance is per agent session, the missing state fits in its closure:

```ts
const missedTargets = new Set<string>();
…
const alreadyMissed = q ? missedTargets.has(q) : false;
if (q) missedTargets.add(q);
if (bindingNameConfirmed || alreadyMissed) { /* terminal */ }
```

A target that has already come back empty is never offered the retry again, whatever the flag says — that is the part a caller cannot forge. So "at most one retry per target" now holds against both a repeated unresolved alias and a first-call assertion of the flag, while an unrelated target still gets its own retry rather than inheriting another target's exhaustion.

Tests added for the two cases that were missing: a repeated unconfirmed miss is terminal, and the memory is per target (a different alias still gets its retry).

The design doc no longer claims the flag bounds the retries. It now says the flag is a declaration of what the caller resolved — not an authorization assertion, and not by itself a bound — and that the bound is enforced per target alongside it. The invariants list calls out the repeated-target case explicitly, and the Portal-mirror invariant is recorded there too.

---

Verified: `npm test` 235 files / 4700 passed, `portal-web` 20 files / 193 passed (ran that explicitly since this touches the frontend), `npx tsc --noEmit` clean.

Note that the Portal mirror fix can't be verified in our staging namespace — it has no Portal deployment — so that one rests on the new consistency test rather than a live check. The tool-side retry bound is going out to staging now.

LikiosSedo:
Re-reviewed at `b2784533`. The Portal description/capability mirror is now synchronized, and repeating the exact same missing query is terminal. One retry issue remains in the actual alias-resolution path.

`missedTargets` is keyed by the normalized query string, but a successful helper resolution changes that string:

```text
query="local-alias"                 -> miss, retry offered
helper resolves a canonical name
query="canonical-but-uncovered"    -> miss
```

If the caller omits `binding_name_confirmed=true` on the second call, `alreadyMissed` is false because the canonical name differs from the alias, so the tool offers alias resolution again. I verified this with a temporary regression test: after the two calls above, the second result still contained `consult a routing-helper skill` and `Retry once`.

The new tests cover a repeated miss for the same string, but not alias -> canonical. Therefore the bound is currently "one retry offer per query string", not "one retry per routing attempt", and the real name-changing retry still relies on the model carrying the boolean correctly. Please bind the first miss to one consumable subsequent retry (an opaque retry token/attempt id or equivalent) and add the alias -> canonical-with-missing-flag regression case.

There is also a smaller Portal copy issue: the new guidance says to attach routing helpers and knowledge in the `(Skills tab)`, while knowledge repositories are configured in the separate **Knowledge** tab. This should point operators to the Skills and Knowledge tabs; otherwise a default-no-skills Coordinator can still be left unable to answer from its knowledge base.

Verification on this head: backend suite 4700 passed / 2 skipped, Portal suite 193 passed, and `tsc --noEmit` passed. GitHub CI is also green; the missing alias -> canonical case is simply not covered by the committed tests yet.

LikiosSedo:
Re-reviewed at `eb9b4d8c`. The Portal copy is fixed, and the opaque token closes the alias -> canonical name-change hole from the previous review. One routing-state leak remains before this is safe to merge.

### [P1] Consume the retry token on success and scope it to a routing attempt

`pendingRetryToken` is only inspected and cleared inside `if (total === 0)`. A successful canonical retry therefore leaves the token outstanding:

```text
query="local-alias" + miss                         -> token issued
query="cluster-alpha" + confirmed name + token    -> delegate found
query="unrelated-alias" + miss                     -> incorrectly terminal
```

I reproduced this with a temporary regression test. The third call returned:

> the one alias-resolution retry for this attempt was already offered

instead of offering the unrelated routing question its own helper lookup and token.

There is a second manifestation of the same lifetime mismatch: if the first miss says to consult a helper but no helper is attached, the model correctly ends the turn without calling `list_delegates` again. The outstanding token then survives into the user's next routing question. The tool instance is created once for the managed agent session, so this state is session-scoped, not turn- or routing-attempt-scoped.

The committed success test does not catch this because it never supplies the newly required `retry_token`, and it makes no subsequent routing query.

Please:

1. consume a matching token independently of whether the retry result is empty or successful;
2. retire an abandoned offer at the user-turn/routing-attempt boundary so it cannot poison a later question; and
3. add generic regression cases for successful canonical retry -> unrelated miss, and abandoned first miss -> next user routing attempt.

Current verification: the committed backend suite passes `4702` tests with `2` skipped, Portal passes `193`, TypeScript build passes, and GitHub CI is green. The temporary successful-retry regression above fails, so the green suite does not cover this state leak yet.

LikiosSedo (approved):

PR #467, opened 2026-08-03
Fix Feishu Topic routing to follow group mode

LikiosSedo:
## Summary
- remove the hidden SICLAW_LARK_THREAD_MODE and thread_mode rollout gates
- make personal/per-user groups use Feishu Topics deterministically
- keep team/shared groups on the main-group reply path
- update Topic routing tests and channel documentation

## Deployment contract
- no Helm values, manifests, runtime environment variables, or database migrations are added or changed
- the existing production Helm values remain valid as-is
- personal/per-user session keys become Topic-scoped by design; team/shared session behavior is unchanged
- existing pre-Topic personal-session history is intentionally not migrated into new Topic roots

## Validation
- npx vitest run src/gateway/channels/lark.test.ts src/gateway/channels/lark-card.test.ts src/gateway/channels/lark-image.test.ts (237 passed)
- npm test (243 files, 5016 passed, 2 skipped)
- npm run build

### DIFF
--- a/src/gateway/channels/lark.ts
+++ b/src/gateway/channels/lark.ts
@@ -1270,7 +1270,9 @@ export async function handleLarkMessage(
       return;
     }
     const current: GroupContextMode = modeBinding.contextMode === "shared" ? "shared" : "per_user";
-    const modeReplyInThread = current === "per_user";
+    // /mode changes the whole group. Keep a root invocation visible on the
+    // main-group path; only retain the card inside an already-established Topic.
+    const modeReplyInThread = isThreadFollowup && current === "per_user";
     rememberGroupMode(groupChannelId, chatId, current);
     const sent = await sendModeCard(larkClient, messageId, current, groupChannelId, chatId, locale, modeReplyInThread);
     if (!sent) {
@@ -1370,7 +1372,7 @@ export async function handleLarkMessage(
     return;
   }
 
-  const personalTopicMode = contextMode === "per_user";
+  const personalTopicMode = isGroupMessage && contextMode === "per_user";
   const conversationKey = personalTopicMode ? topicConversationKey : undefined;
   const replyInThread = personalTopicMode;
   const effectiveSessionKey = binding.sessionKey ?? "";
--- a/src/gateway/channels/lark.ts
+++ b/src/gateway/channels/lark.ts
@@ -242,10 +242,6 @@ export interface LarkChannelConfig {
   domain?: "feishu" | "lark";  // feishu = China (default), lark = Global
   app_id: string;
   app_secret: string;
-  /** Reply to group root messages as Feishu topics and scope sessions to the
-   *  topic root. Default-off; the test runtime can enable it with
-   *  SICLAW_LARK_THREAD_MODE=1 without changing stored channel config. */
-  thread_mode?: "group";
   group_channel_id?: string;
   verification_token?: string;
   encrypt_key?: string;
@@ -462,12 +458,6 @@ function getLarkSenderType(data: any): string | null {
 
 function buildLarkSessionKey(senderOpenId: string | null, chatId: string): string {
   return senderOpenId ? `open_id:${senderOpenId}` : `chat:${chatId}`;
-}
-
-function larkGroupThreadModeEnabled(config?: LarkChannelConfig): boolean {
-  if (config?.thread_mode === "group") return true;
-  const env = process.env.SICLAW_LARK_THREAD_MODE?.trim().toLowerCase();
-  return env === "1" || env === "true";
 }
 
 /**
@@ -985,10 +975,10 @@ export async function handleLarkMessage(
   const senderOpenId = getLarkSenderOpenId(data);
   const senderType = getLarkSenderType(data);
   const sessionKey = buildLarkSessionKey(senderOpenId, chatId);
-  // Rollout gate only: whether personal group conversations MAY use Feishu
-  // topics. The binding's server-authoritative contextMode decides whether
-  // this specific message actually uses the topic path.
-  const topicFeatureEnabled = chatType === "group" && larkGroupThreadModeEnabled(channelConfig);
+  // Every group message carries a provider-native Topic candidate. The
+  // server-authoritative contextMode decides the product behavior after binding
+  // resolution: per_user uses the Topic; shared stays on the main-group path.
+  const isGroupMessage = chatType === "group";
   const eventRootMessageId = typeof message.root_id === "string" && message.root_id.trim()
     ? message.root_id.trim()
     : messageId;
@@ -999,7 +989,7 @@ export async function handleLarkMessage(
   // that this event belongs to a Topic; otherwise an explicit @ starts a new
   // bot conversation rooted at the current message.
   const rootMessageId = threadId ? eventRootMessageId : messageId;
-  const topicConversationKey = topicFeatureEnabled ? `lark_thread:${rootMessageId}` : undefined;
+  const topicConversationKey = isGroupMessage ? `lark_thread:${rootMessageId}` : undefined;
 
   // Raw receipt log: fires for EVERY delivered event before any drop, so a
   // group message that arrives but is filtered (non-text, empty after @-strip)
@@ -1246,7 +1236,7 @@ export async function handleLarkMessage(
 
   // Computed here rather than at the @-gate below: /mode needs both.
   const botMentioned = isBotMentioned(message, botOpenId);
-  const isThreadFollowup = topicFeatureEnabled && threadId !== null && rootMessageId !== messageId;
+  const isThreadFollowup = isGroupMessage && threadId !== null && rootMessageId !== messageId;
 
   // /mode — summon the context-mode switch card. Command words are exact, and
   // the bot must be @-mentioned: this switches the mode for the WHOLE group, and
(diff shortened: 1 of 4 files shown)


src/gateway/channels/lark.ts
```diff
@@ -1380,7 +1370,7 @@ export async function handleLarkMessage(
     return;
   }
 
-  const personalTopicMode = topicFeatureEnabled && contextMode === "per_user";
+  const personalTopicMode = contextMode === "per_user";
```
jacoblee-io: The old `topicFeatureEnabled` bundled two conditions: the rollout gate **and** `chatType === "group"`. The replacement `isGroupMessage` is applied at the two `topicConversationKey` / `isThreadFollowup` sites, but not here — so `personalTopicMode` now derives from `contextMode` alone and no longer asks whether this is a group message at all.

This is unreachable for `p2p` today, because that branch returns unconditionally ~350 lines above. But for any `chat_type` that is neither `"group"` nor `"p2p"` — the field absent, or a value Feishu adds later — on a channel with a `group_channel_id`, the flow reaches here with `contextMode` defaulting to `per_user`, and we then set `replyInThread = true` while `conversationKey` is `undefined` (since `topicConversationKey` *is* group-gated). We would reply into a Topic we are not scoping the session to. Before this diff, `personalTopicMode` was false in that case.

One token fixes it and keeps the invariant local instead of resting on a distant early return:

```ts
const personalTopicMode = isGroupMessage && contextMode === "per_user";
```

Worth doing precisely because this is how the guard got dropped in the first place — it was folded into a compound flag, so replacing that flag lost it silently.

src/gateway/channels/lark.ts
```diff
@@ -1280,7 +1270,7 @@ export async function handleLarkMessage(
       return;
     }
     const current: GroupContextMode = modeBinding.contextMode === "shared" ? "shared" : "per_user";
-    const modeReplyInThread = topicFeatureEnabled && current === "per_user";
+    const modeReplyInThread = current === "per_user";
```
jacoblee-io: Same missing `chatType` half as the `personalTopicMode` line below, and it has a visible product consequence: this flag went from off-by-default to on for every `per_user` group.

`/mode` typed at group root is not in a thread, so replying with `reply_in_thread: true` makes Feishu **create a new Topic** just to render a config card. That reads oddly against what the command does — the comment right above says this "switches the mode for the WHOLE group", and putting a group-wide setting change in a side thread hides it from the group.

No test covers where the card lands: the existing `/mode` cases assert the @-gate and that no topic session is allocated (`lark.test.ts:1568`), not the reply target. So either this is intended and wants an assertion, or `/mode` should stay on the main-group path.

jacoblee-io (commented):
Right direction — deriving Topic behavior from `contextMode` alone is the correct end state, and the removal is thorough (no `SICLAW_LARK_THREAD_MODE` / `thread_mode` leftovers anywhere: helm, k8s, portal-web all clean). Renaming the tests from the gate to the mode (`thread mode creates…` → `personal mode always creates…`) is also the right instinct: a test should assert product behavior, not a switch.

One thing I think blocks the merge, plus two smaller ones inline.

## Legacy `context_mode IS NULL` groups flip to Topics on deploy, with new session keys

`normalizeContextMode` is `value === "shared" ? "shared" : "per_user"` (`src/portal/adapter.ts:59`), and `channel_bindings.context_mode` was added later as `VARCHAR(20) DEFAULT NULL` (`src/portal/migrate.ts:685`). So:

- **New groups are fine** — both PAIR (`adapter.ts:267`) and open auto-bind (`adapter.ts:522`) seed `'shared'` explicitly.
- **Pre-column group bindings are NULL**, and NULL normalizes to `per_user`.

For those groups, removing the gate changes what `buildGroupSessionKey` returns: `open_id:ou_x` becomes `open_id:ou_x:lark_thread:<root>`, and `resolveChannelBindingParticipantSession` mints a **brand-new session** for a key it has not seen. Concretely: after this deploys, every member of a legacy group gets an empty session on their next `@bot`, their group history is gone from the agent's point of view, and each subsequent Topic root is yet another new session.

The env gate was exactly what let this be staged, so removing it needs the migration answered in the same PR:

1. Count the affected rows first — `SELECT COUNT(*) FROM channel_bindings WHERE route_type='group' AND context_mode IS NULL`.
2. If that is not zero, backfill NULL → `'shared'` here. Then only groups that explicitly chose `per_user` via `/mode` — the ones that actually want Topics — enter the new behavior, and legacy groups are untouched.

Also worth correcting in the description: "Sicore conversation/session contracts are unchanged" is true of the *contract*, but the *keys* change, and the keys are what users feel. `2026-07-11-lark-group-context-mode.md:220` does list "migrating existing session history across a mode switch" as a non-goal — but that non-goal is about a user deliberately switching, not about one deploy switching it for groups that chose nothing.

## Stale non-goal in the design doc

`docs/design/2026-07-11-lark-group-context-mode.md:219` still says:

> - Thread-level scoping (Lark group events carry no thread/root id).

That is now false — this code reads `message.root_id` and `thread_id`. This PR is the moment to drop that line and record the new contract in the design doc (Topic behavior is derived from `contextMode`; there is no separate switch). `channels.mdx` covers the user-facing half well, but per the repo convention the contract belongs in `docs/design/`.

LikiosSedo:
Thanks — I addressed the necessary runtime and documentation issues:

- restored the group-chat guard on personal Topic routing, including an unknown `chat_type` regression test;
- kept root `/mode` cards on the main-group path while preserving `/mode` replies inside an established Topic;
- documented the deterministic `per_user`/`shared` Topic contract and removed the stale no-thread non-goal.

For item 1, we are intentionally not adding a legacy-session backfill or compatibility layer. The deployment owner confirmed there is no meaningful legacy Siclaw personal-session inventory worth migrating, so preserving that history does not justify an unstable compatibility path. Mapping legacy `NULL` to `shared` would also contradict the existing grandfathering contract and could merge users' previously isolated contexts. We explicitly accept that pre-Topic `per_user` history is not attached to a new Topic root; each new root starts its own scoped conversation.

No database, Helm, environment-variable, or deployment changes are introduced.

jacoblee-io:
Re-reviewed `d4dac849`. The two code findings are fixed properly and the doc one is done. One item is still open.

## Verified

I checked out the head and ran it rather than reading only:

- `npx vitest run src/gateway/channels/lark{,-card,-image}.test.ts` → 237 passed.
- Both new tests are load-bearing. Reverting `isGroupMessage &&` on `personalTopicMode` fails **only** `never enables Topic delivery for an unknown chat type`; reverting `isThreadFollowup &&` on `modeReplyInThread` fails **only** `/mode at the group root stays on the main-group path`. Each guard has exactly one test pinning it, and neither test passes without its guard — which is the part that usually goes wrong when a guard is added with a test.

`modeReplyInThread = isThreadFollowup && current === "per_user"` is a better answer than the one I suggested: `isThreadFollowup` already carries `isGroupMessage`, so it restores the chat-type half at the same time, and "keep the card where the conversation already is, otherwise keep a group-wide switch visible to the group" is the right rule rather than a special case.

The design-doc block is the right shape too — it states the derivation as a contract instead of describing the code.

## Still open: legacy `context_mode IS NULL` groups

The new doc bullet says:

> Existing pre-Topic `per_user` session history is not migrated into a new Topic root. Each new root intentionally starts its own scoped conversation.

That answers it for groups that **chose** `per_user`. It does not separate them from the population I was pointing at: rows written before the `context_mode` column existed (`migrate.ts:685`, `VARCHAR(20) DEFAULT NULL`), which are NULL and only *normalize* to `per_user` at `adapter.ts:59`. Those groups never chose anything — and because both PAIR and open auto-bind seed `'shared'` explicitly, NULL is precisely the set of oldest, longest-running groups.

For them this deploy is not "a new root starts a new conversation", it is "your existing conversation is replaced by one", because `buildGroupSessionKey` starts returning `open_id:ou_x:lark_thread:<root>` where it returned `open_id:ou_x`, and `resolveChannelBindingParticipantSession` mints a fresh session for the unseen key.

So the question is just the size of that set:

```sql
SELECT COUNT(*) FROM channel_bindings WHERE route_type = 'group' AND context_mode IS NULL;
```

If it is 0, say so and this is settled — nothing to do. If it is not, backfilling NULL → `'shared'` in this PR keeps those groups exactly where they are while every explicitly-`per_user` group still gets Topics, and it costs one idempotent UPDATE. Either outcome is fine; deploying without knowing which one it is is the part I would not do.

jacoblee-io (approved):
Approving.

Both code findings are fixed and I verified them on `d4dac849` by running the suites (237 passed) and mutation-checking each guard — reverting either one fails exactly the test that pins it. The `/mode` resolution (`isThreadFollowup && per_user`) is better than what I proposed, and the design doc now states the Topic-from-`contextMode` derivation as a contract.

The remaining item — legacy `context_mode IS NULL` group bindings normalizing to `per_user`, and so changing session key on first message after deploy — is not a code defect in this PR and is left to the deploy owner rather than held against it. It is a one-line check before rollout:

```sql
SELECT COUNT(*) FROM channel_bindings WHERE route_type = 'group' AND context_mode IS NULL;
```

Non-zero means those groups start fresh sessions; backfilling NULL → `'shared'` avoids that if it is not wanted.

PR #469, opened 2026-08-03
fix(agentbox): never invert podNamePrefix into BoxProfile

LikiosSedo:
## Summary

Production `siclaw-runtime:v0.3.2` fails every KB compile with:

```text
capability.start rejected: unknown BoxProfile: kbc-box
```

even though SiCore correctly sends `profile=kb-compile` (and persists it on `siclaw_capability_runs`).

**Root cause (PR #466 pool resilience):** `AgentBoxManager.podName` asked `K8sSpawner.boxIdFor` after running `profileForPrefix(prefix)`. For compile boxes that prefix is `kbc-box`, and `profileForPrefix` returned it unchanged — then `boxIdFor` called `getBoxProfile("kbc-box")`, which is not a registered profile.

```text
kb-compile → prefix kbc-box → profileForPrefix("kbc-box") → getBoxProfile("kbc-box") ✗
```

`kb-compile` and `kb-compile-codex` **share** the `kbc-box` prefix, so prefix→profile is not invertible and must not exist.

## Fix

- Naming is **one-way only**: profile → `podNamePrefix` → pod name.
- `podName(agentId, profile, instance)` passes the **real profile** into `boxIdFor`.
- Delete `profileForPrefix`.
- Regression test with a `boxIdFor` FakeSpawner (production path); assert `kb-compile` / `kb-compile-codex` never pass `"kbc-box"` as a profile.

## Multi-replica design note

Capability/compile boxes stay **replicas=1** (per-run jobs). Chat multi-box pooling is unchanged and still keys pool membership by agent + labels; this fix only corrects how a single pod name is derived from a profile.

## Test plan

- [x] `vitest run src/gateway/agentbox/manager.test.ts` (92 tests)
- [x] siflow-test: deploy `siclaw-runtime@sha256:20630fb4…` (`pr469-4ad56bea`)
- [x] smoke KB `8bb5de66-…` compile.generate → pod `kbc-box-8cc784d8-…-0` Running; agent progress ("reading the workspace…"); no `unknown BoxProfile: kbc-box`

## Reviewer focus

- Is one-way naming sufficient for multi-replica chat pools? (compile path is replicas=1; pool path uses listForAgent + spawn with profile, not profileForPrefix)
- No new BoxProfile registration for `kbc-box` (prefix only)

### DIFF
--- a/src/gateway/agentbox/manager.test.ts
+++ b/src/gateway/agentbox/manager.test.ts
@@ -1,5 +1,6 @@
 import { describe, it, expect, beforeEach, afterEach, vi } from "vitest";
 import { AgentBoxManager } from "./manager.js";
+import { getBoxProfile } from "./box-profile.js";
 import type { BoxSpawner } from "./spawner.js";
 import type { AgentBoxConfig, AgentBoxHandle, AgentBoxInfo } from "./types.js";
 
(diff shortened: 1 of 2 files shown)


jacoblee-io (approved):
Correct fix for a bug I introduced in #466, and worse than the title suggests.

I added `profileForPrefix` so the manager would stop deriving pod names itself, but that
direction is not invertible: `kb-compile` and `kb-compile-codex` both spawn under the
`kbc-box` prefix. The chain on `main` today is

```
podName(agentId, prefixForProfile("kb-compile"))   → prefix = "kbc-box"
  → profileForPrefix("kbc-box")                    → "kbc-box"   (not a profile)
    → boxIdFor(…, "kbc-box") → getBoxProfile("kbc-box")
      → throw: unknown BoxProfile: kbc-box   (box-profile.ts:176)
```

and it holds at all three call sites — `getOrCreateK8s`, `getAsync`, `stop` — so capability
boxes cannot be spawned, found or stopped on v0.3.2. Chat agents are unaffected
(`"agentbox" → "agent"` resolves fine), and the pooled path never derives a name at all: it
discovers pods by label. That is why the multi-box work verified clean in a cluster while
this was broken the whole time.

Verified locally on this branch: all three call sites now pass the profile through, no
inversion is left anywhere in the tree, the fallback for spawners without `boxIdFor` still
goes profile → prefix, `tsc --noEmit` is clean and `manager.test.ts` passes (92).

The test is the more valuable half. My tests missed this because the fake spawner had no
`boxIdFor`, so the manager took its local fallback — the one path that never calls
`getBoxProfile`. Giving the fake the same contract as the real spawner, and asserting that
what reaches it is `kb-compile` and never `kbc-box`, pins the rule rather than the symptom.
A stub more forgiving than the implementation it stands in for is exactly how this class of
bug survives a green suite.

Nothing blocking. One optional thought for later: since prefix → profile has now proved
non-invertible in production, `getBoxProfile` could reject an argument that is a known pod
prefix with a message saying so — but the assertion here already locks the contract.

PR #471, opened 2026-08-04
feat(kb): harden compile reliability and align OKF v0.2

LikiosSedo:
## Summary

This is the unified Siclaw delivery for KB compile reliability, the OKF v0.2 provenance contract, and standalone compiled-Wiki packaging.

### Compile diagnostics and recovery

- Persist structured AgentBox failure diagnostics even when a box event only carries a safe short message.
- Normalize failure tokens before logging/persistence and classify non-resumable batch failures.
- Preserve the initial attempt candidate-page byte baseline across batch client rebuilds.
- Sanitize in-flight Candidate snapshots before periodic, final, shutdown, and reconnect replay sync. Changed bytes can no longer carry `verified` records belonging to the previous bytes across a process restart, and sync does not race the model's local file writer.

### Incremental compilation

- Emit compact, model-useful source diffs and enforce a byte-exact final budget for `CHANGESET.json`.
- Preserve authoritative execution scope first, then bounded diffs, then optional logical hints.
- Preserve the unaffected-page count when redundant page names are omitted.
- Resolve legacy bare-path citations against the authoritative modified/deleted source set as well as the current Raw inventory, so deleting a source still invalidates every dependent page.
- Emit new ledger-derived provenance with the explicit `raw/` namespace.

### OKF v0.2 clean cut

- Replace `compiled_from` with structured `sources[].resource` across generation, validation, coverage, media verification, incremental impact analysis, fixtures, and prompts.
- Treat only explicit `raw/` / `drop/` resources, plus exact legacy Raw inventory or changed-source matches, as managed source provenance. URLs, bundle paths, references, and arbitrary scope descriptors remain valid OKF provenance without creating false Raw diffs or dangling citations.
- Validate emitted trust, lifecycle, source-credibility, usage-window, and Attested Computation fields; Attested Computation must use exactly one of a `computation` path or an inline `Computation` code fence.
- Deterministically stamp changed concepts with `generated.by: process:siclaw-kbc`, remove verification tied to old bytes, and mark new concepts `status: stable` when omitted or null.
- Replace machine-owned YAML fields through parser source spans while preserving comments, quoting, anchors, and unknown extensions byte-for-byte; ambiguous flow mappings, duplicate keys, or retained aliases fail atomically instead of corrupting the document.
- Do not retain a legacy `compiled_from` compatibility reader.

### Standalone compiled Wiki import artifact

- Add an import-specific OKF v0.2 contract that validates required frontmatter without applying authoring-only producer preferences to an external Wiki.
- Preserve legitimate external verification records, wikilinks, bundle-root links, and unknown fields; require an explicit root `okf_version: "0.2"` marker.
- Require `generated.at` and `verified.at` to use timezone-qualified RFC 3339, matching the direct-import receiver.
- Add `kbc/tools/package_okf.py` to create the deterministic, safe tar.gz used by the direct-import UI without creating Raw, Candidate, or an Agent SDK session.
- Reject symlinks, unsupported entries, invalid UTF-8 Markdown, unsafe sizes/counts, and invalid OKF before writing the destination archive.

The latest review correction is `27d4f25a`: deleted-source impact resolution, durability-safe trust invalidation, RFC 3339 parity, and CI coverage for incremental/emitter regressions.

## Verification

- [x] `test_compile_box.py` full protocol suite, including transport rebuild, shutdown sync, same-process retry, reconnect replay, and cross-process no-op recovery
- [x] `test_selfcheck.py`
- [x] `test_incremental.py`
- [x] `test_okf_package.py`
- [x] `test_emit.py`
- [x] `test_destream.py`, `test_codex_engine.py`, `test_office_ingest.py`, `test_redblue.py`, and `test_mtls_auth.py`
- [x] Python 3.11 syntax compilation for changed KBC modules
- [x] `git diff --check`
- [x] GitHub Type Check, Test, Portal Web Test, and KBC Box Test on `27d4f25a`

## Deployment

No deployment is included. Any environment validation requires new commit-SHA-tagged artifacts for the current head.


kbc/platform/pod/selfcheck.py
```diff
@@ -129,86 +129,97 @@ def has_assets_segment(rel: str) -> bool:
...
-                in_list = False
-    return sources, derived, has_key
+                    resources.append(entry)
+                continue
+            if _explicit_external_source(resource):
+                continue
+            # Clean v0.2 packages use an explicit managed namespace. Preserve
+            # the old bare-path dialect only when it exactly identifies a Raw
+            # inventory entry; every other value is an OKF scope descriptor.
+            if managed_sources is not None:
```
jacoblee-io, line 219: **Resolving a bare `resource` against the *live* Raw inventory silently drops three classes of legitimate provenance. This is the root cause of the `test_incremental.py` failure.**

Bare (non-`raw/`-prefixed) resources are only honoured when still present in `managed_sources`, which `candidate_pages` computes from the *current* raw tree. Verified against the real function:

```
resource='snap/three.md'  inventory={'snap/three.md'}  -> (['snap/three.md'], False, True)
resource='snap/three.md'  inventory={}                 -> ([], False, True)
resource='./live.csv'     inventory={'live.csv'}       -> ([], False, True)
resource='raw//live.csv'  inventory={'live.csv'}       -> (['/live.csv'], False, True)
```

Three consequences, each silent:

1. **A deleted source's citation vanishes.** Consumer deletes `raw/snap/three.md` and sends `{"deleted":["snap/three.md"]}`. `managed_sources` is computed post-delete, so the page citing it parses to `sources: []`, `_pages_citing` finds nobody, and `resolve_affected_pages` returns it in `unaffected_pages` — where the byte-integrity guard forbids touching it. The deleted source's content can never be removed from the library. This is exactly `test_resolve_affected_pages` failing at `test_incremental.py:38`.
2. **Dangling citations fail open.** A citation to a nonexistent bare source is dropped at parse time rather than reported, so `coverage()['dangling_citations']` is `[]` and `closed` can go True with a provenance row pointing at a file that isn't there. Pre-PR, `parse_compiled_from` returned the entry regardless of inventory and `cited - source_set` flagged it.
3. **`_explicit_external_source` (`:164`) classifies `./`- and `../`-prefixed resources as external *before* normalization**, so `./live.csv` is discarded — reintroducing the convergence wedge that `_norm_source_entry`'s own docstring at `:148` still claims to have fixed. And `raw//live.csv` normalizes to the absolute `/live.csv`.

The realistic false-positive direction is clean, for the record — `rawhide/x.md`, `raw-notes.md`, and `https://raw.githubusercontent.com/o/r/main/raw/live.csv` are all correctly *not* treated as managed Raw paths, because the managed test requires the `raw/` slash and the `://` check runs first.

Fix direction: normalize (`posixpath.normpath`) before classifying, and treat a bare resource as a Raw citation on *shape* rather than on presence in the current inventory — presence is what `coverage()` is for, and it's the check that should report the miss.

.github/workflows/kbc-ci.yml
```diff
@@ -45,6 +45,8 @@ jobs:
         working-directory: kbc/platform/pod
       - run: python test_selfcheck.py
         working-directory: kbc/platform/pod
+      - run: python test_okf_package.py
```
jacoblee-io, line 48: **`test_incremental.py` is red on this head and no job runs it.**

This PR adds `test_okf_package.py` to the gate but leaves out the suite it grows the most. On this head:

```
$ cd kbc/platform/pod && python3 test_incremental.py
Traceback (most recent call last):
  File "test_incremental.py", line 383, in <module>
    test_resolve_affected_pages()
  File "test_incremental.py", line 38, in test_resolve_affected_pages
    assert aff == ["a.md", "b.md"], aff
AssertionError: []
```

It aborts on the first assertion, so none of the new byte-budget tests at `test_incremental.py:225-244` — the regression cover for `_fit_changeset_optional_payload`, `_compact_change_summary`, `_changeset_max_bytes`, all added by this PR — execute either, locally or in CI. `test_batching.py`, `test_mediaverify.py` and `test_source_snapshot.py` are also missing from the job, and `mediaverify.py` is touched here.

The body's `[x] Incremental and red-blue suites covered by the full local KBC runs` doesn't hold — the incremental suite does not pass on this head.

Two things needed: fix the underlying regression (see my `parse_okf_sources` comment — the fixture is right, the code changed under it), and add `test_incremental.py` to this workflow so it can't regress silently again.

kbc/platform/pod/compile_box.py
```diff
@@ -2675,6 +2746,16 @@ async def _emit_message(run: CompileRun, msg) -> None:
             # turn. Park the reply, keep the durability sync, skip selfcheck
             # (the final full-corpus pass owns it) and skip turn_done.
             run._last_turn_reply = reply
+            turn_pages_before = getattr(run, "_turn_page_hashes", None)
+            run._turn_page_hashes = None
+            if isinstance(turn_pages_before, dict):
+                turn_pages_after = incremental.page_hashes(run.workdir)
+                turn_changes = incremental.changed_pages(turn_pages_before, turn_pages_after)
+                selfcheck.stamp_siclaw_generated_metadata(
```
jacoblee-io: **Model-authored YAML can kill an entire batch compile, and the retry re-raises identically — a poison pill.**

This batch-path call is unguarded. `stamp_siclaw_generated_metadata` raises `ProvenanceStampError` for flow-style top-level frontmatter (`---\n{type: Topic, title: t}\n---`) or a duplicate top-level key — both reproduced directly. `_drive_batch_session` catches only `(ModelStallError, CLIConnectionError, ProcessError)`, so it escapes to `_run_batch_compile`'s handler and the run ends `batch_failed`.

The damage is worse than a failed turn: the raise sits **above** `await _sync_workspace(run, sent)` in the same `_suppress_turn_done` branch, so every page that batch just wrote is never persisted, and the batch is never marked done in `BATCH_PLAN.json`. Auto-resume re-runs it, the model writes the same page, and it crashes the same way — forever.

The sibling interactive call site at `:2448` is wrapped by the caller's try/except with the comment *"Fail-open: a self-check crash must not kill the turn (§4.5)"*. Model-controlled input must not be able to terminate a run; this call needs the same fail-open treatment.

Related on the interactive path: `_post_turn_selfcheck` clears `run._turn_page_hashes` (`:2443`) *before* stamping, so a stamp failure destroys the baseline and the broad `except Exception` at `:2772` then skips all of `normalize_exclusions_file` / `run_layer1` / coverage / `SELFCHECK.json` / the repair prompt for that turn. One malformed page disables the convergence gate with no error event, and the baseline is gone so no later turn recovers the delta. Also note a spurious abort: the duplicate-key test compares `key_node.value`, so the YAML-distinct keys `1:` and `'1':` are reported as duplicates.

kbc/platform/pod/selfcheck.py
```diff
@@ -580,8 +592,338 @@ def _okf_log_violations(rel: str, text: str) -> list[dict]:
...
+        if not target.is_file():
+            continue
+        try:
+            text = target.read_bytes().decode("utf-8")
+        except (OSError, UnicodeDecodeError):
+            continue
+        lines = text.splitlines(keepends=True)
+        if not lines or lines[0].strip() != "---":
+            continue
+        end = next((i for i, line in enumerate(lines[1:], start=1)
```
jacoblee-io: **The frontmatter terminator scan matches `---` inside a block scalar, and the resulting corruption is valid YAML — so nothing reports it.**

The scan accepts any line stripping to `---`/`...`. A block scalar containing such a line (common in pasted Chinese log excerpts) ends the frontmatter early. Ran `stamp_siclaw_generated_metadata` on:

```yaml
---
title: 节点重启流程
summary: |
  日志摘要:
  ---
  已重启完成
sources:
  - resource: raw/日志.md
verified:
  - by: person:alice
    at: 2026-01-01T00:00:00Z
---
正文
```

`end` resolves to the `  ---` line *inside* the scalar, so `generated:` is inserted at column 0 in the middle of it and `sources:` / `verified:` are pushed out of the frontmatter into the body. `parse_okf_frontmatter` on the result returns `err=None` — `summary` is truncated to `日志摘要:\n`, and `parse_okf_sources` returns `([], False, False)`.

So the page flips to `no_provenance`, every citation is lost, `raw/日志.md` becomes unaccounted, and `coverage.closed` can never be reached — with **no** "invalid frontmatter" diagnostic anywhere, because the output parses fine. It re-triggers every turn.

The terminator has to come from the parser, not a line scan: you already `yaml.compose` the frontmatter here, so use the document end mark rather than searching for `---`.

kbc/platform/pod/okf_package.py
```diff
@@ -0,0 +1,146 @@
...
+        raise OKFPackageError(f"Wiki directory does not exist: {root}") from error
+    if not root.is_dir():
+        raise OKFPackageError(f"Wiki path is not a directory: {root}")
+
+    files: list[tuple[str, bytes]] = []
+    markdown_pages: dict[str, dict] = {}
+    total_bytes = 0
+    for entry in sorted(root.rglob("*"), key=lambda item: item.relative_to(root).as_posix()):
+        rel = entry.relative_to(root).as_posix()
+        if entry.is_symlink():
```
jacoblee-io: **Hard links bypass the symlink guard, and `.json` gets zero validation — together they package arbitrary host-readable bytes into the import archive.**

Verified end to end:

```
leak.json is_symlink=False is_file=True nlink=3
collect ACCEPTED the hardlink
  archive member: leak.json  content: TOP SECRET OUTSIDE THE WIKI
```

`os.link('/etc/hosts', wiki/'leak.json')` passes `is_symlink() == False` and `is_file() == True`, so content from outside the Wiki directory is read and shipped inside the tar.gz that the import UI then indexes as knowledge — the exact escape the symlink check exists to prevent. Actor: anything that can write into the Wiki dir (the authoring agent, a compile-box tool call, a user-supplied tree).

`.json` is the ideal carrier because it is in `ALLOWED_SUFFIXES` (`:24`) but receives no validation at all — only the `.md` branch (`:76`) goes through the OKF gate. A `.json` file containing `\xff\xfe not json at all` is accepted and packaged, so "reject … invalid UTF-8 … and invalid OKF before writing" covers half of the allowed suffixes. Either drop `.json` or `json.loads(data.decode('utf-8'))` it here.

Credit where due — the **symlinked directory** case *is* correctly rejected (`OKFPackageError: Symbolic links are not allowed: linked`), and I confirmed determinism holds properly (`gzip.GzipFile(filename="", mtime=0)` plus pinned mode/uid/gid/uname/gname and sorted order — byte-identical across runs including a 60-char CJK path), as does validation-before-write (temp file under `destination.parent`, `os.replace` only after the size check, `finally` unlink on every failure path).

Three smaller things on the same function:
- `st_nlink > 1` is the check for the leak above; note it also needs `follow_symlinks=False`.
- `entry.read_bytes()` (`:65`) runs *before* the `MAX_FILE_BYTES` check, so an 8 GB page OOM-kills the pod instead of raising `OKFPackageError`; the CLI's `except (OKFPackageError, OSError)` doesn't catch `MemoryError` either. Gate on `entry.stat(follow_symlinks=False).st_size` first.
- The arcname is the raw relative name with no character validation, so a POSIX file literally named `..\..\evil.md` is stored verbatim as that tar member. `relative_to()` keeps POSIX `..` out, but any extractor treating `\` as a separator resolves it outside the root. A newline in a name also breaks anything parsing `tar -tzf` line-by-line. Whitelist arcnames at the collect step.

kbc/platform/pod/selfcheck.py
```diff
@@ -129,86 +129,97 @@ def has_assets_segment(rel: str) -> bool:
...
+    lowered = value.lower()
+    return (
+        "://" in value
+        or lowered.startswith(("mailto:", "urn:", "doi:"))
+        or value.startswith(("/", "./", "../"))
+        or lowered.startswith("references/")
+    )
+
+
+def parse_okf_sources(
```
jacoblee-io, line 183: **The `compiled_from` clean cut has no migration, so an already-compiled library re-entering the box loses all provenance at once.**

`grep -rn compiled_from` over the repo returns zero hits — no reader, no migrator, and no diagnostic that distinguishes "legacy v0.1 library" from "the model forgot provenance".

A library compiled before this deploy has `compiled_from:` and `okf_version: "0.1"`. Replay it into a box workspace for an incremental / resume / repair round on the new image and `parse_okf_sources` returns `([], derived, has_key=False)` for every page: `candidate_pages` records zero sources, `lint_candidate` emits `no_provenance` + `okf_index_frontmatter` corpus-wide, and `coverage()` marks every raw file unaccounted. That is the state that gates `closed` and fires the bounded repair loop, so the run gets a repair order to re-cite every source it already cited and ends unconverged — the "permanent convergence wedge" the docstring at `:187` warns about, arriving via format instead of via path normalization.

Note the `managed_sources` bare-path fallback does not rescue this: it only rescues bare values *inside* a `sources:` list, which a legacy page doesn't have.

This PR already ships the machinery for the deep fix — `stamp_siclaw_generated_metadata` plus span-preserving frontmatter rewriting mutates machine-owned YAML byte-faithfully. A one-shot `compiled_from` → `sources: [{resource: …}]` rewrite on first touch per workspace is a small addition on top of code that's already here, and it converts a silent corpus-wide wedge into a one-time migration. If a clean cut really is intended, it needs at minimum a distinct diagnostic naming the legacy format so the operator knows to recompile rather than chase a phantom provenance bug.

kbc/platform/pod/incremental.py
```diff
@@ -120,20 +127,181 @@ def build_changeset(
...
+    fallback. Logical hints use only the remaining budget and report omissions.
+    """
+    max_bytes = _changeset_max_bytes()
+    if _serialized_changeset_size(changeset) > max_bytes:
+        omitted = len(changeset.get("unaffected_pages", []))
+        changeset["unaffected_pages"] = []
+        if omitted:
+            changeset["unaffected_pages_omitted"] = omitted
+    if _serialized_changeset_size(changeset) > max_bytes:
+        raise ValueError("incremental execution scope exceeds durable CHANGESET byte budget")
```
jacoblee-io: **This `ValueError` escapes to the HTTP handler and wedges the incremental route permanently.**

`materialize_changeset` documents "invalid → `None` → fall back to full compile", and `_start_incremental` (`compile_box.py:3423`) relies on that: the `cs is None` branch two lines below is the full-compile fallback. But a scope-over-budget changeset *raises* instead, and there's no try/except on the call.

Verified: with `KBC_MAX_CHANGESET_BYTES=1024` (the code explicitly invites deployments to lower it) and 30 modified sources across 40 pages, `materialize_changeset` raises `ValueError: incremental execution scope exceeds durable CHANGESET byte budget`, and `authoring/RAW_CHANGES.json` is **still on disk** afterwards — the unlink at `:447` is only reached after a successful write.

So the exception surfaces as a 500 from `/message`, and because RAW_CHANGES survives, `_should_route_to_incremental` keeps returning True: every retry, including the owner's and the consumer's auto-resume, takes the same path and fails identically. The KB cannot be recompiled by any route until an operator hand-deletes the file or raises the env var. Pre-PR an oversized changeset was merely skipped by the sync cap and the turn still ran.

Returning `None` here (and letting the documented fallback fire) is the whole fix.

kbc/platform/pod/incremental.py
```diff
@@ -120,20 +127,181 @@ def build_changeset(
...
+        if isinstance(item, dict) and isinstance(item.get("path"), str)
+    }
+    retained_diffs = 0
+    for path in sorted(diffs):
+        diff = diffs[path]
+        item = modified_by_path.get(path)
+        if item is None or not diff:
+            continue
+        item["diff"] = diff
+        if _serialized_changeset_size(changeset) > max_bytes:
```
jacoblee-io, line 255: **A budget-dropped diff is indistinguishable from "the consumer sent no diff".**

Verified: `build_changeset(..., diffs={'sa.md': '@@ ' + 'x'*8000})` with an 8 KB budget emits

```json
{"path": "sa.md", "affected_pages": ["a.md"], "diff": ""}
```

and nothing else — no `change_summary`, no counter, no omission flag. The diff was well under `KBC_MAX_DIFF_BYTES` (64 KB), so the documented per-source degrade never fired; the budget loop dropped it.

The only markers (`modified_with_diffs` / `modified_without_diffs`) live inside `change_summary`, which `_compact_change_summary` omits entirely when the consumer sends no `summary` — i.e. the backward-compatible default path (`raw.get("summary")`). The model then reads `"diff": ""` as the documented "no diff supplied, re-read the source" case and cannot tell it apart from "the machine computed a surgical diff and threw it away".

Compare `unaffected_pages_omitted` at `:179`, which does mark its own omission — that's the pattern this needs. A truncated payload must not read like a complete one.

jacoblee-io (changes requested):
## Review

This bundles four independent efforts: structured box failure diagnostics in the TS capability layer, an incremental `CHANGESET.json` byte budget, the OKF v0.2 `compiled_from` → `sources[].resource` clean cut, and a new standalone wiki tar.gz packager. CI is green on all four checks.

**CI being green is misleading here.** `kbc/platform/pod/test_incremental.py` **fails on this head**, and it is not wired into `kbc-ci.yml`. I ran it:

```
$ cd kbc/platform/pod && python3 test_incremental.py
AssertionError: []
  test_incremental.py:38 in test_resolve_affected_pages
    assert aff == ["a.md", "b.md"], aff
```

Zero `OK` lines print — it aborts on the first assertion. The same command passes on `origin/main`. That failure is not a stale fixture; it is the symptom of a real behaviour regression in `parse_okf_sources` (first inline comment), and the suite this PR grows by +105 lines to cover the new byte budget is never executed by any job.

I verified the central mechanism directly against the real function:

```
resource='snap/three.md'  inventory={'snap/three.md'}  -> (['snap/three.md'], False, True)
resource='snap/three.md'  inventory={}                 -> ([], False, True)   # source deleted → citation vanishes
resource='./live.csv'     inventory={'live.csv'}       -> ([], False, True)   # dot-slash form dropped
resource='raw/live.csv'   inventory={'live.csv'}       -> (['live.csv'], ...)  # works
resource='raw//live.csv'  inventory={'live.csv'}       -> (['/live.csv'], ...) # absolute-path leak
```

and the packaging leak:

```
leak.json is_symlink=False is_file=True nlink=3
collect ACCEPTED the hardlink
  archive member: leak.json  content: TOP SECRET OUTSIDE THE WIKI
symlinked DIR rejected: OKFPackageError Symbolic links are not allowed: linked
```

Requesting changes on the inline findings. Ranked most severe first; the first four are the ones I'd block on.

### Scope

Four independently landable and independently revertable efforts arrive as one 34-file / +2065 / −494 diff across two languages. Recent PRs in this repo (#464–#469) are each 1–3 commits on a single concern. The OKF clean cut alone is `selfcheck.py` +562 plus prompts and fixtures, and it carries the migration question below — it warrants its own review and its own revert boundary. The PR body's stated reason for bundling (test-runner conflicts) justified it for #453; it doesn't obviously apply across a TS/Python split.

### Smaller items

**`selfcheck.py` — the YAML rewriter contradicts two of its own claims.** The PR says it preserves "comments, quoting, anchors, and unknown extensions byte-for-byte" and that ambiguity "fail[s] atomically instead of corrupting the document". Two gaps:
- `value_node.end_mark.index` for a block collection points at the *start of the next key* (PyYAML emits BlockEnd only after skipping comments and blank lines), so a comment sitting between a machine-owned key and the following key is inside the replaced span and gets deleted. A `# NOTE: 人工维护，勿删` after `verified:` does not survive a stamp.
- The retained-alias guard collects only `value` nodes (`selfcheck.py:879-880`), so an anchor declared on a machine-owned **key** (`&g verified: 1` with `other: *g`), or an alias used as a retained **key** (`verified: &v hello` with `*v: y`), escapes detection. Both write an unparseable document to disk (`found undefined alias`) instead of aborting. The `<<: *g` merge-key case *is* caught, which shows the guard was meant to cover this.

**`selfcheck.py:920` — stamping is not atomic across pages.** The validation is all-or-nothing, but the write loop is in-place `target.write_bytes(...)` with no temp+rename. An `OSError` on the third of five pages leaves two stamped, one possibly truncated, and no rollback. `_write_text_atomic` already exists in this file and is what `_write_exclusions_canonical` uses.

**`selfcheck.py:874` — `verified:` is removed unconditionally, and the stamper runs for cosmetic-only changes.** `remove_keys = {"generated", "verified"}` plus the `compile_box.py:2565` call after `normalize_body_source_annotations` (a pure formatting fix) means a page carrying `verified:\n  - by: person:alice` loses that human attestation and is re-signed `process:siclaw-kbc`, with nothing recording that a human sign-off was dropped. Also `add_stable` uses `fm.get("status") is None`, which cannot distinguish an explicit `status: null` from an absent key — an author who writes `status: null` gets `status: stable`.

**`src/gateway/capability/failure.ts` — three issues in 109 lines.**
- `:26` — the control-char class `/[\u0000-\u0008\u000B\u000C\u000E-\u001F]/` skips LF, CR and TAB, so `failure.message` carries newlines into the persisted checkpoint and on to `operation.error_message` / upstream logs. The docstring says it strips controls, and the paired `asFailureToken` has a test named "rejects tokens with whitespace or newlines (log-injection)" — the `message` channel is the one gap.
- `:28` — `cleaned.slice(0, max)` truncates by UTF-16 code unit and can split a surrogate pair, emitting a lone surrogate into JSON. The Python producer caps at 256 *code points*, so a message the box thinks is under the cap can still exceed 256 UTF-16 units here. `truncateAtBoundary` (`src/agentbox/delegation-summary.ts`) exists for exactly this.
- `:36` — `normalizeFailure` now defaults `code`/`stage` instead of returning `undefined`, and `typeof [] === "object"`, so `checkpoint = {"failure": {}}` hydrates a concrete `runtime_failure` diagnostic on a healthy run. Combined with `rec.failure` never being cleared and being spread into *every* persist including the terminal `done` write, a phantom failure can end up on a successful run's store row.
- Also `structuredBoxFailure` re-implements the whole message-fallback ladder and then calls `normalizeFailure`, which runs it again — it collapses to `normalizeFailure({...evt, code: asFailureToken(evt.code) ?? "box_error"})`.

**`src/gateway/server.ts:~739`** builds the diagnostic from `err.name`, which is `"Error"` for nearly every throw, so the persisted failure reads `relay_failed:Error` — as uninformative as the empty checkpoint this PR set out to fix. For a non-`Error` throw it still asserts `:Error` while omitting `exception_class`.

**`okf_package.py` has no in-repo consumer.** `grep -rn 'okf_package\|package_okf\|collect_import_files\|write_import_archive'` over `src/` and `portal-web/` returns nothing; the validator it must agree with (`validateAdoptionOKFStructure`, named in the `selfcheck.py:986` docstring) lives in another repo. 276 new lines plus a permanent CI step for a feature this repo cannot exercise end to end — worth stating as an explicit cross-repo dependency in the body.

**PR body is stale in two ways.** The last verification box — "[ ] GitHub Type Check, Test, Portal Web Test, and KBC Box Test on the current head" — is unchecked though all four are green, so anyone gating on the checklist blocks on nothing. And "Deployment: No deployment is included." sits against root `CLAUDE.md`'s rule for this path: *"Behavior changes need a `siclaw-kbc-box` image rebuild + redeploy (runtime env `SICLAW_COMPILE_BOX_IMAGE`); a rolled-out runtime does NOT pick up a new box image for existing live sessions."* This PR changes the box's SSE error-frame vocabulary (`_error_event`) and the paired TS side now consumes it — a runtime deployed with the new `session-driver.ts` against an un-rebuilt box gets bare `error` frames again. Worth spelling out the rebuild ordering for whoever rolls it out.

**`.github/workflows/kbc-ci.yml`** — besides `test_incremental.py`, `test_batching.py`, `test_mediaverify.py` and `test_source_snapshot.py` are also absent from the job. `mediaverify.py` is touched by this PR.

jacoblee-io:
Note on timing: my review was submitted at 12:22:31 against `e832d4df`, and `27d4f25a` ("close OKF retry correctness gaps") landed at 12:22:10 — 21 seconds earlier. So the two crossed in flight and the review was **not** written against the current head. I've re-verified all of it on `27d4f25a` so you don't have to guess which comments are stale.

**Already fixed by `27d4f25a` — please ignore that comment.**

The `.github/workflows/kbc-ci.yml` finding is resolved. On the new head:

```
$ cd kbc/platform/pod && python3 test_incremental.py
ALL OK  test_incremental
```

and both `test_incremental.py` (line 50) and `test_emit.py` (line 56) are now in the workflow. Good — that was the one I cared most about, since it was the reason CI could be green over a red suite.

**Partially fixed.** The `parse_okf_sources` comment splits in two. The deleted-source half is addressed: `candidate_pages` now takes `additional_managed_sources` and the incremental lookup feeds deleted Raw paths back in, which is a reasonable narrow fix. But `parse_okf_sources` itself is unchanged, so two of the three cases I listed still reproduce identically on `27d4f25a`:

```
resource='./live.csv'     inventory={'live.csv'}  -> ([], False, True)      # still dropped
resource='raw//live.csv'  inventory={'live.csv'}  -> (['/live.csv'], ...)   # still absolute
```

`_explicit_external_source` still classifies `./`-prefixed resources as external before normalization. Normalizing (`posixpath.normpath`) before the external test fixes both in one line.

**Still reproduce on `27d4f25a`, verified:**

- **`compile_box.py` batch-path `stamp_siclaw_generated_metadata`** — still unguarded, and still sits above `sent = getattr(run, "_sync_sent", None)`, so a `ProvenanceStampError` still skips the durability sync and poisons the retry.
- **Block-scalar frontmatter terminator** — `generated:` is still injected inside the scalar (confirmed: its offset precedes the scalar's own trailing content).
- **Hardlink in the import packager** — still accepted and still shipped:
  ```
  ACCEPTED -> leak.json : OUTSIDE THE WIKI
  ```
- **`compiled_from` migration** — `grep -rn compiled_from` over the repo is still 0 hits: no reader, no migrator, no distinguishing diagnostic.
- **`incremental.py` `raise ValueError`** — still raised at the byte-budget check, and `_start_incremental` still calls `materialize_changeset` with no try/except, so it still escapes as a 500 with `RAW_CHANGES.json` left on disk.

I did not re-check the unmarked-truncated-diff comment on the new head; nothing in `27d4f25a` looks aimed at it, but treat that one as unverified against the current head rather than confirmed.

Happy to re-review whenever you push — and if any of the remaining findings look wrong to you, push back, several of them are judgement calls about the right depth rather than clear bugs.

LikiosSedo:
P2 package/import contract follow-up is fixed in `f84b1073`.

The packager now mirrors the receiver for:
- UTF-8 validation of every packaged Markdown/JSON file
- the 255-Unicode-character `candidate/` path limit
- the `authoring/EXCLUSIONS.json` and `authoring/CONTRADICTIONS.json` allowlist

Regression coverage includes the three rejected packages, the exact 255-character boundary, multibyte path counting, and both allowed authoring ledgers.

Validation: `test_okf_package.py`, `test_incremental.py`, `test_batching.py`, `test_emit.py`, `git diff --check`, and all four GitHub checks pass.

kbc/platform/pod/selfcheck.py
```diff
@@ -580,8 +617,411 @@ def _okf_log_violations(rel: str, text: str) -> list[dict]:
...
+    seen: set[str] = set()
+    for key_node, value_node in root.value:
+        key = key_node.value
+        if key in seen:
+            raise ProvenanceStampError(
+                f"{rel}: duplicate top-level YAML key {key!r} is ambiguous")
+        seen.add(key)
+        pairs.append((key, key_node, value_node))
+
+    remove_keys = {"generated", "verified"}
```
jacoblee-io, line 910: **Still open from my first review: this deletes human attestation records on a runtime-only cosmetic change.**

`remove_keys` is unconditional, and the stamper runs over `mechanically_changed` — which includes pages touched only by `normalize_body_source_annotations`, a pure formatting fix. So a page carrying

```yaml
verified:
  - by: person:alice
    at: 2026-01-01T00:00:00Z
```

loses that record and is re-signed `process:siclaw-kbc`, with nothing anywhere recording that a human sign-off was dropped, and no way to distinguish "the model rewrote this page" from "the runtime reflowed one annotation".

What makes me think this isn't intended: the import profile added by this same PR goes out of its way to **preserve** external verification records ("Preserve legitimate external verification records, wikilinks, bundle-root links, and unknown fields"). So `verified:` is treated as meaningful, human-owned data on the way in, and as machine-owned scratch on the authoring side. One of the two is wrong.

If authoring-side `verified:` really is machine-owned, a sentence saying so in the docstring would settle it. If it isn't, the fix is to scope the removal — drop `verified:` only when the page's *content* changed (which is what invalidates an attestation), not when the runtime reflowed its own annotation.

Same line, smaller: `add_stable` uses `fm.get("status") is None`, which still cannot distinguish an explicit `status: null` from an absent key, so an author who deliberately writes `status: null` gets `status: stable`.

jacoblee-io (approved):
## Re-review (`40f81420` + `f84b1073`) — approving, dismissing my earlier changes-request

Seven of my eight inline findings are fixed, and several are fixed deeper than I asked. Re-verified every one by execution on `f84b1073`.

**Fixed:**

| Finding | Evidence on `f84b1073` |
|---|---|
| bare-resource resolution | `'./live.csv'` → `(['live.csv'], …)`, `'raw//live.csv'` → `(['live.csv'], …)`. Both were `[]` / `['/live.csv']` before. |
| `test_incremental` red + not in CI | `ALL OK`; wired in, along with `test_emit`. |
| batch-path `ProvenanceStampError` | All four call sites now go through `stamp_siclaw_generated_metadata_best_effort` returning `(stamped, failures)`. |
| block-scalar frontmatter terminator | `generated:` no longer lands inside the scalar. |
| hardlink into the archive | `OKFPackageError: Hard links are not allowed: leak.json` |
| `.json` unvalidated | `OKFPackageError: Knowledge file is not UTF-8 text: bad.json` |
| `ValueError` poison pill | Now `ChangesetScopeTooLarge`, caught in `materialize_changeset`, unlinks **both** RAW_CHANGES and CHANGESET, returns `None` → the documented full-compile fallback. |
| unmarked truncated diff | `modified_diffs_omitted` reserved *before* optional content is admitted. |

Plus, from the review body: `failure.ts` control class widened to the full C0+C1 range with `\s+` collapse and code-point slicing (both TS findings closed); `server.ts` uses `asFailureToken(err.name) ?? "Error"` with `"NonErrorThrown"` for non-`Error` throws; the stamp write is now `_write_text_atomic`; and the alias guard covers key nodes as well as values.

Two of these are better than what I proposed. Turning the raising stamper into a `_best_effort` API that returns `(stamped, failures)` fixes the class of problem rather than the one unguarded call site I pointed at — the interactive path no longer needs its blanket `except Exception` to compensate either. And catching `ChangesetScopeTooLarge` *inside* `materialize_changeset` and consuming the one-shot input is the right place: the caller's documented `cs is None` fallback now actually gets to run, instead of the caller learning about a new exception type.

Local runs on this head: `test_incremental`, `test_okf_package`, `test_redblue`, `test_batching` all pass. `test_selfcheck` only fails on a missing local `aiohttp` (import error, not an assertion) — CI's KBC Box Test covers it and all four checks are green.

### Two things still open — follow-ups, not blockers

**1. The `compiled_from` migration** (my `selfcheck.py:182` thread, still unresolved). `grep -rn compiled_from --include=*.py` is still 0 hits repo-wide: no reader, no migrator, no diagnostic distinguishing a legacy v0.1 library from "the model forgot provenance". A pre-PR compiled library replayed into a box still loses all provenance at once and burns its repair budget unconverged. If the clean cut is deliberate because no such library exists in the wild yet, say so in the PR body and I'll stop raising it — it just isn't stated anywhere, and it's the one finding where I can't tell intent from the diff.

**2. `verified:` is still removed unconditionally** (inline comment below).

### Smaller, unchanged

- Still no `docs/design/` doc and no CLAUDE.md Matrix row for `report_domain` / `channel`-style OKF contract / `okf_package`. The root Matrix row for `kbc/platform/pod/*` still names only `test_compile_box.py` as the verification command, while this PR has grown the gate to eight suites.
- `okf_package.py` still has no in-repo consumer (`grep` over `src/` and `portal-web/src` is empty) — worth one line in the body naming it as a cross-repo dependency so the next reader doesn't hunt for the importer.
- PR body's CI checkbox now reads `[x] … on 27d4f25a`, two commits behind the current head; and the Deployment section still says only "No deployment is included" without the `siclaw-kbc-box` rebuild + `SICLAW_COMPILE_BOX_IMAGE` ordering that root CLAUDE.md requires for box behaviour changes. Both are one-line edits.
- The four-efforts-in-one-PR scope point stands, but at this stage splitting would cost more than it saves — noting it for next time rather than asking for it here.

PR #472, opened 2026-08-04
fix: compact incremental knowledge diffs

LikiosSedo:
## Summary

Incremental knowledge compiles currently expose path-level source lists without enough document-level context, causing the compiler to infer intent from attachment churn and oversized diffs. This adds a bounded, validated logical-source summary to CHANGESET and makes the scoped directive use it for orientation while retaining path arrays and diffs as the exact execution scope.

The summary is optional and backwards-compatible. Opaque provider identities are dropped, logical changes are capped, and diff coverage explicitly distinguishes surgical diffs from scoped source re-reads.

## Test Plan

- [x] python3 kbc/platform/pod/test_incremental.py
- [x] python3 -m py_compile for incremental.py and test_incremental.py
- [x] npm test (243 files, 5017 passed, 2 skipped)

### DIFF
--- a/kbc/platform/pod/incremental.py
+++ b/kbc/platform/pod/incremental.py
@@ -94,6 +94,7 @@ def build_changeset(
     changed_sources: dict,
     *,
     diffs: dict[str, str] | None = None,
+    summary: dict | None = None,
     baseline_fingerprint: str | None = None,
     snapshot_fingerprint: str | None = None,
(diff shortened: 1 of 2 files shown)


LikiosSedo:
Superseded by #471. The compact incremental knowledge diff changes were integrated into that branch and verified there.

PR #475, opened 2026-08-05
fix: keep federation source text-readable

LikiosSedo:
## Summary

- replace the literal NUL byte in the federation aggregator source with the equivalent TypeScript escape
- preserve the runtime separator while allowing Git and code knowledge ingestion to treat the file as text

## Validation

- `npm test -- src/gateway/prom-federation-aggregator.test.ts src/gateway/metrics-aggregator.test.ts src/gateway/federation-self-metrics.test.ts src/gateway/internal-api.test.ts` (71 tests)
- `npm run build`
- verified a fresh `git archive` contains valid UTF-8 source with zero literal NUL bytes


LikiosSedo:
Moved this fix into #474 as commit 417889cb, so this standalone PR is no longer needed.

PR #477, opened 2026-08-06
feat(lark): add personal webchat command

chent1996:
## Summary

Adds a personal-chat /webchat command that asks the connected frontend for a short-lived browser chat link. Admission and redemption remain frontend-owned, while unsupported portals degrade to a readable error.

### Problem

Users who can reach an agent through a Feishu personal bot have no direct way to move the conversation into a browser window. Repeated Feishu delivery can also invalidate an earlier one-shot link unless the portal receives a stable request identity.

### Solution

- Handle the exact /webchat command before agent routing and keep group chats silent.
- Deliver one-shot links through a validated HTTP(S) card button with a safe text fallback.
- Preserve unrelated prefixes such as /webchats as ordinary agent input.
- Forward the inbound Feishu message ID as request_id for portal-side durable deduplication.
- Share denial rendering with /apikey while keeping command-specific copy.
- Register a graceful unsupported handler in the standalone Portal adapter.
- Document the protocol, expiry, delivery, and command-boundary invariants.

The corresponding upstream portal must implement channel.issueWebChatLink and persist request_id deduplication. Older portals return the unavailable response without affecting ordinary chat.

## Test Plan

- [x] npm test — 5037 passed, 2 skipped
- [x] npm run build
- [x] Focused Lark/channel/adapter suites — 413 passed
- [ ] End-to-end redemption with the corresponding upstream portal branch


src/gateway/channels/lark.test.ts
```diff
@@ -4470,3 +4473,141 @@ describe("handleLarkMessage — /apikey", () => {
...
+  it("drops the actual command in groups, but not a different slash command sharing its prefix", async () => {
+    const exact = makeLarkClient();
+    for (const text of ["/webchat", "/webchat open"]) {
+      await handleLarkMessage(
+        makeTextEvent(text, { chat_type: "group" }), exact, "group-channel-1",
+        makeAgentBoxManager("a1") as any, undefined, {} as any, "zh-CN",
+      );
+    }
+    expect(exact.im.message.reply).not.toHaveBeenCalled();
+    expect(issuePersonalWebChatLinkMock).not.toHaveBeenCalled();
```
jacoblee-io, line 4836: **This test passes with the production group-drop block deleted.**

Mutation-verified: removing `if (parseWebChatCommand(text))` at `lark.ts:1320-1323` entirely leaves all 7 `/webchat` tests green.

The reason is that the loop sends `makeTextEvent(text, { chat_type: "group" })` with **no mention**, so the @-gate already swallows the message before the drop block would ever matter. The three assertions (no reply, no RPC, no prompt) hold either way, so the one behaviour the block exists for — `@bot /webchat` dropped *before* `botMentioned` is computed — is never exercised.

That ordering is the security-relevant part, and your own comment at `lark.ts:1316` states why: the reply carries a single-use link that signs the redeemer into a chat session, so *"whoever opened it first would own the session — answering at all, even 'DM me', is worse than silence"*. If a later refactor moves that block below the @-gate, `@bot /webchat` in a group falls through to the agent and nothing fails.

One line fixes it — add the mention, as the `/webchatops` case just below already does:

```ts
makeTextEvent(text, {
  chat_type: "group",
  mentions: [{ key: "@_user_1", id: { open_id: "ou_bot_self" } }],
}),
// …and pass "ou_bot_self" as botOpenId
```

Worth noting the `/apikey` group test at `:4453` has the identical gap — same missing mention — so this is an inherited pattern rather than a new sin. Both are one line, and the credential one matters at least as much.

src/gateway/channels/lark.test.ts
```diff
@@ -4470,3 +4473,141 @@ describe("handleLarkMessage — /apikey", () => {
...
+      "zh-CN",
+      makePersonalConfig("sicore_authorized"),
+    );
+  }
+
+  const replyText = (lark: ReturnType<typeof makeLarkClient>) =>
+    lark.im.message.reply.mock.calls[0][0].data.content as string;
+
+  it("issues a link with the stable inbound message id and never reaches the agent", async () => {
+    issuePersonalWebChatLinkMock.mockResolvedValue({
```
jacoblee-io, line 4607: **The whole denial-rendering path is untested — ~90 new lines with no coverage.**

Every `issuePersonalWebChatLinkMock` fixture in this PR returns `success: true`; none carries `denied`. Mutation-verified: stubbing `webchatDenialCopy` to `return undefined` — which kills every localized refusal, the generic fallback, and the denial card — leaves all 200 `lark.test.ts` tests passing.

So `WEBCHAT_DENIAL_COPY_BY_LOCALE`, `webchatDenialCopy`, `formatWebChatDenialReply` and `buildWebChatDenialCard` ship unverified. Concretely, none of these would be caught:
- a `binding_required` refusal rendering an empty or garbled body
- an expired `denied.actionUrl` reaching a card button
- the `resume` line telling a blocked user to resend `/apikey` instead of `/webchat` — which is precisely the mistake the separate copy tables exist to prevent, so the table split is currently protected by nothing

The `/apikey` sibling has ~10 denial tests at `lark.test.ts:3733-4229` covering exactly these shapes; they should port over almost verbatim with the fixture swapped.

src/gateway/channels/lark.ts
```diff
@@ -2170,6 +2258,173 @@ function parseApiKeyCommand(text: string): { subcommand: string } | null {
  */
 const apiKeyRequestsInFlight = new Set<string>();
 
+const WEBCHAT_COMMAND_RE = /^\/webchat(?:\s|$)/i;
```
jacoblee-io, line 2379: **This inverts the sibling's namespace rule, and the stated justification names a prompt that doesn't exist.**

`parseApiKeyCommand` is `/^\/apikey/i` — no trailing boundary. The design doc is explicit about why: *"`API_KEY_COMMAND_RE` deliberately carries **no trailing word boundary**: with `\b`, `/apikeys` escapes the namespace and reaches the model."* The command claims the whole namespace so a malformed subcommand can never fall through to the agent.

`/^\/webchat(?:\s|$)/i` requires a boundary, so the same class of typo now escapes: `/webchats` matches nothing, sails past the group drop-gate at `:1320` (`parseWebChatCommand` returns null), and reaches the agent as a raw prompt in both personal and group chat.

The JSDoc justifies this as protecting *"an existing agent prompt such as `/webchatops`"*. I grepped the repo:

```
src/gateway/channels/lark.ts:2266   ← this comment
src/gateway/channels/lark.test.ts:4555, 4604   ← the two tests this PR added
```

There is no such prompt — the rationale is a hypothetical written as a fact, and the two tests assert the hypothetical rather than a requirement. Meanwhile the plural typo the sibling's rule was actually written for is untested and unhandled.

Not severe: unlike `/apikeys`, a leaked `/webchats` mints nothing and costs a wasted LLM turn. But it's a silent divergence from a documented invariant. Either drop the boundary (and let `/webchatops` get usage help, exactly as `/apikeyfoo` does today), or keep it and say in the comment that the plural typo is a knowingly accepted leak — the current comment implies neither.

src/gateway/channels/lark.ts
```diff
@@ -1137,6 +1151,71 @@ export async function handleLarkMessage(
...
+        // second Runtime replica.
+        const issued = await issuePersonalWebChatLink(personalChannelId, senderOpenId, frontendClient, messageId);
+        const reply = formatWebChatIssueReply(issued, locale);
+        // Audit line: this path bypasses chat persistence, so without it there is no runtime-side
+        // record of who asked for a session and when. NEVER log actionUrl — it is a bearer token.
+        console.log(`[lark] /webchat channel=${personalChannelId} sender=${senderOpenId} ok=${issued.success} agent=${issued.agentId ?? "?"}`);
+        // Card with a button rather than a bare URL: the client unfurls text links, and handing a
+        // one-time token to a preview fetcher is not something to rely on the endpoint tolerating.
+        let cardDelivered: boolean | null = null;
+        if (issued.success && issued.actionUrl) {
```
jacoblee-io: **Two issues on the success path, both about a value that came from the frontend.**

**1. The minted `actionUrl` skips the scheme gate every refusal link goes through.** `isRenderableActionUrl` (`:1986`) restricts a link to `http:`/`https:` precisely because *"this value reaches a Feishu `open_url` button, where other schemes resolve as deeplinks"* — but it's only reached via `rendersActionLink` on the denial path. Here `issued.actionUrl` goes straight to `url:`. A frontend misconfiguration (or a non-string value) that yields `lark://applink.feishu.cn/...` or a relative path lands on the button unchecked. The same gap exists for `/apikey`'s `pickupUrl`, so it's symmetric rather than a regression — but this is the more sensitive link, and reusing the existing predicate is one call.

**2. A freshly-minted link that is already expired renders a self-contradicting reply.** If `issued.expiresAt <= Date.now()` by delivery time — ordinary clock skew between the issuing frontend and this pod, plus Feishu latency — `deliverSingleUseLink` takes its `remaining === "expired"` branch and the user gets:

> ✅ 网页对话链接已就绪
> 打开后是一个全新的空白对话，不会延续这里的聊天记录
> 之前的链接已过期，请再发一条消息获取新链接。

"Ready" and "expired" in one message, no URL at all — and the recovery instruction is inherited from the *binding-refusal* flow, where any next message makes the frontend mint a fresh `actionUrl`. For `/webchat` that's false: an ordinary next message goes to the agent as a prompt and mints nothing. The user follows the instruction, gets an LLM answer, and reports the command as broken. `buildWebChatDenialCard` already guards `minutesUntil(...) === "expired"`; the success path doesn't. Worth either checking it here or passing a `/webchat`-specific expired string.

src/gateway/channels/lark.ts
```diff
@@ -2170,6 +2258,173 @@ function parseApiKeyCommand(text: string): { subcommand: string } | null {
...
+function formatWebChatIssueReply(result: PersonalWebChatLinkResult, locale: LarkLocale): string {
+  if (!result.success || !result.actionUrl) {
+    const localized = result.denied ? formatWebChatDenialReply(result.denied, locale) : null;
+    if (localized) return localized;
+    const reason = truncateDenialProse(result.denied?.message)
+      ?? result.error
+      ?? (locale === "en-US" ? "unknown error" : "未知错误");
+    return locale === "en-US"
+      ? `❌ Could not create a web chat link: ${reason}`
+      : `❌ 生成网页对话链接失败: ${reason}`;
```
jacoblee-io, line 2462: **`result.error` is interpolated unbounded, and an oversized reply turns "always answer" into total silence.**

The line above caps `denied.message` via `truncateDenialProse` (`PERSONAL_DENIAL_MESSAGE_MAX_CHARS`), but `result.error` — the other frontend-supplied string — goes in raw. A frontend returning a stack trace or a SQL error dump in `error` produces a reply Feishu rejects for size; `replyToLark` swallows the non-zero code and returns false; `cardDelivered` is null so the `??` falls through to that same failing send; the sender gets nothing.

That's the one outcome this command explicitly rules out — your own comment at `:1207`: *"the user is waiting on a link, and silence reads as a broken bot."*

Running `error` through the same `truncateDenialProse` cap closes it. Mirrored in `/apikey` at `:2481`, so one fix covers both.

jacoblee-io (commented):
## Review

Adds a `/webchat` personal-chat command that asks the frontend for a one-shot browser-chat link, delivered as a card button with a text fallback, with a `success:false` stub in the standalone Portal adapter. It is a deliberate near-clone of `/apikey`, and it's clear the sibling was studied rather than copy-pasted — the comments explain each divergence instead of just restating what the code does.

**The runtime behaviour is right.** I checked it against every invariant the `/apikey` Change Impact Matrix row and `docs/design/2026-07-28-feishu-apikey-command.md` record, and all of them hold:

- graceful `success:false` adapter stub — present, and I traced it end to end: it renders as `❌ 生成网页对话链接失败: Web chat links are only available in Upstream mode`, not `undefined`
- group drop sits at `lark.ts:1321`, `botMentioned` is computed at 1326 — correctly **before** the @-gate
- both dispatch sites share one parser
- bare command only may mint; a subcommand → usage help with zero RPC
- single-flight per `(channel, sender)`, released in `finally`, bypassing the per-binding queue
- never model-callable — deterministic parse, every branch returns
- `actionUrl` never logged (the audit line takes `ok`/`agent` only)
- both the thrown-RPC and missing-frontend-client paths reply
- no PAIR-style authorized-mode gate
- timestamps go through `minutesUntil`, sidestepping the `Intl.format` RangeError that bit `/apikey`

402 tests pass across the three suites; CI is green. Not blocking — but the test coverage does not hold up under mutation, and there are three small correctness items I'd fix. Details inline, most valuable first.

### Three things I'd land before merge

1. A group test that actually exercises the drop (first comment) — the current one passes with the production block deleted.
2. Decide the `/webchats` question (third comment) — either drop the regex boundary or record the leak as deliberate.
3. The expired-link reply (fourth comment) — it currently says "✅ ready" and "expired" in the same message.

### Harness

No `docs/design/` doc and no Change Impact Matrix row. The sibling has both, and the row at `CLAUDE.md:126` is what a contributor greps before touching `lark.ts` — it currently names only `channel.issueApiKey`/`channel.apiKeyStatus` as the RPCs requiring an adapter stub, so `channel.issueWebChatLink` is now a third such RPC that appears nowhere in the harness. The invariants this PR depends on live only as inline comments: one-live-link-per-requester (`lark.ts:2286`), the `request_id` durable-dedup contract the frontend owes (`channel-manager.ts` JSDoc), and the group-drop-before-@-gate ordering. Per the repo's own protocol — *"every surprise improves the harness"* — a row naming the new RPC and those three rules is cheap and is exactly what keeps the next refactor from quietly breaking the ordering that finding #1 shows the tests won't catch.

Cheapest half of that fix: widen the comment above `CHANNEL_API_KEY_UNSUPPORTED` (`adapter.ts:3218`) so it states the general rule — *every unconditionally-called `channel.*` RPC needs a stub* — instead of reading as if it governs exactly two API-key methods.

For the record, `adapter-rpc.test.ts`'s net −9 is a comment rewrite plus the handler count 59→60; no case dropped, no assertion weakened. And the new tests use `sicore.example` / `sicore_authorized` only where the surrounding file already does, so no new internal-naming leakage.

src/gateway/channels/lark.ts
```diff
@@ -2170,6 +2258,173 @@ function parseApiKeyCommand(text: string): { subcommand: string } | null {
...
+const WEBCHAT_OPEN_BUTTON_BY_LOCALE: Record<LarkLocale, string> = {
+  "zh-CN": "打开网页对话",
+  "en-US": "Open web chat",
+};
+
+/**
+ * Copy for a `/webchat` refusal: the reason's own wording when known, else the generic pair — but
+ * only when there is actually a link to offer. Mirrors apiKeyDenialCopy.
+ */
+function webchatDenialCopy(
```
jacoblee-io: **These three helpers are token-for-token clones of their `/apikey` twins.**

Your argument that the copy *tables* must be separate is right, and I'm not disputing it — `"关联完成后回来重发 /webchat"` vs `/apikey` is a real per-command difference, and getting it wrong would send a blocked user to the wrong command. That justifies the 6 strings.

It doesn't reach the three functions built on them:
- `webchatDenialCopy` (2323) vs `apiKeyDenialCopy` (2577) — identical bodies, `WEBCHAT_*` for `API_KEY_*`
- `formatWebChatDenialReply` (2333) vs `formatApiKeyDenialReply` (2587) — identical, down to the nested-not-chained `rendersActionLink` branch
- `buildWebChatDenialCard` (2351) vs `buildApiKeyDenialCard` (2543) — identical, same button lookup and neutral fallback

~60 lines carrying zero command-specific logic. The cost is specific: those bodies encode bug fixes whose *explanations* survive only on the `/apikey` side — *"Falling back to the GENERIC notice rather than null when a link must still be delivered: returning null stranded a refusal that carried a usable `actionUrl` but no prose"* and *"Chaining the expired notice onto `actionUrl` alone told `access_denied` senders their live link had expired."* The webchat copies have the shape but not the reasoning, so the next person fixing that logic patches one side and silently misses the other — and per the finding above, no test would catch it.

One factory taking `(copyTable, genericTable)` and returning the trio keeps the per-command `resume` wording you correctly insisted on.

src/gateway/channels/lark.ts
```diff
@@ -1137,6 +1151,71 @@ export async function handleLarkMessage(
...
+            url: issued.actionUrl,
+            expiresAtMs: issued.expiresAt,
+          }, reply);
+        } else {
+          const denialCard = buildWebChatDenialCard(issued.denied, locale);
+          if (denialCard) {
+            cardDelivered = await deliverSingleUseLink(larkClient, messageId, locale, denialCard, reply);
+          }
+        }
+        // No rotation-style escalation here, unlike /apikey: a link that never arrives costs the
```
jacoblee-io: This comment contradicts the one at `:2286`.

Here: *"a link that never arrives costs the sender nothing they held — resending /webchat mints another."*

There: *"Minting replaces the sender's outstanding link (one live link per requester), so a double-tap would hand back two links of which only the last works."*

If minting replaces the outstanding link, then a `/webchat` whose reply is lost *has* destroyed something the sender held — structurally the same committed-mutation-then-lost-reply window the design doc devotes a section to for rotation, just with a cheaper asset.

The decision to skip the retry is probably right (a session link isn't a credential, and the command is trivially re-sendable). But the reason as written is factually wrong per this PR's own code, so a future reader reconciling the two comments can't tell which describes the frontend's actual semantics. Rewording to "the asset is not a credential and the command is trivially retryable" says the same thing and stays true.

chent1996: Fixed in b158711c. The comment now states the actual trade-off: no configured credential is invalidated, the command is safely retryable, and a high-signal audit line records a total delivery failure.

src/gateway/channels/lark.test.ts
chent1996: Fixed in b158711c. The group fixtures now carry a bot mention and botOpenId, and assert that binding resolution is never reached. I applied the same mutation-resistant coverage to the inherited /apikey group test as well.

src/gateway/channels/lark.ts
chent1996: Fixed in b158711c. Both webchat and API-key result.error values now pass through the shared prose cap, with regression tests using oversized upstream errors.

src/gateway/channels/lark.ts
chent1996: Kept the exact token boundary intentionally and documented the trade-off in the new webchat design contract and Change Impact Matrix row. Repository grep cannot discover downstream or user-provided slash commands; unlike /apikey, no link can be minted unless the exact parser accepts the command. The regression test now names /webchats explicitly so this remains a deliberate compatibility choice.

src/gateway/channels/lark.ts
chent1996: Fixed in b158711c. The copy tables remain command-specific, while copy selection, text rendering, and denial-card construction now share one implementation. The new webchat denial tests exercise every important branch.

src/gateway/channels/lark.ts
chent1996: Fixed both issues in b158711c. Success URLs now pass the same absolute HTTP(S) gate before reaching a card or fallback, and an expired success uses a command-specific reply that withholds the URL and tells the sender to rerun /webchat. I applied the same URL and expiry hardening to the /apikey sibling.

src/gateway/channels/lark.test.ts
chent1996: Fixed in b158711c. Added webchat denial tests for binding_required text fallback, access_request_required card delivery, unknown future reasons, access_denied without self-service, and expired links. These now protect the command-specific /webchat resume copy.

src/gateway/channels/lark.ts
```diff
@@ -1088,23 +1112,34 @@ export async function handleLarkMessage(
           // idempotent across a redelivery or a second replica — the in-process guard above
           // cannot span either.
           const issued = await issuePersonalApiKey(personalChannelId, senderOpenId, frontendClient, messageId);
-          reply = formatApiKeyIssueReply(issued, locale);
-          carriesCommittedRotation = Boolean(issued.success && issued.pickupUrl);
+          const pickupUrl = isRenderableActionUrl(issued.pickupUrl) ? issued.pickupUrl : null;
+          const invalidSuccess = issued.success && !pickupUrl;
```
jacoblee-io, line 1131: **A committed rotation now reports as "service temporarily unavailable", so the user never learns their old key is dead.**

```ts
const invalidSuccess = issued.success && !pickupUrl;
reply = invalidSuccess
  ? API_KEY_UNAVAILABLE_NOTICE_BY_LOCALE[locale]
  : formatApiKeyIssueReply(issued, locale);
carriesCommittedRotation = Boolean(issued.success);   // ← true: the rotation DID commit
```

The URL gate itself is right — rejecting `/siclaw/api-key/pickup/token`, `lark://…` and `javascript:…` is exactly what I asked for. The problem is the reply chosen for that path.

`API_KEY_UNAVAILABLE_NOTICE_BY_LOCALE` is "❌ 暂时无法处理 API Key 请求，请稍后重试。" — **the identical string** used at ~1085 for the `!frontendClient` case, where no RPC was issued and nothing was rotated. So one string now means both "nothing happened, retry" and "your credential was just destroyed". The user's MCP client starts 401-ing while the only message they received says the service was busy.

And it doesn't self-correct: every retry rotates again and returns the same notice, so a control plane emitting relative pickup URLs puts users in a loop that repeatedly kills their key while telling them nothing happened.

Root `CLAUDE.md`, `/apikey` row: *"`rotated:true` MUST warn the old key died"*.

The new test at `lark.test.ts:4102` asserts `toContain("暂时无法处理 API Key 请求")` for a `{success:true, rotated:true}` fixture — so it pins the degraded copy in place rather than catching it.

Fix: give `invalidSuccess` its own string that states the old key is invalid when `issued.rotated`, distinct from the no-RPC notice.

src/gateway/channels/lark.ts
```diff
@@ -1966,13 +2148,15 @@ async function deliverSingleUseLink(
...
   textFallback: string,
+  expiredReply?: string,
 ): Promise<boolean> {
   const remaining = minutesUntil(card.expiresAtMs);
   // Re-checked HERE, not just when the reply was composed: a link can lapse between the two, and
   // the render decision earlier says nothing about the state at the send boundary. An expired link
   // must reach neither the button nor the text — hand over the resend instruction instead, which is
   // the only thing that actually helps (the frontend mints a fresh link on the next message).
   if (remaining === "expired") {
+    if (expiredReply) return replyToLark(larkClient, messageId, expiredReply);
```
jacoblee-io, line 2181: **Same invariant, second path: the expired short-circuit returns before the body that carries the rotation warning.**

```ts
if (remaining === "expired") {
  if (expiredReply) return replyToLark(larkClient, messageId, expiredReply);
  const urlFreeBody = card.body.includes(card.url) ? … : card.body;
  return replyToLark(larkClient, messageId, `${urlFreeBody}\n${PERSONAL_DENIAL_LINK_EXPIRED_BY_LOCALE[locale]}`);
}
```

Withholding an expired URL is correct and was my request. But on `origin/main` there was no `expiredReply`, so an expired `/apikey` success fell into the `urlFreeBody` branch and still sent `card.body` — which is `formatApiKeyIssueCardBody(issued, locale)`, whose own comment says *"`rotated` still MUST be stated"*. Now the early return replaces the whole message with `API_KEY_LINK_EXPIRED_NOTICE` ("❌ API Key 领取链接已过期，请重新发送 /apikey 获取新链接。"), which mentions only a *link*.

Trigger is the ordinary one this boundary re-check exists for: frontend/pod clock skew, or delivery latency, with `{success:true, rotated:true, expiresAt: now-1}`.

It also silences the backstop: this returns `replyToLark(...)`'s result, so `cardDelivered` is `true` and the `carriesCommittedRotation` escalation below never fires.

Fix: when `issued.rotated`, prepend the rotation warning to `expiredReply` rather than replacing the body with it. (`/webchat` needs none of this — nothing is destroyed there, so the plain expired notice is right for that command.)

src/gateway/channel-manager.ts
```diff
@@ -390,6 +390,55 @@ export async function getPersonalApiKeyStatus(
   });
 }
 
+/**
+ * Result of `/webchat` — a one-shot link to a browser chat window for the personal bot's agent.
+ *
+ * `actionUrl` is short-lived and single-use, and unlike the API-key pickup it yields no
+ * credential at all: redeeming it creates a chat session owned by whoever is signed in, and the
+ * resulting page URL is authorized per request rather than by bearing a secret.
```
jacoblee-io: **Three places in this PR still describe the minted link three different ways, and the difference is exactly what decides the security blast radius.**

- here (`:396-398`): *"unlike the API-key pickup it **yields no credential at all**: redeeming it creates a chat session **owned by whoever is signed in**, and the resulting page URL is **authorized per request rather than by bearing a secret**"* → **not** a bearer token; the redeemer must already be authenticated
- `lark.ts:1351-1352`: *"a single-use link that **signs the redeemer into a chat session** … **Whoever opened it first would own the session**"* → **is** bearer authority
- `adapter.ts:3228-3229`: *"redeeming the link **signs the visitor in** and creates a chat session **owned by their account**"*

Why it matters concretely: `/webchat` is dispatched before `resolvePersonalBinding`, so the Runtime doesn't apply the binding gate — a sender who would be refused for ordinary chat still causes a mint, and the frontend is the only thing standing between them and a link. If the first description is right, a wrongly-approved mint hands out something of little value. If the second is right, it hands a **session-granting bearer link to someone who never passed the binding gate** — strictly worse than `/apikey`, where the pickup at least yields a credential scoped to the requester's own account.

The new design doc is the natural home for this and gets closest with *"Opening its preview does not consume it; the upstream frontend owns the confirmation and consumption semantics"* — but it also says *"the first person to open it could claim the resulting session"*, and these two comments keep asserting the semantics in opposite directions.

Ask: state it **once**, in the doc, as a contract the frontend must satisfy ("redemption requires an authenticated session and binds to that account" or "the URL is bearer authority until confirmed"), and reduce all three comments to a pointer. If the honest answer is "the Runtime doesn't know", then this comment shouldn't claim "authorized per request rather than by bearing a secret" — that's a statement about the frontend it can't make.

Worth confirming against the paired frontend MR while you're there: does `channel.issueWebChatLink` apply the same admission check as `channel.resolvePersonalBinding`? If it doesn't, `/webchat` is a way around the binding gate — and the same question applies to `channel.issueApiKey`.

jacoblee-io (changes requested):
## Review round 2 (`b158711c`)

All seven findings from round 1 are fixed, and the response went well past what I asked for: the URL gate, expired-link handling and prose cap were applied to `/apikey` as well, the denial helpers were unified without merging the copy tables, a design doc was added, and both Matrix rows (new `/webchat` **and** an updated `/apikey`) landed. `npx tsc --noEmit` is clean and 413 tests pass.

I re-ran the two mutations that survived last time, and both are now caught — plus the two new shared helpers:

| mutation | round 1 | now |
|---|---|---|
| delete the group-drop block | 7 passed (undetected) | **1 failed** |
| `actionDenialCopy` → undefined | n/a | **10 failed** |
| `formatActionDenialReply` → undefined | n/a | **7 failed** |
| `buildActionDenialCard` → null | n/a | **3 failed** |

The helper extraction itself I checked against `origin/main` and it is behaviour-preserving: `actionDenialCopy` is the old `apiKeyDenialCopy` with the tables injected, `formatActionDenialReply` keeps the nested (not chained) `rendersActionLink` → expired → hint/URL structure so the `access_denied`-with-live-link dead end stays fixed, and `buildActionDenialCard`'s `[lead, resume].filter(Boolean).join("\n")` is equivalent to the old ternary for every input. `isRenderableActionUrl(url: unknown): url is string` narrows correctly at all three call sites, `minutesUntil` is sound for `undefined`/`0`/negative/out-of-range, and `cardDelivered ?? await replyToLark(...)` correctly does not let `false` fall through.

**But the hardening introduced a regression in `/apikey`, which is shipped production code this PR now modifies.** Both new reject paths drop the rotation warning, so a user whose key was just destroyed is told something that reads as "nothing happened". Root `CLAUDE.md` states the invariant verbatim: *"`rotated:true` MUST warn the old key died"*. Two inline comments below; they're the same bug in two places and the fix is small — carry the rotation warning into both reject replies when `issued.rotated`.

What makes it worth blocking rather than noting: the new tests **assert** the degraded copy, so this locks in as correct-by-test.

### Still unresolved from our side discussion — the security question

`/webchat` sits before `resolvePersonalBinding`, so the Runtime does not apply the binding gate; admission is entirely the frontend's. That is the same posture as `/apikey` and the design doc now states it explicitly (*"The Runtime … does not decide whether the sender may use the agent"*), which is the right call — I'm not asking to change it.

What I do want pinned down is **what the minted link actually confers**, because it decides the blast radius when the frontend gets an admission decision wrong. Three places in this PR still say three different things (fourth inline comment). The doc punts to *"the upstream frontend owns the confirmation and consumption semantics"* — which is honest, but two of the three code comments assert the semantics anyway, in opposite directions. Please make the doc state it once and have the comments defer to it.

### Smaller

**`/apikey status` is the one place the new prose cap wasn't applied.** `formatApiKeyStatusReply` (`lark.ts:2641`, unchanged so I can't anchor it) still does:

```ts
const reason = result.error ?? (locale === "en-US" ? "unknown error" : "未知错误");
```

Same untrusted field, same command family, same failure the cap was added for: a frontend returning a 2 KB stack trace or an HTML error body on `channel.apiKeyStatus` produces a reply Feishu drops, and the sender gets silence. `truncateDenialProse` also covers what the bare `??` doesn't — a non-string `error` currently interpolates as `[object Object]`. One line, to match the two paths you already fixed.

jacoblee-io (approved):
## Re-review (`b4e7b141`) — approving, dismissing my changes-request

All four round-2 items are fixed, and the two blocking ones are fixed at the right depth.

**The rotation-warning regression.** `invalidSuccess` no longer borrows the "service unavailable" string — it gets its own `API_KEY_UNUSABLE_LINK_NOTICE_BY_LOCALE`, and both reject paths compose through `withRotationWarning(notice, issued.rotated, locale)`. The expired reply is now built at the **call site** rather than inside `deliverSingleUseLink`, with the reason stated where the next reader will need it: *"only this call site knows a rotation committed behind it."* That's the correct seam — the shared deliverer stays rotation-unaware instead of growing a special case, which is what I'd have gotten wrong if I'd specified the fix myself.

I mutation-tested it in **both** directions, because the mirror-image error is as bad as the original:

| mutation | result |
|---|---|
| `withRotationWarning` never warns | **2 failed** |
| `withRotationWarning` always warns | **2 failed** |

So a requester with no previous key can't be told one died, either — and your `2026-07-28` doc update calls that case out explicitly (*"The mirror-image error is equally real… both cases are worth a test"*).

**`/apikey status` prose cap** — `truncateDenialProse(result.error)` applied, with the `[object Object]` type-check rationale in the comment. The doc now states the cap applies to *every* path carrying that field and notes which one went a release without it.

**The semantic contradiction.** This is the answer I was hoping for. Rather than picking whichever description was most convenient, the doc now has a `### What the link confers — the Runtime's single assumption` section that:

- states one answer — **the Runtime treats `actionUrl` as bearer authority** — and says plainly that it is an assumption, not knowledge;
- explains why that's the safe direction to be wrong in (right under a true bearer link, merely conservative under an authenticated one);
- connects it to the thing that actually made me ask: `/webchat` is dispatched before `resolvePersonalBinding`, so the frontend is the only admission control, and a wrongly-approved mint hands a session-granting link to someone who never passed that gate;
- leaves redemption semantics to the frontend without letting that become an excuse not to decide the Runtime's own posture.

All three code comments now defer to it instead of restating it, with a note saying why. That's the right shape for a cross-repo contract the Runtime can't verify.

Verified on this head: `npx tsc --noEmit` clean, 417 tests pass across the three suites, CI green. `withRotationWarning` is called only from the two `/apikey` reject paths, and the `/webchat` diff is comment-only — so the blast radius of this round is exactly the credential command it was meant to fix.

Nothing further from me. The remaining open item is the cross-repo one I flagged last round and can't check from here: confirm in the paired frontend MR that `channel.issueWebChatLink` applies the same admission check as `channel.resolvePersonalBinding` — under the bearer assumption you've now written down, that check is the whole gate. The same question applies to `channel.issueApiKey`.

PR #478, opened 2026-08-06
fix(gateway): preserve HTTPS stream query parameters

LikiosSedo:
## Summary

Preserve SSE query parameters in the mTLS AgentBox client so Runtime reattachment can actually request knowledge-workspace replay. Add a regression test that exercises the real HTTPS and mutual-TLS transport path.

### Problem

`streamPathHttps` sent only the URL pathname to `https.request`. Query parameters such as `?replay=1` were therefore omitted, so a reattached knowledge compile run could stream normally without replaying workspace artifacts that may have been dequeued before the previous Runtime persisted them.

### Solution

Send `pathname + search` for HTTPS SSE requests, matching the existing non-stream HTTPS request behavior. The regression test starts a mutual-TLS HTTPS server and asserts that it receives the exact replay URL.

### Rollout Note

Workspace replay is currently one full-workspace event. Collection caps each file at 1 MiB but has no aggregate workspace cap, and the Runtime-to-consumer RPC uses a 30-second default timeout. Measure replay with a representative large workspace before deployment; protocol batching would change the existing one-event/one-transaction atomicity and is intentionally out of scope here.

## Test Plan

- [x] `npm run build`
- [x] `npm test` — 243 files passed, 5018 tests passed, 2 skipped
- [x] HTTPS SSE regression test verifies `/events/r1?replay=1`
- [ ] Pre-deployment large-workspace replay/load validation

## Architecture Checklist

- [x] **Deployment mode**: LocalSpawner and K8sSpawner share this HTTPS/mTLS transport; the regression test covers that common path.
- [x] **Security model**: No execution or trust-boundary changes.
- [x] **DB parity**: No schema or persistence implementation changes.
- [x] **Tool protocol**: No tool changes.

## General Checklist

### DIFF
--- a/src/gateway/agentbox/client.ts
+++ b/src/gateway/agentbox/client.ts
@@ -476,17 +476,26 @@ export class AgentBoxClient {
     }
 
     console.log(`[agentbox-client] SSE open (HTTPS) path=${path}`);
+    const decoder = new TextDecoder();
     let buffer = "";
+    // The replay frame can be one large data line. Remember the prefix already
+    // checked for a newline so each chunk scans only newly-decoded text instead
+    // of repeatedly splitting the entire accumulated frame.
+    let newlineSearchFrom = 0;
     let eventCount = 0;
 
     try {
       for await (const chunk of res) {
-        buffer += chunk.toString();
+        // TLS chunks are arbitrary byte boundaries. A stateful decoder preserves
+        // multi-byte UTF-8 characters that straddle two chunks.
+        buffer += decoder.decode(chunk, { stream: true });
 
-        const lines = buffer.split("\n");
-        buffer = lines.pop() || "";
+        let newlineIndex = buffer.indexOf("\n", newlineSearchFrom);
+        while (newlineIndex !== -1) {
+          const line = buffer.slice(0, newlineIndex);
+          buffer = buffer.slice(newlineIndex + 1);
+          newlineSearchFrom = 0;
 
-        for (const line of lines) {
           if (line.startsWith("data:")) {
             // SSE spec: the value is everything after the colon, minus ONE
             // optional leading space — "data:x" is as valid as "data: x".
@@ -503,8 +512,14 @@ export class AgentBoxClient {
             // be reaped as stale); it is never yielded as an event.
             opts?.onComment?.();
           }
+
+          newlineIndex = buffer.indexOf("\n", newlineSearchFrom);
         }
+        newlineSearchFrom = buffer.length;
       }
+      // Flush decoder state at EOF. SSE dispatch requires a terminating newline,
+      // so any remaining unterminated text stays buffered as before.
+      buffer += decoder.decode();
     } catch (err) {
       console.error(`[agentbox-client] SSE stream error path=${path}:`, err instanceof Error ? err.message : err);
       throw err;
--- a/src/gateway/agentbox/client.ts
+++ b/src/gateway/agentbox/client.ts
@@ -460,7 +460,7 @@ export class AgentBoxClient {
         {
           hostname: urlObj.hostname,
           port: urlObj.port,
-          path: urlObj.pathname,
+          path: urlObj.pathname + urlObj.search,
           method: "GET",
           headers: { Accept: "text/event-stream" },
           agent: this.httpsAgent!,
--- a/src/gateway/agentbox/client.test.ts
+++ b/src/gateway/agentbox/client.test.ts
@@ -1,12 +1,14 @@
 import { describe, it, expect, beforeAll, afterAll, beforeEach, afterEach, vi } from "vitest";
 import http from "node:http";
+import https from "node:https";
 import type { AddressInfo } from "node:net";
+import { CertificateManager } from "../security/cert-manager.js";
 import { AgentBoxClient } from "./client.js";
 
 /**
  * Tests for AgentBoxClient — Gateway's HTTP client for reaching an AgentBox.
- * We spin a tiny http.Server to act as the AgentBox. mTLS is orthogonal and
- * invariant §3 (K8s-only), so we exercise plain HTTP here.
+ * Most cases use a tiny plain-HTTP server; HTTPS transport behavior is covered
+ * separately with an mTLS server so the production request path is exercised.
  */
 
 // ── Test HTTP server ──────────────────────────────────────────────────
@@ -398,6 +400,43 @@ describe("AgentBoxClient — error paths", () => {
 });
 
 describe("AgentBoxClient — streamEvents (SSE)", () => {
+  it("preserves query parameters over HTTPS with mTLS", async () => {
+    const certManager = await CertificateManager.create();
+    const serverCert = certManager.issueAgentBoxCertificate("agentbox-test", "org-test", "box-test");
+    const clientCert = certManager.issueServerCertificate("runtime.test");
+    const urls: string[] = [];
+    const server = https.createServer({
+      cert: serverCert.cert,
+      key: serverCert.key,
+      ca: serverCert.ca,
+      requestCert: true,
+      rejectUnauthorized: true,
+    }, (req, res) => {
+      urls.push(req.url ?? "");
+      res.writeHead(200, { "Content-Type": "text/event-stream" });
+      res.end(`data: {"ok":true}\n\n`);
+    });
+    await new Promise<void>((resolve) => server.listen(0, "127.0.0.1", resolve));
+    const port = (server.address() as AddressInfo).port;
+
+    try {
+      const client = new AgentBoxClient(`https://127.0.0.1:${port}`, 30_000, {
+        cert: clientCert.cert,
+        key: clientCert.key,
+        ca: serverCert.ca,
+      });
+      const events: unknown[] = [];
+      for await (const event of client.streamPath("/events/r1?replay=1")) {
+        events.push(event);
+      }
+
+      expect(events).toEqual([{ ok: true }]);
+      expect(urls).toEqual(["/events/r1?replay=1"]);
+    } finally {
+      await new Promise<void>((resolve) => server.close(() => resolve()));
+    }
+  }, 60_000);
+
   it("yields parsed JSON events from data: lines", async () => {
     const srv = await startServer((req, res) => {
       if (req.url?.startsWith("/api/stream/")) {
--- a/src/gateway/agentbox/client.test.ts
+++ b/src/gateway/agentbox/client.test.ts
@@ -437,6 +437,52 @@ describe("AgentBoxClient — streamEvents (SSE)", () => {
     }
   }, 60_000);
 
+  it("preserves UTF-8 data split across HTTPS chunks", async () => {
+    const certManager = await CertificateManager.create();
+    const serverCert = certManager.issueAgentBoxCertificate("agentbox-test", "org-test", "box-test");
+    const clientCert = certManager.issueServerCertificate("runtime.test");
+    const expected = {
+      type: "syncArtifacts",
+      artifacts: [{ path: "候选/页面.md", content: "集群运维知识库：节点排障与网络诊断" }],
+    };
+    const frame = Buffer.from(`data: ${JSON.stringify(expected)}\n\n`);
+    const marker = frame.indexOf(Buffer.from("候选"));
+    expect(marker).toBeGreaterThanOrEqual(0);
+    // Split one byte into the first three-byte CJK character. Decoding either
+    // chunk independently produces replacement characters while valid JSON
+    // still parses, which would silently persist corrupted workspace data.
+    const splitAt = marker + 1;
+    const server = https.createServer({
+      cert: serverCert.cert,
+      key: serverCert.key,
+      ca: serverCert.ca,
+      requestCert: true,
+      rejectUnauthorized: true,
+    }, (_req, res) => {
+      res.writeHead(200, { "Content-Type": "text/event-stream" });
+      res.write(frame.subarray(0, splitAt));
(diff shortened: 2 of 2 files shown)


src/gateway/agentbox/client.ts
```diff
@@ -460,7 +460,7 @@ export class AgentBoxClient {
         {
           hostname: urlObj.hostname,
           port: urlObj.port,
-          path: urlObj.pathname,
+          path: urlObj.pathname + urlObj.search,
```
jacoblee-io, line 463: **This line is correct — but it switches on a path whose reader silently corrupts non-ASCII text, and the frame it switches on is close to the worst case for that bug.**

`session-driver.ts:79` is the only caller that passes a query, so `?replay=1` is exactly and solely what this enables. That request returns one SSE frame carrying the entire workspace (`compile_box.py:6979-6989` — a single `data:` write with every artifact inline, serialized with `ensure_ascii=False`, so CJK stays as raw multi-byte UTF-8).

Twenty lines below, the mTLS reader decodes each TLS chunk independently:

```ts
buffer += chunk.toString();            // client.ts:484
```

while the fetch/HTTP branch this PR is bringing to parity with uses a *stateful* decoder:

```ts
const decoder = new TextDecoder();
buffer += decoder.decode(value, { stream: true });   // client.ts:409, :413
```

Any multi-byte character straddling a chunk boundary becomes U+FFFD on both sides of the cut. I ran both strategies over the same bytes, splitting a small CJK payload at every width from 8 to 40 bytes:

```
原始: {"page":"集群运维知识库","body":"节点排障与网络诊断"}

chunk=8B   toString()      -> {"page":"���群运维知��库","body":"��点排���与网络诊��"}
           TextDecoder     -> {"page":"集群运维知识库","body":"节点排障与网络诊断"}

29 of 33 splits differ — and all 29 still JSON.parse successfully.
```

That last part is what makes it serious. The replacement characters land *inside* JSON string values, so `JSON.parse` succeeds, no error is logged, and `session-driver` base64s the corrupted text straight into `capability.persistArtifacts` — overwriting good rows in the consumer's durable store with mojibake. A replay exists to *repair* state that a previous Runtime failed to persist; this makes it a corruption vector instead.

Why it hasn't bitten yet: with the query dropped, the mTLS path only ever carried small single-chunk `log`/`turn_done` frames that rarely cross a boundary. A multi-MB replay frame arrives as hundreds of chunks, so it crosses boundaries constantly. And it fires on a *few-MB* Chinese workspace — well below any size the "measure with a representative large workspace" step in your Rollout Note would flag.

The fix is two lines, and makes this reader match the one at `:409` it is otherwise being aligned with:

```ts
const decoder = new TextDecoder();
...
buffer += decoder.decode(chunk, { stream: true });
```

Worth a test too — the existing HTTPS test asserts the URL, which is the right assertion for the fix itself, but a second case that streams a CJK payload in deliberately small chunks would pin the decoder. Given this PR is what makes the path reachable, I'd rather see both land together than ship the enabler first.

jacoblee-io (changes requested):
## Review

One-line fix: `streamPathHttps` sent only `urlObj.pathname` to `https.request`, so query parameters were dropped on the mTLS SSE path. Now it sends `pathname + search`, matching `httpsRequest` at `client.ts:566`.

**The fix itself is correct, and the test is better than most.** I verified:

- `httpsRequest` (`:566`) already used `pathname + urlObj.search`, so the stream path really was the odd one out — the change makes the two consistent rather than inventing a convention.
- The non-TLS branch builds `` `${this.endpoint}${path}` `` and hands it to `fetch`, so the query has always survived there. No third variant.
- Mutation-verified the regression test: reverting the fix produces `expected [ '/events/r1' ] to deeply equal [ '/events/r1?replay=1' ]`. It stands up a real mTLS HTTPS server rather than mocking the transport, which is exactly the level this bug needed — a mock would have passed either way.
- `client.ts` / `client.test.ts` are untouched on `main` since your merge base, so the 27-commit lag is not a conflict risk.
- Every factual claim in your Rollout Note checks out (details below). Writing it was the right call.

**But I can't approve it, because of what it switches on.** `session-driver.ts:79` is the only caller that passes a query, so `?replay=1` is precisely and solely what this one-liner activates — and the mTLS reader has a latent UTF-8 bug that the replay frame is close to a worst case for. Detail inline.

### Also worth fixing while you're in here

**`client.ts:486` — the same reader is quadratic, and this PR gives it its first big input.**

```ts
buffer += chunk.toString();
const lines = buffer.split("\n");
buffer = lines.pop() || "";
```

The whole replay event is one `data:` line, so `buffer` grows to the full payload and every chunk re-splits and re-allocates the entire accumulated string. ~800 chunks over a 50 MB workspace is ~20 GB of string scanning on the runtime's single event loop — blocking every other tenant's SSE relay and WS heartbeats, not just this run. Past V8's ~512 MB max string length the `buffer +=` throws outright. Scanning from the last newline instead of re-splitting fixes both.

**`session-driver.ts:144` — the retry loop makes a permanently-failing payload invisible.** `persistArtifacts` retries in an unbounded `for(;;)` with backoff and calls `manager.touchHeartbeat(runId)` on every failure, so a payload that can *never* succeed (30 s RPC timeout, or a frame over the `ws` default 100 MiB `maxPayload`) becomes a hot loop the watchdog will not reap while the run reports itself healthy. Worth a bounded attempt count or a terminal failure after N tries, now that replay actually reaches it.

**`compile_box.py:872` — replay is not authoritative, despite being documented as "re-send every current file".** `_collect_workspace_artifacts` does `if f.stat().st_size > MAX_SYNC_FILE_BYTES: continue` — over-cap files are skipped with no log and no event. The consumer correctly keeps its stale row rather than tombstoning, so the operator sees a successful replay over a workspace that is quietly one page out of date. A count of skipped files in the replay event would make that visible.

**`src/agentbox/gateway-client.ts:182` — the identical asymmetry, in the counterpart client.** `delegateStream` uses `path: url.pathname` while `request()` at `:274` uses `url.pathname + url.search`. It is **not** a live bug — the URL is the literal `new URL("/api/internal/delegate", ...)` with no query — but it is the same shape you just fixed, and `+ url.search` on an always-empty search is free. Making it uniform means the next person adding a query param there doesn't rediscover this the hard way.

### Rollout Note — verified

All four claims hold: replay is one SSE frame (`compile_box.py:6979-6989`, single `data:` write with every artifact inline); the per-file cap is `MAX_SYNC_FILE_BYTES = 1 MiB` applied via `stat()` before reading — with the qualifier that it's env-overridable via `KBC_MAX_SYNC_FILE_BYTES`, so a deployment that raised it has no per-file bound either; there is no aggregate byte or file-count bound in `_collect_workspace_artifacts`; and the RPC is `capability.persistArtifacts` on `frontend-ws-client.ts:49`'s `DEFAULT_TIMEOUT_MS = 30_000`, which `session-driver.ts:147` never overrides.

Two mitigations your note doesn't claim, both real: replay is requested **only on re-attach to an already-live box** (`materialize.ts:232`, the `isBoxAlreadyLive` branch), not on every stream open — so this is a runtime-restart path, not a steady-state one. And server-side replay is not untested code, since the plain-HTTP branch has always sent the query; what is genuinely new is only the mTLS transport. Which is exactly where the decoder differs.

On sizing, I'd say the note is if anything pessimistic: a few hundred compiled pages plus authoring JSON is roughly 400 files × ~15 KB ≈ 6 MB, ~8 MB after base64 on the consumer hop — two orders of magnitude under the WS ceiling, and a 400-row upsert in one transaction well inside 30 s. The timeout and the missing aggregate cap are latent cliffs rather than near-term ones. The decoder is the near-term one, and it bites far below any size you'd flag as "large".

src/gateway/agentbox/client.ts
```diff
@@ -476,17 +476,26 @@ export class AgentBoxClient {
...
+        // multi-byte UTF-8 characters that straddle two chunks.
+        buffer += decoder.decode(chunk, { stream: true });
 
-        const lines = buffer.split("\n");
-        buffer = lines.pop() || "";
+        let newlineIndex = buffer.indexOf("\n", newlineSearchFrom);
+        while (newlineIndex !== -1) {
+          const line = buffer.slice(0, newlineIndex);
+          buffer = buffer.slice(newlineIndex + 1);
+          newlineSearchFrom = 0;
```
jacoblee-io, line 497: **This reset is correctness-critical and has no test — I removed it and all 36 tests stayed green.**

It only matters when a chunk ends mid-line and the next chunk carries a complete, shorter line. Running the loop both ways over `["data: {\"long\":\"…40 chars…\"}", "\ndata: {\"b\":2}\n"]`:

```
with reset (current):  [{"long":"…"}, {"b":2}]
without:               [{"long":"…"}]           ← second event silently lost
```

Mechanism: `newlineSearchFrom` is ≤ `newlineIndex` when the line is found, but after `buffer = buffer.slice(newlineIndex + 1)` the remaining buffer can be *shorter* than that retained offset. `indexOf("\n", tooLargeOffset)` then returns -1 permanently, the outer `newlineSearchFrom = buffer.length` locks it in, and every subsequent event on the stream is dropped with no error and no parse warning.

That's precisely the traffic shape replay produces — one large frame, then ordinary small events on the same connection.

The code is correct as written. I'm flagging it because this is new hand-rolled index arithmetic in the exact path the PR is hardening, and its load-bearing line is unprotected: any future "optimization" that drops or hoists it reintroduces silent event loss with a green suite. A two-chunk test shaped like the above pins it in a few lines.

src/gateway/agentbox/client.ts
```diff
@@ -503,8 +512,14 @@ export class AgentBoxClient {
...
             opts?.onComment?.();
           }
+
+          newlineIndex = buffer.indexOf("\n", newlineSearchFrom);
         }
+        newlineSearchFrom = buffer.length;
       }
+      // Flush decoder state at EOF. SSE dispatch requires a terminating newline,
+      // so any remaining unterminated text stays buffered as before.
+      buffer += decoder.decode();
```
jacoblee-io, line 522: This statement has no effect: `buffer` is never read after the `for await` loop — the `finally` below only logs the event count. Removing the line leaves 36/36 tests green.

The reasoning in the comment is right (SSE dispatch needs a terminating newline, so trailing unterminated text is correctly dropped), but the assignment itself just writes to a variable that goes out of scope. The comment carries the whole meaning; the `buffer +=` reads as if it feeds something downstream.

If the intent is to flush decoder state so a truncated final character can't leak into a *later* read, note there is no later read — the decoder is per-invocation. `decoder.decode()` on its own (discarding the result) would at least not imply an append that matters, but dropping the line and keeping the comment is cleaner.

jacoblee-io (approved):
## Re-review (`3d91cc2e`) — approving, dismissing my changes-request

The blocking finding is fixed, and fixed properly on both halves I raised.

**The decoder.** `chunk.toString()` → a module-level `TextDecoder` with `decode(chunk, { stream: true })`, matching the fetch branch at `:409` that this PR was aligning with all along. Mutation-verified — reverting it to `chunk.toString()` fails 2 tests including the exact assertion.

**The test that pins it is genuinely good.** It doesn't just stream some UTF-8 and hope for a boundary — it locates the CJK character, splits one byte into it (`marker + 1`), and forces the split across two TCP writes with a delay so the chunking is deterministic rather than incidental. The comment names the failure mode precisely ("valid JSON still parses, which would silently persist corrupted workspace data"). That's the level this bug needed; a mocked transport or a whole-frame write would have passed either way.

**The quadratic scan too**, which I'd only listed as a follow-up: `split("\n")` per chunk is replaced by `indexOf("\n", newlineSearchFrom)` with the scanned prefix carried forward, so the large single-line replay frame is scanned once end to end instead of re-split on every chunk.

I also checked for flakiness, since two mTLS servers per file with freshly-issued CAs is the kind of thing that goes green locally and red in CI: three consecutive runs, 36/36 each time. Clean.

Approving. Two small things below — neither is a defect in the current code, so they're yours to take or leave.

### The one line I'd add a test for

`newlineSearchFrom = 0` at `:497` is correctness-critical and **completely uncovered** — I deleted it and all 36 tests stayed green.

It matters when a chunk ends mid-line and the *next* chunk carries a complete short line. Running the loop both ways over `["data: {\"long\":\"xxx…40 chars…\"}", "\ndata: {\"b\":2}\n"]`:

```
with the reset (current code): [{"long":"xxx…"}, {"b":2}]
without it:                    [{"long":"xxx…"}]          ← second event silently lost
```

Once the retained offset exceeds the length of the post-slice buffer, `indexOf` returns -1 forever and every subsequent event on that stream is dropped with no error. That is exactly the shape a replay produces: one large frame, then ordinary small events on the same connection.

The code is right — I'm flagging it because it's *new hand-rolled index arithmetic* whose critical line has no test, in the very path this PR is hardening. A two-chunk case shaped like the above would pin it, and it's cheaper to write now than to rediscover from a report of missing events.

### Dead line

`buffer += decoder.decode();` at `:522` has no effect — `buffer` is never read after the loop (the `finally` only logs). Removing it leaves 36/36 green. The intent is clear and the comment is accurate about *why* unterminated text is dropped, but as written the statement itself does nothing; the comment alone would say it.

### Still open from last round (not in this commit's scope)

Unchanged, and all outside `client.ts`, so I'm not gating on them — but they're the reason the Rollout Note's "measure before deployment" step still matters: the unbounded `persistArtifacts` retry loop at `session-driver.ts:144` that keeps calling `touchHeartbeat`, the silent >1 MiB skip in `_collect_workspace_artifacts` (`compile_box.py:872`) that makes a "full" replay quietly partial, and the same `pathname`-without-`search` shape at `gateway-client.ts:182` (harmless today — literal path, no query — but free to make uniform).

PR #480, opened 2026-08-09
fix: three runtime failures found while testing a v0.3.4 rolling deploy

jacoblee-io:
## Summary

Three independent fixes found while reading production logs and then testing a rolling deploy of v0.3.4 in a test environment. The first is a lost-work bug; the other two are log noise that had a correctness cost of its own — and that noise is why the first one went unnoticed for so long.

### 1. A Runtime restart strands every in-flight turn

**Problem.** Observed during a rolling deploy: a conversation was mid-turn, the pod was replaced, and the chat sat at "still working" until the user reloaded. Reproducible by construction, not a race — it happens to every turn that is streaming when the process exits.

`close()` dropped the consumer WS without a word. Termination is `prompt_done` (`docs/design/2026-08-02-error-surfacing-contract.md`): a consumer renders an error but keeps the turn open until that arrives. No later event can ever arrive — the process that owned the turn is gone, and its replacement does not adopt turns it did not start. Every other exit path from a turn already emits a terminal; only shutdown did not.

**Solution.** Report each turn in `activeStreamAborts` as a retriable `STREAM_INTERRUPTED`, then `prompt_done`, while the WS is still up; abort the consumer afterwards so its own finalization runs (partial text persisted, running tool rows closed) instead of leaving a row at "running" for a reload to re-paint as live. The abort makes the consumer take its own terminal path, so a `shuttingDown` flag suppresses that second report — exactly one terminal per turn.

This is the graceful half only. A Runtime that dies without running `close()` (OOM, eviction, kill -9) still needs the consumer to fail in-flight turns on disconnect; that belongs on the consumer side and is not in this change.

### 2. The federation pull loop 404s on every tick

The pod list it reads is selected by the `app=agentbox` label, which every spawned box carries — including the KB compile boxes, whose Python HTTP contract has no `/api/internal/metrics-snapshot` route. One 404 error line every 30s per compile box, forever, and each 404 scored as a pull failure, so the federation's own health counter reported a fault that never cleared.

Judge pull-eligibility by box profile — already on `AgentBoxInfo` from the `boxType` label. Filtering at the consumer rather than the pod-list layer is deliberate: the same list feeds pool reconciliation, drain and the orphan GC, all of which need every box. Liveness also keeps reading the full list — a box can reach federation by the SIGTERM flush, which keys on cert identity rather than profile, and evicting it while it runs would deflate cross-pod gauge sums. ADR-014 gains the scope rule.

### 3. The cron reconcile re-arms every job every 15s

`addOrUpdate` cancelled and re-armed each timer and logged two lines per job per pass. On a runtime with a handful of cron jobs that was the bulk of the log.

Leave an armed job alone when `schedule`/`status` did not change. Three consequences of no longer re-arming constantly, each handled:

- The fire path reads the row back from the registry instead of the one captured when the timer was armed, so a changed prompt/agent/name still takes effect.
- `scheduleNext` replaces the job's timer instead of adding one. Both callers can reach a job that already has a timer and `timers` holds one handle per job, so setting without clearing dropped a live timer out of the map — where neither `cancel()` nor `stop()` could reach it, and it fired the job a second time. Reachable before this change too, just rarer.
- A job mid-fire has no timer, so a reconcile pass arms one. That is what gets a job whose fire never settles back on a schedule.

A bogus cron expression also no longer reports `Scheduled job` — `scheduleNext` armed nothing and already said so.

## Test Plan

- [x] `npm test` — 245 files / 5061 passing; `npx tsc --noEmit` clean
- [x] Rolling deploy of v0.3.4 exercised end to end in a test environment: gateway replaced, agentbox pool drained and rebuilt one box at a time (5-replica pool never dropped below 4 accepting), zero old-image pods left, no errors in the new gateway log
- [x] Fix 1 found by that deploy interrupting a live turn; the two regression tests reproduce it
- [x] Every regression test confirmed to fail without its fix, not just pass with it: shutdown terminal + no-double-report, orphan timer reachable by `stop()`, hung-fire re-arm

### DIFF
--- a/docs/design/decisions.md
+++ b/docs/design/decisions.md
@@ -527,6 +527,15 @@ loop. The agentbox `/api/internal/metrics-snapshot` pull and the SIGTERM
 `/api/internal/metrics-flush` push share one `MetricsFlushPayload { incarnation, prom }`.
 Per-message `ttft_ms` (still consumed by the chat timing badge) is kept.
 
+**Scope — which boxes are pulled**:
+Only boxes running the default `agent` profile serve `/api/internal/metrics-snapshot`.
+The pod list the loop reads is selected by the `app=agentbox` label, which every spawned
+box carries including the KB compile boxes (Python, a different HTTP contract), so
+pull-eligibility is judged by profile, not by the label. Liveness — what
+`retainInstances` grace-evicts against — deliberately stays the FULL box list: a box that
+reaches federation by some other route (the SIGTERM flush, keyed on cert identity rather
+than profile) must not be evicted while it is still running.
+
 **Consequences**:
 - ✅ One metrics path instead of two; no in-memory counter state on the gateway
 - ✅ External Grafana unaffected — every federated `siclaw_*` series is unchanged
--- a/src/agentbox-main.ts
+++ b/src/agentbox-main.ts
@@ -26,6 +26,38 @@ import { startCapacityMetrics } from "./shared/capacity-metrics.js";
 const config = loadConfig();
 const PORT = config.server.port;
 
+/**
+ * How long the gateway-dependent startup work gets before the box stops waiting on it.
+ *
+ * Everything before `listen()` — settings, resource bundles, the tool whitelist — is a
+ * round trip to the Runtime, and the moment they all happen at once is a rolling deploy:
+ * the Runtime is itself restarting while every box of every agent asks it for the same
+ * things. Whatever is not back by then, the box comes up without and picks up on the next
+ * push, because a box that answers no probe is judged crashed at 60s and replaced by
+ * another one that will queue behind the same slow gateway.
+ *
+ * Deliberately under that 60s so the box loses the sync, not its life.
+ */
+const STARTUP_DEADLINE_MS = 30_000;
+
+/** Run gateway-dependent startup work, giving up (loudly) rather than delaying `listen`. */
+async function withStartupDeadline<T>(label: string, work: () => Promise<T>): Promise<T | undefined> {
+  let timer: ReturnType<typeof setTimeout> | undefined;
+  try {
+    return await Promise.race([
+      work(),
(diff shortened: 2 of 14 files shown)


jacoblee-io:
## End-to-end verification in a test environment

Both interruption paths were staged against a real deployment (5-replica agent pool + 2-replica coordinator, consumer driving turns over A2A), not just unit tests.

### `runtime_restart` — a restart under a live turn

```
02:35:11.970  [sse-consumer] <session>: message_start  role=assistant   ← still streaming
02:35:12.956  [runtime] Shutting down...
```

Consumer-side task went `WORKING` → `FAILED`, carrying the reason text verbatim:

```
message: "Runtime restarted; this turn was interrupted"
```

Before this change the task stayed `WORKING` until the consumer's own 30-minute stream cap — which on screen is a turn that never ends.

### `box_rolled` — a box cut at the drain deadline

First attempt did NOT reproduce it, and that is worth recording: the roll **waited** for the in-flight turn.

```
02:48:18  Draining …-1 (image v0.3.4 != v0.3.4-turnterm)
02:52:58  Removing drained box …-1        ← 4m40s later, no deadline, turn completed normally
```

So a rolling deploy does not interrupt ordinary turns; only one that outlives the 5-minute drain deadline is cut. Reproduced by keeping a turn working past it:

```
03:00:30  Draining …-1
03:05:30  Removing drained box …-1 (deadline reached)
03:05:30  …-1 still held 1 session(s); reporting them interrupted
```

```
message: "The agentbox running this turn was replaced; the turn was interrupted"
```

### Not verified end to end

`activeSessions` on `runtime.register` is unit-tested only. Exercising it needs the WS to drop **without** the Runtime dying, i.e. restarting the consumer side, which was out of scope for this environment.

nightmeng (approved):

PR #482, opened 2026-08-10
fix: route delegated peers to their owning runtime

LikiosSedo:
## Root cause

Direct chat routes an agent to its configured Runtime, but coordinator delegation always called the source Runtime's local `AgentBoxManager`. When the coordinator and peer belonged to different Runtimes, the peer was created in the coordinator's network environment and received the wrong MCP/tool surface.

## What changed

- resolve and authorize peer placement before creating an AgentBox
- preserve the existing local fast path for same-Runtime delegation
- route cross-Runtime starts and aborts through the management-plane Runtime mesh
- relay the target Runtime's live chat events back to the coordinator
- fail closed when placement or delegated-session persistence cannot be proven
- declare and reassert `skipInitialPersistence` so the target Runtime cannot duplicate the delegated user row
- always derive the final remote answer from durable session history, so partial live-frame loss cannot return a truncated success
- treat the remote timeout as an event-silence watchdog, configurable with `SICLAW_REMOTE_DELEGATION_IDLE_TIMEOUT`
- prevent Stop received during session-lock contention or target cold-start from dispatching a remote turn afterward
- document the Runtime placement, persistence, relay, and rollout contracts

## User impact

A coordinator can delegate to a peer assigned to another Runtime without changing that peer's MCP bindings, cluster reachability, session history, live progress, or Stop behavior.

## Validation

- `npm test` — 245 test files passed; 5085 tests passed, 2 skipped
- affected delegation/runtime regression tests passed after final hardening
- `npm run build`
- `git diff --check`
- companion management-plane `go test ./internal/siclaw/...`
- companion relay concurrency packages pass under `go test -race`

The management-plane routing support should be deployed before rolling out this Runtime change; older management planes fail closed instead of creating a peer in the wrong Runtime.


src/gateway/delegate-api.ts
```diff
@@ -260,7 +355,82 @@ export async function handleDelegate(
...
+      const unsubscribe = deps.frontendClient.subscribe("delegation.event", (data) => {
+        const envelope = data as DelegationRelayEnvelope;
+        if (envelope?.delegationId !== delegationId || envelope.sessionId !== peerSessionId || !envelope.event) return;
+        observePeerEvent(envelope.event, false);
+        const type = String((envelope.event as any)?.type ?? "");
+        if (type === "prompt_done" || type === "done") resolveRemoteDone?.();
+      });
+      try {
+        remoteStartRequested = true;
+        await deps.frontendClient.request("delegation.start", {
```
jacoblee-io: **P1 — nothing in this repo sets `skipInitialPersistence`, so the duplicate-persistence fix may not actually engage.**

This payload carries `sessionId`, `userId`, `text`, `agentId`, model fields, `origin` and `delegation` — but not `skipInitialPersistence`. The only consumer is `server.ts`, which requires `params.skipInitialPersistence === true`. That means the flag has to be injected by the management plane when it turns `delegation.start` into a `chat.send`.

If the management plane forwards `prompt` verbatim (which is the natural implementation, and what this payload shape suggests), the target Runtime will still run `ensureChatSession` + `appendMessage` + `incrementMessageCount` — overwriting the coordinator ownership/lineage this function just wrote and duplicating the delegated task row. That is exactly the failure the "avoid duplicate session/user persistence" bullet claims to remove.

Suggest setting it explicitly here:

```ts
prompt: {
  sessionId: peerSessionId,
  userId: ownerUserId,
  skipInitialPersistence: true,
  ...
}
```

It is self-documenting, idempotent, and does not depend on another repo's forwarding behaviour. `server-chat-system-prompt.test.ts` covers "flag present ⇒ skip", but by construction it cannot cover "who sets the flag".

src/gateway/delegate-api.ts
```diff
@@ -260,7 +355,82 @@ export async function handleDelegate(
...
+      const timeout = setTimeout(() => rejectRemoteDone?.(new Error("remote delegation timed out")), REMOTE_DELEGATION_TIMEOUT_MS);
+      timeout.unref?.();
+      const onAbort = () => rejectRemoteDone?.(new Error("delegation stopped"));
+      peerAbort.signal.addEventListener("abort", onAbort, { once: true });
+      const unsubscribe = deps.frontendClient.subscribe("delegation.event", (data) => {
+        const envelope = data as DelegationRelayEnvelope;
+        if (envelope?.delegationId !== delegationId || envelope.sessionId !== peerSessionId || !envelope.event) return;
+        observePeerEvent(envelope.event, false);
+        const type = String((envelope.event as any)?.type ?? "");
+        if (type === "prompt_done" || type === "done") resolveRemoteDone?.();
```
jacoblee-io: **P1 — a lost relay frame becomes a successful, empty delegation result.**

On the remote path `finalText`, `steps` and `artifact` are reconstructed purely from what arrives on this channel. If the assistant `message_end` is dropped (WS reconnect on either side, or a relay gap) but `prompt_done` still lands, `resolveRemoteDone()` fires and the function falls through to the final frame with `ok: true, status: "done", finalText: undefined`. The coordinator then relays an empty answer as a completed delegation.

The local path shares the shape but consumes in-process SSE, so the risk is not comparable.

The good news is that the target Runtime already persisted the assistant rows to the shared DB for this `peerSessionId`. Suggest: when the remote turn completes with no `finalText` and no `artifact`, read the assistant messages back from the DB instead of returning an empty success (or, failing that, report `status: "failed"` rather than `done`). That also makes the "session history preserved" claim in the PR description hold under a relay gap.

src/gateway/delegate-api.ts
```diff
@@ -39,6 +39,19 @@ import type {
  * back" in a long-running (never-switched) conversation.
  */
 const RECENT_DELEGATION_LIMIT = 8;
+const REMOTE_DELEGATION_TIMEOUT_MS = 10 * 60 * 1000;
```
jacoblee-io: **P2 — asymmetric timeout semantics between local and remote delegation.**

The local path has no wall-clock cap: it drains the peer's SSE for as long as the turn runs. The remote path gets a hard 10 minutes, after which the source rejects, aborts the target turn, and the coordinator sees `failed`.

So the same delegated investigation succeeds when the peer happens to live on the coordinator's Runtime and fails when it does not — which is the class of behavioural difference this PR is otherwise trying to eliminate. Deep diagnostic delegations exceeding 10 minutes are realistic.

At minimum make it env-tunable (the codebase already has `parsePositiveIntEnv` for exactly this in `src/core/subagent-registry.ts`), and consider a liveness-based deadline (reset on each relayed event) rather than a total-duration one — a hung relay is what you actually want to catch, not a long-running turn.

src/gateway/delegate-api.ts
```diff
@@ -260,7 +355,82 @@ export async function handleDelegate(
...
+            origin: "api",
+            delegation: {
+              delegationId,
+              parentSessionId: body.parentSessionId,
+              parentAgentId: coordinatorAgentId,
+              readOnly: false,
+            },
+          },
+        });
+        if (peerAbort.signal.aborted) {
```
jacoblee-io: **P2 — the remote path lacks the pre-dispatch abort bail the local path deliberately has, and the abort listener cannot fire for an already-aborted signal.**

Two interacting details:

1. `sessionTurnLocks.acquire(peerSessionId)` can block for a long time behind another in-flight turn on the same peer session. A Stop during that window aborts `peerAbort` *before* this block runs.
2. `peerAbort.signal.addEventListener("abort", onAbort, { once: true })` does **not** fire when the signal is already aborted, so the only thing covering that window is this post-`delegation.start` check.

The net effect is that a Stop that arrived before the lock was granted still causes a remote turn to be dispatched and then immediately aborted. Worse, `delegation.abort` can now race ahead of the target actually starting its stream; if the router treats an abort for a not-yet-started delegation as a no-op, the turn runs headless — precisely the failure mode the preserved comment on the local path (`Cancellation during cold spawn…`) was written to prevent.

Suggest mirroring the local path: right after acquiring the lock and before `delegation.start`, bail if `peerAbort.signal.aborted`.

src/gateway/delegate-api.ts
```diff
@@ -128,6 +141,26 @@ export async function handleDelegate(
...
   }
 
+  // Direct chat already routes agent_id → runtime_id in Sicore. Delegation must
+  // make the same placement decision before touching the local AgentBoxManager;
+  // otherwise a coordinator Runtime can create a correctly configured peer in
+  // the wrong network environment. Fail closed if the control plane cannot
+  // prove the route — a local fallback would recreate the incident.
+  let route: DelegationRoute;
+  try {
+    route = await deps.frontendClient.request("delegation.resolveRoute", {
```
jacoblee-io, line 383: **P2 — two notes on this gate.**

**Deploy coupling.** Every delegation now goes through this RPC, including single-Runtime deployments where cross-Runtime placement cannot occur. A management plane without `delegation.resolveRoute` therefore breaks delegation entirely (503), not just the cross-Runtime case. Failing closed is the right default, but it makes the rollout strictly ordered — worth pinning in a Change Impact Matrix row / release note rather than only in the PR description, so a future operator upgrading in the wrong order gets a documented cause instead of "delegation is down".

**Validation.** The guard checks `route.local` and `route.targetRuntimeId` but not `sourceRuntimeId`, even though the interface declares it required. Either validate it or drop it from the required set — a half-validated envelope invites a caller to trust a field that may be absent.

src/gateway/delegate-api.ts
```diff
@@ -260,7 +355,82 @@ export async function handleDelegate(
...
+      if (remoteErrorMessage) {
+        finished = true;
+        writeFrame({
+          type: "delegate_result",
+          result: { ok: false, peerAgentId, peerName: member.name, status: "failed", steps, peerSessionId, error: remoteErrorMessage } satisfies DelegateResponse,
+        });
+        res.end();
+        return;
+      }
+    } else {
```
jacoblee-io: **P3 — the `else` block is not re-indented, and `handleDelegate` is now well past a readable size.**

Everything from here to line ~496 sits one level short (the original 4-space body under a new `} else {`). It keeps this diff small, but it makes the file misleading to read and every future diff in this region noisy.

Since the two branches now share only `observePeerEvent` and the result-emitting tail, extracting `runLocalDelegation(...)` and `runRemoteDelegation(...)` would fix the indentation, cut a ~400-line function down, and give the remote path a place to hold its own timeout/abort/readback logic.

src/gateway/server.ts
```diff
@@ -410,6 +410,11 @@ export async function startRuntime(opts: StartRuntimeOptions): Promise<RuntimeSe
     // concierge) delegated this turn over the mesh. Forwarded to the agentbox so
     // the worker gates its toolset read-only and stamps the result artifact.
     const delegation = params.delegation as PromptOptions["delegation"];
+    // A cross-Runtime delegation session is created by the coordinator Runtime
+    // before Sicore routes this chat.send to the target Runtime. Re-inserting the
+    // session/user row here would overwrite ownership/lineage and duplicate the
+    // delegated task. Only honor the flag on an authenticated delegation turn.
+    const skipInitialPersistence = params.skipInitialPersistence === true && Boolean(delegation?.delegationId);
```
jacoblee-io, line 727: This is the only place `skipInitialPersistence` is read, and no code in this repo writes it (see the comment on `delegate-api.ts` — the `delegation.start` payload omits it). The gate itself is sound: requiring `delegation?.delegationId` keeps a bare `chat.send` from suppressing persistence.

Two asks:

1. Have the source set the flag explicitly so the behaviour is not contingent on how the management plane forwards `prompt`.
2. Document the field as part of the cross-Runtime delegation contract — a wire flag whose only producer lives in another repository is exactly the kind of thing that silently regresses.

Minor: on this path `promptMessageId` stays `undefined`, so anything downstream keyed on the user-message row id (trace binding, message updates) degrades. That state was already reachable via the persist-failure branch, so it is presumably tolerated — but here it becomes the *normal* path for every cross-Runtime delegation rather than an error case. Worth a sentence confirming that is intended.

jacoblee-io (commented):
Reviewed the routing change. Direction looks right — making the placement decision *before* touching the local `AgentBoxManager` and failing closed is the correct shape, and reusing the reverse WS lane avoids exposing private Gateway addresses.

For the record on scope: this is **not a regression**. `src/gateway/delegate-api.ts` has exactly one prior commit (`641b898a`, "feat(delegation): siclaw-native agent-to-agent delegation"), and that first version already did `deps.agentBoxManager.getOrCreate(peerAgentId)` unconditionally. There has never been any Runtime-placement logic on the delegation path, and `docs/design/coordinator-routing.md` never states the single-Runtime assumption. So every multi-Runtime deployment has had this since delegation shipped.

Two things I think should be fixed before merge (details inline):

1. **`skipInitialPersistence` has no producer in this repo.** `server.ts` consumes it, but the `delegation.start` payload built here does not set it. The de-duplication the PR describes only happens if the management plane injects the flag while forwarding to `chat.send`.
2. **The remote path can return a successful-but-empty result.** `finalText` / `steps` / `artifact` are rebuilt *only* from relayed events, so a dropped frame plus a delivered `prompt_done` yields `ok:true, status:"done", finalText: undefined`, which the coordinator relays as a real answer.

Also: this adds a new cross-Runtime contract (`delegation.resolveRoute` / `delegation.start` / `delegation.abort`, the reverse `delegation.event` lane, and `skipInitialPersistence`) with no documentation change. Per CLAUDE.md's documentation rule these are contracts, not implementation details — worth a section in `docs/design/coordinator-routing.md` plus Change Impact Matrix rows for `delegate-api.ts` and `frontend-ws-client.ts`, including the deploy-ordering requirement.

src/gateway/delegate-api.ts
```diff
@@ -252,15 +424,95 @@ export async function handleDelegate(
...
+      }
+      throw err;
+    } finally {
+      if (idleTimer) clearTimeout(idleTimer);
+      peerAbort.signal.removeEventListener("abort", onAbort);
+      unsubscribe();
+    }
+
+    if (remoteErrorMessage) return { error: remoteErrorMessage };
+    if (!finalText && !artifact && !inputQuestion) {
```
jacoblee-io: **The recovery net only covers total loss, but partial loss is the more likely shape.**

The reverse event lane is best-effort by construction: each subscriber has a bounded queue, and when a consumer falls behind the management plane drops **individual events** for that subscriber rather than blocking its read loop. That is the right trade on its side, but it means the failure mode is not "the relay went silent" — it is "three assistant frames arrived, one didn't".

This guard only fires when `finalText`, `artifact` and `inputQuestion` are *all* empty. A turn that loses one frame out of several keeps a non-empty `finalText`, skips the readback entirely, and returns a silently truncated result as `status: "done"`. That is harder to notice than the empty case this now catches: an incomplete report still reads like an answer, and the coordinator relays it as one.

Suggest making the durable history authoritative on the remote path rather than a fallback: use the relay for live `peer_event` frames and `steps`, and always derive the returned `finalText` from `recoverRemoteResult`. The target Runtime already persisted those rows and (per the ordering above) they are committed before `prompt_done` can arrive, so the readback is strictly better evidence than the reassembled stream. That also makes the result independent of how many frames the lane happened to deliver, instead of correct only at the two extremes.

If you would rather keep the reassembled text as the primary source, the guard at least needs to detect *incompleteness* rather than emptiness — but I don't see a cheap signal for that from the event stream alone, which is the argument for just trusting the rows.

jacoblee-io (commented):
Re-reviewed against `0c4570b1`. All six points are addressed, several better than what I suggested — the idle watchdog instead of a total-duration cap, and the `route.local !== (source === target)` consistency assertion, are both improvements on the original comments.

**Retracting my first P1.** I claimed nothing sets `skipInitialPersistence`. I had only read this repo; the management plane injects it at its trust boundary. The source now sets it explicitly too, which is the right call — the flag is asserted on both sides rather than depending on one.

I also checked the two risks the new readback could have introduced, and both are clear:

- `recoverRemoteResult`'s boundary scan is safe: `getMessages` reverses internally, so `limit: 500` is the *most recent* 500 rows, not the oldest.
- No read-too-early race: `prompt_done` is emitted only after `consumeAgentSse` resolves, and every persistence flush inside it is awaited (`sse-consumer.ts:1085`). The assistant rows are committed before the terminal event can reach the source.

One finding left, inline.

src/gateway/delegate-api.ts
```diff
@@ -252,15 +430,99 @@ export async function handleDelegate(
...
+          systemPrompt: binding.systemPrompt ?? undefined,
+          origin: "api",
+          delegation: {
+            delegationId,
+            parentSessionId: body.parentSessionId,
+            parentAgentId: coordinatorAgentId,
+            readOnly: false,
+          },
+        },
+      });
```
jacoblee-io, line 640: **P1 — Stop can still be acknowledged before the target Runtime has anything abortable.**

The new post-lock `peerAbort.signal.aborted` check closes the window while the source waits for `sessionTurnLocks`, but a second window remains after this `delegation.start` request is acknowledged. On the target Runtime, `chat.send` returns its async ack immediately, while the background task does not install `activeStreamAborts[sessionId]` until after `getOrCreate()` and `client.prompt()` succeed (`server.ts`, around lines 549–606). During a cold spawn, a Stop therefore reaches `chat.abort` with no controller registered and no session in the box yet; `abortSession()` treats “session not found” as already stopped. Sicore can then remove the relay even though the background `chat.send` task is still alive and will dispatch `prompt()` later, leaving a headless peer turn.

The post-request signal check here cannot close that race because the abort RPC has already returned success before the target becomes runnable. The target needs a pre-start cancellation/tombstone registered before the `chat.send` ack (and checked before `prompt()`), or the start/abort protocol must otherwise guarantee that a successful abort prevents a later dispatch. A regression test can hold `getOrCreate()`, abort after the start ack, then assert that `client.prompt()` is never called.

src/gateway/delegate-api.ts
```diff
@@ -226,15 +351,68 @@ export async function handleDelegate(
...
+  };
+
   // Propagate Stop: when the coordinator aborts, its delegateStream destroys the
   // HTTP request → this response's socket closes. Break the drain loop AND cancel
   // the peer's own turn (otherwise the peer keeps running headless).
   const peerAbort = new AbortController();
   let finished = false;
   let peerClient: AgentBoxClient | undefined;
+  let remoteStartRequested = false;
   res.on("close", () => {
```
jacoblee-io: **P1 — disconnects before the SSE phase are never observed.**

`peerAbort` and this `close` listener are created only here, after the handler has already awaited the roster lookup, model binding, route resolution, parent validation, and delegated-session persistence. If Stop destroys the delegate HTTP request during any of those waits, `close` has already fired and EventEmitter will not replay it when this listener is attached. The handler can then write to a dead response and still call the local `prompt()` or remote `delegation.start`, leaving a headless peer turn.

Please register the cancellation signal immediately after reading the request (and account for an already-aborted/destroyed request), then check it after the async validation/persistence boundaries and immediately before dispatch. A regression test can hold `delegation.resolveRoute`, close the response, release the RPC, and assert that neither `getOrCreate` nor `delegation.start` is called.

src/gateway/server.ts
```diff
@@ -1355,6 +1411,7 @@ export async function startRuntime(opts: StartRuntimeOptions): Promise<RuntimeSe
     // the consumer — so consumeAgentSse runs its abort-finalization (in-flight tool
     // rows → "stopped", partial assistant text persisted) instead of exiting as a
     // normal completion that leaves the tool row stuck "running" → "resumes on refresh".
+    for (const ctrl of pendingStartAborts.get(sessionId) ?? []) ctrl.abort();
```
jacoblee-io: **P1 — the new Gateway cancellation latch can leave the AgentBox pre-spawn abort latch armed for the next send.**

For a pending-only cold start, this aborts `turnAbort`, so the background `chat.send` exits before calling `/api/prompt`. The code then still calls `abortSession()` below. If the box is ready but has not created this session yet, AgentBox treats that as a pre-spawn Stop and records a three-minute pending abort (`http-server.ts`, lines 1339–1347). The cancelled send never reaches `/api/prompt` to consume it, so the next intentional send with the same session id consumes the stale latch and is silently returned as `aborted: true`.

The current regression test mocks `abortSession`, so it cannot observe this interaction. Please either avoid the tombstone-producing AgentBox abort when only a Gateway pending start exists (the post-`prompt()` branch already handles the in-flight handoff), or add a non-tombstoning abort/explicit latch cleanup. The regression should cold-start → Stop → send again on the same session and assert that the second prompt actually runs.

src/gateway/delegate-api.ts
```diff
@@ -252,15 +430,99 @@ export async function handleDelegate(
...
+
+    const onAbort = () => rejectRemoteDone?.(new Error("delegation stopped"));
+    peerAbort.signal.addEventListener("abort", onAbort, { once: true });
+    const unsubscribe = deps.frontendClient.subscribe("delegation.event", (data) => {
+      const envelope = data as DelegationRelayEnvelope;
+      if (envelope?.delegationId !== delegationId || envelope.sessionId !== peerSessionId || !envelope.event) return;
+      armIdleWatchdog();
+      observePeerEvent(envelope.event, false);
+      const type = String((envelope.event as any)?.type ?? "");
+      if (type === "prompt_done" || type === "done") resolveRemoteDone?.();
```
jacoblee-io: **P1 — an interrupted target turn can still be returned as a successful durable result.**

The target Runtime emits `prompt_done` with `aborted: true` and a `reason` when the Runtime restarts or its AgentBox is replaced, but this handler looks only at the event type. Because the reverse lane is explicitly best-effort, the preceding `stream_error` may be the frame that gets dropped while this terminal still arrives. In that case `remoteErrorMessage` stays empty, `recoverRemoteResult()` can recover partial assistant text, and the coordinator receives `status: "done"` for an interrupted turn.

Please make `event.aborted === true` authoritative here and preserve `reason` as the failure/interruption message before resolving `remoteDone`. A regression test should deliver only `{ type: "prompt_done", aborted: true, reason: "box_rolled" }` after a persisted partial assistant row and assert that the delegation fails rather than returning that partial row as success.

src/gateway/server.ts
```diff
@@ -1355,7 +1411,17 @@ export async function startRuntime(opts: StartRuntimeOptions): Promise<RuntimeSe
...
+    const active = activeStreamAborts.get(sessionId);
+    for (const ctrl of pending ?? []) ctrl.abort();
+    active?.abort();
+
+    // A pending-only turn has not established an AgentBox session. Its background
+    // chat.send checks the Gateway controller before prompt(), and checks again
+    // immediately after an in-flight prompt is accepted. Calling AgentBox abort
+    // here would create a pre-spawn tombstone that this cancelled send never
+    // consumes, causing the next intentional send on the same session to abort.
+    if (pending?.size && !active) return { ok: true };
```
jacoblee-io: **P1 — A lost `/api/prompt` ACK still makes Stop a no-op.**

AgentBox starts the prompt before returning its HTTP acknowledgement (`http-server.ts`: the run starts at line 1037 and the ACK is sent at line 1072). If the request reaches AgentBox and starts the turn but that ACK is lost or times out, this pending-only branch returns without calling `abortSession()`. `client.prompt()` then rejects, so the post-accept compensation path at lines 594–601 is never reached and the real AgentBox turn continues without a consumer.

Please distinguish “not dispatched” from “dispatch in flight / outcome unknown”, or make prompt and abort generation-aware with a request/turn ID so an uncertain start can be compensated without leaving a stale session tombstone. A regression test should let AgentBox start the prompt, drop the ACK, issue Stop, and assert that the real turn is aborted.

src/gateway/server.ts
```diff
@@ -1355,7 +1422,43 @@ export async function startRuntime(opts: StartRuntimeOptions): Promise<RuntimeSe
...
+      // A failed probe is not evidence of absence. Fall through to the abort attempt
+      // rather than assume nothing is running — a headless turn burns model quota
+      // until it finishes, while a tombstone at worst costs the next send a retry.
+      const holdsSession = await dispatched.listSessions()
+        .then(({ sessions }) => sessions.some((s) => s.id === sessionId))
+        .catch((err) => {
+          console.warn(`[runtime] abort: could not probe box for session=${sessionId}; aborting anyway:`, err);
+          return true;
+        });
+      if (!holdsSession) return { ok: true };
```
jacoblee-io: **P1 — The one-shot session probe still leaves the ambiguous-start window open.**

This only closes the case where `listSessions()` already sees the session. `/api/prompt` may already have reached AgentBox and be awaiting `sessionManager.getOrCreate()` while the session is not yet visible. Stop can therefore observe `holdsSession === false` and return here; the request can then create and start the session, lose its ACK, and never reach the post-accept compensation path, leaving the turn headless.

The new tests cover “already present” and “never present”, but not “absent at probe, then present”. Also, when no Stop is issued, a lost prompt ACK still follows the background error path without compensating the possibly started AgentBox turn. This needs a generation/attempt-scoped cancellation handshake rather than a one-shot presence snapshot.

jacoblee-io:
All 14 threads are resolved. Summary of where this landed, since the shape of the fix changed materially after the first round.

## The root cause behind most of the findings

Nearly every remaining issue traced to one thing: **an abort was addressed by `(agentId, sessionId)`**. A session id names a *conversation*, and delegation deliberately reuses a peer session for follow-ups in one investigation — so "abort session S" becomes ambiguous the moment a turn ends, and a supervisor's abort can be delayed past that point by a lease expiry or a retry.

That single ambiguity produced the `listSessions()` probe (guessing whether a dispatch had landed), a pre-spawn latch a later prompt could consume, a cold-start Stop skipped entirely to avoid arming it, and a control-plane abort landing on a successor turn.

Aborts now name the turn they mean:

- a prompt carries a `turnId`, the box records it as the running turn, and an abort naming any other turn is answered as already stopped;
- the pre-spawn latch records the turn that armed it and is consumed only by that turn's prompt, so an orphan cannot cancel a deliberate retry;
- `chat.abort` takes an optional `turnId` and ignores a stale one — the Stop button sends none and still means "stop what is running";
- the id comes back in the `chat.send` ack so a supervisor can name it later.

This **removed** machinery rather than adding it: the probe is gone, and so is the special case that skipped a cold-start Stop. Compensation on a failed prompt is now unconditional, because an abort naming a turn the box never started is a no-op it cannot confuse with a later one — which is what closes the lost-ack window in both the Stop and no-Stop cases.

Both fields are optional, so this half of the rollout is additive in either order.

## Reliability of the result path

- **Live relay text is never authoritative.** Final remote text always comes from durable assistant rows after the current delegation boundary, so an individually dropped frame cannot yield a truncated success. Readback widens a window over the newest rows instead of walking a timestamp cursor — `created_at` is second-granular, so a cursor skips rows sharing a second.
- **Both legs of terminal delivery are acknowledged.** The reverse leg already was; the target Runtime now also reports a delegated turn's terminal over an acknowledged RPC. Losing that first hop used to strand the source until its idle window elapsed, which it then reported as a *failure* for a turn that had finished.
- **A re-delivered terminal is acknowledged, not rejected.** Its consumer is gone precisely because it consumed the original; rejecting it kept the sender retrying, which kept supervision alive over a finished turn — and that supervision's expiry was the abort that could hit a successor.
- **Supervision ends when a terminal is observed**, whether or not it could be handed onwards.
- **An unconfirmed Stop is reported as a failure**, never as success: answering `ok` tells the control plane to stop retrying and tear down supervision.
- The idle window is event-silence based and clamped to the control plane's relay lease — waiting past it cannot succeed, since those events no longer have a route.

## Rollout

One strict ordering and one that is not:

- **Strict:** the control plane implementing route/start/abort and the reverse lane must be deployed *before* this Runtime. An older one makes every delegation route lookup fail closed, including same-Runtime delegation.
- **Not strict:** turn addressing spans the **AgentBox image** (`http-server.ts`, `session.ts`). Each half degrades to the older session-wide meaning on its own, so that half can roll either way — but it does mean building the agentbox image, repointing `SICLAW_AGENTBOX_IMAGE`, and recycling pods.

Cross-Runtime delegation also assumes a control plane whose Runtime connections all terminate in one process. Placement, command delivery and event subscription resolve against process-local connections, so a multi-replica control plane fails closed on a split pair rather than misrouting.

## Contracts are documented

`docs/design/coordinator-routing.md` now states the turn-addressing and two-leg delivery contracts, the readback rule, the idle-window bound, and both rollout facts. Change Impact Matrix rows were added/extended for `delegate-api.ts`, the reverse-delivery lane, and the agentbox turn protocol.

## One knowingly-deferred item

The `settledDelegations` memory is per-process, so a source Runtime restart forgets it. With turn-scoped aborts the consequence is now just a warning line — the restart already ended the delegate stream, and a late abort names a turn that is gone.

## Verification

`npm test` — 245 files, 5098 passed, 2 skipped. `npx tsc --noEmit` clean.

One test was replaced rather than dropped: `does not dispatch prompt after Stop is acknowledged during cold spawn` pinned the retired probe semantics (`abortSessionCalls === []`). `lets a retry run after a cold-start Stop, because the latch is turn-scoped` asserts the same no-dispatch and retry-runs properties, plus the turn tagging.

src/gateway/server.ts
```diff
@@ -476,14 +551,28 @@ export async function startRuntime(opts: StartRuntimeOptions): Promise<RuntimeSe
...
+      } catch (persistErr) {
+        console.warn(`[runtime] failed to persist the user message session=${sessionId}; continuing without a row id:`, persistErr);
+      }
     }
 
+    // Registered before the async ack below. This closes the cold-start window in
+    // which chat.abort previously had no local cancellation state to update.
+    const turnAbort = new AbortController();
+    registerPendingStart(sessionId, turnAbort);
+    currentTurnIds.set(sessionId, turnId);
```
jacoblee-io: [P1] Do not let a pending send replace the active turn ID

`currentTurnIds` is keyed only by session, and this assignment happens before `sessionTurnLocks.acquire()`. If turn A is active and a second send B arrives, B replaces A in this map while it waits up to five seconds to become a steer. During that window, a targeted abort for A is returned as stale; an unscoped Stop aborts A’s local consumer but sends B to AgentBox, which correctly rejects it because A is still running. The Runtime then returns `ok` while A continues headless. Please track pending and active turn IDs separately, make Stop select the actual active controller/turn, and add a double-send + Stop regression test.

src/gateway/server.ts
```diff
@@ -430,8 +466,47 @@ export async function startRuntime(opts: StartRuntimeOptions): Promise<RuntimeSe
...
+          }, 10_000);
+          return;
+        } catch (err) {
+          const message = err instanceof Error ? err.message : String(err);
+          if (/unknown method/i.test(message)) return;
+          if (attempt === 3) {
+            console.error(`[runtime] could not confirm delivery of the terminal for delegation=${delegation.delegationId} session=${sessionId}:`, message);
+            return;
+          }
+          await new Promise((r) => setTimeout(r, 250 * attempt));
```
jacoblee-io: [P2] Make terminal delivery survive the reconnect window

When the WS is disconnected, `request()` rejects immediately. These retries sleep only 250 ms + 500 ms, whereas `FrontendWsClient` reconnects no earlier than 1 s plus up to 2 s jitter, and later backs off to 30 s. Therefore all three attempts can finish before the first reconnect, preserving exactly the lost-terminal/false-timeout failure this helper is meant to close. Please use a reconnect-aware or outbox retry bounded by a delivery deadline, not a fixed three-attempt loop.

src/gateway/server.ts
```diff
@@ -642,7 +783,10 @@ export async function startRuntime(opts: StartRuntimeOptions): Promise<RuntimeSe
             });
           }
           // Shutdown, or a box removal, already reported this turn before aborting it.
-          if (!alreadyReported()) context.sendEvent("chat.event", { sessionId: promptResult.sessionId, event: { type: "prompt_done" } });
+          if (!alreadyReported()) {
+            context.sendEvent("chat.event", { sessionId: promptResult.sessionId, event: { type: "prompt_done" } });
+            await deliverDelegationTerminal({ type: "prompt_done" });
```
jacoblee-io: [P2] Report supervisor termination over the acknowledged path

`endTurns()` emits `stream_error` and `prompt_done` fire-and-forget and marks this session in `supervisorEndedTurns`. The consumer then reaches this branch with `alreadyReported()` true and skips `deliverDelegationTerminal()`; `close()` immediately closes `FrontendWsClient`. Runtime rollout and AgentBox replacement can therefore still lose the terminal on the first hop and leave the source waiting for the idle timeout. Please retain `delegationId` and `turnId` with the active controller and make `endTurns()` use the acknowledged delivery path before the WS is closed.

src/agentbox/session.ts
```diff
@@ -521,17 +537,23 @@ export class AgentBoxSessionManager {
       if (fs.existsSync(path.join(this.getBaseSessionDir(), sessionId))) return;
```
jacoblee-io: [P1] Preserve turn-scoped aborts for restored sessions

The on-disk history check is still a session-wide safety guard, but a scoped latch cannot cancel an unrelated successor. For a reused peer session that was released from memory, prompt B can be blocked in `getOrCreate()` while `abort(B)` sees no managed session; this return drops the latch, and B starts after the abort was acknowledged. The same occurs when the abort sees an idle managed session carrying the previous turn ID. Please keep this guard only for legacy unscoped aborts, and record a turn-scoped latch for B across restore/rebuild. Add an ack-lost prompt + restored-session regression test.

jacoblee-io:
Follow-up on the review of the turn-addressing work. Four P1s and three P2s, all addressed; head is `b468e5a9`.

## P1 — two turns can be live at once

`currentTurnIds` was set before the background task takes the session lock, so a second send on the same session immediately replaced the running turn's id. Stop then named the *queued* turn, the box correctly rejected the mismatch, the Runtime consumer was torn down anyway, and the running turn continued headless.

Turns are now tracked as a set per session, and a Stop that names none names all of them: the box stops the one it is running and answers the rest as already stopped. Falling back to a session-wide abort instead would re-arm a session-wide pre-spawn latch, which is what the turn scoping exists to prevent.

Writing the test for this surfaced a second bug in the same code: `chat.abort` held a *reference* to the live set, and breaking the consumer lets a turn settle and remove itself before the reference was read — so it aborted only the queued turn. The set is now snapshotted before the consumer is broken.

## P1 — a lost `chat.send` ack left compensation with nothing to name

Reading the turn id from the acknowledgement failed in exactly the case that needs it: no ack, so no id, so the compensating abort fell back to stopping the session and could hit whatever turn is running by then.

The control plane now mints the id **before** dispatch and passes it in the prompt; `chat.send` accepts a caller-supplied `turnId` and echoes it, minting one only when none is supplied. Every abort for a delegation names a turn, ack or no ack.

## P1 — a reused session never armed a turn-scoped latch

`markPendingAbort` skipped arming when the session already had history on disk. That guard exists only because a *session-wide* latch could cancel an unrelated later prompt; a turn-scoped latch cannot, since only that turn's own prompt consumes it. Since a delegated peer thread always has history, the guard was suppressing the latch in precisely the case it was needed. It is now skipped when the abort names a turn, and kept for the legacy session-wide call.

## P1 — a reported terminal could overtake queued control frames

`delegation.terminal` forwarded and retired the relay straight from the RPC, jumping ahead of artifact, error and clarification frames still queued on that delegation's subscription and dropping them.

A subscription can now be handed a payload for its own queue, so the reported terminal lands behind those frames and the ordinary handler interprets it in arrival order — including ending supervision, which is why nothing else happens in the RPC. The handler also validates the reported turn and accepts only a terminal event type: this path retires a relay, so an arbitrary frame must not be able to trigger it.

Verified the test detects the regression rather than merely passing: reverting to direct forwarding fails it.

## P2s

- **Retry budget.** 3 attempts over ~750 ms could not outlast a WS reconnect, which cannot complete faster than its own backoff. Now 5 attempts over ~15 s.
- **Supervisor terminals.** Shutdown and box removal end a turn on behalf of its own reporting path, which is then suppressed — so that terminal went out fire-and-forget on exactly the path where the transport is about to disappear. Delegated turns are tracked so the supervisor reports their terminals the same acknowledged way, carrying the interruption cause it already names. This also exposed an ordering bug: `close()` called `frontendClient.close()` immediately after ending turns, killing the connection those deliveries need. Shutdown now waits briefly for them first; a box removal does not, since its connection stays up.
- **Terminal handler validation.** Covered above: turn id checked, event type restricted to terminals.

## Verification

`npm test` — 245 files, 5101 passed, 2 skipped. `npx tsc --noEmit` clean.

Docs updated alongside: the turn-addressing section previously read as if one turn were live per session and as if the id could come from the acknowledgement — both were exactly the assumptions that had to be corrected in code.

## Deployment

Unchanged in shape, but worth restating since this now spans three images: build the **agentbox** image, repoint `SICLAW_AGENTBOX_IMAGE`, recycle agent pods, then roll Runtime and the control plane. The control plane must precede the Runtime for route/start/abort; turn addressing degrades to session-wide semantics on either side alone, so that half is order-independent. Assumes a single-process control plane and `SICLAW_REMOTE_DELEGATION_IDLE_TIMEOUT` within the relay lease (values above it are clamped with a warning).

src/gateway/server.ts
```diff
@@ -369,21 +500,79 @@ export async function startRuntime(opts: StartRuntimeOptions): Promise<RuntimeSe
         // Best effort: a consumer that already went away must not stop what caused this.
         console.warn(`[runtime] could not report interrupted turn session=${sessionId}:`, err);
       }
+      // A delegated turn's caller is a machine that will otherwise wait out its idle
+      // window and report a failure. Same terminal, but acknowledged — detached,
+      // because this runs on the shutdown path and must not hold it open.
+      for (const turnId of liveTurnIds.get(sessionId) ?? []) {
```
jacoblee-io: **[P2] Claim the supervisor terminal before starting delivery**

endTurns() can run first for box_rolled and then again for runtime_restart while the real SSE consumer has not settled yet. The first call tracks its delivery but leaves the turn in liveTurnIds/delegatedTurns, so shutdown starts a second authoritative terminal with a different reason. Because either delivery can win after retries, runtime_restart can replace the original box_rolled result. Please atomically claim each turn before starting the first supervisor delivery and make subsequent calls skip terminal emission/delivery for already claimed turns; they can still perform the necessary abort and await the tracked delivery. The new test currently hides this window because its mocked consumer resolves immediately when the signal is aborted.

src/gateway/server.ts
jacoblee-io: Addressed, and the probe is gone entirely rather than patched.

You were right that a presence snapshot cannot answer this: absent-at-probe-then-present is unobservable from outside, and the no-Stop case had no compensation at all. Both are now closed by the generation-scoped handshake you asked for — a prompt carries a `turnId`, the box records it as the running turn, and an abort naming any other turn is answered as already stopped. Compensation on a failed prompt is therefore unconditional: naming the turn makes it a no-op when the box never started it, and the only thing that stops it when it did.

`listSessions()` is no longer called here (`e3d5a218`), and the pre-spawn latch is turn-scoped, so the tombstone hazard that motivated the probe is gone with it.

src/gateway/server.ts
jacoblee-io: Fixed, and the failure was slightly worse than described: since the abort itself was session-keyed, an unscoped Stop also tore down A's consumer while telling the box only about B.

Turns are now tracked as a set per session with per-turn abort controllers (`turnAborts`), so a Stop that names none names all of them — the box stops the one it is running and answers the rest as already stopped — and a Stop that names one touches only that turn's controller. The set is snapshotted before the consumer is broken, since a settling turn removes itself from it.

The double-send + Stop regression test you asked for is in `4bbe5e76` (`stops the RUNNING turn when a second send has already queued behind it`), plus `does not touch the running turn's consumer when the abort names a queued turn`. Both were verified to fail against the previous behaviour rather than merely pass.

src/gateway/server.ts
jacoblee-io: Fixed. I took the numbers from `FrontendWsClient` rather than guessing: `BACKOFF_CAP_MS` is 30s with up to 2s of jitter, so a single reconnect can take 32s. The budget is now 500ms → 32s across seven attempts (~63.5s), which covers a full worst-case reconnect with margin.

Delivery also no longer holds the turn open — that was the reason a long budget looked unaffordable. It runs detached in a tracked set that shutdown flushes, so the turn settles immediately (`b468e5a9`, `182e892d`).

src/gateway/server.ts
jacoblee-io: Fixed in `b468e5a9`. Delegated turns are tracked (`delegatedTurns`, keyed by turn), so `endTurns()` reports their terminals over the acknowledged path carrying the cause it already names — `aborted: true, reason: box_rolled | runtime_restart`.

Your note about `close()` was the other half of it: it closed `FrontendWsClient` immediately after ending turns, killing the transport those deliveries need. Shutdown now waits for them first (bounded — a live connection settles in milliseconds, a dead one must not hold the process open for the whole budget). A box removal does not wait, since its connection stays up.

src/agentbox/session.ts
jacoblee-io: Fixed in `4bbe5e76`. The guard is now skipped whenever the abort names a turn, and kept for a legacy unscoped one — which is the split you described.

Your reasoning was the part I had wrong: I had treated the on-disk check as generally safe, when in fact a delegated peer thread always has history, so the guard was suppressing the latch in exactly the case that needs it.

The restore/rebuild window is covered too, in `182e892d`: an abort naming a turn that is not the one currently running no longer answers "already stopped" — that turn's prompt may still be in flight, or its session rebuilding — it records a latch for that turn instead and returns `pending: true`. Latches are also held per `(session, turn)`; one slot per session let two turn-scoped latches overwrite each other, which is ordinary when one turn is cancelled while another runs.

Coverage: `does not stop the running turn for an abort that names another one` and `scopes a pre-spawn latch to the turn that armed it` in `http-server.test.ts`.

src/gateway/server.ts
jacoblee-io: Fixed in `5c7ca9af`, including your point about the test.

Each turn is claimed before the first supervisor report. Later passes still cancel it and still stop its box, but do not report it again — the delivery they would have started is already tracked for shutdown to flush. So a `box_rolled` interruption keeps that cause instead of being overwritten by `runtime_restart`.

You were right that my test hid the window: the mocked consumer settled the moment its signal aborted, which a real one does not, so the turn always left `liveTurnIds` before the second pass. The mock can now model a consumer that ignores the abort, and with that the new test (`reports a supervisor-interrupted turn once, keeping the first cause`) fails against the previous behaviour with two terminals instead of one.

Two things I found while in there, both in this commit:

- A box removal was asking the *removed* box to stop its turn. That endpoint is dead by definition, so the request only hung out its timeout and logged a misleading warning; skipped for that reason.
- One thing I did **not** change, having convinced myself it was wrong to: clearing `_currentTurnId` when a turn finishes. It looks tidier, but a Stop arriving just after completion would then stop matching the turn it names and would only arm a latch — skipping `clearQueue` + `discardPendingNotifications`, which are what stop a buffered background-job notification from reviving the session after Stop. Keeping the field as "the turn this session last ran" is what makes that path still fire.

src/gateway/server.ts
```diff
@@ -347,43 +462,129 @@ export async function startRuntime(opts: StartRuntimeOptions): Promise<RuntimeSe
...
+
+      // Cancelling on this side does not stop the BOX. The consumer only notices its
+      // signal when the next event arrives, and a dropped SSE subscription just
+      // unsubscribes — neither ends the prompt. In K8s the boxes deliberately outlive
+      // a Runtime roll, so a turn already reported as interrupted would keep running
+      // there with nobody left to read it. Only a dispatched turn has a box to ask,
+      // and `busyOn` is the record of that placement.
+      // Not for a box that was just removed: its endpoint is dead by definition, so
+      // the request would only hang out its timeout and log a misleading warning.
+      const placement = reason === "box_rolled" ? undefined : sessionTurnLocks.busyOn(sessionId);
```
jacoblee-io: **[P1] Keep the scoped box abort on box roll**

The premise here that the endpoint is already dead is not true for the actual manager path. reapDrainedBoxes() invokes the turn terminator at manager.ts:1135 before awaiting spawner.stop() at :1141, and a failed stop is explicitly left for a later retry. In that failure window the Runtime has already emitted an authoritative box_rolled terminal and dropped its consumer, while the AgentBox prompt can keep running headless and producing tool side effects. Please keep abortSession(sessionId, turnId) against the busyOn() placement for box_rolled as well; an already-gone box can remain a best-effort/ignored outcome.

jacoblee-io: You are right, and the premise was mine to check — I asserted the endpoint was dead without reading the reap path. `manager.ts` reports the interruption at the `turnTerminator?.(held, "box_rolled")` call and only afterwards awaits `spawner.stop(boxId)`, whose failure path `continue`s and keeps the drain mark for a later tick. So the box is alive for the whole of that window, with the consumer already dropped and the terminal already sent.

Reverted in `f098a2ec`: the scoped `abortSession(sessionId, turnId)` is issued against the `busyOn()` placement for a box roll as well. A box that has in fact gone answers nothing, which is the outcome that was wanted, so that failure is logged rather than warned; a shutdown, where the box is expected to answer, still warns.

Worth flagging that I had gone further in the wrong direction before your comment landed: I was about to make the skip *survive* across passes, so a shutdown following a roll would not have asked either. That is not in any pushed commit.

`stops the box turn on a box roll too, since the box is not reliably gone yet` pins it, and fails against the skip — its abort simply never arrives. It also uses the ignore-the-abort consumer, so the turn is still live when the terminator runs, which is the state the reap path actually produces.

src/gateway/server.ts
```diff
@@ -347,43 +463,135 @@ export async function startRuntime(opts: StartRuntimeOptions): Promise<RuntimeSe
...
+      // failed stop is left for a later retry — so in that window the prompt keeps
+      // running, and producing tool side effects, with the consumer already dropped.
+      // A box that has in fact gone is simply the outcome we wanted, so failures here
+      // are ignored; only a shutdown, where the box is expected to answer, says so
+      // loudly.
+      const placement = sessionTurnLocks.busyOn(sessionId);
+      if (placement) {
+        const client = new AgentBoxClient(placement.endpoint, 10000, agentBoxTlsOptions);
+        for (const id of sessionTurns) {
+          started.push(client.abortSession(sessionId, id).catch((err) => {
```
jacoblee-io: **[P1] Track box-roll aborts across shutdown**

The box-roll caller discards the array returned by endTurns(), and this abort promise is not added to any central in-flight set. If the local SSE consumer settles on its next event, the turn leaves liveTurnIds; a SIGTERM immediately afterwards sees no live turn and gateway-main calls process.exit(0) as soon as close() returns, potentially cutting off both this abort and the concurrent box removal. K8s shutdown deliberately preserves boxes, so a delete that did not land leaves an already-reported turn running headless. Please track supervisor-started box aborts centrally and include them in the same bounded shutdown wait, or make the manager await a dedicated abort completion. A regression test should block abortSession, let the consumer settle, start close(), and verify shutdown still waits.

jacoblee-io: Fixed in `8f3257c1`. Same gap as the terminal deliveries, in the half I did not fix — the abort was only pushed onto `endTurns`' return value, which the box-removal caller discards.

Both kinds of outliving work now go through one central set that the bounded shutdown wait drains, so what a caller does with the return value no longer matters.

The test is the one you described — block `abortSession`, let the consumer settle, start `close()`, assert shutdown still waits — and it fails against the untracked version, where `close()` completes immediately.

It also needed its own session id, which is worth recording for anyone writing tests here: a test whose consumer never settles holds that session's turn lock for the rest of the file, since `sessionTurnLocks` is process-wide. Reusing the id made the new test hang in `acquire()` instead of testing anything, and only in a full-file run — it passed in isolation.

src/gateway/server.ts
```diff
@@ -347,43 +470,134 @@ export async function startRuntime(opts: StartRuntimeOptions): Promise<RuntimeSe
...
+   * close — so wait for those, but briefly: a live connection settles the first
+   * attempt in milliseconds, and a dead one must not hold the process open for the
+   * full retry budget.
+   */
+  const SHUTDOWN_TERMINAL_GRACE_MS = 3_000;
+
+  async function endInFlightTurns(): Promise<void> {
+    // Streaming turns AND cold-starting ones: both are ours, and both leave a caller
+    // waiting if they simply vanish.
+    const sessionIds = [...new Set([...activeStreamAborts.keys(), ...liveTurnIds.keys()])];
```
jacoblee-io, line 677: **[P1] Fence new work before taking the shutdown snapshot**

This snapshot is taken while every producer is still active: the reverse frontend command lane is closed only after endInFlightTurns() returns, the HTTP(S) servers are still listening, and the manager background loops are stopped even later. A chat.send or box-roll callback during this wait can therefore add a live turn or pendingShutdownWork after both snapshots; an already-entered chat.send can also still be in persistence before addLiveTurn(). The shutdown never rescans, then closes the transports and gateway-main calls process.exit(0), while K8s deliberately preserves the box. That new turn can continue without a terminal or abort. Please gate/quiesce inbound commands and manager callbacks before snapshotting, while keeping the frontend transport available for outbound terminal delivery, and account for handlers already in progress; alternatively drain repeatedly under the same deadline after producers have been fenced.

jacoblee-io: Fixed in `fe0504ae`, with one decision worth your explicit sign-off.

**The fence.** `chat.send` refuses once shutdown has begun, and a turn now registers before the persistence awaits in its own handler — so a handler that got past the check is already visible to the drain rather than appearing while it was in the database. That covers both windows you named: the newly admitted turn, and the already-entered one.

**It reverses a documented decision.** `server-shutdown-turns.test.ts` deliberately admitted a post-`close()` turn, on the stated reasoning that a blanket shutting-down flag would leave it unreported and therefore hanging. I think admitting it is the worse half of that trade — a refusal is an explicit error the caller can act on by placing the turn on a Runtime that will still be there, which is more than an unsupervised turn offers — so that test now asserts the refusal and records why the trade flipped. If you would rather keep the old contract, the alternative is the repeated drain and the fence should come back out; say so and I will swap them.

**Two things I removed rather than kept.**

I wrote the repeated drain first. Under the fence I could not construct a case it catches: a turn that finishes during the wait finds itself already reported and does not report again, and a box roll finds every live turn already claimed and already asked to stop. Failing to write a discriminating test was the argument against keeping it — unexercised machinery on the shutdown path is its own hazard — so the drain takes one snapshot and the reason it is sufficient is in the comment.

Writing that loop did surface a real defect in it, which is now fixed independently: a turn whose consumer never settles stays live, so every pass re-issued its box abort. The supervisor now asks about a given turn once, which also stops a box roll followed by a shutdown from asking twice.

Tests: `npm test` — 245 files, 5109 passed. The fence half of the new test fails without the fence (the late send resolves instead of rejecting).

src/gateway/server.ts
```diff
@@ -347,43 +489,148 @@ export async function startRuntime(opts: StartRuntimeOptions): Promise<RuntimeSe
...
+    //
+    // The fence is what makes one snapshot sufficient. Producers do outlive this drain —
+    // the command lane stays open so terminals can still be delivered, the servers are
+    // still listening, the manager's loops run until later — but after it, none of them
+    // can add work this snapshot would miss: no new turn is admitted; a turn that
+    // finishes now finds itself already reported and does not report again; and a box
+    // removal finds every live turn claimed and already asked to stop. Re-scanning was
+    // tried and removed for exactly that reason — nothing reachable turned up in a
+    // second pass, and unexercised machinery on the shutdown path is its own hazard.
+    shuttingDown = true;
```
jacoblee-io, line 665: [P1] Fence every turn-producing ingress\n\nThis flag currently gates only chat.send, so the invariant used to justify the single snapshot does not hold. The reverse command lane remains open while endInFlightTurns() awaits, and task.fireNow can still complete its RPC and detach executeJob(); the mTLS server is also still accepting POST /api/internal/delegate, whose local path calls AgentBoxClient.prompt() directly and whose remote path starts a relay. None of that work is registered in liveTurnIds, so this snapshot cannot report or abort it, and K8s shutdown preserves its box. Please apply the admission fence to every path that can start AgentBox work, including requests already inside an async pre-dispatch phase, while leaving terminal/control/abort traffic available for draining. A shutdown test through at least one non-chat.send ingress would catch this.

src/gateway/server.ts
```diff
@@ -347,43 +489,148 @@ export async function startRuntime(opts: StartRuntimeOptions): Promise<RuntimeSe
...
+      // running, and producing tool side effects, with the consumer already dropped.
+      // A box that has in fact gone is simply the outcome we wanted, so failures here
+      // are ignored; only a shutdown, where the box is expected to answer, says so
+      // loudly.
+      const placement = sessionTurnLocks.busyOn(sessionId);
+      if (placement) {
+        const client = new AgentBoxClient(placement.endpoint, 10000, agentBoxTlsOptions);
+        for (const id of sessionTurns) {
+          if (boxAbortAsked.has(id)) continue;
+          boxAbortAsked.add(id);
```
jacoblee-io, line 630: [P1] Let shutdown retry a failed box abort\n\nThe turn is inserted into boxAbortAsked before abortSession() resolves, and every non-404 failure is caught into a fulfilled tracked promise without clearing that marker. If a box-roll abort times out, a following shutdown can still see the live turn but skips its only retry; if the consumer settles meanwhile, the failed abort disappears from the remaining bookkeeping altogether. If spawner.stop() also failed—the exact case for which the box-roll abort was added—the prompt remains headless. Please deduplicate only an in-flight or successful abort, or retain failed attempts so shutdown can retry them, and add a regression test where the box-roll abort rejects before close().

src/gateway/server.ts
jacoblee-io: Right on both counts, and the second half of your point changes what I claim rather than only what the code does.

**Fenced the delegation ingress.** `POST /api/internal/delegate` starts AgentBox work of its own — a local peer prompt, or a remote relay — and now answers 503 once the fence is up, before it touches the roster. There is a test through that ingress (`refuses a delegation once the Runtime is shutting down`), which is the non-`chat.send` coverage you asked for.

**Narrowed the invariant instead of overstating it.** You are right that the single snapshot cannot rest on "every producer is fenced", because two are not: the task coordinator's scheduled and fire-now jobs (`task-coordinator.ts` keeps its own `AgentBoxClient` and prompts at :294) and capability runs both start box work and are registered nowhere in `liveTurnIds`. Neither was ever reported or aborted by this drain, before this PR or after it. So the comment and the design doc now say what actually holds — one snapshot suffices for the `chat.send` and delegation ingresses, which are the turns this drain has ever known — and name the other two as an open gap rather than leaving a claim that reads as covering them.

I have deliberately not extended the fence to them here. Giving those paths an admission gate *and* registration so the drain can see them is a shutdown-quiescence change across two subsystems this PR does not otherwise touch, and I would rather it be reviewable on its own than appended to a delegation-routing branch. Happy to open it next if you want it before this merges — but it is not a regression this PR introduces.

**The abort retry.** Fixed: the mark is cleared when an attempt is refused, so only an in-flight or successful abort dedupes. Exactly the case you named — a box roll whose `spawner.stop()` also failed — was spending the shutdown's only retry. `retries a box abort that failed, instead of counting it as asked` fails when a refused attempt counts as asked.

Tests: `npm test` — 245 files, 5111 passed. Head is `622d1513`.

src/gateway/server.ts
jacoblee-io: Right on both counts, and the second half of your point changes what I claim rather than only what the code does.

**Fenced the delegation ingress.** `POST /api/internal/delegate` starts AgentBox work of its own — a local peer prompt, or a remote relay — and now answers 503 once the fence is up, before it touches the roster. There is a test through that ingress (`refuses a delegation once the Runtime is shutting down`), which is the non-`chat.send` coverage you asked for.

**Narrowed the invariant instead of overstating it.** You are right that the single snapshot cannot rest on "every producer is fenced", because two are not: the task coordinator's scheduled and fire-now jobs (`task-coordinator.ts` keeps its own `AgentBoxClient` and prompts at :294) and capability runs both start box work and are registered nowhere in `liveTurnIds`. Neither was ever reported or aborted by this drain, before this PR or after it. So the comment and the design doc now say what actually holds — one snapshot suffices for the `chat.send` and delegation ingresses, which are the turns this drain has ever known — and name the other two as an open gap rather than leaving a claim that reads as covering them.

I have deliberately not extended the fence to them here. Giving those paths an admission gate *and* registration so the drain can see them is a shutdown-quiescence change across two subsystems this PR does not otherwise touch, and I would rather it be reviewable on its own than appended to a delegation-routing branch. Happy to open it next if you want it before this merges — but it is not a regression this PR introduces.

**The abort retry.** Fixed: the mark is cleared when an attempt is refused, so only an in-flight or successful abort dedupes. Exactly the case you named — a box roll whose `spawner.stop()` also failed — was spending the shutdown's only retry. `retries a box abort that failed, instead of counting it as asked` fails when a refused attempt counts as asked.

Tests: `npm test` — 245 files, 5111 passed. Head is `622d1513`.

src/gateway/server.ts
jacoblee-io: Thanks — I agree that task/capability/channel quiescence can be split into a follow-up because those paths predate this PR. The delegation part is still not closed, though. isShuttingDown() is sampled only once at delegate-api.ts:310; after that the handler awaits roster, model binding, routing, session/reuse lookups, persistence, the session lock, and (locally) getOrCreate() before it calls delegation.start or prompt(). The handler is never registered in liveTurnIds or another shutdown tracker. It can therefore observe false, pause in any of those awaits, let shutdown take its one snapshot, and then dispatch. A delegation already running when the fence rises is likewise not drained before gateway-main calls process.exit(). The new test holds the flag true from entry, so it misses both cases. Please bind this handler to a shared shutdown signal/registration (including aborting already-dispatched local or remote work), or otherwise make the final dispatch boundary atomic with the gate. A regression test can block a pre-dispatch await, start close(), release it, and assert that neither prompt() nor delegation.start is sent.

src/gateway/server.ts
jacoblee-io: This closes the case where the first abort has already rejected before shutdown takes stock, but not the overlapping case. If the box-roll abort is still pending when close() starts, endTurns() sees boxAbortAsked and skips a second request; alreadyPending only waits for that same request. When it then rejects, the catch deletes the marker, but the one-pass shutdown never revisits the turn. If the local consumer settled meanwhile, dropLiveTurn() has also removed the retry target entirely. The new test uses an immediate rejection, keeps the consumer live, and waits before close(), so it specifically avoids this interleaving. Please retain per-turn abort state/continuation so shutdown can retry a failed in-flight attempt within its grace (or allow the safe turn-scoped duplicate). A regression test should block the first abort, start close(), then release it as a rejection and assert a second abort.

jacoblee-io:
Drawing the scope line explicitly, because this branch has been reviewed well past what it sets out to do.

The PR's purpose is Runtime placement for delegated peers. Everything in the first review round was about that, and it has been settled for several rounds. What followed came from one decision — addressing aborts by turn rather than by session — which is not a delegation requirement but a pre-existing ambiguity that delegation exposed. Fixing it pulled in the shutdown path, and the shutdown path pulled in producers this branch does not otherwise touch.

**This branch stops here.** Head is `622d1513`: `npm test` 245 files / 5111 passed, `npx tsc --noEmit` clean, every thread answered. The paired control-plane change is confirmed reviewable as it stands.

Two known gaps are deliberately left, both pre-existing and now documented in `docs/design/coordinator-routing.md` rather than implied away:

- The task coordinator's scheduled and fire-now jobs, and capability runs, start AgentBox work with their own clients and are registered nowhere in `liveTurnIds`. They are therefore neither fenced at shutdown nor drained — as was true before this branch. Giving them the same admission gate and registration is a shutdown-quiescence change across two subsystems, worth reviewing on its own.
- `settledDelegations` is per-process, so a Runtime restart forgets it. With turn-scoped aborts the consequence is a log line, not a wrong outcome.

Further findings adjacent to cancellation or shutdown are welcome as issues rather than as blockers here. A finding that this PR's own subject — placement, the two-leg terminal contract, or the durable result readback — is wrong is of course still a blocker.

src/gateway/server.ts
jacoblee-io: Both correct, and both were interleavings my own tests stepped around. Fixed in `4cbf7035`.

**The dispatch boundary.** The gate is re-read immediately before dispatch, so the long stretch of awaits between entry and `delegation.start` / `prompt()` can no longer carry a turn past the drain's one look. And the handler now registers a wind-down — the same one a client disconnect triggers, which for a remote peer includes `delegation.abort` — so a delegation already past refusing is stopped by that shutdown instead of outliving it. Shutdown runs those hooks *before* it takes stock, so whatever they start is waited on too.

Test: `does not dispatch when shutdown begins during a pre-dispatch await` blocks `delegation.resolveRoute`, flips the gate while that await is outstanding, releases, and asserts neither `delegation.start` nor `prompt()` was sent. Against the previous code it hangs waiting for a relay it should never have started.

`registers a wind-down so a shutdown reaches a delegation already under way` covers the other half.

Taking your agreement on the task/capability/channel split — those stay for a follow-up.

src/gateway/server.ts
jacoblee-io: Correct, and the fix belongs where you point rather than in another pass. `4cbf7035`.

Deduplicating on "asked" cannot be recovered by a later caller: it finds the mark set while the first attempt is still outstanding, `alreadyPending` waits on that same request, and when it rejects there is no later pass — with the consumer settled meanwhile, `dropLiveTurn()` has removed the target entirely. So the retry now lives with the attempt, inside the promise shutdown is already awaiting, and the mark is cleared only once every attempt has been refused.

Test: `retries an abort that was still outstanding when shutdown began` blocks the first abort, starts `close()` while it is outstanding, asserts shutdown issued nothing of its own and is still holding the transport, then releases it as a rejection and asserts the retry goes out. Against a single-attempt version it never sees that second request.

src/gateway/delegate-api.ts
```diff
@@ -105,30 +284,103 @@ export async function handleDelegate(
...
+        console.warn(`[delegate-api] failed to abort peer session ${peerSessionId}:`, err);
+      });
+    }
+  };
+  res.on("close", onResponseClose);
+  if (res.destroyed) onResponseClose();
+  // Same wind-down a client disconnect triggers: abort the peer, and for a remote peer
+  // tell the control plane. Registered from here so a shutdown reaches a delegation that
+  // is already under way, not only the ones it can still refuse.
+  const unregisterFromShutdown = deps.shutdownGate?.register(() => onResponseClose()) ?? (() => {});
```
jacoblee-io: [P2] Unregister the shutdown hook on every exit path

This registration is released only by the inner finally reached after session-lock acquisition. Every earlier return after this line, including entry shutdown, authorization, binding, route, parent, persistence failure, or cancellation, leaves a closure in shutdownCancels for the Runtime lifetime. Rejected requests can therefore grow the set and retain request, response, and dependency state, and shutdown later invokes stale hooks. Please cover the whole post-registration handler with an outer try/finally, or move registration to an atomic pre-dispatch scope that always unregisters.

src/gateway/server.ts
jacoblee-io: The dispatch-boundary half is fixed, but the in-flight wind-down is still not awaited. The registered callback invokes onResponseClose, while onResponseClose returns void and fire-and-forgets both delegation.abort and the local abortSession. Therefore the promise tracked by endInFlightTurns settles as soon as those calls are scheduled; frontendClient.close or process exit can still cut off the actual abort. The new test invokes the fake cancel directly and has the abort resolve immediately, so it proves that an abort is started, not that server.close waits for it. Please return and await the selected abort promise from the hook and add a shutdown test that blocks it, starts close, and verifies the transport remains open until the abort is released.

src/gateway/delegate-api.test.ts
```diff
@@ -221,3 +262,461 @@ describe("handleDelegate — model-failure propagation (P1)", () => {
...
+        isShuttingDown: () => false,
+        register: (fn: () => void) => { cancel = fn; return () => { cancel = undefined; }; },
+      },
+    };
+    deps.frontendClient.request = vi.fn(async (method: string, params: any) => {
+      if (method === "config.getDelegates") return { members: [{ id: PEER, name: "peer", description: "", clusters: [], hosts: [] }] };
+      if (method === "delegation.resolveRoute") return { local: false, sourceRuntimeId: "shanghai", targetRuntimeId: "aries" };
+      if (method === "delegation.start") {
+        // Under way now: shutdown can only wind it down.
+        cancel?.();
```
jacoblee-io, line 435: [P1] Handle cancellation while delegation.start is pending

This cancellation fires while runRemoteDelegation is still awaiting delegation.start. Its abort listener rejects remoteDone, but no rejection handler is attached until the later await remoteDone, after delegation.start returns. I reproduced this on the PR head: all 50 targeted assertions pass, but Vitest reports Unhandled Rejection: delegation stopped from this line and exits 1. Please handle remoteDone immediately, or race startup with cancellation, and keep the regression test free of unhandled errors.

src/gateway/delegate-api.ts
jacoblee-io: All three correct, and the second one matters beyond its own bug: `npm test` was exiting 1 while I was reporting the suite as clean, because I had been reading its summary lines and not its exit code. That is on me, and it is why the rejection survived several rounds. `npm test` now exits 0 (245 files, 5114 passed).

**Awaiting the wind-down.** The hook returned void and fire-and-forgot both aborts, so the promise `endInFlightTurns` tracked settled as soon as the request was scheduled. It now returns the abort it issues — a client disconnect can fire and forget, a shutdown cannot. `hands the shutdown a wind-down it can WAIT for, not one it merely starts` blocks `delegation.abort`, drives the hook the way shutdown does, and asserts it has not settled until the abort is released; against a fire-and-forget hook it fails immediately.

**The unhandled rejection.** `remoteDone` is marked handled at creation, so cancellation during `delegation.start` no longer rejects an unobserved promise. The later `await remoteDone` still observes it, so behaviour is unchanged.

**The leaked hooks.** Registration moved inside the scope whose `finally` always runs, so a request rejected before dispatch cannot leave a closure behind. Nothing before that point has peer work to wind down — the gate re-read is what covers that window — so moving it costs no coverage.

src/gateway/delegate-api.test.ts
jacoblee-io: All three correct, and the second one matters beyond its own bug: `npm test` was exiting 1 while I was reporting the suite as clean, because I had been reading its summary lines and not its exit code. That is on me, and it is why the rejection survived several rounds. `npm test` now exits 0 (245 files, 5114 passed).

**Awaiting the wind-down.** The hook returned void and fire-and-forgot both aborts, so the promise `endInFlightTurns` tracked settled as soon as the request was scheduled. It now returns the abort it issues — a client disconnect can fire and forget, a shutdown cannot. `hands the shutdown a wind-down it can WAIT for, not one it merely starts` blocks `delegation.abort`, drives the hook the way shutdown does, and asserts it has not settled until the abort is released; against a fire-and-forget hook it fails immediately.

**The unhandled rejection.** `remoteDone` is marked handled at creation, so cancellation during `delegation.start` no longer rejects an unobserved promise. The later `await remoteDone` still observes it, so behaviour is unchanged.

**The leaked hooks.** Registration moved inside the scope whose `finally` always runs, so a request rejected before dispatch cannot leave a closure behind. Nothing before that point has peer work to wind down — the gate re-read is what covers that window — so moving it costs no coverage.

src/gateway/server.ts
jacoblee-io: All three correct, and the second one matters beyond its own bug: `npm test` was exiting 1 while I was reporting the suite as clean, because I had been reading its summary lines and not its exit code. That is on me, and it is why the rejection survived several rounds. `npm test` now exits 0 (245 files, 5114 passed).

**Awaiting the wind-down.** The hook returned void and fire-and-forgot both aborts, so the promise `endInFlightTurns` tracked settled as soon as the request was scheduled. It now returns the abort it issues — a client disconnect can fire and forget, a shutdown cannot. `hands the shutdown a wind-down it can WAIT for, not one it merely starts` blocks `delegation.abort`, drives the hook the way shutdown does, and asserts it has not settled until the abort is released; against a fire-and-forget hook it fails immediately.

**The unhandled rejection.** `remoteDone` is marked handled at creation, so cancellation during `delegation.start` no longer rejects an unobserved promise. The later `await remoteDone` still observes it, so behaviour is unchanged.

**The leaked hooks.** Registration moved inside the scope whose `finally` always runs, so a request rejected before dispatch cannot leave a closure behind. Nothing before that point has peer work to wind down — the gate re-read is what covers that window — so moving it costs no coverage.

src/gateway/delegate-api.ts
```diff
@@ -105,30 +284,106 @@ export async function handleDelegate(
...
+  let peerSessionId = "";
+  let remoteStartRequested = false;
+  /**
+   * Wind this delegation down, RETURNING the abort it issues.
+   *
+   * A client disconnect can fire and forget; a shutdown cannot — it has to wait for the
+   * abort to land before the transport closes and the process exits, so this hands the
+   * promise back instead of scheduling it and returning void.
+   */
+  const cancelPeerWork = async (): Promise<void> => {
```
jacoblee-io: [P1] Keep the real wind-down tracked through failure and disconnect

The returned promise is now awaited, but both abort branches still catch a refusal and resolve, so endInFlightTurns treats an immediate failed abort as a completed wind-down. For a local delegation there is no Sicore relay or lease fallback, and K8s preserves the box, so the peer prompt can still remain headless. Please put a bounded self-retry inside this returned promise, as stopTurnOnBox does, rather than converting failure to success immediately.

There is also a disconnect-to-SIGTERM window: onResponseClose starts cancelPeerWork fire-and-forget; peerAbort can let the handler reach its finally and unregister the shutdown hook while that abort is still pending. A following shutdown then sees neither the hook nor centrally tracked work and can exit underneath it. Please memoize the cancellation promise and keep the hook registered until that promise settles, or track the promise centrally regardless of whether disconnect or shutdown started it. A regression test can block the local abort, trigger response close, let the handler settle, then start Runtime close; shutdown must still wait, and a rejected first attempt should be retried.

jacoblee-io: Both halves confirmed and fixed in `9690c535`.

**The refusal.** The attempt now retries, bounded, instead of catching and resolving — the same shape as the supervisor's own box abort. You are right that the local path has no fallback: no relay lease, and the box outlives the Runtime, so a refused abort left the prompt running with nothing else on the way.

**The disconnect-to-SIGTERM window.** The cancellation is memoized, so a disconnect and a shutdown wait on the same attempt rather than the shutdown finding nothing; and the hook stays registered until that attempt settles, instead of being released when the handler reaches its finally.

Worth recording how the test got there, because two earlier versions of it were worthless and I only caught that by reverting each half:

- Counting aborts on the remote path proves nothing — the handler issues a `delegation.abort` of its own from its catch, so the count reaches two whether or not the wind-down retried.
- Asserting the hook is still registered right after the retry proves nothing either — the handler had not reached its finally yet, so it was still registered for an unrelated reason.

It now drives the local path, where the wind-down is the only source of `abortSession`, and asserts after `await handling`. Reverting the retry gives `expected 1 to be 2`; reverting the hook lifetime gives `expected undefined to be defined`.

`npm test` exits 0 — 245 files, 5115 passed. I am checking the exit code now, not the summary lines.

jacoblee-io (approved):

PR #479, opened 2026-08-07
feat(kbc): understand multi-repository code domains

LikiosSedo:
## Summary

- treat an optional `raw/CODE_SOURCES.json` as the control manifest for a multi-repository code domain
- expose repository boundaries and immutable source identities to the code-knowledge compiler
- preserve repository-local component boundaries in deterministic coverage and incremental recompilation
- keep the manifest out of content accounting, batch routing, and compilation batches while leaving it readable through source inspection
- guide the compiler to model cross-repository interfaces, deployment topology, state ownership, and failure propagation with concrete source citations
- validate and render server-owned knowledge renew plans so the runtime follows the platform-selected objective, scope, and OKF contract transition

## Why

A canonical code-domain archive can contain multiple repositories under stable top-level paths. The compiler needs to understand those boundaries without treating the aggregate archive commit as shared repository history or collapsing all components inside one repository into a single coverage unit.

Knowledge refresh also needs one convergence intent: the service decides whether a frozen Raw delta is scoped incremental work or a full output-contract migration, while the compiler receives an explicit bounded task contract.

## Validation

- `python kbc/platform/pod/test_incremental.py`
- `python -m py_compile kbc/platform/pod/compile_box.py`
- targeted renew command normalization/rendering cases in `test_compile_box.py` (covered by KBC Box CI)
- `git diff --check`

PR #483, opened 2026-08-10
feat(kbc): honor server-owned knowledge renew plans

LikiosSedo:
## Summary

- validate the server-owned knowledge renew strategy and bounded plan fields
- append the authoritative renew plan to localized KBC compile instructions
- cover OKF contract migration context in the typed-command protocol tests

## Why

Sicore sends a typed renew plan when recompiling a knowledge base or migrating its output contract. Without this change, KBC accepts the command but does not expose the strategy, scope, or OKF version transition to the compiler.

This is the generic renew-plan commit extracted from #479. Multi-repository Code Raw behavior is intentionally excluded from this PR.

## User impact

Commands without a `renew` plan are unchanged. When a plan is present, KBC validates it and tells the compiler exactly which scope and OKF contract transition the platform authorized.

## Validation

- `/private/tmp/siclaw-kbc-renew-py312/bin/python kbc/platform/pod/test_compile_box.py`
- `git diff --check origin/main...HEAD`
- extracted patch-id matches original commit `e8b38265`


kbc/platform/pod/prompts/en/commands.json
```diff
@@ -4,6 +4,7 @@
   "compile.regenerate": "Regenerate the complete knowledge-base draft in this staging workspace. The owner has authorized execution without another plan-confirmation stop. Read the structured compile brief, cover the complete source set with provenance, handle uncertainty through best-guess plus tickets, and finish through candidate/index.md.",
   "compile.approve_plan": "The owner approved the exact current proposed plan. Execute it now, preserve the structured compile brief, handle uncertainty through best-guess plus tickets, and finish through candidate/index.md.",
   "compile.incremental": "Apply the structured source changeset as a scoped incremental compilation.",
+  "compile.renew_context": "Platform renew plan (authoritative): strategy={strategy}; objective={objective}; scope={scope}; output contract {current_okf_version} -> {target_okf_version}. Do not widen the content-edit scope beyond this plan. Preserve owner governance artifacts and unknown OKF fields.",
```
jacoblee-io: The standing compiler prompts still require every candidate page to be an OKF v0.1 concept document and require `candidate/index.md` to contain only `okf_version: "0.1"` (`box_role.md` and `playbook.md`). Appending a target transition such as `0.1 -> 0.2` here conflicts with those stronger constraints and does not tell the compiler what the v0.2 structure is, so a `contract_migration` can still produce v0.1 output and fail to converge. Please update or parameterize the base OKF contract instructions as part of this change, and add a test that asserts the resulting artifact version rather than only checking that the command text contains `0.2`.

jacoblee-io (approved):
Re-reviewed the updated implementation. The OKF v0.2 contract conflict is resolved, renew targets are validated against the compiler-supported version, and the scoped supersession path retains its per-round declaration and integrity guard. Incremental tests and CI pass.

PR #489, opened 2026-08-18
fix(knowledge): wipe agentbox disk when the next release has no libraries

LikiosSedo:
## Summary

A successful empty `config.getKnowledgeBundle` means the type unbound every library. AgentBox used to keep the previous on-disk wiki when `repos: []` arrived, so a delete-and-republish left library A visible.

Fetch failures never reach `materialize` (the reload handler returns 500 first), so the old preserve guard was protecting the wrong case.

## Test plan

- [x] `npx vitest run src/agentbox/sync-handlers.test.ts` (57 passed)
- [ ] After merge: publish a type with knowledge A, then publish again with knowledge unbound; a new tenant session must not see A

Made with [Cursor](https://cursor.com)

### DIFF
--- a/docs/design/invariants.md
+++ b/docs/design/invariants.md
@@ -38,9 +38,10 @@ TUI has two sub-modes: **standalone** (no Portal in the cwd) and **Portal-paired
 - Any code that writes/deletes files in `./skills/` affects ALL users simultaneously
 - `skillsHandler.materialize()` is **NOT safe** in local mode — it wipes `skills/global/`, `skills/skillset/`, and `skills/user/` subdirectories (not `core/`), which in a shared filesystem destroys ALL users' personal skills. This is designed for K8s pods with isolated filesystems.
 - Per-user skill sync in local mode must write only to `skills/user/<userId>/` without touching `skills/core/` (global + personal skills from the bundle are both written into the user's directory)
+- Knowledge sync in local mode must materialize and read from an agent-scoped directory (`.siclaw/knowledge/<agentId>/`). Empty and non-empty bundles both replace their target, so pointing the canonical handler at the shared knowledge root would let one agent erase another agent's libraries.
 - Local SQLite (via `node:sqlite`) uses WAL mode with a shared process — local mode is single-process by design; production K8s uses MySQL and has no such constraint
 
-**Source**: `src/gateway/agentbox/local-spawner.ts`, `src/agentbox/resource-handlers.ts:82-97`
+**Source**: `src/gateway/agentbox/local-spawner.ts`, `src/agentbox/sync-handlers.ts`, `src/agentbox/session.ts`, `src/core/agent-factory.ts`
 
 ### 1.3 K8s Pod Isolation
 
@@ -316,6 +317,7 @@ postReload(context)  Notify active sessions to pick up changes
 |---------|----------------------|------------------|-------|
 | `mcpHandler.materialize()` | ✅ Yes | ✅ Yes | Merges, does not wipe |
 | `skillsHandler.materialize()` | ❌ No | ✅ Yes | Wipes `global/` + `skillset/` + `user/` subdirs (not `core/`) |
+| `knowledgeHandler.materialize()` | ✅ With agent-scoped target | ✅ Yes | Replaces its whole target; LocalSpawner binds `.siclaw/knowledge/<agentId>/` |
 
 For local mode skills sync, write directly to `skills/user/<userId>/` without delegating to `skillsHandler.materialize()`. Global and personal skills from the bundle are both placed under the user's directory.
 
--- a/src/agentbox/http-server.test.ts
+++ b/src/agentbox/http-server.test.ts
@@ -62,6 +62,7 @@ vi.mock("./sync-handlers.js", () => ({
   getSyncHandler: () => undefined,
   createClusterHandler: () => ({ type: "cluster", fetch: async () => 0, materialize: async (n: number) => n }),
   createHostHandler: () => ({ type: "host", fetch: async () => 0, materialize: async (n: number) => n }),
+  createKnowledgeHandler: () => ({ type: "knowledge", fetch: async () => ({ repos: [] }), materialize: async () => 0 }),
   createToolsHandler: (target: { allowedToolsState: string[] | null }) => ({
     type: "tools",
     fetch: async () => ({ allowedTools: null }),
--- a/src/agentbox/http-server.test.ts
+++ b/src/agentbox/http-server.test.ts
@@ -62,7 +62,12 @@ vi.mock("./sync-handlers.js", () => ({
   getSyncHandler: () => undefined,
   createClusterHandler: () => ({ type: "cluster", fetch: async () => 0, materialize: async (n: number) => n }),
   createHostHandler: () => ({ type: "host", fetch: async () => 0, materialize: async (n: number) => n }),
-  createKnowledgeHandler: () => ({ type: "knowledge", fetch: async () => ({ repos: [] }), materialize: async () => 0 }),
+  createKnowledgeHandler: () => ({
+    type: "knowledge",
+    fetch: async () => ({ repos: [] }),
+    materialize: async () => 0,
+    getLastKnowledgeSyncStatus: () => null,
+  }),
   createToolsHandler: (target: { allowedToolsState: string[] | null }) => ({
     type: "tools",
     fetch: async () => ({ allowedTools: null }),
--- a/src/agentbox/http-server.ts
+++ b/src/agentbox/http-server.ts
@@ -27,6 +27,7 @@ import {
   createHostHandler,
   createKnowledgeHandler,
   createToolsHandler,
+  type KnowledgeSyncHandler,
 } from "./sync-handlers.js";
 import { GATEWAY_SYNC_DESCRIPTORS, type AgentBoxSyncHandler, type GatewaySyncType } from "../shared/gateway-sync.js";
 import { detectLanguage } from "../shared/detect-language.js";
@@ -456,6 +457,13 @@ async function abortBrainForHttp(
 
 export interface CreateHttpServerOptions {
   /**
+   * Knowledge handler already bound to this box's target directory. LocalSpawner
+   * supplies the same instance used by its initial sync so status and reloads
+   * share one per-box lifecycle. Other entrypoints create their handler here.
+   */
+  knowledgeHandler?: KnowledgeSyncHandler;
+
+  /**
    * If true, skip the idle self-destruct entirely. Intended for LocalSpawner,
    * which runs AgentBox in-process with the Portal — shutting down from
    * the idle timer would take the whole `siclaw local` process down with it.
@@ -518,9 +526,8 @@ export function createHttpServer(
     perServerHandlers.host = createHostHandler(sessionManager.credentialBroker);
   }
   if (sessionManager.knowledgeDir) {
-    perServerHandlers.knowledge = createKnowledgeHandler({
-      knowledgeDir: sessionManager.knowledgeDir,
-    });
+    perServerHandlers.knowledge = options.knowledgeHandler
+      ?? createKnowledgeHandler({ knowledgeDir: sessionManager.knowledgeDir });
   }
   // tools handler — same per-box rationale as cluster/host. It writes the
   // resolved allowedTools into THIS box's sessionManager and fetches with THIS
--- a/src/agentbox/http-server.ts
+++ b/src/agentbox/http-server.ts
@@ -21,7 +21,13 @@ import { checkMetricsAuth } from "../shared/metrics.js"; // also registers metri
 import { GatewayClient } from "./gateway-client.js";
 import { CredentialBroker } from "./credential-broker.js";
 import { HttpTransport } from "./credential-transport.js";
-import { getSyncHandler, createClusterHandler, createHostHandler, createToolsHandler } from "./sync-handlers.js";
+import {
+  getSyncHandler,
+  createClusterHandler,
+  createHostHandler,
+  createKnowledgeHandler,
+  createToolsHandler,
+} from "./sync-handlers.js";
 import { GATEWAY_SYNC_DESCRIPTORS, type AgentBoxSyncHandler, type GatewaySyncType } from "../shared/gateway-sync.js";
 import { detectLanguage } from "../shared/detect-language.js";
 import { stripLanguageDirective } from "../shared/strip-language-directive.js";
@@ -510,6 +516,11 @@ export function createHttpServer(
   if (sessionManager.credentialBroker) {
     perServerHandlers.cluster = createClusterHandler(sessionManager.credentialBroker);
     perServerHandlers.host = createHostHandler(sessionManager.credentialBroker);
+  }
+  if (sessionManager.knowledgeDir) {
+    perServerHandlers.knowledge = createKnowledgeHandler({
+      knowledgeDir: sessionManager.knowledgeDir,
+    });
   }
   // tools handler — same per-box rationale as cluster/host. It writes the
   // resolved allowedTools into THIS box's sessionManager and fetches with THIS
--- a/src/agentbox/resource-sync.test.ts
+++ b/src/agentbox/resource-sync.test.ts
@@ -127,6 +127,7 @@ describe("syncResource — error paths", () => {
     expect(result.message).toBe("boom");
     // mcp descriptor says maxRetries=3
     expect(handler.fetch).toHaveBeenCalledTimes(3);
+    expect(handler.materialize).not.toHaveBeenCalled();
   });
 });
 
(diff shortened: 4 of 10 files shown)


jacoblee-io (commented):
The staleness bug this fixes is real, and treating a successful empty fetch as authoritative is the right instinct. But I don't think this can land as-is: the justification for removing the guard holds for only one of the two paths that reach `materialize`, and in local mode the now-unconditional wipe crosses a tenant boundary.

## Blocking

### 1. In local mode this wipes knowledge shared by *all* users

`knowledgeDir` is `path.resolve(process.cwd(), config.paths.knowledgeDir)` — no user, agent, or session component. `LocalSpawner` runs every AgentBox in-process off one cwd, so all of them share that single directory.

`local-spawner.ts` says so itself, right above the call:

> Skills and MCP are intentionally skipped: their handlers wipe a shared directory and would clobber other users' state (invariant #1 in CLAUDE.md).

…and then calls `await syncResource("knowledge", ...)` a few lines down, fire-and-forget, on every spawn. Knowledge was safe to call there *because of the guard this PR deletes*. Remove it and knowledge gets the same blast radius as `skillsHandler.materialize()`, except knowledge is the one handler that actually runs in local mode.

Scenario: user A's agent has a knowledge library materialized into `.siclaw/knowledge/`. User B spawns an agent with nothing bound → `{repos: []}` → the shared directory is wiped → A loses its knowledge until the session restarts.

This is the 🔴 invariant in CLAUDE.md ("`LocalSpawner` runs ALL AgentBox instances **in-process**, sharing one filesystem"), so it needs an explicit gate rather than an incidental one.

### 2. "Fetch failures never reach materialize" is only true on the HTTP reload path

There are two callers of `handler.materialize(payload)`:

- `src/agentbox/http-server.ts` (~1464) — the reload endpoint. The claim holds here: fetch throws, the handler returns 500, `materialize` is never reached.
- `src/agentbox/resource-sync.ts` (~43) — `syncResource`'s retry loop. It retries only on a **thrown** fetch. A fetch that *succeeds* and returns an empty bundle goes straight into `materialize` and now wipes.

That second path is the one `local-spawner.ts` calls at spawn time, before Gateway-side bindings are reliably readable. And the Gateway side can't tell the two cases apart either: the knowledge bundle handler in `src/portal/adapter.ts` returns `{repos: []}` whenever the `agent_knowledge_repos` query comes back with zero rows — genuinely unbound and not-yet-written look identical. With backoff retries on top, one unlucky ordering is a permanent delete rather than a transient miss.

The deleted comment referred to a Gateway-side primary fix ("empty-bundles only arrive when the agent is genuinely unbound"). If that guarantee exists, it'd be worth pointing at it in the PR — it's the premise for dropping the belt-and-suspenders. If it doesn't, the belt is going too.

## Worth addressing before merge

### 3. The skills guard is left behind with no note saying why

`skillsHandler.materialize` still carries a structurally identical guard with identical comment wording — the knowledge one described itself as "symmetric to skills".

To be clear, I don't think skills has the same bug: core skills are baked into the image and `buildSkillBundle()` packages only global + personal, so an **empty skills bundle is a normal state** and wiping on it would destroy image-provided content. An empty *knowledge* bundle really does mean unbound. The two are legitimately different.

The problem is that the PR's stated reasoning ("a successful empty fetch is authoritative") is broader than the case it applies to, and nothing in the file records the distinction. After this change one file gives two opposite answers to the same question, and the next reader will most likely conclude skills was simply missed. A one-line comment on the skills guard explaining why it stays would settle it.

### 4. Nothing pins the assumption that replaced the guard

With the guard gone, data safety rests entirely on "a failed fetch never reaches materialize". I couldn't find a test asserting that — not in `sync-handlers.test.ts`, `http-server.test.ts`, or `resource-sync.test.ts`. Nothing fails if a future refactor of the `catch` in `http-server.ts` lets a failed fetch through. Trading a guard for an assumption is fine, but the assumption should be tested.

### 5. The new test covers the changed line, not the reported bug

The added test asserts that `materialize({repos: []})` wipes the directory — i.e. it restates the diff. The bug in the description is the end-to-end one (publish library A → unbind → republish → a new session must not see A), and that's the checkbox still unticked in the test plan. Worth automating, since it's the behavior the PR exists for.

### 6. A data-destruction contract changed with no doc update

Per CLAUDE.md: "After changes: run `npm test`, update docs if behavior changed", and design docs record "contracts and rationale". "Does an empty bundle delete on-disk state" is exactly that kind of contract, and it's irreversible. Nothing in `docs/design/` reflects the new semantics.

## Suggested shape

1. Gate the wipe by spawner mode (same treatment as skills/MCP), or scope `knowledgeDir` per user in local mode. This one is the merge blocker.
2. Make the distinction explicit on the wire instead of overloading `repos: []`. `KnowledgeBundlePayload` currently can't express "unbound" separately from "empty", which is what forces the consumer-side guessing. An explicit `unbound: true` (or a nullable bundle) makes both call paths safe and retires the "failures can't reach here" argument entirely.
3. Add the comment on the skills guard, and a test pinning the fetch-failure boundary.

Happy to be wrong on #2 if there's a Gateway-side guarantee I missed — pointer welcome.

LikiosSedo:
Addressed the local-mode isolation concern in a048eaf6.

- LocalSpawner now assigns each agent an isolated `.siclaw/knowledge/<agentId>/` target.
- Initial sync and HTTP reload both use a per-box knowledge handler bound to that target.
- Parent sessions and spawned sub-agents read from the same agent-scoped directory; K8s keeps the existing pod-local default.
- Added regressions proving an empty bundle for agent B cannot wipe agent A, and that thrown fetch failures never call `materialize`.
- Updated the resource-sync invariant documentation.

Local verification: `npx tsc --noEmit`; focused 159 tests; full `npm test` (5117 passed, 2 skipped).

jacoblee-io (commented):
Re-reviewed `a048eaf6`. The blocker is genuinely fixed, and the shape is better than what I suggested — factoring `createKnowledgeHandler({knowledgeDir})` and threading the target through `session.ts` / `agent-factory.ts` (including spawned sub-agents) is the right depth, and putting the skills-vs-knowledge difference into the local-mode safety table in `invariants.md` beats the inline comment I proposed. Nice touch using `resolveUnderDir` for the agent segment rather than a bare `path.join`.

Status of my earlier points: (1) fixed; (3) fixed via the invariants table; (4) fixed by `expect(handler.materialize).not.toHaveBeenCalled()`; (6) fixed. (5) is improved — the new LocalSpawner test spins up two real boxes and runs a real `materialize`, even if it's still not end-to-end. (2) is discussed at the bottom.

Two new things, both edges introduced by this change. Neither is large.

## A. An empty `agentId` silently degrades back to the shared directory

`resolveUnderDir`'s guard is:

```js
if (resolved !== base && !resolved.startsWith(base + path.sep)) throw
```

`path.resolve(base, "")` is exactly `base`, so `resolved !== base` is false, nothing throws, and it returns the base unchanged. With `agentId === ""`, `sessionManager.knowledgeDir` becomes the knowledge **root** — isolation is gone and the original cross-agent wipe is back, silently.

Traversal (`"../.."`) is correctly rejected; degeneration isn't. A non-empty assertion at the call site in `local-spawner.ts` is probably the safest fix — changing `resolveUnderDir`'s semantics would affect its other callers.

## B. The directories are isolated now, but the status slot still isn't — and #491 reads it

`lastKnowledgeSyncStatus` (`sync-handlers.ts:342`) is still module-level, written from both materialize paths (`:390` for the empty-bundle case, `:486` for the normal one). Every per-box handler instance now writes that same single slot:

- agent A materializes → `{targetDir: …/a1, repos: [A]}`
- agent B materializes → overwrites with `{targetDir: …/a2, repos: [B]}`

The struct already carries `targetDir`, which acknowledges that different handlers point at different targets — but there's still only one global slot holding it.

This will collide with #491: its `GET /api/sync-status` reads exactly this variable (via `readBoxSyncStatus`). With both merged, a developer console querying agent B's box in local mode can be shown agent A's library list — disks isolated, reported numbers still crossed.

The fix follows the same move this PR already made: pull the variable into the `createKnowledgeHandler` closure and expose `getLastKnowledgeSyncStatus()` as a method on the handler. The factory is already there, so this is mostly mechanical — and worth doing before #491 lands rather than after.

## On (2): the reasoning is accurate now, the gap is smaller, and I'd split it out

The comment is correctly narrowed to *thrown* fetch failures, and the new test pins that. What remains is the successful-but-empty fetch: `resource-sync.ts` retries only on a throw, and the Portal-side bundle handler returns `{repos: []}` whenever the `agent_knowledge_repos` query yields zero rows, which it can't distinguish from "the binding rows aren't visible yet".

I want to be fair about the severity, though — isolation dropped it a long way:

- a mistaken wipe now costs **one agent's** knowledge, not every local user's
- and on an agent's first spawn the directory is empty anyway, so there's nothing to lose. To actually hit it you need "this agent had knowledge, and a later spawn happened to fetch an empty bundle"

So I no longer think it blocks this PR. It's a design debt worth its own issue: `KnowledgeBundlePayload` can't express "unbound" separately from "empty" on the wire. An explicit `unbound: true` would retire the whole class of "can a failure reach materialize?" arguments instead of re-litigating it per call path.

## Verdict

Fine to merge once A and B are folded in. A is one assertion; B is moving one variable into the closure you already created — and skipping it means #491 ships with crossed reporting. Happy to see (2) tracked separately.

jacoblee-io (commented):
Took another pass at `8921eb0f`, this time reading the PR as a whole rather than checking off my earlier comments — which surfaced one thing I'd walked past three times.

## 1. A single local box creates *two* knowledge handler instances, so the new status isolation only covers half the story

`8921eb0f` moved `lastKnowledgeSyncStatus` into the `createKnowledgeHandler` closure, which is the right move. But one local box ends up with two instances of that closure:

`local-spawner.ts`, for the initial sync:

```js
const knowledgeHandler = createKnowledgeHandler({ knowledgeDir: sessionManager.knowledgeDir });
...
await syncResource("knowledge", gatewayClient.toClientLike(), knowledgeHandler);
```

`http-server.ts`, for HTTP reloads — and `createHttpServer` is called by LocalSpawner itself:

```js
if (sessionManager.knowledgeDir) {
  perServerHandlers.knowledge = createKnowledgeHandler({
    knowledgeDir: sessionManager.knowledgeDir,
  });
}
```

Both point at the same directory, each with its own `lastKnowledgeSyncStatus`. The initial sync records into instance #1; every subsequent reload records into instance #2. Whoever reads the status sees only half the history.

So the granularity that actually shipped is per-handler-instance, not the per-box that the doc comment and the table row describe. The new `sync-handlers.test.ts` case verifies isolation using two handlers on two *different* directories, which is exactly the configuration that doesn't exercise this — hence green.

Nothing reads the status inside this PR, so there's no symptom today. I'd still fix it before merge, because `KnowledgeSyncStatus.targetDir` strongly implies "status follows the target", and the next person to wire something to it will believe that. Building the handler once in LocalSpawner and passing it into `createHttpServer` (or keeping only the http-server one) makes the code match the comment.

Narrow but real side effect of having two: the staging directory is named `.sync-staging-${Date.now()}-${process.pid}`, and in local mode every box shares one pid. Cross-agent collisions are now prevented by the directory split, but the two entrypoints for the *same* agent — the fire-and-forget initial sync racing an immediately-arriving reload — land in the same directory with the same pid, separated only by `Date.now()`. Two independent instances make that race easier to hit, not harder.

## 2. The `http-server.test.ts` mock no longer satisfies the interface

```js
createKnowledgeHandler: () => ({ type: "knowledge", fetch: async () => ({ repos: [] }), materialize: async () => 0 }),
```

`KnowledgeSyncHandler` now requires `getLastKnowledgeSyncStatus()`; this mock doesn't have it. `vi.mock` factories aren't type-checked, so this compiles and passes. Harmless right now because nothing in this PR calls it — but the mock has quietly decoupled from the real signature, so the first call site that does use it will TypeError in production while the suite stays green.

## 3. Still open: an empty `agentId` degrades isolation back to the shared root

`resolveUnderDir`'s guard:

```js
if (resolved !== base && !resolved.startsWith(base + path.sep)) throw
```

`path.resolve(base, "")` equals `base`, so the first clause is false, nothing throws, and the base is returned unchanged. With `agentId === ""`, `knowledgeDir` becomes the shared knowledge root and the wipe this PR is about is cross-agent again. Traversal is rejected; degeneration isn't.

A non-empty assertion at the LocalSpawner call site is safer than changing `resolveUnderDir`'s semantics, since that helper has other callers.

## 4. Two minor ones

- `local-spawner.test.ts` replaces the whole config module with `loadConfig: () => ({ paths: { knowledgeDir: ".siclaw/knowledge" } })`. It works today only because LocalSpawner's other dependencies are mocked too; the first time it needs `isMemoryEnabled` or another `paths` field, the failure will point somewhere unrelated to the cause.
- `.siclaw/knowledge/<agentId>/` has no cleanup path, so directories outlive deleted agents in a long-running local deployment. The credentials directory follows the same pattern, so this is consistent with existing behaviour rather than new — low priority.

## Things I re-checked and found correct

- **K8s is untouched.** `sessionManager.knowledgeDir` is unset there, so no per-box handler is built and both the reload route and `getSyncHandler`-driven initial sync use the global `createKnowledgeHandler()` reading the config path. Pod filesystems are already isolated.
- **`readAllowedDirs` follows correctly** — the resolved `knowledgeDir` in `agent-factory.ts` is also what gates the Read tool, so one agent can't read another's knowledge directory.
- **The wiki catalog follows too** — `buildKnowledgeWikiCatalog(knowledgeDir ?? …)` receives the same resolved path.
- **`postReload` can't cross boxes** — the HTTP path passes that box's own sessions, and `resource-sync.ts` passes `{}` during initial sync so it returns early on an empty `context.sessions`.

(1) is the one I'd act on, less because it breaks today and more because it makes a freshly-written invariant untrue. (2) and (3) are one or two lines each.

jacoblee-io (approved):
Approving. The bug this set out to fix is fixed, the risk it introduced along the way is fixed with tests holding it in place, and nothing left on my list is something this PR should carry.

## Why this is done

- **The original defect is addressed properly.** A successful empty bundle is now authoritative, so a delete-and-republish no longer leaves the previous library visible.
- **The isolation regression is closed at the right depth.** `createKnowledgeHandler({ knowledgeDir })`, the target threaded through `session.ts` → `agent-factory.ts` into spawned sub-agents, and `91acb8d1` making the initial sync and HTTP reloads share one per-box instance. Two layers of regression tests hold it: agent B's empty bundle can't touch agent A at the handler level, and the LocalSpawner test asserts the same handler *reference* reaches `createHttpServer`.
- **The contract is written down.** `invariants.md` §1.2 gained the agent-scoped rule, the safety table gained a `knowledgeHandler.materialize()` row next to the skills one, and the Source line points at the files that actually implement it.
- CI is green across Test / Type Check / Portal Web Test.

I also re-walked the entrypoints that can trigger a knowledge materialize, and they're all correctly routed: `syncAllResources` (K8s, via `agentbox-main`) uses the global handler against the pod-local config path; LocalSpawner passes its per-box handler to both `syncResource` and `createHttpServer`; the reload route falls back to `getSyncHandler` only when `sessionManager.knowledgeDir` is unset, which is exactly the K8s case. No path is left pointing at the shared root.

## Retracting one of my earlier points

I raised the empty-`agentId` degeneration in `resolveUnderDir` twice. I've now traced it and **it isn't reachable from this PR** — all three `spawner.spawn()` call sites in `manager.ts` funnel through `getOrCreateWithDisposition`, which starts with:

```js
if (!agentId) throw new Error("AgentBoxManager.getOrCreate requires an agentId");
```

`spawnInstances` and `getOrCreateLocal` are both private and only reached from that path. So nothing to do here — sorry for asking twice before checking.

The underlying gap is still real, but it belongs to the helper, not to you: `resolveUnderDir`'s contract says segments are untrusted and the result is "guaranteed to be under base", yet `path.resolve(base, "")` returns `base` itself and passes the guard. Its riskiest existing caller is `sync-handlers.ts` skill cleanup — `fs.rmSync(resolveUnderDir(resolvedDir, skill.dirName), { recursive: true, force: true })` — where an empty `dirName` would target the whole skills directory. I'll open a separate issue for the helper rather than widen this PR into a change that touches `credential-manager.ts` too.

## Worth separate issues, not this PR

- **`repos: []` can't distinguish "unbound" from "the bindings weren't visible yet".** The Portal handler returns an empty bundle whenever `agent_knowledge_repos` yields zero rows, and `resource-sync.ts` retries only on a throw. After isolation this costs one agent's knowledge rather than everyone's, and knowledge is derived data — the next successful sync rebuilds it from the Portal — so it's a temporary degradation, not data loss. An explicit `unbound: true` on the wire would retire the whole "can a failure reach materialize?" question instead of re-arguing it per call path.
- **No automated unbind → republish → fresh-session coverage.** The new LocalSpawner test moved the level up meaningfully; the full path is still manual (the unticked box in your test plan).

Merge-order note: this one should land first — #490 has already adopted this PR's `agent-factory.ts` shape, and #491 needs the per-box handler in place before its `readBoxSyncStatus` can report the right directory.

PR #490, opened 2026-08-18
feat: cite original sources in knowledge answers

LikiosSedo:
## Summary

Knowledge-backed answers now append trusted original-source links only when the agent actually adopts pages it read. The capability follows `read_files`, is independent of Agent Type, and applies to web/API/A2A and Feishu delivery.

### Problem

Compiled knowledge did not carry a safe end-to-end contract for distinguishing pages merely inspected from evidence used in an answer. Asking the model to print URLs would allow invented or stale citations.

### Solution

- Materialize server-owned citation metadata beside the exact knowledge bundle.
- Track successful per-turn knowledge reads and register only adopted pages through `knowledge_cite`.
- Resolve OKF source references against trusted metadata, deduplicate and cap links, then append them at response boundaries.
- Include `knowledge_cite` in the `read_files` capability group, with no Agent Type gate.

Existing published bundles must be recompiled and published once because older frozen manifests contain no origin URL.

## Test Plan

- [x] `npm run build`
- [x] `npm test -- --reporter=dot` (247 files; 5121 passed, 2 skipped)
- [ ] Manual end-to-end verification after the paired server change is deployed

## Architecture Checklist

- [x] Deployment mode: citation metadata is materialized through the existing knowledge sync path.
- [x] Security model: URLs are server-owned, HTTPS-only metadata; the model cannot supply arbitrary URLs.
- [x] Database parity: no schema changes.
- [x] Tool protocol: registered through the Tool Registry with a TypeBox schema.

## General Checklist

- [x] Changes are focused and backward compatible.
- [x] Tests cover read/adopt validation, rendering, sync, tool capabilities, SSE, and Feishu output.
- [x] No new dependency was added.

### DIFF
--- a/portal-web/src/components/CapabilityGroupSelector.test.tsx
+++ b/portal-web/src/components/CapabilityGroupSelector.test.tsx
@@ -38,16 +38,16 @@ describe("CapabilityGroupSelector — render contract", () => {
     const html = render(new Set(["read_files"]))
     expect(html).toContain("Restricted")
     expect(html).not.toContain("Unrestricted")
-    // read_files grants 4 tools; "1 group" must be singular.
-    expect(html).toContain("1 group · 4 tools")
+    // read_files grants 5 tools; "1 group" must be singular.
+    expect(html).toContain("1 group · 5 tools")
     expect(html).toContain("Capability groups (1 / 11)")
     expect(countChecked(html)).toBe(1)
   })
 
   it("pluralizes groups and sums the deduped tool count for a multi-group selection", () => {
     const html = render(new Set(["read_files", "run_commands"]))
-    // 4 + 4 distinct tools, plural "groups".
-    expect(html).toContain("2 groups · 8 tools")
+    // 5 + 4 distinct tools, plural "groups".
+    expect(html).toContain("2 groups · 9 tools")
     expect(html).toContain("Capability groups (2 / 11)")
     expect(countChecked(html)).toBe(2)
   })
--- a/portal-web/src/lib/toolCapabilities.test.ts
+++ b/portal-web/src/lib/toolCapabilities.test.ts
@@ -101,17 +101,17 @@ describe("countToolsForSelection", () => {
   })
 
   it("counts a single group's tools", () => {
-    expect(countToolsForSelection(new Set(["read_files"]))).toBe(4)
+    expect(countToolsForSelection(new Set(["read_files"]))).toBe(5)
     expect(countToolsForSelection(new Set(["scheduling"]))).toBe(1)
   })
 
   it("counts the deduped union across multiple groups", () => {
-    // read_files (4) + search_memory (2), no shared tools → 6 distinct.
-    expect(countToolsForSelection(new Set(["read_files", "search_memory"]))).toBe(6)
+    // read_files (5) + search_memory (2), no shared tools → 7 distinct.
+    expect(countToolsForSelection(new Set(["read_files", "search_memory"]))).toBe(7)
   })
 
   it("ignores unknown keys in the selection", () => {
-    expect(countToolsForSelection(new Set(["read_files", "ghost"]))).toBe(4)
+    expect(countToolsForSelection(new Set(["read_files", "ghost"]))).toBe(5)
     expect(countToolsForSelection(new Set(["ghost"]))).toBe(0)
   })
 
--- a/portal-web/src/lib/toolCapabilities.ts
+++ b/portal-web/src/lib/toolCapabilities.ts
@@ -18,7 +18,7 @@ export interface CapabilityGroup {
 }
 
 export const CAPABILITY_GROUPS: CapabilityGroup[] = [
-  { key: "read_files", name: "Read files", description: "Read & search files and knowledge pages", tools: ["read", "grep", "find", "ls"] },
+  { key: "read_files", name: "Read files", description: "Read & search files and knowledge pages", tools: ["read", "grep", "find", "ls", "knowledge_cite"] },
   { key: "write_sandbox", name: "Write & author skills", description: "Write/edit scratch files and author skills (sandboxed to user-data)", tools: ["write", "edit", "skill_preview"] },
   { key: "inspect_infra", name: "Inspect infrastructure", description: "Read-only discovery of bound clusters and hosts", tools: ["cluster_list", "host_list"] },
   { key: "run_commands", name: "Run commands", description: "Execute whitelisted shell commands (kubectl read-only)", tools: ["bash", "node_exec", "pod_exec", "host_exec"] },
--- a/src/agentbox/sync-handlers.test.ts
+++ b/src/agentbox/sync-handlers.test.ts
@@ -1087,6 +1087,20 @@ describe("knowledgeHandler multi-repo identity", () => {
     expect(index).toContain(`[[repos/${knowledgeRepoDirName("平台知识", "repo-b")}/index]] - 平台知识 v1\n`);
   });
 
+  it("materializes server-owned citation metadata beside the exact repo roots", async () => {
+    const repos = [
+      { id: "repo-a", name: "A", version: 1, sizeBytes: 10, dataBase64: packageBase64("a"),
+        citationSources: [{ resource: "feishu/a.md", title: "A 原文", url: "https://docs.feishu.cn/wiki/a" }] },
+      { id: "repo-b", name: "B", version: 1, sizeBytes: 10, dataBase64: packageBase64("b"), citationSources: [] },
+    ];
+    await knowledgeHandler.materialize({ version: "v1", repos });
+    const manifest = JSON.parse(fs.readFileSync(path.join(knowledgeTmpDir, ".citation-manifest.json"), "utf8"));
+    expect(manifest).toEqual({ version: 1, repos: [
+      { id: "repo-a", root: `repos/${knowledgeRepoDirName("A", "repo-a")}`, sources: repos[0].citationSources },
+      { id: "repo-b", root: `repos/${knowledgeRepoDirName("B", "repo-b")}`, sources: [] },
+    ] });
+  });
+
   it("says the entries are libraries, since the prompt around it says pages", async () => {
     // The system prompt introduces this file as a page catalog — true when one
     // library unpacks at the root, false here. An agent reading these ten lines
--- a/src/agentbox/sync-handlers.ts
+++ b/src/agentbox/sync-handlers.ts
@@ -281,6 +281,7 @@ interface KnowledgeBundlePayload {
     sha256?: string | null;
     sizeBytes: number;
     fileCount?: number | null;
+    citationSources?: Array<{ resource: string; title: string; url: string }>;
     dataBase64: string;
   }>;
 }
@@ -389,6 +390,11 @@ export function createKnowledgeHandler(
     const stagingDir = path.join(knowledgeDir, `.sync-staging-${Date.now()}-${process.pid}`);
     fs.mkdirSync(stagingDir, { recursive: true });
     const syncedRepos: KnowledgeSyncStatus["repos"] = [];
+    const citationRepos: Array<{
+      id: string;
+      root: string;
+      sources: Array<{ resource: string; title: string; url: string }>;
+    }> = [];
 
     try {
       if (repos.length === 1) {
@@ -404,6 +410,7 @@ export function createKnowledgeHandler(
         }
         syncedRepos.push({ id: repos[0].id, name: repos[0].name, version: repos[0].version,
           sha256: info.sha256, expectedSha256: repos[0].sha256 ?? null, fileCount: info.fileCount, sizeBytes: repos[0].sizeBytes });
+        citationRepos.push({ id: repos[0].id, root: "", sources: repos[0].citationSources ?? [] });
       } else {
         const repoRoot = path.join(stagingDir, "repos");
         fs.mkdirSync(repoRoot, { recursive: true });
@@ -445,6 +452,7 @@ export function createKnowledgeHandler(
           }
           syncedRepos.push({ id: repo.id, name: repo.name, version: repo.version,
             sha256: info.sha256, expectedSha256: repo.sha256 ?? null, fileCount: info.fileCount, sizeBytes: repo.sizeBytes });
+          citationRepos.push({ id: repo.id, root: `repos/${dirName}`, sources: repo.citationSources ?? [] });
           // This line is the whole of what the agent knows about a library
           // before deciding to open it: the catalog goes into the system prompt,
           // there is no search tool, and everything else costs a Read of that
@@ -476,6 +484,8 @@ export function createKnowledgeHandler(
 
       fs.writeFileSync(path.join(stagingDir, ".sync-manifest.json"),
         JSON.stringify({ syncedAt, version: payload.version ?? "1", repos: syncedRepos }, null, 2) + "\n");
+      fs.writeFileSync(path.join(stagingDir, ".citation-manifest.json"),
+        JSON.stringify({ version: 1, repos: citationRepos }, null, 2) + "\n");
       await replaceDirectoryContentsFromStaging(knowledgeDir, stagingDir);
       lastKnowledgeSyncStatus = { syncedAt, targetDir: knowledgeDir, repoCount: syncedRepos.length, repos: syncedRepos };
       return repos.length;
--- a/src/core/agent-factory.test.ts
+++ b/src/core/agent-factory.test.ts
@@ -1,3 +1,5 @@
+import fs from "node:fs";
+import path from "node:path";
 import { describe, it, expect } from "vitest";
 
 // NOTE: We deliberately do NOT import "./agent-factory.js" here.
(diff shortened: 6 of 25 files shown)


jacoblee-io (commented):
I spent most of this review trying to break the trust boundary, and it holds: the model passes **page paths**, never URLs; `knowledge_cite` refuses any page not in `readPages` for the current turn; `noteRead` only accepts `.md` files resolved under the knowledge root; and the URL itself comes from `.citation-manifest.json` via a resource lookup. `normalizeKnowledgeSourceCitations` then enforces `https:`, dedupes by `href`, strips newlines from titles and caps at 3. The `javascript:alert(1)` test is the right test to have written. So the central claim in the description — the model cannot supply arbitrary URLs — is accurate.

What follows is everything I found around that boundary. The first two are the ones I'd fix before merge.

## 1. The tool is registered without a citation manifest, while the prompt that explains it isn't

Two gates guard the same feature and they don't agree.

Prompt side:

```js
if (knowledgeCitationsEnabled && knowledgeDir && hasKnowledgeCitationManifest(knowledgeDir))
```

Tool side:

```js
const citationSupport = opts?.sessionEventEmitter ? createKnowledgeCitationSupport({...}) : undefined;
...
if (citationSupport &&
    (!Array.isArray(allowedTools) || allowedTools.includes("read")) && ...) {
  customTools.push(citationSupport.tool);
}
```

The tool registration never checks `hasKnowledgeCitationManifest`. So on a mount with no manifest the model is handed a `knowledge_cite` tool whose description promises "the runtime validates their published original-source metadata and appends trusted links", with **no** accompanying system-prompt guidance, and every call it makes returns "No trusted original-source metadata is available for this knowledge mount."

This matters more than it looks because of your own note: *"Existing published bundles must be recompiled and published once because older frozen manifests contain no origin URL."* Until every bundle is recompiled, "no manifest" is the **normal** state — so during the whole rollout, every agent with `read` gets a tool that can only fail, and burns a round-trip discovering that.

You already reasoned about exactly this failure shape for the missing-sink case ("otherwise the tool would promise that links are appended while CLI/child sessions silently drop the event"). Manifest-absent deserves the same treatment: add `hasKnowledgeCitationManifest(knowledgeDir)` to the registration condition so both gates agree.

## 2. This collides with #489 on the same line in `agent-factory.ts`, and the wrong merge order silently reverts its isolation fix

Both PRs rewrite the `knowledgeDir` declaration in `createSiclawSession`.

\#489 (`a048eaf6`) changes it to accept an explicit override, which is what makes per-agent isolation work in local mode:

```js
const knowledgeDir = opts?.knowledgeDir
  ?? (opts?.portalKnowledgeDir && fs.existsSync(opts.portalKnowledgeDir)
    ? opts.portalKnowledgeDir
    : path.resolve(cwd, config.paths.knowledgeDir));
```

This PR moves the same declaration earlier (so `citationSupport` can use it) but keeps the **pre-#489** form, with no `opts?.knowledgeDir`:

```js
const knowledgeDir = opts?.portalKnowledgeDir && fs.existsSync(opts.portalKnowledgeDir)
  ? opts.portalKnowledgeDir
  : path.resolve(cwd, config.paths.knowledgeDir);
```

Git will flag this as a conflict, but the resolution is easy to get wrong in the direction that compiles and passes tests while dropping `opts?.knowledgeDir` — at which point LocalSpawner's agent-scoped directory is ignored and one agent's empty bundle wipes another's knowledge again. Worth calling out in whichever PR merges second, and worth a test that asserts `opts.knowledgeDir` wins, so a bad resolution goes red.

Related, same interaction: this PR's `citationSupport` is built from the *pre-override* `knowledgeDir`, so once #489 lands, citation lookups in local mode would read the shared root while `Read` serves the agent-scoped path — `noteRead`'s root check would then reject every page and citations would silently never fire. Threading the resolved directory through fixes both.

## 3. The URL isn't escaped for the Markdown context it's interpolated into

```js
const lines = citations.map((source) => `- [${source.title.replace(/[\[\]]/g, "")}](${source.url})`);
```

The title is escaped for `[` and `]`; the URL is not escaped at all. `(`, `)`, `[` and `]` are legal in a URL path and `new URL()` does not percent-encode them, so a URL containing `)` closes the link early and the remainder becomes document text — including a second, attacker-chosen link:

```
url = https://ok.example/)[Login](https://evil.example/
→ - [Title](https://ok.example/)[Login](https://evil.example/)
→ renders as two links, the second one arbitrary
```

This is not reachable by the model — it requires influence over the manifest's `url`, which is server-owned. But the manifest is generated from compiled knowledge, and origin URLs ultimately derive from scraped/imported page metadata, so I wouldn't treat it as fully trusted input either. It's cheap to close: percent-encode `(` and `)`, or use the angle-bracket destination form `[title](<url>)`, which Markdown defines for exactly this. Given these links land in Feishu cards that users click, defense in depth seems worth the two lines.

## 4. `pageResources` reads the page with no `try`/`catch`, unlike `readManifest`

`readManifest` wraps its `readFileSync` + `JSON.parse`. `pageResources` doesn't:

```js
const body = fs.readFileSync(pagePath, "utf8");
```

A page recorded by `noteRead` can be gone by the time `knowledge_cite` runs — a knowledge reload wipes and re-materializes the directory, which is precisely what #489 makes unconditional. Then `readFileSync` throws ENOENT out of `execute` and the whole tool call errors instead of degrading to "no trusted sources". `yaml.load` can also throw on malformed frontmatter. Same `try`/`catch` treatment as `readManifest` would keep a citation miss from taking down the turn.

## 5. `repos.find` treats `root: ""` as a wildcard and takes the first match

```js
const repo = manifest.repos.find((candidate) => {
  const root = candidate.root.replace(/\/$/, "");
  return root === "" || rel === root || rel.startsWith(root + "/");
});
```

An entry with an empty `root` matches every page, so in a multi-repo manifest whichever such entry appears first in the array swallows all lookups and pages get attributed to the wrong repo's sources — a wrong-but-plausible citation, which is the worst failure mode for this feature. Your own fixture uses `root: ""`, so this isn't hypothetical shape. Sorting candidates by root length and taking the most specific match would mirror the "most specific wins" rule the repo already uses for nested skills.

## 6. Smaller things

- **`resource` and `page` pass through unvalidated.** `normalizeKnowledgeSourceCitations` carefully validates `url` and sanitizes `title`, then copies `resource`/`page` on a bare `typeof === "string"` check — no newline stripping, no length cap. Nothing renders them today, but they travel in the same event payload, so any future consumer inherits an unsanitized field from a struct that otherwise looks sanitized.
- **The cap of 3 is hardcoded twice.** `MAX_KNOWLEDGE_CITATIONS = 3` in `knowledge-citations.ts`, and two bare `citations.length >= 3` breaks in `knowledge-citation-tool.ts`. Import the constant, or the two drift on the first change.
- **Two near-duplicate append sites.** `lark.ts` clears `pendingKnowledgeSources` on `model_route_start`/`model_route_rollback`; `sse-consumer.ts` clears it inside `discardRoutedAttempt()`. Both are correct — I checked — but in the `lark.ts` path the clear only happens on the successful-append branch, so a turn ending with `stopReason === "error"` or empty text leaves the pending sources set. Clearing unconditionally at the turn boundary would make the invariant local instead of depending on which route events fired.

## Overall

The security design is sound and the test coverage around read/adopt validation is real. (1) and (2) are what I'd want fixed before this merges — (2) especially, since it's cross-PR and fails silently in the direction of losing a security fix.

jacoblee-io (commented):
Re-reviewed `e9169a99`. Five of six are fixed, and on two of them your approach is better than what I asked for — one because you knew something about the session lifecycle that I didn't.

## Where you were right and I wasn't

**The gate mismatch.** I asked you to add `hasKnowledgeCitationManifest` to the tool-registration condition. You instead kept the tool always registered and made the *prompt* two-state, with this comment:

> This text is rebuilt by `session.reload()`, while custom tools remain attached to the session.

That's the part I missed. Custom tools are fixed at session creation while the system prompt is rebuilt on reload, so gating **registration** on the manifest would mean a warm session can never pick up citations after a manifest is first published — it'd need a full session rebuild. Your version lights up on the next reload. Adding "Use only when the current system prompt says knowledge source citations are available" to the tool description covers the gap in the other direction, and the manifest-appears test pins the transition. Better outcome; please disregard my original suggestion.

**The cross-PR collision.** Adopting #489's `opts?.knowledgeDir ?? …` form and feeding the resolved directory into `citationSupport` closes both halves of what I raised. And the structural guard in `agent-factory.test.ts` is exactly the "make a bad conflict resolution go red" test I wanted, with a clear note about why source-text matching is the only option available in that workspace. `\s*` in the pattern shows you thought about formatting drift.

## The rest

- **URL escaping** — `(`/`)` percent-encoded, with the WHATWG-URL root cause written down, and the test uses the exact payload from my comment with `expect(rendered.match(/\]\(/g)).toHaveLength(1)`. Confirmed the remaining literal `[`/`]` inside the destination are harmless: link-text parsing is already past by then and bracket characters don't terminate a destination.
- **`pageResources`** — wrapped, and the two degradation tests (page deleted, malformed frontmatter → `cited: 0`) are the right pair.
- **Repo matching** — `findCitationRepo` with longest-root-wins, and the fixture deliberately orders the `root: ""` fallback first so a regression to `.find()` fails. Good test design.
- **`MAX_KNOWLEDGE_CITATIONS`** — exported and threaded through `maxItems` and the description text, so the cap can't drift.

## Two small leftovers, plus one new note

- `resource` and `page` still pass through on a bare `typeof === "string"` check while `url` and `title` are validated and sanitized. Nothing renders them today; the risk is that a later consumer reasonably assumes the whole struct was sanitized because most of it was.
- The two append sites are still near-duplicates, and in `lark.ts` the clear still only happens on the successful-append branch — a turn ending with `stopReason === "error"` or empty text leaves `pendingKnowledgeSources` set. Clearing unconditionally at the turn boundary makes that invariant local instead of dependent on which route events fired.
- **New, minor:** the structural guard in `agent-factory.test.ts` is deliberately a stopgap, and it's format-sensitive by nature — a legitimate refactor (say, extracting `resolveKnowledgeDir(opts, cwd, config)`) would turn it red while the semantics stay correct. Worth a "retire once #489 lands" note on it, so the next person knows it's a merge-window guard rather than a permanent contract.

Nothing here blocks. The only real gate left is ordering: this and #489 touch the same `agent-factory.ts` line and now agree on its shape, so landing #489 first makes this a clean follow-on.

jacoblee-io (commented):
Another pass at `e9169a99`, this time reading the parts I'd never opened — the manifest producer, the capability layer, and the "why did this test need updating" files. Two new things, neither of which breaks today.

## 1. The manifest is written with no validation; all of it lives in the renderer

I'd only checked the consumer side before. The producer in `sync-handlers.ts`:

```js
citationRepos.push({ id: repos[0].id, root: "", sources: repos[0].citationSources ?? [] });
...
fs.writeFileSync(path.join(stagingDir, ".citation-manifest.json"),
  JSON.stringify({ version: 1, repos: citationRepos }, null, 2) + "\n");
```

`citationSources` comes straight off the Gateway payload and lands on disk verbatim — no protocol check on `url`, no newline stripping on `title`, no count bound, no shape check on `resource`.

So "URLs are server-owned" rests on the Gateway keeping them clean *plus* `normalizeKnowledgeSourceCitations` filtering to `https:` at render time. The defense in depth is real — a `javascript:` URL in the manifest never reaches a rendered link — but there is exactly one validation point and it sits at the very bottom of the pipeline.

The risk isn't today's code path, it's the next consumer. Anything that reads the manifest without going through `normalizeKnowledgeSourceCitations` — a developer console reading it directly, a URL echoed into `knowledge_cite`'s result text, an export — gets unfiltered data out of a file that looks like validated server-owned metadata. Running the same normalize at write time (or just reusing that function) is cheap and puts the boundary where the data enters the system.

**Correction to my previous comment while I'm here:** I claimed a `root: ""` entry would swallow lookups in a multi-repo manifest. Reading the producer, single-repo takes `root: ""` and multi-repo takes `repos/${dirName}` in mutually exclusive branches, so this code never emits a mixed manifest. Your longest-root-wins fix is still right — it guards hand-edited and future-shape manifests — but the reachable trigger is narrower than I described.

## 2. Registration keys off `includes("read")` while the capability system says `read_files`

`tool-capabilities.ts` now declares it as part of the group:

```js
read_files: ["read", "grep", "find", "ls", "knowledge_cite"],
```

But `agent-factory.ts` gates registration on something else:

```js
if (citationSupport &&
    (!Array.isArray(allowedTools) || allowedTools.includes("read")) && ...) {
  customTools.push(citationSupport.tool);
}
```

That tests for `"read"`, not `"knowledge_cite"`. The two are equivalent today because they're in the same capability group and always resolve together — but it means `allowedTools` no longer governs this tool. An agent whose resolved list does *not* contain `knowledge_cite` still gets the tool registered, as long as `read` is there.

That's out of step with how this repo treats that list elsewhere. From `session.ts`:

> A spawned sub-agent must never be broader than its parent: inherit the parent's per-agent tool whitelist. Without this, a restricted agent that has the `spawn_subagents` capability could escalate…

Two futures where it bites: wanting an agent that can read knowledge but must not append links to answers (removing `knowledge_cite` from its list would have no effect), and any future regrouping that separates `knowledge_cite` from `read` (the condition silently becomes wrong).

Changing the test to `allowedTools.includes("knowledge_cite")` lines it up with the capability system. The "intrinsic, side-effect-free companion to Read" reasoning in the comment explains why it originally tracked `read`, but once the tool is a declared member of a capability group, the group membership should be the gate.

## 3. Minor

`tool-capabilities-coverage.test.ts`'s stub now always supplies `knowledgeCitationTool` and `sessionEventEmitter`, so `available()` is unconditionally true inside that probe and it can't verify the real availability rule. The comment explains why the sentinel is needed for a static catalog walk, and `knowledge-cite.test.ts` covers the actual `available()` logic, so there's no gap — just noting that the coverage probe isn't the thing guarding it.

## Re-verified and correct

- **The manifest is written into `stagingDir` and promoted with it**, so the manifest and the pages it describes are always the same version. That's the easiest thing to get wrong in this feature and it's right.
- **An empty bundle's wipe removes the manifest too**, so `hasKnowledgeCitationManifest` flips to false and the prompt switches to its unavailable branch on its own. The unbind path is self-consistent end to end.
- **The capability mirror is in sync** — `src/core/tool-capabilities.ts` and `portal-web/src/lib/toolCapabilities.ts` were both updated, no one-sided drift.
- `local-spawner.test.ts`, `internal-api.test.ts` and `cli-snapshot-api.test.ts` only needed expectation updates from the capability-group change; I'd wondered whether this PR was also touching local mode, and it isn't.

Neither (1) nor (2) is a today-bug. (2) is the one I'd change — one line, and it decides whether `allowedTools` can still be read as an authorization list for this tool.

jacoblee-io (approved):
Approving. The rebase onto merged #489 is clean, the trust boundary holds under the specific attacks I tried, and the two items I still have are future-proofing rather than gates.

## The cross-PR risk I flagged first is now retired

The force-push onto merged #489 (`7b6d0560`) kept 489's form intact while moving the declaration up for `citationSupport`:

```js
const knowledgeDir = opts?.knowledgeDir
  ?? (opts?.portalKnowledgeDir && fs.existsSync(opts.portalKnowledgeDir)
    ? opts.portalKnowledgeDir
    : path.resolve(cwd, config.paths.knowledgeDir));
```

This was the resolution I was worried would silently drop `opts?.knowledgeDir` and re-open 489's cross-agent wipe. The structural guard you added in `agent-factory.test.ts` is what would have caught it, and it's exactly why that test was worth writing.

## Why I'm satisfied with the security model

Recording the reasoning so a future reader can re-check it rather than re-derive it:

- The model passes **page paths**, never URLs. `knowledge_cite` rejects any page not in `readPages` for the current turn, `noteRead` only accepts `.md` files resolved under the knowledge root, and the URL is looked up from `.citation-manifest.json` by resource.
- The manifest is written into `stagingDir` and promoted with it, so the manifest and the pages it describes are always the same version.
- Rendering enforces `https:`, dedupes by `href`, strips newlines from titles, drops `[`/`]` from link text, percent-encodes `(`/`)` in the destination, and caps at 3. I tried the Markdown-injection payload `https://ok.example/)[Login](https://evil.example/` and it resolves to a single link.
- An empty bundle's wipe removes the manifest too, so `hasKnowledgeCitationManifest` flips false and the prompt switches to its "do not call" branch on its own.
- Degradation paths are covered: page deleted after `Read`, malformed frontmatter, no manifest at all — each yields `cited: 0` instead of a failed turn.

## Two items I'm explicitly not gating on

**1. `allowedTools.includes("read")` instead of `includes("knowledge_cite")`.** I traced this rather than leave it as a suspicion: `resolveCapabilities(groupKeys)` takes capability *group keys*, `agents.tool_capabilities` stores those keys as JSON, and the tool list is expanded at runtime. So `read_files` always yields `read` and `knowledge_cite` together and the two can't come apart today — an `allowedTools` containing `read` but not `knowledge_cite` isn't reachable. Worth aligning the condition with the capability group anyway, because the day someone splits the group or introduces a hand-written tool list, this reads as correct while being wrong. Not a blocker.

**2. The manifest is written with no validation.** `citationSources` lands on disk verbatim off the Gateway payload; the only validation in the pipeline is `normalizeKnowledgeSourceCitations` at render time. That's real defense in depth for today's single consumer. The exposure is the next consumer — anything reading the manifest directly, or echoing a URL into tool result text, inherits unfiltered data from a file that looks like validated server-owned metadata. Running the same normalize at write time would put the boundary where the data enters.

Both are the kind of thing I'd rather see in a follow-up than hold a 25-file PR for.

## Heads-up on #491

It will now fail to compile against `main`: `readBoxSyncStatus()` reads the module-scoped `lastKnowledgeSyncStatus`, which #489 turned into a per-handler closure variable. Its rebase needs to read the status off the box's own handler (and take that box's `knowledgeDir`), not the process-global accessor — otherwise local mode reports an empty inventory.

PR #491, opened 2026-08-18
Expose AgentBox observed inventory through sync-status

LikiosSedo:
## Summary
- AgentBox now serves `GET /api/sync-status` from the materialized knowledge manifest, skill directories, and MCP config.
- Runtime `agent.syncStatus` reads that list from a running box so Sicore can show what the test box actually loaded.

## Test plan
- [ ] `npx vitest run src/agentbox/sync-handlers.test.ts src/agentbox/http-server.test.ts src/gateway/server-sync-status.test.ts`
- [ ] With a running box, `GET /api/sync-status` returns the knowledge repos on disk
- [ ] Pair with the Sicore developer-console MR that displays this on test deploy

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/agentbox/http-server.test.ts
+++ b/src/agentbox/http-server.test.ts
@@ -294,9 +294,10 @@ describe("http-server — /health + /api/sessions + /api/models", () => {
     expect(r.data.sessions).toEqual([]);
   });
 
-  it("GET /api/sync-status reports the observed inventory", async () => {
+  it("GET /api/sync-status uses the box's agent-scoped knowledge directory", async () => {
     const root = fs.mkdtempSync(path.join(os.tmpdir(), "http-sync-status-"));
     const knowledgeDir = path.join(root, "knowledge");
+    const sharedKnowledgeDir = path.join(root, "shared-knowledge");
     const skillsDir = path.join(root, "skills");
     fs.mkdirSync(path.join(skillsDir, "resolved", "k8s-debug"), { recursive: true });
     fs.mkdirSync(knowledgeDir, { recursive: true });
@@ -304,10 +305,15 @@ describe("http-server — /health + /api/sessions + /api/models", () => {
       syncedAt: "2026-08-18T08:00:00.000Z",
       repos: [{ id: "kb-1", name: "hardware", version: 2, sha256: "abc", fileCount: 12 }],
     }));
-    mockConfigState.knowledgeDir = knowledgeDir;
+    mockConfigState.knowledgeDir = sharedKnowledgeDir;
     mockConfigState.skillsDir = skillsDir;
     mockConfigState.mcpServers = { incidents: { transport: "http" } };
     try {
+      await new Promise<void>((resolve) => (server as http.Server).close(() => resolve()));
+      (sm as any).knowledgeDir = knowledgeDir;
+      server = createHttpServer(sm as any);
+      port = await startServer(server);
+
       const r = await getJson(port, "/api/sync-status");
       expect(r.status).toBe(200);
       expect(r.data).toEqual({
--- a/src/agentbox/http-server.test.ts
+++ b/src/agentbox/http-server.test.ts
@@ -77,6 +77,11 @@ vi.mock("./sync-handlers.js", () => ({
     },
     postReload: async () => {},
   }),
+  readBoxSyncStatus: () => ({
+    knowledge: { syncedAt: null, repos: [] },
+    skills: { names: [] },
+    mcp: { names: [] },
+  }),
 }));
 
 vi.mock("./credential-broker.js", () => ({
@@ -284,6 +289,16 @@ describe("http-server — /health + /api/sessions + /api/models", () => {
     const r = await getJson(port, "/api/sessions");
     expect(r.status).toBe(200);
     expect(r.data.sessions).toEqual([]);
+  });
+
+  it("GET /api/sync-status reports the observed inventory", async () => {
+    const r = await getJson(port, "/api/sync-status");
+    expect(r.status).toBe(200);
+    expect(r.data).toEqual({
+      knowledge: { syncedAt: null, repos: [] },
+      skills: { names: [] },
+      mcp: { names: [] },
+    });
   });
 
   it("GET /api/internal/box-status reports drained when the box holds nothing", async () => {
--- a/src/agentbox/http-server.test.ts
+++ b/src/agentbox/http-server.test.ts
@@ -1,4 +1,7 @@
 import { describe, it, expect, beforeEach, afterEach, vi } from "vitest";
+import fs from "node:fs";
+import os from "node:os";
+import path from "node:path";
 import type { AddressInfo } from "node:net";
 import type http from "node:http";
 import type https from "node:https";
@@ -17,6 +20,9 @@ import type https from "node:https";
 const mockConfigState = vi.hoisted(() => ({
   modelRouting: undefined as unknown,
   memoryEnabled: true,
+  skillsDir: "skills",
+  knowledgeDir: "knowledge",
+  mcpServers: {} as Record<string, unknown>,
 }));
 
 // Silence metrics auth side effects.
@@ -43,8 +49,8 @@ vi.mock("../core/config.js", () => ({
   loadConfig: () => ({
     paths: {
       userDataDir: "/tmp/siclaw-test-user-data",
-      skillsDir: "skills",
-      knowledgeDir: "knowledge",
+      skillsDir: mockConfigState.skillsDir,
+      knowledgeDir: mockConfigState.knowledgeDir,
       credentialsDir: ".siclaw/credentials",
     },
     providers: {
@@ -53,12 +59,14 @@ vi.mock("../core/config.js", () => ({
       },
     },
     modelRouting: mockConfigState.modelRouting,
+    mcpServers: mockConfigState.mcpServers,
   }),
   isMemoryEnabled: () => mockConfigState.memoryEnabled,
 }));
 
 // Make sync-handlers a no-op registry.
-vi.mock("./sync-handlers.js", () => ({
+vi.mock("./sync-handlers.js", async (importOriginal) => ({
+  ...(await importOriginal<typeof import("./sync-handlers.js")>()),
   getSyncHandler: () => undefined,
   createClusterHandler: () => ({ type: "cluster", fetch: async () => 0, materialize: async (n: number) => n }),
   createHostHandler: () => ({ type: "host", fetch: async () => 0, materialize: async (n: number) => n }),
@@ -76,11 +84,6 @@ vi.mock("./sync-handlers.js", () => ({
       return target.allowedToolsState ? target.allowedToolsState.length : 0;
     },
     postReload: async () => {},
-  }),
-  readBoxSyncStatus: () => ({
-    knowledge: { syncedAt: null, repos: [] },
-    skills: { names: [] },
-    mcp: { names: [] },
   }),
 }));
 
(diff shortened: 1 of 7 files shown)


jacoblee-io (commented):
Clean idea and the split is right — AgentBox reports what it observed, Runtime just relays. I didn't find a correctness bug in the happy path. My concerns are almost all on the test side, where the mocking is loose enough that the parts most likely to break in production aren't covered.

One thing I checked and want to save the next reviewer the trouble: **the new endpoint is properly authenticated.** The global gate in `http-server.ts` (~1671) rejects any request without a client cert whose `OU` is `Gateway` or `Runtime`, for every route except `/health`, and local mode binds `127.0.0.1`. The missing `/api/internal/` prefix is a naming inconsistency with its sibling, not an access-control gap — nothing in the codebase keys off that prefix.

## 1. The mock drops the third constructor arg, which is the mTLS credential

Real signature:

```ts
constructor(endpoint: string, timeoutMs = 30000, tlsOptions?: AgentBoxTlsOptions)
```

Production call site: `new AgentBoxClient(box.endpoint, 8_000, agentBoxTlsOptions)`. The mock in `server-sync-status.test.ts` takes only `(endpoint)`.

`agentBoxTlsOptions` is exactly what the gate above validates. Drop it or pass it wrong and every call gets a 403 while the suite stays green. The failure mode is also mislabelled in that case: a 403 message contains no `404`, so `/\b404\b/.test(message)` is false and the RPC reports `query_failed` rather than `unsupported` — the console shows "query failed" when the real cause is a cert problem.

The deliberately-narrowed `8_000` timeout (vs. the 30s default) is likewise untested.

## 2. The cross-process contract is typed as `unknown[]`, so neither side breaks when it drifts

```ts
const status = await client.getJson<{
  knowledge?: { syncedAt?: string | null; repos?: unknown[] };
  ...
}>("/api/sync-status");
```

AgentBox defines `ObservedKnowledgeRepo` (`id`/`name`/`version`/`sha256`/`fileCount`); Gateway declares the same payload as `unknown[]` and forwards it. Box image and runtime ship independently, so renaming or dropping a field compiles fine on both sides and the console just renders blanks.

The new test demonstrates the hole: its mocked repo is `{ id, name, version }` with **no `sha256`**, which `ObservedKnowledgeRepo` requires, and the assertion passes anyway. Moving `ObservedKnowledgeRepo` into `src/shared/` and having both sides import it would weld this shut at compile time.

## 3. `boxes: 0` is wrong in the `no_running_box` branch

`targets` is "boxes for this agent that are running", but that branch hardcodes `boxes: 0` while every other branch returns `targets.length`. Same field, two meanings.

Concretely: an agent whose boxes are all still `pending` (image pull) reports `boxes: 0`, which reads as "no box exists" when the truth is "not up yet, check again shortly" — different action for whoever is looking at the console. The neighbouring `agent.reload` RPC returns `boxes: targets.length`; matching it would be enough.

## 4. With multiple boxes, the more informative failure reason is discarded

```ts
if (targets[targets.length - 1] === box) {
```

If box #1 fails with a 404 (old image) and box #2 with a network error, the result is `query_failed` and the "one of your boxes is running an old image" signal is lost — and which one wins depends on iteration order. Since the loop already sees every failure, collecting the reasons and picking by priority (`unsupported` over `query_failed`) would report the actionable one.

Minor, same line: this tests "is this the last element" by object identity. `index === targets.length - 1` says it directly and doesn't assume `list()` returns the same object references each call.

## 5. The new doc comment landed between box-status's comment and its route

The new `/** GET /api/sync-status … */` block is inserted immediately after the comment block that documents `/api/internal/box-status` (the one ending "…and only the box can see that."). The two blocks now sit back to back and box-status's documentation reads as if it belongs to the new route. Moving the new block above that comment fixes it.

Also in that block:

> `* Placement uses box-status (sessions / drain).`

This doesn't parse — looks like an unfinished edit. I'd guess the intent is "placed next to box-status, which covers sessions / drain", but worth rewriting to whatever you meant.

## 6. Each layer is mocked at its own boundary, so nothing exercises two together

`http-server.test.ts` mocks `readBoxSyncStatus` and then asserts the route returns that mock — it proves the route is wired and the JSON round-trips, but no part of the real projection runs. `server-sync-status.test.ts` mocks `AgentBoxClient`.

Credit where due: the three new cases in `sync-handlers.test.ts` do drive the **real** `readBoxSyncStatus` (memory-preferred, `.sync-manifest.json` fallback after restart, skill-directory listing) and they're good. What's missing is one case putting the real `readBoxSyncStatus` behind the real HTTP handler — the thing the PR actually ships.

Relatedly, all three boxes in the test plan are unticked, including the vitest run. `CONTRIBUTING.md` asks for `npm test` before submitting.

## 7. Two small ones

- `sync-handlers.ts`: `resetKnowledgeSyncStatusForTests` is exported from a production module. The file already has the `_mockKnowledgeDir` / `_mockSkillsDir` hook pattern for this; following it would keep the surface consistent.
- `readBoxSyncStatus` reads `lastKnowledgeSyncStatus`, a module-level variable, while local mode runs every box in one process — so querying agent B's box can return agent A's knowledge. This predates the PR (in local mode `knowledgeDir` is shared too), but this endpoint is what first surfaces it in a UI. Worth a comment noting the data is process-scoped, not agent-scoped, in local mode.

jacoblee-io (commented):
Re-reviewed `ad75ca73`. Five of my seven points are properly fixed, and two of them came out better than what I proposed:

- The client mock now takes all three constructor args and records them, with `expect(clientCalls).toEqual([{ endpoint, timeoutMs: 8_000, tlsOptions: server.agentBoxTlsOptions }])`. That pins the mTLS credential and the narrowed timeout, which was the gap I cared most about.
- `src/shared/agentbox-sync-status.ts` is exactly the fix, and using `satisfies BoxSyncStatus` on the fixture means future drift fails to compile rather than passing silently. You also caught the missing `sha256` in the mock.
- The `sawUnsupported ||= …` rewrite kills both the object-identity check and the lost-reason problem at once, with a test covering 404 + ETIMEDOUT together — and it's shorter than what it replaced.
- `http-server.test.ts` switching to `importOriginal()` with a real temp dir, a real `.sync-manifest.json` and a real `skills/resolved/` makes that the genuine cross-layer test the suite was missing.
- The doc comment is back above `box-status`, and the garbled sentence reads properly now.

Two things left. The second is new, and it's the one I'd act on.

## A. `boxes: 0` in the `no_running_box` branch

Still hardcoded. An agent whose boxes are all `pending` (image pull) reports `boxes: 0`, which reads as "no box exists" rather than "not up yet" — different action for whoever is reading the console. `agent.reload` returns `targets.length`; matching it is enough.

## B. The new doc comment contradicts #489, and merging both yields an empty inventory in local mode

This revision added:

> LocalSpawner **intentionally shares one process and filesystem** across its logical boxes, so local-mode inventory is process-scoped.

But #489 (`a048eaf6`) is in the middle of removing that premise for knowledge specifically, and has already written the opposite into `docs/design/invariants.md`:

> Knowledge sync in local mode must materialize and read from an agent-scoped directory (`.siclaw/knowledge/<agentId>/`).

So the two open PRs now assert opposite things about the same fact. Once both land this isn't just a stale comment — it's a functional break in the feature this PR exists for:

- `readBoxSyncStatus()` resolves `path.resolve(process.cwd(), config.paths.knowledgeDir)` — the **shared root**.
- After #489, the manifest is written to `<root>/<agentId>/.sync-manifest.json`.
- `readKnowledgeInventoryFromDisk(root)` therefore looks for `<root>/.sync-manifest.json`, doesn't find it, and returns `{ syncedAt: null, repos: [] }`.

Net effect in local mode: while `lastKnowledgeSyncStatus` is warm the endpoint reports whichever agent materialized last (cross-talk), and after a process restart it reports an empty inventory. Both are wrong for an endpoint whose whole purpose is "show me what this box actually loaded after a test deploy".

The fix rides on the machinery #489 already built:

1. Have `readBoxSyncStatus()` accept a `knowledgeDir` (the route already has `sessionManager.knowledgeDir` in scope — it's what gates `createKnowledgeHandler` a few lines up).
2. Move `lastKnowledgeSyncStatus` into the `createKnowledgeHandler` closure and read it off the handler, so per-box status follows per-box materialization instead of a single module-level slot.

Skills aren't affected — local mode deliberately skips skills materialization — so knowledge is the only axis that needs threading.

Worth coordinating across the two PRs regardless of merge order: whichever lands second inherits the mismatch, and it fails quietly (an empty list looks like "nothing is bound" rather than an error).

jacoblee-io:
Correction to my previous comment — I was reading #489 at `a048eaf6` and missed that `8921eb0f` (pushed ~30 min earlier) had already done half of what I asked for. Sorry for the noise.

What's already done in #489, so please ignore step 2 of my last comment:

```ts
export interface KnowledgeSyncHandler extends AgentBoxSyncHandler<KnowledgeBundlePayload> {
  getLastKnowledgeSyncStatus(): KnowledgeSyncStatus | null;
}

export function createKnowledgeHandler(options): KnowledgeSyncHandler {
  let lastKnowledgeSyncStatus: KnowledgeSyncStatus | null = null;   // now per-handler
```

`lastKnowledgeSyncStatus` is already a closure variable with a per-handler accessor, plus a module-level `getLastKnowledgeSyncStatus()` kept for backwards compatibility that delegates to the default global handler.

Two consequences for this PR, and the second is the one that matters:

**This PR will fail to compile once #489 lands, which is the good outcome.** `readBoxSyncStatus()` reads the module-scoped `lastKnowledgeSyncStatus` directly, and that binding no longer exists at module scope. Git will also conflict in that region, so whoever resolves it can't miss it — unlike the `agent-factory.ts` collision with #490, which compiles fine either way.

**Switching to the compatibility accessor is not sufficient.** `getLastKnowledgeSyncStatus()` delegates to the *default global* handler, but in local mode every box uses its own `createKnowledgeHandler({ knowledgeDir })` instance. The default instance therefore never materializes anything, returns `null`, and `readBoxSyncStatus()` falls through to the disk path — which resolves the shared root while the manifest now lives at `<root>/<agentId>/.sync-manifest.json`. Result is the empty inventory I described before.

So the remaining work here is narrower than I said: thread the box's own handler (and its `knowledgeDir`) into `readBoxSyncStatus()` rather than reaching for module state. `http-server.ts` already has both in scope at the route — `sessionManager.knowledgeDir` is what gates `createKnowledgeHandler` a few lines above the `/api/sync-status` registration, and `perServerHandlers.knowledge` is that instance.

jacoblee-io (approved):
Approving. The rebase onto merged #489 handled the breakage I predicted, and `3372bff8` closes the last two items.

## The #489 interaction is resolved, both halves of it

`readBoxSyncStatus` now takes the box's directory *and* its handler, with the reason stated in the comment:

> boxes must pass their agent-scoped handler and directory to avoid cross-talk

```js
const knowledgeDir = options.knowledgeDir
  ? path.resolve(options.knowledgeDir)
  : path.resolve(process.cwd(), config.paths.knowledgeDir);
… .getLastKnowledgeSyncStatus()
  : readKnowledgeInventoryFromDisk(knowledgeDir);
```

So the status comes off the per-box handler rather than the process-global accessor, and the disk fallback reads the agent-scoped path. That was the failure mode where local mode would have reported an empty inventory after a restart, and it's gone.

The test for it is the part I want to call out: putting `.sync-manifest.json` **only** in the agent directory while pointing `config.paths.knowledgeDir` at a separate empty `shared-knowledge` dir means an implementation that reads the config path returns empty and the test fails. That's a test that catches the wrong behaviour rather than one that describes the right one.

## `boxes` semantics

```js
const agentBoxes = boxes.filter((b) => b.agentId === agentId);
const targets = agentBoxes.filter((b) => b.status === "running");
```

All three exits now report `agentBoxes.length`, so the field consistently means "boxes belonging to this agent". The three tests pin exactly the distinction I was worried about — one pending box reports `boxes: 1` with `no_running_box`, no boxes at all reports `boxes: 0`, and the success path with a running plus a pending box reports `boxes: 2`. A console can now tell "still pulling the image" from "nothing is bound".

## Everything from my earlier passes

Recording it so the history is readable: the client mock now asserts all three constructor args including `agentBoxTlsOptions`; the cross-process shape lives in `src/shared/agentbox-sync-status.ts` with `satisfies BoxSyncStatus` on the fixture; the multi-box loop uses `sawUnsupported ||= …` with a 404-plus-timeout test; `http-server.test.ts` drives the real `readBoxSyncStatus` through `importOriginal()` against real files; the doc comment is back above `box-status`; and the test-only reset is prefixed to match the file's existing `_mock*` convention.

## Two things I checked myself rather than assume

- `listSkillNames` reads `skillsDir/resolved/` and lists directories. That matches production: `sync-handlers.ts` materializes skills into `path.join(skillsDir, "resolved")`, `agent-factory.ts` reads `path.join(skillsBase, "resolved")`, and `script-resolver.ts` resolves `resolved/<skill>/scripts`. The fixture layout isn't inventing a shape.
- The `/\b404\b/` check on the error message does match what `AgentBoxClient` actually throws (`AgentBox request failed: ${status}`), so the `unsupported` branch is reachable in production and not just against the mock.

CI is green across Test / Type Check / Portal Web Test.

## Merge note

This and #490 both touch `src/agentbox/sync-handlers.ts` — #490 in the knowledge materialize path (writing `.citation-manifest.json`), this one adding `readBoxSyncStatus` and moving types to `src/shared/`. Different regions, so whichever lands second should rebase cleanly, but worth a look rather than an assumption.

PR #495, opened 2026-08-19
fix(kbc): make compile recovery checkpointed and reconnect-safe

LikiosSedo:
## Summary
- treat Raw Head and Stable Wiki Head as content heads, while each sealed Candidate + task ledger is durable compilation progress
- require a Runtime-to-KBC acknowledgement after the durable backend atomically persists every batch checkpoint; the next batch cannot start before that ACK
- replay the exact unacknowledged sync across Runtime reconnects and make ACK delivery idempotent
- prevent periodic sync from leaking a half-written current batch
- seal a batch-zero plan/handoff checkpoint so a failure before batch 1 can recover autonomously
- keep full frozen Raw + current Candidate visible to each batch and preserve complete handoff metadata
- reduce the representative 125-section shape to 42 batches while retaining the 1M Claude context window and an internal 500K planning target
- add checkpoint ACK lifecycle logs without adding Helm values, environment variables, or resource changes

## Verification
- KBC compile-box full protocol suite: passed, including atomic ACK barrier and reset/replan preservation
- Runtime focused Vitest: 24 passed
- representative batching case: 125 sections -> 42 batches
- Python byte compilation and git diff check: passed
- broad npm run build remains blocked by pre-existing pi-coding-agent type drift outside this diff; GitHub Type Check is green

### DIFF
--- a/kbc/design/compiler-agent-environment.md
+++ b/kbc/design/compiler-agent-environment.md
@@ -78,7 +78,8 @@ It is not a page per file and not a line-by-line code reference.
   is explicitly prohibited from starting a second full compilation.
 - All access remains read-only, path-confined, bounded, and covered by regression
   tests.
-- Default planning uses the discovered 1M context window but targets no more
-  than 500K estimated tokens per turn. Actual per-batch usage and high-water
-  context are stamped in the checkpoint for later tuning.
+- The compiler keeps the selected model's 1M context window and uses an internal
+  500K planning target per turn; neither number is a Helm value or environment
+  variable. Actual per-batch usage and high-water context are stamped in the
+  checkpoint for later tuning.
 - A persisted commit-phase checkpoint can finish with no new model session.
--- a/kbc/design/compiler-agent-environment.md
+++ b/kbc/design/compiler-agent-environment.md
@@ -40,6 +40,17 @@ It is not a page per file and not a line-by-line code reference.
 7. **Gates produce actionable truth.** Deterministic checks protect confinement,
    provenance, coverage, and output shape. They do not reward deleting citations
    or bulk-excluding ordinary code merely to make a score green.
+8. **Handoff is durable and globally scoped.** Each stateless batch starts from
+   `authoring/COMPILATION_STATE.json`, which names the phase, full Candidate page
+   set, source-to-page map, tree hash, full Raw inventory, and Candidate index.
+   The assigned batch limits responsibility, not what the agent may discover.
+9. **Delivery is its own resumable phase.** After final review the compiler
+   checkpoints `phase=commit` before emitting provenance commit input. A relay
+   failure at this point replays only delivery; it never repeats map, reduce, or
+   final semantic work.
+10. **Checkpoints prove their identity.** Every persisted `BATCH_PLAN.json`
+    carries a monotonically advancing revision, update time, and SHA-256 digest.
+    Corruption pauses with `plan_integrity`; it never looks like an absent plan.
 
 ## Phase access matrix
 
@@ -67,3 +78,7 @@ It is not a page per file and not a line-by-line code reference.
   is explicitly prohibited from starting a second full compilation.
 - All access remains read-only, path-confined, bounded, and covered by regression
   tests.
+- Default planning uses the discovered 1M context window but targets no more
+  than 500K estimated tokens per turn. Actual per-batch usage and high-water
+  context are stamped in the checkpoint for later tuning.
+- A persisted commit-phase checkpoint can finish with no new model session.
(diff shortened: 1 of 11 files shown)


jacoblee-io (approved):
Approving. This is a distributed-durability change spanning two processes and two languages, so I went looking for the failure modes that usually bite in that shape — one-sided timeouts, non-idempotent replay, deadlock with no escape hatch, deploy-order coupling — and the protocol closes all of them. Notes at the end are suggestions, not gates.

## What I checked, and what holds

**Deploy-order coupling is designed out, not documented around.** `artifact_ack: true` on the `/session` body with `run._artifact_ack_required = body.get("artifact_ack") is True` on the box means all four version combinations degrade safely: an old box ignores the field, a new box only waits when the runtime advertises support. That's better than the sequenced rollout CLAUDE.md's kbc row would otherwise require, and the comment says so at the call site.

**Both sides of the barrier have an escape hatch, and they compose.** The box waits `ARTIFACT_ACK_TIMEOUT_SECS = 10 * 60` and then raises; the runtime retries forever but exits on `!rec || isTerminalCapabilityStatus(rec.status)`. So a box timeout terminates the run, which in turn releases the runtime's loop. Neither side can hang alone — that's the part I most expected to find broken.

**ACK handling is genuinely idempotent, including the replay path.**

```python
future = run._pending_sync_acks.pop(sync_id, None)
run._pending_sync_events.pop(sync_id, None)
if future is not None and not future.done():
    future.set_result(True)
```

`pop` with a default makes an unknown or duplicate `sync_id` a no-op instead of a `KeyError`, the `done()` check prevents a second `set_result` from raising `InvalidStateError`, and popping both dicts means a completed sync doesn't accumulate. The test covers the duplicate delivery explicitly ("Lost ACK responses are harmless"), which is the case a reconnect actually produces.

**`asyncio.shield(ack)` is the right call** — without it `wait_for`'s timeout cancels the future itself, and a cancelled future is `done()`, so a later ACK would silently skip `set_result`. Good catch to make; see note 4.

**`_sync_lock` serializes syncs**, which is what actually delivers "periodic sync can't leak a half-written batch" — the barrier alone wouldn't.

**`onConnected` is lifecycle-only by design**, and the comment states the reason: observers re-read durable state rather than trusting anything cached across the disconnect. Registering while already connected still fires (via `queueMicrotask`), it returns an unsubscribe, and a throwing or rejecting observer can't break the connection path. `server.ts` uses it as `frontendClient.onConnected?.(…)` with a matching optional teardown, so an older client without the method is fine.

**Removing `manager.touchHeartbeat(runId)` from the persist-retry loop is correct**, and I checked why rather than assuming: the box emits `: heartbeat` SSE comments that already drive the same clock via `onComment`, and the file's own comment notes a turn that only heartbeats is still reaped at `dataStaleMs`. So the manual touch was redundant while the box is alive, and actively unhelpful when it isn't — dropping it makes a dead box get reaped sooner instead of being masked by the runtime's own retries.

## Four suggestions

**1. The box image rebuild isn't mentioned.** Per CLAUDE.md's kbc row, behaviour changes need a `siclaw-kbc-box` rebuild plus a `SICLAW_COMPILE_BOX_IMAGE` bump, and a rolled-out runtime doesn't pick up a new image for existing live sessions. The `artifact_ack` negotiation removes the *ordering* hazard, but the box half of this PR still does nothing until a new image ships — worth a line in the description so nobody deploys the runtime, sees `artifact_ack` going out, and assumes the barrier is live.

**2. The description links an internal GitLab MR** (`http://gitlab.scitix-inner.ai/…`). PR bodies are public repository metadata and outlive the branch, and #499 just scrubbed exactly this class of reference from the tree ahead of open-sourcing. Worth replacing with a neutral reference (also note it's `http://`, not `https`).

**3. The two timeout budgets are a contract that isn't written down.** 10 minutes on the box has to exceed the runtime's worst-case reconnect-plus-retry window (10s per POST, backoff capped at 5s, unbounded attempts). That relationship is correct today and invisible in the code — someone lowering `ARTIFACT_ACK_TIMEOUT_SECS` or raising the runtime's cap could break replay with every test still green. A one-line comment on the constant naming what it must outlast would pin it.

**4. Nothing locks the `shield`.** `test_durable_workspace_sync_waits_for_consumer_ack` ACKs within a second, so it never reaches the timeout path — remove `asyncio.shield` and the suite stays green while cross-timeout replay quietly stops working. A case that lets the wait time out and then delivers the ACK would protect the subtlety you deliberately handled.

CI green across Test / Type Check / Portal Web Test / KBC Box Test, and the kbc and gateway halves moved together as that contract requires.

PR #496, opened 2026-08-19
feat(knowledge): support cited OKF packages

LikiosSedo:
Adds safe standalone OKF citation handling within Siclaw. Runtime extraction removes uploader-owned .okf-citations.json while the server-owned citation manifest and knowledge_cite flow remain authoritative. It also adds a migration utility that converts compiled_from into OKF v0.2 sources[].resource and embeds only explicit trusted mappings. Validation: focused Vitest tests, the migration utility test, and all PR checks pass.


jacoblee-io (commented):
The direction here is right and the PR is a net improvement — dropping uploader-supplied citation metadata at extraction, keeping the server-owned manifest authoritative, and the README paragraph states the trust model cleanly. The Python migration tool is careful: `_safe_archive_path`, explicit `schema_version` validation, per-row type checks.

I'm holding the approval for one reason, and it isn't a defect in this diff. While auditing the function this PR touches, I found a reachable arbitrary-file-read path in the same file, and the fix is the rule your own Python tool in this PR already enforces. Details first, then the two smaller items.

## 1. Knowledge packages can ship symlinks, and the Read guard doesn't resolve them

`validateKnowledgePackage` only skips directories:

```js
if (entry.type === "5") continue;
```

Nothing rejects tar type `2` (symlink) or `1` (hardlink), and the path checks look only at the **member name** — absolute paths and `..` are rejected there, but a symlink's *target* is never examined.

I verified the extraction end of this rather than reasoning about it. A package with these three members:

```
./index.md          regular file
./notes.md          symlink -> /etc/passwd
./relative.md       symlink -> ../../../secret.txt
```

Member names are `./index.md`, `./notes.md`, `./relative.md` — no `..`, nothing absolute, so `assertPathTraversal`-style checks have nothing to catch. After `tar -xzf … --exclude=.okf-citations.json`:

```
-rw-r--r--  index.md
lrwxr-xr-x  notes.md -> /etc/passwd
lrwxr-xr-x  relative.md -> ../../../secret.txt
```

Both links are created, including the relative one that escapes the knowledge tree.

The second half is the Read guard. `assertPathAllowed` in `agent-factory.ts` normalizes **lexically**:

```js
const resolved = path.resolve(absolutePath);
const allowed = allowedDirs.some(dir => resolved === dir || resolved.startsWith(dir + path.sep));
```

`path.resolve` does not follow symlinks — only `fs.realpath` would. So `knowledgeDir/notes.md` passes the allow-list because the *link path* is inside `knowledgeDir`, and `readFile` then follows it to the target.

Chain: someone with knowledge-upload rights publishes a package → the symlink materializes into the agent's knowledge tree → the model reads it as an ordinary knowledge page → contents of any file readable by the agentbox process come back. `.siclaw/credentials/` (kubeconfig, SSH keys) lives in that same container, and a relative link reaches it without the member name ever looking suspicious.

**None of this is introduced by this PR** — both functions predate it. I'm raising it here because (a) `extractKnowledgePackageToDir` is the function this PR modifies, (b) the PR's subject *is* which parts of an uploaded package may be trusted, and (c) `attach_okf_citations.py` in this very diff already enforces the missing rule:

```python
if not member.isfile():
    raise CitationMigrationError(f"unsupported archive entry: {member.name}")
```

The offline tool is stricter than the runtime, and the runtime is the attack surface. Applying the same rule in `validateKnowledgePackage` — accept only regular files and directories, reject every other typeflag — closes it at the entry point and keeps the two paths consistent. That's a few lines in a file you're already editing.

If you'd rather keep this PR narrow, that's reasonable — but I'd want it as an immediate follow-up rather than a backlog item, given what's in `credentials/`. Say which you prefer and I'll approve accordingly. (Hardening `assertPathAllowed` to use `realpath` is a separate, broader question — it affects every tool path check, not just knowledge.)

## 2. The exclude is only tested at the archive root

The new test places the sidecar at top level. Nested copies aren't covered:

```js
{ name: OKF_CITATION_SIDECAR, content: … }   // root only
```

Exclusion of `repos/x/.okf-citations.json` depends on tar matching the pattern unanchored. I checked it does — with bsdtar 3.5.3 locally, sidecars at the root, one level down and two levels down were all dropped — and GNU tar's default is `--no-anchored` too. But that default is a tar behaviour, not a property of this code, and GNU tar reads `TAR_OPTIONS` from the environment, so `--anchored` there would silently reduce this control to root-only.

A fixture with a nested sidecar is a one-line addition and pins the behaviour you actually depend on. Filtering during validation instead — where the code already walks every member — would remove the dependency on tar's matching semantics altogether.

## 3. `.citation-manifest.json` isn't excluded; today's safety comes from write ordering

An uploaded package containing `.citation-manifest.json` currently loses to the server's copy only because of ordering in `knowledgeHandler.materialize`:

```
484/526:  extractKnowledgePackageToDir(…)                     ← unpacks first
564:      writeFileSync(stagingDir/".citation-manifest.json")  ← server overwrites after
```

For the single-repo case the archive unpacks straight into `stagingDir`, which is exactly where the manifest lands — so the two collide, and the server happens to write second. That's correct today and fragile tomorrow: anything that reorders those steps, or defers extraction, flips it silently and hands the uploader control of the citation URLs that #490 renders as trusted links.

Since this PR is already establishing an exclude list for uploader-owned citation files, adding `.citation-manifest.json` to it turns "safe because of ordering" into "safe because we said so", at no cost.

## Verified fine

- The trust model in the README matches the code: uploaded copy dropped at materialization, server-owned manifest authoritative, import service validates and freezes.
- The Python tool refuses non-file members, validates `schema_version`, and type-checks every `sources[]` row before emitting a mapping.
- CI green across Test / Type Check / Portal Web Test / KBC Box Test.

jacoblee-io (approved):
Approving — but first, a correction I owe you, because my previous review made a serious claim on a wrong premise.

## My symlink finding was a false positive

`validateKnowledgePackage` already rejects link entries, and has all along:

```js
if (entry.type === "5") continue;
if (entry.type !== "0" && entry.type !== "") {
  throw new Error(`Unsupported tar entry type for ${name}`);
}
```

Only regular files (`"0"` and the legacy `""`) get through, so a package containing a symlink or hardlink is rejected outright at validation. The chain I described breaks at its first step — the attack I claimed was reachable is not reachable.

How I got it wrong: I grepped for `type ===` among other patterns, matched line 52 (`entry.type === "5"`), concluded "only directories are skipped", and never read the next line, which uses `!==`. Then what I called verification only exercised the two *later* links in the chain — that `tar` creates the symlinks, and that `assertPathAllowed` normalizes lexically rather than via `realpath`. Both of those are true in isolation. I never actually ran `validateKnowledgePackage` against a link-bearing package, which is the one check that would have settled it. Absence of a grep hit isn't absence of the guard, and I treated it as such — on a finding severe enough that I withheld approval and named `credentials/` as the blast radius. Sorry for that.

Your new `it.each([["symlink", "2"], ["hardlink", "1"]])` case is still a real gain: the guard existed but nothing pinned it, so a future refactor of that loop could have dropped it silently. Adding coverage to an existing protection is worth doing on its own terms.

## The other two are fixed, and one is better than what I asked for

**Nested exclusion.** I suggested a nested fixture. You did that and closed the underlying dependency instead of just testing around it:

```js
// GNU tar reads TAR_OPTIONS before argv. Do not let ambient options change
// the matching semantics of security-sensitive exclusion patterns.
const env = { ...process.env };
delete env.TAR_OPTIONS;
```

plus explicit `--exclude=*/${name}` patterns alongside the bare ones, so the behaviour no longer rests on tar's unanchored-matching default at all. And the test *sets* `TAR_OPTIONS="--anchored"` before extracting and still asserts every nested copy is gone — that asserts the hardening rather than the happy path. Restoring the previous value in `finally` keeps it from leaking into other tests.

**`.citation-manifest.json`.** Now in `UPLOADER_ONLY_CITATION_FILES` next to the sidecar, so the server-owned manifest is protected by an explicit exclusion instead of by `materialize()` happening to write after extraction. That's the "safe because we said so" version.

## What I checked this time

- The exclusion covers the root, one level down, and two levels down, with `TAR_OPTIONS=--anchored` actively set against it.
- Both citation filenames are excluded, and `SERVER_CITATION_MANIFEST` is exported so the runtime writer and this filter can't drift apart on the name.
- The README trust model still matches the code: uploaded copy dropped at materialization, server-owned manifest authoritative, import service validates and freezes.
- `attach_okf_citations.py` stays stricter than it needs to be (rejects non-file members, validates `schema_version`, type-checks every `sources[]` row) — appropriate for a migration tool whose output feeds the trusted manifest.
- CI green across Test / Type Check / Portal Web Test / KBC Box Test.

One item genuinely remains open and it isn't yours: `assertPathAllowed` resolves lexically rather than through `realpath`. With link entries rejected at package validation there's no path from a knowledge upload to it, so nothing here is blocked on it — but it's the kind of guard worth revisiting on its own, and I'll raise it separately rather than attach it to this PR.

PR #497, opened 2026-08-19
fix(node-logs): report why a log query came back empty

jacoblee-io:
Five post-mortems flagged this skill: the script swallowed every underlying failure with `|| true` and reported one identical "No logs found" for all of them, exit 0. An agent that asked for `--grep "id1|id2"` got that message and read it as evidence the node was fine.

## What was wrong

- **`--grep` was a basic regexp.** `grep -i` treats `|` literally, so `id1|id2` matched nothing and exited 0 — a false negative that hides real kubelet/containerd lines.
- **`|| true` wrapped the whole pipeline**, and `journalctl … 2>&1` merged the source's stderr into stdout before the filter ran, so a failing journalctl became "log content" that was then silently dropped. The tool layer already forwards a non-zero exit *and* stderr to the model (`node-script.ts`), so the script was throwing away a signal it gets for free.
- **An unknown flag printed help and exited 0.** Two other skills (`image-pull-debug`, `node-health-check`) invoked this one with a `--node` flag it never had, via `bash skills/core/...` rather than a script tool — so an agent following them got a page of usage text and a success.
- **No `--until`, no boot selection.** `--since` only accepted journalctl's own syntax, while every alert and Kubernetes timestamp is RFC3339, which journalctl rejects outright.
- Absence and failure were indistinguishable: missing unit, no journalctl on the node, unreadable file, filter matched nothing, window genuinely empty.

## What this changes

**One `status:` line, always the last line of the output**, with any explanation above it — so "read the last line" is a rule that holds for every outcome:

| `status:` | Meaning |
|---|---|
| `ok` | source read, lines matched |
| `no_match` | source read fine, filter matched nothing — bounded non-evidence, nothing about the node |
| `source_error` | the log source itself failed (its stderr is passed through untouched) |
| `filter_error` | the pattern was rejected, so *nothing was searched* |
| plus `not_found` / `not_a_log_file` / `empty_source` / `query_unsupported` / `forbidden` on the API path |

The verdict cannot be printed by the counting stage — whether an empty stream means `no_match` or `source_error` is only knowable once the source has exited, which happens after that stage has already written its output. So it reports "nothing matched" through its exit code and the shell prints the single verdict.

Also:

- `--grep` is an ERE, repeatable and OR-joined; `--grep-fixed` for literals (image refs, paths, UUIDs). An invalid pattern is `filter_error`, never "nothing found".
- `--until`, `--boot`, `--list-boots`. RFC3339 and epoch windows are converted for journalctl. A default `--since` no longer survives alongside `--boot` or an explicit `--until`, where it silently clipped the requested window.
- `--include-rotated` for numeric rotations; per-file absence reporting; `--tail` clamped at 20000 with a visible note, since the ring buffer lives on the node.
- One streaming pass counts `scanned:` and `matched:` in constant memory, so a long window costs one journalctl read no matter how many patterns are asked about.
- New `get-node-logs-via-api.sh` (tier 3, `local_script`): reads a node's logs through the kubelet log endpoint when the node is neither a bound SSH host nor able to start a debug pod — the state where the previous two tiers both fail.
- Corrected the two skills that invoked this one with a flag it never had, and the `node_script` / `host_script` tool descriptions that advertised `--lines` and a script name that does not exist.

## Verification

Fixture tests (52) run the scripts for real against a PATH containing only the utilities they may use plus fake `journalctl` / `systemctl` / `kubectl`, which is what makes "journalctl is absent" and "journalctl failed" reproducible states rather than properties of the host. The test lives under `src/` because everything inside a skill directory is packaged and shipped to agents.

Six defects were found by running on a real node and a live cluster, none of which a fixture could have produced:

1. **`-- No entries --` counted as a log line** — journalctl's placeholder for an empty query made an empty journal report `scanned: 1 / matched: 1 / status: ok` with that placeholder as the evidence. Skipped now; `-- Reboot --` is kept, since that one is a real boundary.
2. **`systemctl list-unit-files -- <name>` exits 1 both for "no such unit" and for "cannot reach the system bus"** — keying off its exit code invented a missing `docker.service` on a host that runs docker. The inventory is fetched once and its emptiness decides whether systemctl can be trusted at all.
3. **The similar-unit hint searched for the full requested name**, which by definition is not in the inventory when the name is wrong. It searches on a prefix now, limited to `.service` units — on a real node "kubel" also matches a page of `var-lib-kubelet-pods-….mount`.
4. **journalctl shows the system journal only to root and to `adm` / `systemd-journal` members**; any other account gets its own messages and exit 0. Over `host_script` that is the normal case, so an empty result there is a permission artifact — now reported as a note and repeated in the `no_match` explanation.
5. **A rejected pattern read as a failed log source**: grep dies instantly on an invalid ERE, the pipe closes, and journalctl — still streaming megabytes — dies of SIGPIPE with 141. The ladder judged the source first, so a bad regex was reported as "the log source failed, see stderr". The filter's exit code is read first now; the regression test writes 200k lines to make the SIGPIPE real.
6. **`--head-bytes` on the API path** stops the transfer by closing the pipe, so kubectl dies of SIGPIPE by design — which was being filed as `source_error`.

Behaviour confirmed on a live node (via the same nsenter path `node_script` uses) and a live cluster:

| Case | Result |
|---|---|
| `--unit kubelet --grep "a\|b"` | matched 678 / 8394 — the query that previously returned "No logs found" |
| `--since 2026-08-18T13:00:00Z --until …13:01:00Z` | converted to `@1787058000`, window honoured |
| window with no entries | `scanned: 0` → `no_match` (previously `ok` with the placeholder) |
| `--list-boots`, `--boot 0` | default `--since` correctly suppressed |
| `--unit kubelettt` | `not a known systemd unit (similar: kubelet.service)` |
| `--file /var/log/syslog --include-rotated` | 5 files, 385M lines scanned, 62M matched |
| invalid ERE | `filter_error`, exit 2 |
| tier 3: `--list` / `--file syslog` / `--query kubelet` / `--file messages` / `--file journal` | `ok` / `ok` / `query_unsupported` / `not_found` / `not_a_log_file` |

The kubelet log endpoint's own behaviour drove the tier-3 design: an unsupported `?query=` returns the `/var/log` HTML **directory index** with HTTP 200 rather than an error (verified on kubelet 1.29, where `NodeLogQuery` is off by default), so every response is classified by kubectl's exit code and stderr — never by body content, which would misread a syslog full of the word `error:`.

Not covered by anything but fixtures, stated for the record: a node without journalctl (exit 3), a node without gzip (the skipped-`.gz` note), and a proxy refused for lack of `nodes/proxy` (exit 6).


LikiosSedo:
## Review (comment, not blocking)

**Verdict: approve, with two should-fix notes.** 0 runtime 🔴. The product direction is right: the old script wrapped the fetch in `|| true` and printed one "No logs found" at exit 0, so an agent could read every empty outcome as "the node is fine."

### Self-fix pattern (process note, not a blocker)

Five follow-up commits fixed live-host bugs the first rewrite missed. The final tree includes them; this is also evidence the script was actually run on a real systemd host.

| Commit | Intermediate bug |
|---|---|
| `59a3263` | journalctl `-- No entries --` counted as scanned=1 / status=ok; systemctl failure treated as "not a unit"; non-privileged account reads the system journal at exit 0 |
| `1b0ae2c` | unreachable system bus and a typo'd unit both exit 1, so a typo was explained as "systemctl could not answer" |
| `fd00f7c` | similar-unit hint depended on `tr`, which vanishes under the private-PATH harness |
| `de8e897` | invalid ERE kills grep, source dies of SIGPIPE 141, reported as `source_error` |

Squash to 1–2 commits, or leave a sentence in the PR that these were found on-box and are in the final tree.

### Should fix

**[1] Empty `--grep` matches every line** — `get-node-logs.sh` / `get-node-logs-via-api.sh`

`[[ $# -ge 2 ]]` only checks that a second argument exists. `--grep ""` becomes `grep -e ""`, matches the whole file, and reports `status: ok`. An agent that passes an empty pattern would treat a no-op query as "everything matched."

```bash
[[ -z "$2" ]] && die_usage "--grep pattern cannot be empty"
```

Same check on `--grep-fixed`. This is the one change I would make before merge if you are still iterating.

**[2] BSD `date` fallback only accepts `Z`** — `normalize_time` in `get-node-logs.sh`

GNU `date -d` runs first and accepts `+08:00`. The BSD fallback is hardcoded to `%Y-%m-%dT%H:%M:%SZ`. Nodes are GNU, and conversion failure exits 2 (fails closed), so production risk is low. Tests can keep using `Z` / epoch.

### Architecture notes (not merge-blocking)

- The seven-status contract is heavy for a skill script, but a false "node is fine" is more expensive than a long script. I would not collapse it to three states just to ship.
- Duplicated filter/status logic between the two scripts is a constraint, not sloppiness: skill scripts are transmitted and executed one file at a time, so there is nothing to `source`. Do not extract `common.sh` unless the transmission model changes.
- `SKILL.md` is long; the call-site updates in `image-pull-debug` and `node-health-check` are the part that matters. They no longer invoke `bash skills/core/... --node` (a flag this script never had, previously exit 0).

### What looks good

- Empty-result contract is a real fix: `no_match` (read succeeded, filter hit nothing) vs `source_error` (the source failed). stderr is no longer merged into stdout and dropped by the filter.
- Filter exit is judged before source exit. The 200k-line SIGPIPE regression is the kind of case fixtures cannot invent.
- Tier-3 classification uses kubectl's exit / stderr, never body content. That is the right answer to kubelet returning the `/var/log` HTML index with HTTP 200 for an unsupported `?query=`.
- Security looks intact: node name is constrained to Kubernetes DNS, tier-3 paths reject `/` and `..`, only `kubectl get --raw`, arguments go through arrays.

Tests living under `src/tools/script-exec/` is correct — everything inside a skill directory is packaged and shipped (`collectSkillDirectoryFiles`).

jacoblee-io:
Thanks — both should-fixes are real. Fixed in `06d2858`.

**[1] Empty pattern.** Confirmed before fixing: `grep -E -i -a -e "^__MARK__:" -e ""` matches every line, so `--grep ""` came back as a full hit at `status: ok` — a no-op query that reads like a filter that succeeded, which is exactly the failure mode this PR exists to remove. Both scripts now reject an empty `--grep` / `--grep-fixed` with `an empty pattern matches every line, which is not a filter`, and `SKILL.md` states it. Verified on a live node: refused before anything is read.

**[2] BSD `date` fallback.** Also fixed rather than left as a caveat — a value that converts on a node and fails on a workstation is a trap for whoever debugs this next. Three differences from GNU are absorbed now: the colon RFC3339 puts in `%z` (BSD strptime wants `+0800`), the absent `%OS` (fractional seconds are dropped — a log window needs no sub-second precision), and a zoneless value being local time. Conversion still fails closed with exit 2.

Verified on a live node that both spellings of the same instant agree:

```
--since 2026-08-18T13:00:00Z      → journalctl: @1787058000
--since 2026-08-18T21:00:00+08:00 → journalctl: @1787058000
```

New fixture cases cover the empty pattern (both flags) and an offset window with fractional seconds.

**On the commit history.** Keeping the six commits rather than squashing: each one carries *why* a live-host finding was not visible in fixtures, which is the part worth having in `git log` when this script is touched again — e.g. `de8e897`'s message records that a few lines of fake output fit in the pipe buffer, so the source never gets signalled and the SIGPIPE case cannot be reproduced without writing 200k lines. Happy to squash if you would rather have one commit; say the word. The PR body already lists all six findings and confirms they are in the final tree.

Two notes on the architecture points, since they match what I ran into:

- The duplicated filter/status logic is exactly the transmission constraint you name — `resolveScript` reads one file and pipes it to `bash -s`, so there is nothing to `source`. Both copies carry a comment saying so, to stop a future reader "fixing" it.
- The seven statuses are not symmetric by design: `ok` and `no_match` are the only ones that mean the source was read. Everything else is a statement about the *tool or the access*, which is why they carry distinct exit codes — an agent that only reads the last line still cannot mistake one for the other.

LikiosSedo:
## Re-review

Both should-fixes are in `06d2858`, and both are real.

**Empty pattern.** Confirmed in both scripts (`[[ -n "$2" ]]` on `--grep` / `--grep-fixed`), stated in `SKILL.md`, and pinned for `get-node-logs.sh` (exit 2, stderr contains `matches every line`). I also checked the via-api copy: same reject, no dedicated test — fine, it is the same one-liner and dies before any transfer.

**BSD date.** I ran `normalize_time` under bash on a Mac (so the GNU `date -d` branch is dead and the fallback is what runs):

| input | output |
|---|---|
| `2026-08-18T14:00:00+08:00` | `@1787032800` (06:00Z) |
| `2026-08-18T15:00:00.500+08:00` | `@1787036400` (frac dropped) |
| `2026-08-18T13:00:00Z` / `2026-08-18T21:00:00+08:00` | both `@1787058000` |

Matches the live-node numbers in your comment. Conversion still fails closed on garbage. Fine to keep the six commits — the SIGPIPE / pipe-buffer note in `de8e897` is the kind of thing you want in `git log`.

Nothing left from my side. LGTM.

LikiosSedo (approved):
LGTM. Empty `--grep` / `--grep-fixed` now fail closed, and the BSD date fallback accepts the same RFC3339 shapes GNU does (verified off-node). The empty-result status contract is the real fix.

PR #498, opened 2026-08-19
fix(utf8): decode streams once, not per chunk

jacoblee-io:
A multibyte character split across two `data` events was silently corrupted into two U+FFFD wherever a reader accumulated a stream with `body += chunk` or `chunk.toString()` per event: each fragment is decoded on its own, so the three bytes of one character become two invalid sequences.

## How it surfaced

While verifying a skill deploy, the `SKILL.md` materialized inside an agentbox was 3 bytes larger than the source and differed on exactly one line:

```
source:  ... **Start here** [e2 80 94] file names differ ...    ← U+2014 EM DASH
in pod:  ... **Start here** [ef bf bd ef bf bd] file names ...  ← two U+FFFD
```

The DB row was byte-identical to the source (verified by md5), so the corruption was in transit. The skill bundle is fetched by `GatewayClient.request`, which decoded per chunk. Of the 29 em dashes in that file exactly one was destroyed — whichever happened to straddle a chunk boundary, which is why this survived unnoticed: it does not throw, it does not log, and it only bites when the split lands inside a character.

## Scope

Same defect, same one-line fix, at every site that accumulated a stream into a string:

| Site | What flows through it |
|---|---|
| `agentbox/gateway-client.ts` | every sync payload — skills, knowledge, tools, credentials — plus the delegation SSE stream and its error body |
| `agentbox/http-server.ts` | request bodies, i.e. **prompts**: a long non-ASCII prompt could be rewritten at a chunk boundary |
| `gateway/agentbox/client.ts` | gateway → agentbox responses |
| `gateway/credential-proxy.ts` | request bodies |
| `tools/infra/exec-utils.ts`, `tools/cmd-exec/restricted-bash.ts`, `tools/script-exec/local-script.ts` | command output, where a CJK log line was equally exposed |

## The fix

Decode on the stream (`setEncoding("utf8")`), so Node's own StringDecoder holds the partial bytes across events. This is what `core/background-bash-runner.ts` already does for child output — with a comment naming this exact failure — and what `rest-router.ts`, `internal-api.ts`, `delegate-api.ts` and `siclaw-api.ts` already achieve by concatenating Buffers and decoding once at the end. The HTTP paths simply never caught up.

`local-script.ts`'s 10 MB cap now measures `Buffer.byteLength(chunk)`: on a string, `.length` counts UTF-16 units and would let CJK-heavy output run past the limit.

`credential-proxy`'s hand-rolled `IncomingMessage` double gained the `setEncoding` it was missing, so it exercises the path production takes. The production code deliberately does **not** call `setEncoding?.()` — an optional call would silently skip the fix wherever the stream is not a real `IncomingMessage`, which is how the bug would come back.

## Verification

Two regression tests split a body / response mid-character across two TCP writes and assert the far side sees the original text. Both were confirmed to fail before the fix:

- `http-server.test.ts` — POST `/api/prompt` with `查一下 kubelet — 日志`; before the fix `brain.prompt` received mangled text.
- `gateway-client.test.ts` — the server writes a JSON response in two chunks; before the fix the parsed object failed deep equality.

End-to-end on a live deployment: after rolling out this change, the same skill materialized into a fresh agentbox is byte-identical to the source (`b315fe50` / 11771 bytes, previously `96354eed` / 11774 bytes).

**Deploy note:** this spans two images — `gateway/*` ships in **runtime**, `agentbox/*` and `tools/*` ship in **agentbox**. Rolling out only one leaves the other half of the paths uncorrected.


LikiosSedo:
## Review (comment, not blocking)

**Verdict: approve.** Root cause, fix, and regression tests all line up. 1 🟡, 0 🔴.

The UTF-8 mid-character split is real (`body += chunk` / `chunk.toString()` per `data` event). `setEncoding("utf8")` is the right Node fix and matches `background-bash-runner.ts` / the Buffer.concat paths. The two TCP-split tests pin the failure that actually shipped.

Deploy reminder from the PR body still holds: this spans **runtime** (`gateway/*`) and **agentbox** (`agentbox/*`, `tools/*`). Rolling only one leaves the other half uncorrected.

### Should fix

**[1] credential-proxy has no mid-character split test** — `src/gateway/credential-proxy.test.ts`

`FakeReq` now implements `setEncoding`, and production `readBody` is correct, but unlike `gateway-client.test.ts` / `http-server.test.ts` there is no test that splits an em dash across two chunks and asserts the parsed body is not `�`.

Worth adding when convenient. Not a merge blocker.

### What looks good

- `local-script` size cap now uses `Buffer.byteLength`. After `setEncoding`, `.length` is UTF-16 units and would let CJK-heavy output run past 10 MB.
- `http-server` `parseJsonBody` already accounts string chunks with `Buffer.byteLength`, so the cap did not regress.
- Security model untouched: no new commands, no kubectl widening, credential-proxy still reads JSON.

jacoblee-io:
Added in `9b83cdb` — and writing it turned up something worth flagging, because the test would otherwise have been theatre.

`FakeReq.setEncoding` (added in the first commit so the 18 existing cases kept passing) recorded the encoding and then decoded **each chunk on its own**. That means the double corrupted the text regardless of what production did: the new split test would have failed even with the fix in place, and "fixing" it by relaxing the assertion would have produced a green test that pins nothing.

The double now runs ONE `StringDecoder` across all chunks, which is what a real stream does with an encoding set. With that, the new case behaves like the other two readers' tests — it fails on the old reader:

```
AssertionError: expected '诊断 kubelet ��� 日志' to be '诊断 kubelet — 日志'
```

The assertion targets `getClusterCredential(identity, source_id, purpose)`'s third argument, so it pins the corruption where it would actually matter here: the purpose string a credential is requested and audited under.

Deploy reminder noted and it is already the state of things: `agentbox:test` and `runtime:test` were both rebuilt from this change and pushed (`6646b07d` / `fa4d4f7b`). End-to-end on a live deployment, a skill materialized into a fresh agentbox is now byte-identical to source (`b315fe50`, 11771 bytes; previously `96354eed`, 11774 — the 3 extra bytes being one em dash replaced by two U+FFFD).

One environment where half of this is still uncorrected, for the record: a deployment whose runtime is pinned to an older tag keeps the `gateway/*` half of the paths until its runtime is redeployed. The agentbox half (sync payloads, prompts, command output) applies as soon as a new session spawns.

LikiosSedo:
## Re-review

The follow-up is real, and better than the note I left.

`9b83cdb` does not just add the missing split test. The first-commit `FakeReq.setEncoding` decoded **each chunk on its own**, so a split-character case would have gone red even with the production fix (or gone green only by weakening the assertion). One `StringDecoder` across chunks is what a real `IncomingMessage` does; the new case asserts `getClusterCredential(..., purpose)` and would fail on the old reader with `诊断 kubelet ��� 日志`.

Nothing left from my side. LGTM.

LikiosSedo (approved):
LGTM. The UTF-8 stream decode is correct, and `9b83cdb` makes the credential-proxy split test actually pin production behavior (one StringDecoder across chunks, assertion on `purpose`).

PR #499, opened 2026-08-19
chore: replace internal hostnames and registry references with neutral placeholders

chent1996:
## Summary

Scrub internal infrastructure references from the codebase ahead of community open-sourcing: test fixtures now use reserved `example.org`/`example.com` domains, docs use a `<your-registry>` placeholder, and code comments reference generic incident notes instead of internal project names.

### Solution

- Test fixtures (`image-url-ingest.test.ts`, `client.test.ts`, `lark.test.ts`): internal OSS hostnames → `example.org` family. Fixture semantics preserved — the mixed allowlist test keeps two distinct entries, and the apex-vs-wildcard case still exercises the subdomain boundary.
- `brain-session.test.ts`: provider patch value → `"other-provider"` (still differs from the base fixture, so the rebind assertion is unchanged).
- `image-url-ingest.ts`: comment example domain only.
- `services/ocr-backend/README.md`: internal registry in the build command → `<your-registry>` placeholder.
- `kbc/platform/pod/{redblue,test_selfcheck}.py`: comments now describe the incident/workflow generically instead of naming internal projects.
- `CHANGELOG.md`: registry replication note genericized (drops internal region/registry naming).

Fixture/comment/doc-only — no runtime behavior change.

## Test Plan

- `git grep` over tracked files confirms the scrubbed strings are gone
- Full `npm test`: 249 files, 5144 passed / 2 skipped / 0 failed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ Renamed the KB compile box so operators can tell it apart from chat agentboxes a
 - **Pod-name prefix**: compile boxes now spawn as `kbc-box-<id>` instead of `agentbox-<id>`, declared on the BoxProfile (`podNamePrefix`) for the `kb-compile` / `kb-compile-codex` profiles. Derived resources (cert Secret, hostname) follow the prefix. Chat agentboxes and the read-only `kb-test` box keep the `agentbox-` prefix.
 - **Upgrade behavior**: on the first compile spawn after upgrade, the spawner reaps any pod left under the old `agentbox-<id>` name for that agent (guarded to compile boxes only — a chat box under the same name is never touched), so the old and new pods do not coexist.
 - **Transition double-push**: `make push-kbc` republishes the same image digest under the legacy name `kbc-compile-box` for one release cycle so runtimes still pinned to the old name keep resolving. Remove the legacy tag/push next release.
-- **DevOps action required**: the registry replication rule (ap-southeast → cn-shanghai) must gain an entry for `siclaw-kbc-box` (the old `kbc-compile-box` was not in the rule and had to be pushed manually). Until the rule is added, push `siclaw-kbc-box` to cn-shanghai manually for production.
+- **DevOps action required**: the internal registry replication rule must gain an entry for `siclaw-kbc-box` (the old `kbc-compile-box` was not in the rule and had to be pushed manually). Until the rule is added, push `siclaw-kbc-box` to the production registry manually.
 
 ### Added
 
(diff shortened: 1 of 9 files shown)


jacoblee-io (approved):
Approving. Fixture-only substitutions like this are easy to get subtly wrong — the failure mode is a test that still passes while no longer exercising the boundary it was written for — and the two places where that could have happened here are handled correctly.

## The two substitutions that could have broken semantics

**Mixed allowlist.** The pair became `oss.example.org` (exact) + `*.example.com` (wildcard). Keeping those on *different* TLDs is what preserves the test: had the exact entry been written as `oss.example.com`, the `*.example.com` wildcard would have covered it and the exact-match path would silently stop being tested. Two distinct entries, one exact and one wildcard, same as before.

**Apex vs wildcard.** `*.example.org` with the apex `example.org` rejected keeps the original relationship intact — the case still proves an apex is not a subdomain of its own wildcard, which is the actual security property.

Also checked: `client.test.ts` swapped the fetch-mock predicate (`String(url).includes(...)`) in step with the URLs it matches, so the mock still intercepts what the test fetches, and the protocol case still has an allowlisted host with an `http://` scheme so it exercises the protocol rejection rather than the allowlist.

**`brain-session.test.ts`.** The patch value `"other-provider"` still differs from the fixture's `provider: "scitix"`, so `modelNeedsRebind` fires exactly as before. Worth stating explicitly because that test is the one guarding the rebind rule — a substitution that accidentally matched the base fixture would have turned it green-and-useless.

## Scrub completeness

I verified this rather than taking the test plan's word for it: every tracked file on `main` containing `siflow` is one of the nine files in this PR, so there's no leftover. I also checked the adjacent internal identifiers — `scitix-inner`, `registry-ap-southeast`, `jump-cl`, `cl-ops` — and all are already absent from the repo, so the scope here is complete rather than partial.

Keeping `scitix` itself is the right call: it's the public org this repository lives under, not internal infrastructure.

Using `example.org` / `example.com` is also a small improvement beyond the scrub — they're RFC 2606 reserved names that can never be registered, so a test that accidentally makes a real request can't reach a third party's server.

The comment edits keep the useful part while dropping the identifier. `test_selfcheck.py` becoming "real incident 2026-07-07" retains the date and the specific symptom (需/础/成 mangled), so the note is still traceable internally and still explains why the check exists.

## One suggestion, not blocking

A scrub is a point-in-time fix; drift is continuous. A CI grep guard over a small denylist (`siflow`, `scitix-inner`, `registry-<region>`, internal repo paths — explicitly *not* `scitix`) would keep this from eroding as new fixtures and comments land. The `git grep` in your test plan proves the state today; a check in CI would keep proving it.

## Deploy note

`kbc/platform/pod/*` carries a rebuild requirement for behaviour changes, but both edits there are comment-only, so no `siclaw-kbc-box` image rebuild is needed. KBC Box Test passing alongside Test / Type Check / Portal Web Test.

PR #501, opened 2026-08-19
feat(exec): classify exit codes, and project JSON without jq on the target

jacoblee-io:
Second batch from the `node_exec` retro feedback. **Stacked on #493** (`base: fix/node-exec-batch1`) because it touches the same result-shaping code; the diff to review is the two commits here.

Each item was checked against the code before being treated as real, and every new test was verified by reverting its fix.

## 1. What a non-zero exit MEANS (`58c80b13`)

All four exec tools reduced "did not exit 0" to one bit — `details.error: true` plus a bare `[exit code: N]`. That merged situations calling for opposite next steps, including the retro's single most frequent complaint.

| class | meaning | `details.error` |
|---|---|---|
| `channel_error` | The exec path failed; the target never ran the command, so this status is NOT its answer | yes |
| `dependency_missing` | Not on the target (127, or the runtime's "executable file not found"); retrying cannot help | yes |
| `not_executable` | Found but not runnable (126) | yes |
| `no_match` | Ran and matched nothing (`grep`/`pgrep`/`test` exiting 1) — **not a failure** | no |
| `target_reported_failure` | Ran and reported this status; the target's own answer | yes |
| `interrupted` | Signalled; partial output is still a result | only with no output |

**A dead channel is now distinguishable from a command that failed.** `kubectl exec` reports its own failures — pod gone, connection not upgradable, API server refusing — through the same exit status it uses to relay the remote command's, so the two were indistinguishable without reading stderr, and nothing in the tree looked. `host_exec` never had this problem (an SSH-level failure throws and was already reported as "a connection failure, not a command error"); it now uses the same vocabulary.

Detection is deliberately conservative: markers are line-anchored and only consulted when stdout is empty, since a channel that failed produced no command output — so a command whose own output contains `error:` is not mistaken for a broken channel. `command terminated with exit code N` is **not** a marker: it means the command DID run. The runtime's "executable file not found" is checked first, because it also matches a channel marker and the specific reading is the useful one.

**The command's own failure text survives.** stderr already bypassed the sanitizer, which is what carries the target's words. What did NOT survive was the classification: folding the annotation into the body fed it to the output sanitizer, and a structural (JSON) sanitizer replaces anything it cannot parse with a suppression notice — precisely what a failed command's output is. A test on the `crictl inspectp` path caught that; `appendAnnotation` now runs after sanitizing AND truncating.

Two contracts, neither derivable from the other: the class must be in the **text** (`details` is stripped before the model sees a result), and `details.error` must stay accurate (it drives the Trace outcome) — which is why `no_match` sets neither.

Also here: the not-Ready node message no longer says "cordoned". Cordoning sets `spec.unschedulable` and leaves `Ready=True`, so it cannot produce that condition, and naming it sent the reader to check a field that was never the cause. Ready=False already prevents a debug pod — all three call sites return on the check.

## 2. `json_path`, and refusals that say something (`da0228c8`)

**The agent can now process JSON on a target without `jq`.** Removing our own dependency on host `jq` (#493, the `pod=` netns resolution) left the agent's problem untouched: the diagnostics that matter emit JSON (`crictl inspect`, `nvidia-smi -q -x`, `ip -j`) and `jq` is usually absent from a node, so the whole document came back and was truncated exactly when it was large. `node_exec`, `pod_exec` and `host_exec` take an optional `json_path`, evaluated in the AgentBox.

Not jq, and not a subprocess: jq would hand agent-authored program text to a process running as `agentbox` — the user that OWNS the credentials — and jq's `$ENV` reads the environment. So this is a path evaluator with **no expression language**: no filters, conditions, functions or I/O. It cannot reach outside the parsed document by construction rather than by allow-list, and a filter expression is rejected rather than ignored, so a jq-looking path cannot silently project something else.

Grammar in full: `.a.b`, `.a[0]` (negative counts from the end), `.a[]` / `.a[*]`, `.["quoted name"]`, `.`.

Where it runs is a contract:

- **after** sanitization — projecting first would strip the shape a structural sanitizer matches on (crictl's `info.config.envs`) and could hand back a value it would have redacted. There is a test that fails if the order is swapped.
- **before** truncation — truncated JSON does not parse, so a later projection would fail on precisely the documents worth projecting.

Two things the tests forced out: a structural sanitizer appends a redaction notice, which makes its own output un-parseable (so the projector parses the JSON span and ignores trailing prose), and it re-appends that notice, because a projection that drops it presents edited content as verbatim.

**A flag value beginning with `-` is not a flag.** `journalctl -b -1` was refused with `"-1" is not allowed` — a message naming the value rather than the parse, so an agent drops the `-1`, loses the intent, and never discovers that `-b-1` works. Commands may now declare `valueFlags`. Deliberately narrow: only a negative NUMBER directly after a declared value-flag is consumed, so `journalctl -b -1 -f` and `journalctl -x -1` are still refused — this cannot become a way to smuggle a token past the whitelist.

**A sensitive-path refusal names what matched and what to do instead.** "Accessing sensitive paths is not allowed" identified neither the offending argument nor whether the command itself was the problem, so the usual response was to try another command and be refused again. It now returns `matched`, `rejected_by: "sensitive_path"` and a hint keyed to the kind of secret — read the certificate's public half, use host_exec rather than reading a key, check that a secret is mounted rather than reading it. The hint is guidance, not a way through: every path stays refused, with a test pinning that.

Because the refusal is machine-readable now, ~46 tests that pinned the exact sentence assert `rejected_by` instead. A phrase assertion only restates the wording and breaks on any rewrite — which is how it broke here.

## Found while reviewing my own work

- **`json_path` with `run_in_background` was silently ignored** — the background branch returns before the projection runs, so the agent would ask for one field, get the whole stream, and have no way to tell its parameter did nothing. That is the same class of problem this PR is about. The combination is refused, and the check sits with parameter validation rather than in the background branch, because resolving a cluster first answered with a kubeconfig error and hid the real mistake.
- The projector hook had been copy-pasted into all three tools; it now lives beside `projectJson`.
- `details.exit_class` was missing from `pod_exec`'s and `restricted-bash`'s success paths.
- `node_exec`'s description still advised preferring a command's own output flags "over piping to a JSON processor" — written before `json_path` existed, and now steering the agent away from the better option.

## Verification

`npm test` exit 0 — 252 files, 5237 passed (judged by exit code, not by reading the summary). `npx tsc --noEmit` clean. Nine fixes across the two commits were revert-checked individually.

Worth recording: one revert-check was initially **invalid** — a crude edit broke the file so vitest reported "no tests" rather than a failure, which looks like success at a glance. Redone with an exact replacement. And one test first passed for the wrong reason: `analyzeOutput(binary, args)` was called with the whole command as `binary`, so the sanitizer never ran and the "projects after sanitization" test proved nothing until the resolved action was asserted non-null — at which point it immediately caught the redaction-notice/JSON-parse interaction above.

**Not verified live.** These two commits have unit tests only; the deployed test image predates them. The parts worth exercising against real data are `json_path` on a real `crictl inspect` document and the channel-vs-command split on real `kubectl exec` errors.

**Deployment:** agentbox image (`src/tools/`), so merging changes nothing until the image is rebuilt, `SICLAW_AGENTBOX_IMAGE` is repointed, and the pods recycle.


jacoblee-io:
Folded into #493 — the same result-shaping code, and one PR is easier to review than a stack. The two commits are now on `fix/node-exec-batch1`; #493's description covers them.

Also flagging an overlap found while consolidating: #494 already fixes the "exit code spliced into the body before sanitization" bug that the second commit here fixes independently, and it does so with a better shape (`exitCode`/`signal`/`notes` options on `postExecSecurity`, so the annotation logic lives in one place instead of at each call site). That duplicate will be resolved in favour of #494's version rather than kept in both — see the note on #493.

PR #502, opened 2026-08-19
feat: preserve structured MCP results across turns

LikiosSedo:
## Summary

- preserve the MCP-standard `structuredContent` field in internal tool results
- correlate Runtime `chat.event` envelopes with the existing per-turn `turnId`
- keep the transport generic: no product-support schema or completion policy lives in Siclaw

## Why

API facades need a machine-verifiable MCP result for one exact turn without parsing assistant text or confusing events from a long-lived session. Siclaw already receives both values; this change stops dropping `structuredContent` and carries the existing turn identity across the Runtime boundary.

## Validation

- `npx vitest run src/core/mcp-client.test.ts`
- `npx vitest run src/gateway/server-chat-abort.test.ts`
- `npx tsc --noEmit`
- `npm run build`

### DIFF
--- a/src/core/mcp-client.test.ts
+++ b/src/core/mcp-client.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect } from "vitest";
+import { describe, it, expect, vi } from "vitest";
 import { jsonSchemaToTypebox, normalizeMcpInputSchema, buildMcpToolName, isMcpTool, MCP_TOOL_PREFIX, mcpContentToAgentContent, McpClientManager } from "./mcp-client.js";
 
 describe("jsonSchemaToTypebox", () => {
@@ -166,6 +166,25 @@ describe("createToolDefinition server description", () => {
   it("applies the server context to the fallback description too", () => {
     const def = makeDef("Monitoring tenant ID: t-123");
     expect(def.description).toBe('[Server "grafana" context: Monitoring tenant ID: t-123]\nMCP tool query from grafana');
+  });
+
+  it("preserves official MCP structuredContent for host-side consumers", async () => {
+    const structuredContent = { label: true, info: { summary: "ready" } };
+    const def = (manager as any).createToolDefinition(
+      "product-support-result",
+      undefined,
+      { name: "submit_product_support_result" },
+      {
+        callTool: vi.fn(async () => ({
+          content: [{ type: "text", text: JSON.stringify(structuredContent) }],
+          structuredContent,
+        })),
+      },
+    );
+
+    await expect(def.execute("call-1", {})).resolves.toMatchObject({
+      details: { structuredContent },
+    });
   });
 });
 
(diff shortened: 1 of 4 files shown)


jacoblee-io (approved):
Approving. Both halves are additive and I verified the two things that additive changes like these usually get wrong — a missed branch, and a new payload path that bypasses a budget.

## `turnId` on `chat.event`: no send site missed

I enumerated every `sendEvent("chat.event", …)` in `server.ts` on `main` — six of them (the `onEvent` relay, `prompt_done` on the success path, `stream_error` + `prompt_done` on the catch path, and `stream_error` + `prompt_done` on the supervisor path). All six carry `turnId` after this diff. That's the failure mode I was looking for: a consumer correlating by `turnId` would silently drop whichever events came from the one branch that didn't set it, and error paths are exactly where that hurts.

The change is field-only, so the abort contract is untouched — `abortSessionTurnIds` and the existing assertions in `server-chat-abort.test.ts` are unchanged, and the new `toMatchObject({ sessionId: "orphan", turnId: ack.turnId })` pins the envelope shape. If anything this reinforces "cancellation is addressed by turn, not by session": a consumer can now tell which turn an event belongs to instead of inferring it from the session.

## `structuredContent`: correctly budgeted, not a bypass

The spread keeps `error` and `structuredContent` independent rather than mutually exclusive, and gating on `!== undefined` preserves legitimately falsy values (`null`, `false`, `0`), which matters because the MCP spec allows any JSON there.

Since this puts server-supplied data onto a new path, I traced where `details` ends up. It does enter message history — `compaction.ts` deletes it explicitly with a `SECURITY: toolResult.details can contain untrusted/verbose payloads` note — and `tool-result-context-guard.ts` charges it:

```js
const details = (msg as { details?: unknown }).details;
chars += estimateUnknownChars(details);
```

So it's inside the context budget rather than sneaking past it, and compaction still strips it. That was the answer I needed before being comfortable with the change.

## Two light notes

**1. `structuredContent` usually duplicates `content`.** The spec encourages servers to send both — a human-readable rendering plus the machine-readable object — and your own fixture reflects that (`content: [{ text: JSON.stringify(structuredContent) }]` alongside `structuredContent`). So for any server that follows the convention, the same payload is now charged twice against the tool-result budget (which carries its own `TOOL_RESULT_CHARS_PER_TOKEN` weighting) until compaction runs. That's inherent to serving two consumers rather than waste — the model reads `content`, the host reads `details` — but if any bound MCP server returns large structured payloads, this makes those turns hit compaction sooner than before. Worth knowing rather than worth changing.

**2. The fixture names undercut the genericity claim.** The summary says "no product-support schema or completion policy lives in Siclaw", and the code holds to that — but the test uses `"product-support-result"` and `"submit_product_support_result"`, which names the specific consumer this exists for. Neutral names (`example-server`, `some_tool`) would keep the transport-level test from reading like it knows about a product feature. Cosmetic, and it's in a test only.

PR #503, opened 2026-08-19
fix(gateway): correlate supervisor terminal events

LikiosSedo:
## Summary

Follow up #502 by preserving `turnId` when the Runtime supervisor terminates live turns during shutdown or AgentBox replacement. Strict `/run` consumers now receive the real interruption for the affected turn instead of misclassifying it as an unsupported protocol.

### Solution

Emit one terminal pair per unreported live turn, carrying that turn ID. Keep the existing session-only fallback for the cold-start window before a turn ID exists.

## Test Plan

- [x] `npm test` — 250 files, 5209 passed, 2 skipped
- [x] `npm run build`
- [x] focused gateway shutdown/abort and MCP transport tests — 52 passed

### DIFF
--- a/src/gateway/server-shutdown-turns.test.ts
+++ b/src/gateway/server-shutdown-turns.test.ts
@@ -96,6 +96,12 @@ function chatEvents(client: any, sessionId: string): Array<Record<string, unknow
     .map(([, payload]: [string, any]) => payload.event);
 }
 
+function chatEventPayloads(client: any, sessionId: string): Array<Record<string, unknown>> {
+  return client.emitEvent.mock.calls
+    .filter(([channel, payload]: [string, any]) => channel === "chat.event" && payload?.sessionId === sessionId)
+    .map(([, payload]: [string, any]) => payload);
+}
+
 afterEach(() => {
   capturedSignal = undefined;
   vi.clearAllMocks();
@@ -107,7 +113,10 @@ describe("startRuntime — shutdown ends in-flight turns", () => {
     const server = await bootRuntime(client);
     const send = server.rpcMethods.get("chat.send")!;
 
-    await send({ agentId: "a", userId: "u", text: "hi", sessionId: "S" }, { sendEvent: vi.fn() });
+    const ack = await send(
+      { agentId: "a", userId: "u", text: "hi", sessionId: "S" },
+      { sendEvent: vi.fn() },
+    ) as { turnId: string };
     await waitFor(() => capturedSignal !== undefined);
     expect(capturedSignal!.aborted).toBe(false);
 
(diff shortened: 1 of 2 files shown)


jacoblee-io (approved):
Approving. Clean follow-up to #502, and the refactor preserves the reporting-ownership rules rather than just moving the emit into a loop.

## Behaviour boundaries check out

I walked the three cases against the old `ownsTheReport` logic:

- no live turns → old: one session-only pair; new: `[undefined]` → one session-only pair. Same.
- live turns, none unreported → old: `ownsTheReport === false`, nothing emitted; new: `turnsToReport = []`, loop doesn't run. Same.
- live turns with unreported ones → old: **one** pair for the session; new: **one pair per unreported turn**, each carrying its own id. This is the intended change.

`supervisorEndedTurns` still absorbs *all* of `sessionTurns` while only `unreported` gets reported, so each turn's own terminal stays suppressed — the invariant that keeps a second, plain `prompt_done` from reading as a turn that completed normally.

And the abort side still covers everything: `for (const id of sessionTurns) turnAborts.get(id)?.abort()` aborts the full set, which is a superset of what gets reported, so there's no turn that gets declared interrupted without being aborted.

Moving the `try`/`catch` inside the loop is a real improvement beyond the correlation work — one turn's failed emit no longer swallows the remaining turns' terminals, and the warning now names which turn (`turn=${turnId ?? "pending"}`).

The truthiness guard in `...(turnId ? { turnId } : {})` reads slightly loose next to a `string | undefined` type, but it degrades in the right direction: an empty id couldn't be correlated anyway, so falling back to the session-only envelope is the correct behaviour rather than emitting `turnId: ""`.

## One gap: the per-turn behaviour isn't the part under test

The new assertion is exact and good for what it covers:

```js
expect(chatEventPayloads(client, "S")).toEqual([
  expect.objectContaining({ turnId: ack.turnId, event: expect.objectContaining({ type: "stream_error" }) }),
  expect.objectContaining({ turnId: ack.turnId, event: expect.objectContaining({ type: "prompt_done" }) }),
]);
```

But it drives a single `chat.send`, and with one turn the old and new code emit the same two events — the only difference is the added field. So this pins the correlation, not the fan-out that is the actual subject of the change.

The interesting case is two live turns on one session, which is a real state rather than a hypothetical: `sessionTurns` comes from `[...(liveTurnIds.get(sessionId) ?? [])]`, a Set per session, and a second send registers its turn before it can take the session lock. A test with two sends asserting four payloads — a `stream_error`/`prompt_done` pair per turn id, each pair carrying its own — would cover the case where a regression to the old single-pair emit is invisible today. It would also pin the ordering, since `toEqual` on the array is already strict about length and position.

Not blocking, since the shape is right and the boundaries above hold, but that's the test I'd want guarding this.

PR #504, opened 2026-08-20
fix(models): resolve a model's wire settings on every path, not just ours

jacoblee-io:
> **History rewritten** (2026-08-20): squashed from eight commits into one. The branch contained two dead ends — throws that a later commit deleted, and a descriptor-only compat fix that did not move the live failure — and both would lead a reviewer down them. Their reasoning is kept in the commit message and below. The two review threads are unaffected (they are issue comments, not line-anchored); the shas they name are gone, and §1/§2/§3 here map to them.

Four failures with one shape: a per-model **wire attribute** that siclaw assembles itself, where the keys nobody thought about silently take pi's default. Each reaches the operator as a provider 400 several layers from its cause.

## 1. The Anthropic thinking shape — the live 400

```
400: "thinking.type.enabled" is not supported for this model.
     Use "thinking.type.adaptive" and "output_config.effort"
```

pi sends the adaptive shape only when `compat.forceAdaptiveThinking` is true, otherwise the legacy `{type:"enabled", budget_tokens}` — which Claude 4.6+ and the 5 family reject. **We never set the key**, so every reasoning model on those generations fails the moment thinking is on.

**Why here and not in the control plane**, and why that is the opposite of the `maxTokens` answer:

| | `maxTokens` | `forceAdaptiveThinking` |
|---|---|---|
| depends on | the **endpoint** — a gateway can rewrite the ceiling | the **model's API contract** — identical wherever served |
| answerable from the id alone | no (the reason we declined to fill it) | yes |
| pi holds the authority | a table, but a gateway overrides it | yes |

**Why a table lookup alone does not fix it.** pi 0.80.7 knows 14 claude ids and `claude-opus-5` — the one that failed — is not among them. Its table is a release snapshot, so it lags the API *exactly during a launch*, which is when a model is most likely to be added. Resolution order:

| # | source |
|---|---|
| 1 | per-model `compat_overrides` — always wins (pi documents `false` as the opt-out) |
| 2 | pi's own `ANTHROPIC_MODELS` — authoritative where it answers; an entry with no compat means pi's defaults, which for those legacy models are correct |
| 3 | a **generation rule** — a rule, not a mirror: nothing is added when a model ships |
| 4 | the latest generation — deliberately the **inverse** of pi's default |

Step 4 is the judgement call: pi's default protects an old config file, while this boundary receives a live id from a control plane, and the set requiring adaptive is open and growing while the legacy set is closed and shrinking. Failing toward adaptive fails safe; failing toward legacy fails on launch day.

A missing minor version means `.0`, never "unknown" — `claude-opus-4` is real, pre-adaptive and absent from pi's table, so reading it as current generation would 400 it the other way. That inverse risk is stated, not hidden, and is what `compat_overrides` exists for: whitelisted keys, **rejected on write** (an operator typo is reportable exactly once) and **ignored on read** (a value from a newer build must not take a model out of service).

### Applied twice, because there are two paths

The first version of this change touched only `buildProviderModelDescriptor` — the configs assembled here from `model_entries` — **and the 400 did not move.** A control plane answering `config.getModelBinding` supplies the whole `modelConfig`, compat included, and the runtime forwards it verbatim, so that function never runs on the path the failing traffic takes. This is the split CLAUDE.md already warned about in prose ("a rule fixed here is NOT fixed there") and `maxTokensField` had already been fixed twice for it.

`withResolvedModelCompat` closes it at the last point both paths share before pi is handed anything: the turn's config **and every routing candidate's own config** — each candidate is registered from its own, so resolving only the binding moves the failure one candidate down. Only keys the caller left unstated are filled; the control plane's object is copied, not mutated.

`forceAdaptiveThinking` also joins `modelNeedsRebind` — the third field of that shape after `api` and `maxTokensField`, and the loudest: correcting a model to adaptive changes nothing else about it, so skipping the rebind keeps sending the rejected shape. Absent stays `undefined` rather than folding to `false`, or every model without the key rebinds every turn.

`supportsTemperature` rides along because pi reads it from the same object. **Labelled honestly as defensive**: pi's agent path never sets `options.temperature` today, so it is not a fix for an observed failure.

## 2. Registration and lookup belong to the routing runner

The prompt path had its own copy of "register the provider, find the model, set it" above `runPromptWithModelRouting`, and it swallowed both failures: a `registerProvider` throw was a `console.warn`, and `findModel` returning nothing hit an `if (found)` with no `else`. The turn continued on whatever model the session already had — the operator's binding silently ignored.

That block is **deleted rather than hardened**. Registration is inherently per candidate: candidates carry their own `modelConfig`, a caller above the runner can only register the primary's, and a fallback candidate may be on another provider entirely. Having registered it, that caller must invent what a refusal *means* — the judgement the policy already encodes.

An earlier version of this PR resolved the binding early and answered 500. Review caught that it broke ordered fallback for exactly the case fallback exists for. The binding is now passed to `resolveEffectivePolicy` as the single candidate, and merged into the matching candidate of a configured policy — `ModelRouteCandidate.modelConfig` is optional while the top-level one is the documented registration config, so an identity-only policy is valid here, and dropping the binding left the primary unregistered. A candidate's own config always wins and is never lent to a candidate on another provider.

A config refusal classifies as `unknown`, in `DEFAULT_NO_FALLBACK_ON` — **kept deliberately**: a misconfiguration is not transient, and answering from a second provider hides it. An operator who disagrees adds the kind to `fallbackOn`, in the policy rather than hardcoded above it.

Preserved tolerance: a binding with **no** `modelConfig` naming a model absent from the registry keeps the session on its current model — nothing to register means the caller meant an already-existing model.

## 3. Runtime params are applied after the candidate's `setModel`

pi's `setThinkingLevel` clamps to the **current** model, a clamp to `off` on a non-reasoning model is not persisted, and `setModel` then restores that untouched default. So `reasoning_effort` applied before the switch is discarded: a non-reasoning → reasoning binding asking for `high` runs its first turn at the previous level. This is also the order `BrainSession.applyModelParams` documents. Moving `setModel` into the runner without moving the params reversed it; review caught it with a direct probe.

Params now travel with the attempt, per candidate, unconditionally — the level is session state a previous candidate may have moved. Candidate params win, the turn's top-level params are the fallback: hydrated candidates carry *provider* config with no `params`, so reading only the candidate drops the caller's effort on every routed turn.

## 4. The `max_tokens` default was above a live model's ceiling

pi's Anthropic table puts `claude-haiku-4-5`, `claude-opus-4-5` and `claude-sonnet-4-5` at **64000**; the default was **65536**, and pi sends the stored value verbatim where Anthropic rejects rather than clamps. The replacement is pi's own number (16384 / 128000), not a ceiling table of ours — that would be a second source of truth destined to drift.

**A default is only reachable where nobody supplies a value, so changing the constant fixed nothing on its own.** The first version did only that, while the Portal create form pre-filled 65536 and always sent it. Every layer that can name the number now agrees:

- the constants;
- the **form** — blank, and the field omitted unless the operator filled it in (`parseInt("")` would put NaN on the wire, which is not the same as absent). `context_window` stays pre-filled: it already equals the backend default, and too low there costs preflight headroom, not a turn;
- the import dialog's label, which described the backend wrongly;
- the **DDL**, interpolated from the constants rather than restated;
- `setColumnDefault` for an **existing** MySQL table, which nothing else could reach — `CREATE TABLE IF NOT EXISTS` skips it, `safeAlterTable` only ADDs, and `widenColumn` compares COLUMN_TYPE where INT vs INT reads as "no change". SQLite would need a table rebuild, so an existing file keeps its old default; the value is a backstop no writer reads.

Stored **rows** are deliberately not migrated: 65536 can be a considered choice for a model whose ceiling really is that high, and a migration cannot tell that from an unlucky default. Affected rows: `SELECT model_id, max_tokens FROM model_entries WHERE max_tokens > 64000`.

Accepted cost: a 128000-capable model is held to 16384 until someone fills the field in — a truncated answer, not a failed turn. The inverse fails every turn.

## Verification

`npm test` — 251 files, **5251 passing**, 2 skipped. `portal-web` — 23 files, 229 passing. `tsc --noEmit` clean.

Every test pinning new behaviour was confirmed failing without its fix, and **which baseline it fails on differs** — stating that matters, because they are not interchangeable:

| Test | Fails on |
|---|---|
| `claude-opus-5` resolves to adaptive | pre-PR `main` (key absent → the live 400) |
| a control-plane config gets the key filled in | the descriptor-only version (function did not exist) |
| `claude-opus-4` reads as legacy | the version before the bare-minor rule |
| bound model missing → must not answer on the previous model | pre-PR `main` (the old model answered) |
| missing primary + healthy secondary → secondary answers | the early-resolution version (500) |
| `reasoning_effort` survives non-reasoning → reasoning | the version that moved setModel but not params |
| identity-only policy + top-level config → primary registered | same |
| create / PUT / batch persist 16384 when absent | pre-PR `main` (65536, all three) |
| legacy MySQL column default corrected | before `setColumnDefault` existed |

The params tests required the fake brain to **model** pi's clamp / non-persistence / restore-on-switch semantics — a fake that records the last call cannot observe an ordering bug at all, which is how the regression got past me.

### Built artifact

Verified inside digest-pinned pods by **evaluating the resolver against a control-plane-shaped config**, not by grepping for strings — and never by tag, since `:test` is shared and gets rewritten by others:

| input | resolved compat |
|---|---|
| `{api:"anthropic-messages", models:[{id:"claude-opus-5"}]}`, no compat | `{forceAdaptiveThinking:true, supportsTemperature:false}` |
| same with an explicit `forceAdaptiveThinking:false` | `false` — the caller's decision survives |
| `claude-sonnet-4-5` | `false` / `true` — legacy untouched |
| `gpt-5` on `openai-completions` | no thinking keys emitted at all |

Also confirmed the `@earendil-works/pi-ai/providers/anthropic.models` subpath resolves inside the production image (`--omit=dev`), and that the resolver is actually wired into the prompt handler rather than merely exported.

### End to end, on the configuration that failed

```
[model-route] attempt on sicore-custom-e7d3b77f…/claude-opus-5
              (api=anthropic-messages, maxTokensField=max_tokens,
               adaptiveThinking=true, reasoning=true)
[agentbox-http] Prompt completed for session 5b0fde08…
```

| claim | evidence |
|---|---|
| ran on the model that failed | `claude-opus-5` |
| took the **control-plane** path — the one a descriptor-only fix never touches | provider `sicore-custom-…`, minted by the control plane, not a Portal row |
| the fix is in effect | `adaptiveThinking=true` — the key did not exist on this path before |
| the turn succeeded | `Prompt completed`, no `not supported` anywhere in the pod's log |

The pod was created after the deploy and runs the digest under test.

`runAttempt` logging the model per attempt is part of this change, and closes an observability regression the change itself introduced: model identity used to be logged by the prompt path's `setModel` call, which moved into the runner, after which nothing logged it — leaving "which model was this turn on?" unanswerable from stdout, the first question a provider-4xx triage asks. Attempt-scoped, since a fallback runs on a different model than the turn started with. It is also what made this verification conclusive rather than inferential.

## Deploy note

`src/core/model-compat.ts`, `brain-session.ts`, `brains/pi-agent-brain.ts` and `agentbox/http-server.ts` are in the **agentbox** image — that is where the thinking fix takes effect. `src/portal/*` ships in **runtime** (it runs the migrations) and **portal**; `portal-web` in **portal** only. A deployment that only recycles agentbox pods therefore gets the compat and routing fixes but not the column-default migration.

**`imagePullPolicy: Always` only pulls on pod CREATE.** A live agentbox pod keeps running the old code, so testing against an existing session after a deploy proves nothing — the first verification attempt here failed for exactly that reason, on pods that were three days old.


LikiosSedo:
I found four merge-blocking gaps in the current implementation:

1. **The normal Portal add-model flow still persists `65536`.** `portal-web/src/pages/Models.tsx` initializes and resets `modelForm.max_tokens` to `"65536"`, and `handleAddModel` always parses and sends it. Because the API preserves any positive supplied value, the new `DEFAULT_MAX_TOKENS` fallback is never reached on the primary user path. Adding a 64K-ceiling Claude model through Portal therefore still reproduces the failure this PR is intended to fix. The import dialog also still describes the backend fallback as `66K out (default)`. Please align the form/default label with `16384` (or omit an unset value so the API owns the default) and add a create-flow assertion.

2. **The eager setup failure bypasses ordered fallback.** The new throws in `src/agentbox/http-server.ts` return HTTP 500 before `resolveEffectivePolicy` / `runPromptWithModelRouting`. The routing runner already converts a missing model into `model_not_found`, which is a default fallback condition. With the production combination of `modelProvider` + `modelId` + `modelConfig` + `modelRouting`, an invalid or missing primary now prevents a healthy secondary candidate from running. When automatic multi-candidate routing is active, provider registration and lookup should remain owned by `runAttempt`; please add an HTTP-level primary-missing / secondary-success regression test.

3. **The unknown-API test does not model pi 0.80.7 behavior.** Against the exact pinned `@earendil-works/pi-coding-agent` version, dynamically registering a model with `api: "misspelled-api"` still makes `ModelRegistry.find(provider, modelId)` return that model; rejection occurs later when pi dispatches the stream. I reproduced this as `{"found":true,"api":"misspelled-api","all":1}`. The new test forces `findModel()` to return `undefined`, so it only covers an ID/config mismatch and does not prove the claimed unknown-API fix. Please validate against the real registry/API provider behavior and replace the stub-only counterexample with an integration-level regression.

4. **The persistence default and upgrade story remain inconsistent.** `src/portal/migrate.ts` still declares `model_entries.max_tokens DEFAULT 65536`, and existing rows are not remediated. This leaves a second source of truth and means a fresh omitted-column insert or an already-affected deployment retains the unsafe value. At minimum, change the fresh-schema default to `16384` and test it. For existing rows, either implement a narrowly safe remediation or make the required operator correction an explicit, verifiable release step.

The fail-loud provider-registration change and prompt-lock release are good improvements, and the focused tests/CI are green, but the current patch does not yet close the real user path or preserve the established routing contract.

jacoblee-io:
All four confirmed and fixed in `cc1dd74e` + `8d68b697`. I verified each against the code rather than taking or dismissing them, and two of them corrected me on something I had asserted as fact — details below, because the corrections matter more than the patch.

## 3 — you are right, and my premise was wrong

I had written, in code comments, commit messages and CLAUDE.md, that pi's `parseModels` drops a model whose `api` it does not recognise. That is false. In 0.80.7:

- `parseModels` skips a model only when `api`/`baseUrl` is **falsy** — `if (!api) continue`. A misspelled value is not falsy.
- Portal-managed providers do not take that path at all. `registerProvider` → `applyProviderConfig` sets `api: modelDef.api || config.api` and then pushes **unconditionally**.
- `find()` is a plain `Array.find`, so it returns the model. Rejection lands later, in pi-ai's `resolveApiProvider` (`No API provider registered for api: …`).

Which matches your `{"found":true,"api":"misspelled-api","all":1}` exactly.

Two consequences I want stated rather than quietly patched: the test that stubbed `findModel` to `undefined` pinned an **id/config mismatch** and proved nothing about an unrecognised api — your "stub-only counterexample" is the right description — and the fix never covered the unknown-api case, while its error text pointed the operator at `models[].api`, i.e. at the wrong thing. The false claim is corrected in all four places.

A related defect your point exposed: the tolerance test stubbed `findModel` to return `undefined` for **every** id, which also broke resolution of the model the tolerance path falls back to. So the turn exhausted and never ran, and the test passed anyway. It now stubs only the named id and asserts the turn actually ran.

Incidentally, the same reading confirms the *other* half of the story: `applyProviderConfig` passes `contextWindow`/`maxTokens` through with **no** defaults, while the static path applies `?? 128000` / `?? 16384`. pi's two paths genuinely disagree; only the api claim was wrong.

## 2 — you are right, and the routing contract wins

Confirmed: `DEFAULT_FALLBACK_ON` contains `model_not_found` and `runAttempt` already produces it, so failing at 500 above the runner takes away the fallback for exactly the case it exists for.

Checking how to hand it back turned up something stronger than an ordering bug. Candidates carry their **own** `modelConfig` and the gateway hydrates each one (`resolveAgentModelRouting`), so registration is inherently per-candidate: a caller above the runner can only ever register the *primary's* config, and a fallback candidate may live on a different provider entirely. My placement was structurally wrong, not just early.

So both blocks are gone. `resolveEffectivePolicy` now takes the request's binding — with its `modelConfig` — as the single candidate, and `runAttempt` owns registration, lookup and `setModel`. A refusal becomes that candidate's classified setup failure **carrying pi's own message**; a missing model becomes `model_not_found` and the chain proceeds.

One boundary worth naming explicitly rather than leaving implicit: a config refusal classifies as `unknown`, which is in `DEFAULT_NO_FALLBACK_ON`, so it does **not** fall back by default. I kept that: a misconfiguration is not transient, and silently answering from a second provider hides it. An operator who disagrees adds the kind to `fallbackOn` — the decision stays in the policy instead of being hardcoded above it. Tell me if you would rather it fall back by default and I will move it, but that is a routing-policy change, not something to bury in this fix.

Regression tests you asked for:

| Test | Result on the code you reviewed | On pre-PR `main` |
|---|---|---|
| missing bound primary + healthy secondary → secondary answers | **fails (500)** | passes |
| bound model missing, no fallback → must not answer on the previous model | passes | **fails — the old model answered** |
| refused provider config → surfaced with pi's reason, no default fallback | **fails (500)** | passes |

The first row is the regression you caught; the second is the original defect. Both were confirmed failing before the fix, not asserted after it.

Also updated: the `setModel`-throws test. Its subject is "does not deadlock the session", which still holds — it now expects the failure through the route events rather than an HTTP 500, since `setModel` runs inside the runner.

## 1 — you are right, and this is the one that matters most

Confirmed: `modelForm` initialised and reset `max_tokens` to `"65536"` and `handleAddModel` always sent it, so with the API preserving any positive value, the new default was unreachable on the primary path. I changed three backend write paths and never looked at who calls them — the fix was real and the bug was still alive on the entry point operators actually use.

The form now starts **blank** and omits the field unless it was filled in, so the backend default stands (`parseInt("")` would put NaN on the wire, which is not the same as absent). `context_window` stays pre-filled — it already equals the backend default, and too low there costs preflight headroom rather than a turn. `IMPORT_DEFAULT_MAX_TOKENS` and the `66K out (default)` label are corrected; the input carries a placeholder and a title explaining that a value **above** the model's ceiling fails every turn. Four assertions added in `Models.test.ts`, including one that fails if anything ever pre-fills a value ≥ 64000.

## 4 — agreed, with one qualification I would rather state than overstate

Fresh-schema default is now `16384`, asserted by a new test in `migrate-sqlite.test.ts` against the constants *and* independently against the 64000 ceiling.

The qualification: all three write paths do name the column explicitly, so I could not find a current code path that reaches the old column default — it was a second source of truth (worth removing, and removed) rather than an actively firing bug. I mention it only so the fix is not credited with more than it does.

On existing rows I took the explicit-release-step option rather than remediation: a stored 65536 can be an operator's considered choice for a model whose ceiling really is that high, and a migration cannot tell that from an unlucky default. This is now stated in `migrate.ts`, in the CLAUDE.md row and in the PR body.

## Verification

`npm test` — 250 files, 5219 passing, 2 skipped. `portal-web` — 23 files, 229 passing. `tsc --noEmit` clean.

Images built from the previous head were verified by digest-pinned pod; I will rebuild from this head before the deploy state means anything, and the PR body's artifact section is stale until I do.

LikiosSedo:
Thanks for the thorough follow-up. I re-reviewed the exact current head `8d68b697`. The four original gaps are substantially addressed: the Portal create path now omits a blank `max_tokens`, the fresh schema default is 16384, missing primaries are handled inside ordered routing, and the unknown-api explanation now matches pi 0.80.7.

I still found one merge-blocking regression introduced by moving model setup into the runner, plus two contract/release gaps:

1. **[P1] `reasoning_effort` is applied before the candidate model switch.**

   `http-server.ts:887-896` calls `applyModelParams`, while `runAttempt` switches the model later at `model-routing.ts:956-958`. This reverses the explicit `BrainSession` contract at `brain-session.ts:206-212`, which says runtime parameters are applied after `setModel`.

   Against the exact pinned pi 0.80.7, a non-reasoning current model clamps an early `high` effort to `off`; `setModel` then restores the configured/default level. A direct probe produced:

   `{"old_set_then_apply":"high","new_apply_then_set":"medium"}`

   So a non-reasoning -> reasoning binding silently loses the requested effort on its first turn. Fallback candidates also never receive their own `modelConfig.params` per candidate. Please parse/apply params inside `runAttempt`, immediately after the candidate `setModel`, and add non-reasoning -> reasoning plus fallback-candidate regression tests.

2. **[P2] Top-level binding config is discarded when a multi-candidate policy is present.**

   `resolveEffectivePolicy` returns `configured` unchanged at `model-routing.ts:331`, dropping `requestBinding.modelConfig`. But the public `PromptOptions` contract documents top-level `modelConfig` as the dynamic registration config, while `ModelRouteCandidate.modelConfig` is optional. Therefore a valid top-level binding plus identity-only primary/fallback policy no longer registers the primary and can silently skip it as `model_not_found`.

   Portal and the current Sicore resolver hydrate every candidate, so their standard paths hide this regression; the AgentBox API/type boundary does not require that invariant. Please merge the matching binding config into the configured primary, or explicitly validate/reject unhydrated or inconsistent policies at `/api/prompt`, with a boundary-level test.

3. **[P2] Existing databases keep the old column default.**

   The fresh DDL is correctly changed to 16384, but `CREATE TABLE IF NOT EXISTS` does not update an existing `model_entries.max_tokens DEFAULT 65536`, and the additive migration block never alters it. A legacy-schema probe after `runPortalMigrations()` still produced `{"default":"65536","inserted":65536}` for an omitted-column insert.

   Leaving existing row values untouched is reasonable because 65536 may be an intentional model-specific choice. The column default itself is unambiguous, though, and should be migrated independently for upgraded databases. Please add a legacy-schema test distinct from the fresh-install assertion. For already affected rows, the release needs an executable and verifiable audit/remediation step rather than only a prose note.

There are two remaining evidence gaps:

- The create, PUT, and batch API writers still lack route-level assertions that the final persisted value is 16384; the new tests cover the Portal helper and fresh DDL separately, not the full write paths.
- The PR body is stale. It still describes synchronous HTTP 500 behavior, direct `releasePromptLockOnSetupFailure`, the 500-character wrapper, and images built from “this head”. Those statements describe the old implementation/artifacts, not `8d68b697`. Please update the body and rebuild/verify current-head artifacts before treating the deployment section as evidence.

Verification on `8d68b697`: backend 5219 passed / 2 skipped, Portal 229 passed, `tsc --noEmit`, Portal build, and `git diff --check` all passed. GitHub CI is green. I did not verify MySQL or a live deployment.

jacoblee-io:
All three confirmed and fixed in `eb0ac65f` + `b8a86603`, and both evidence gaps closed. I verified each against pi's source and the handlers rather than taking the report on trust — every one held, and the P1 is squarely a regression I introduced.

## 1 [P1] — confirmed, and the mechanism is exactly as you describe

Traced it in pinned 0.80.7 rather than reasoning from the contract:

- `setThinkingLevel` (`agent-session.js` ~1253) clamps to the **current** model.
- The persist is gated on `supportsThinking() || effective !== "off"`, so a clamp to `off` on a non-reasoning model does **not** become the default.
- `setModel` then reads `_getThinkingLevelForModelSwitch` (~1302), which — the old model being unable to think — returns that untouched persisted default.

So `high` → `off` (unpersisted) → restored to whatever was there. Your `{"old_set_then_apply":"high","new_apply_then_set":"medium"}` is what those three lines produce, and it is my regression: before this PR `setModel` ran in `http-server` ahead of `applyModelParams`, which is why the documented order held by accident.

`runAttempt` now calls `applyCandidateModelParams` after its own `setModel`, unconditionally — the level is session state a previous candidate may have moved, so "no rebind" is not "no params change".

On per-candidate params: candidate params win, the turn's top-level params are the fallback. Both directions matter. A control plane hydrates candidates with **provider** config, which carries no `params`, so reading only the candidate would drop the caller's effort on *every* routed turn — while reading only the top level would ignore a genuine per-candidate override. That also closes the fallback gap you named.

I did not put params on `ModelRouteCandidate.modelConfig` for identity-only candidates: `runAttempt` passes `modelConfig` to `registerProvider`, so a config carrying only `params` would fail pi's `validateProviderConfig` for want of a `baseUrl`. Hence a separate option rather than smuggling them through the registration path.

Both regression tests you asked for are in. They needed the fake brain to **model** pi's clamp / non-persistence / restore-on-switch rather than record the last call — a recording fake cannot observe an ordering bug at all, which is precisely how this got past me.

## 2 [P2] — confirmed

`ModelRouteCandidate.modelConfig` is optional while the top-level `modelConfig` is the documented registration config for the turn, so an identity-only policy is valid at this boundary and returning `configured` untouched left the primary unregistered and skipped as `model_not_found`. `withBindingConfig` merges the binding into the candidate that names it.

Two rules I made explicit in code, because they are the reason a blanket merge would be wrong: a candidate's **own** config always wins (more specific), and the binding is **never** lent to a candidate on another provider — that would register the wrong provider under its name.

Your point about the API/type boundary not requiring hydration is the part I had wrong: I checked that both control planes in tree hydrate every candidate and treated that as the contract. It is the reason the gap is invisible from their side, not a reason it does not exist.

Boundary test added, plus a routing-level test that a candidate with its own config is untouched and a bare fallback stays bare.

## 3 [P2] — confirmed, with the SQLite half stated rather than glossed

`setColumnDefault` is new because nothing existing could do it: `CREATE TABLE IF NOT EXISTS` skips an existing table, `safeAlterTable` only ADDs, and `widenColumn` compares COLUMN_TYPE where INT vs INT reads as "no change" — the same blind spot the `api_type` nullability fix hit, which is why that one needed `tightenColumnNotNull`. Guarded on COLUMN_DEFAULT so it is idempotent across per-boot migration runs; a default-only MODIFY is metadata-only on MySQL 8.

SQLite is a deliberate no-op — changing a default there means rebuilding the table, for a value no writer reads — so an existing SQLite file keeps its old default. That is in the helper's doc comment and the CLAUDE.md row rather than left for someone to discover.

Legacy-schema test is separate from the fresh-install assertion, as asked: four cases against a recording fake (corrects 65536 → 16384; issues nothing when already correct; nothing when the column is absent; nothing on SQLite). Asserting *no* SQL is issued matters as much as the ALTER — an unguarded MODIFY would run forever.

I also made the DDL **interpolate** the constants instead of restating them. A hand-copied number is the second source of truth this whole thread is about.

On already-affected rows I am keeping the operator-correction step rather than remediating: a stored 65536 can be a considered choice for a model whose ceiling really is that high, and nothing here can distinguish that from an unlucky default. What I can do — and did — is make the step verifiable rather than prose: the exact rows are `SELECT me.model_id, me.max_tokens FROM model_entries me WHERE me.max_tokens > 64000`. Say the word if you want that as a script in the repo instead of a release note.

## Both evidence gaps

**Write paths.** Route-level tests now assert the parameter actually bound for create, PUT (including the cleared-box `null` → default case) and batch — not the constant the handler reads. All three fail on pre-PR `main`, persisting 65536. You were right that the earlier tests would have passed no matter what the write paths did.

**PR body.** Rewritten against the current head: the synchronous 500, the direct `releasePromptLockOnSetupFailure`, and the 500-character wrapper are all gone from the implementation and from the body. The artifact section now names the digests I actually rebuilt and verified (`7d7be884` / `c44caae8`) and says plainly that they predate these two commits, so they are not evidence for them until rebuilt. It also states what cannot be verified end to end and why.

## Verification

`npm test` — 251 files, 5232 passing, 2 skipped. `portal-web` — 23 files, 229 passing. `tsc --noEmit` clean.

Every new test was confirmed failing without its fix, and the PR body now records *which* baseline each one fails on, since they differ (some fail on pre-PR main, some only on the intermediate versions of this PR).

jacoblee-io:
Heads-up on two things: a **new failure** this PR turned out not to cover, and a **history rewrite**.

## New: a live 400 that the first version of this fix did not touch

On sicore-test, a `claude-opus-5` model returned:

```
400: "thinking.type.enabled" is not supported for this model.
     Use "thinking.type.adaptive" and "output_config.effort"
```

Same class as `maxTokensField`: pi decides the thinking shape from
`compat.forceAdaptiveThinking` (`api/anthropic-messages.js:746`) and we never set the key, so every reasoning model on Claude 4.6+ / the 5 family fails the moment thinking is on. It surfaced now rather than earlier because the params-ordering fix in this branch made the requested effort actually stick — the regression had been silently dropping it, which suppressed the symptom.

I fixed it in `buildProviderModelDescriptor`, rebuilt, deployed — **and the 400 did not move.** A control plane answering `config.getModelBinding` supplies the entire `modelConfig`, compat included, and the runtime forwards it verbatim (`agent-model-binding.ts` returns `data.binding` unchanged), so that function never runs on the path the failing traffic takes. CLAUDE.md already says this in prose — "a rule fixed here is NOT fixed there, and both sides had to fix `maxTokensField` separately" — and I fixed only the side I happened to be reading. The working fix applies the resolution again at the last point both paths share before pi is handed anything, for the turn's config *and* each routing candidate's own config.

Details, including why this one belongs in siclaw when `maxTokens` did not, why a lookup in pi's table is not sufficient on its own (`claude-opus-5` is not among its 14 claude ids), and what the residual inverse risk is, are in §1 of the body.

Two things there are worth your eye specifically, since both are judgement calls rather than facts:

- **the default for an id nobody recognises is adaptive** — deliberately the inverse of pi's own default, on the reasoning that pi's protects an old config file while this boundary receives a live model id, and the adaptive set is open while the legacy set is closed. If you would rather it default to legacy and require an explicit opt-in, that is a one-line change and a defensible call in the other direction;
- **a config refusal still does not fall back** (`unknown` ∈ `DEFAULT_NO_FALLBACK_ON`), which I flagged last round and am keeping unless you say otherwise.

I also found one defect in my own rule while reviewing it: `claude-opus-4` / `claude-sonnet-4` are real, pre-adaptive, and absent from pi's table, and my generation rule read a missing minor version as "unknown" → current generation → would have 400ed them in the opposite direction. A missing minor now means `.0`. Test added.

## Verification, and a correction about the previous round

Artifacts are now verified by **evaluating the resolver inside a digest-pinned pod against a control-plane-shaped config**, not by grepping `dist/` for strings — see the table in the body. That change was forced by the failure above: string checks confirmed the code was present while it sat on a path production does not take.

A correction to how I reported the earlier deploy: `imagePullPolicy: Always` only pulls on pod **create**, and the agentbox pods I asked for a retest against were three days old. So the "still 400" result from that round was not evidence about the fix at all. Live agentbox pods now run the current image, and end-to-end confirmation is still outstanding as of writing.

Also corrected in the body: it previously described a synchronous HTTP 500 and a 500-character wrapper that no longer exist.

## History rewrite

Squashed the eight commits into one, and force-pushed. The branch had two dead ends — throws that a later commit deleted, and the descriptor-only compat fix above — and a reviewer walking the series would be led down both. Their reasoning is preserved in the commit message and the body rather than as commits.

Your two comment threads are unaffected (they are issue comments, not line-anchored), but the shas they name are gone. The mapping:

| old | now in |
|---|---|
| `f7770669` + `cc1dd74e` (refuse-as-error, then handing the binding to the runner) | §2 |
| `eb0ac65f` (params after setModel, binding config merge) | §3 |
| `6cfac49b` + `8d68b697` + `b8a86603` (max_tokens across layers, column default) | §4 |
| `38042970` + `63d0c450` (thinking shape; then the control-plane path) | §1 |

I kept it as one commit rather than splitting by concern because six files carry hunks from more than one part, so a split would need hunk surgery with intermediate commits at risk of not passing tests. If you would rather review four separate commits, say so and I will do the surgery properly.

`npm test` — 251 files, 5251 passing, 2 skipped. `portal-web` — 229 passing. `tsc --noEmit` clean.

jacoblee-io:
End-to-end confirmed on sicore-test, on the exact configuration that produced the 400.

```
[model-route] attempt on sicore-custom-e7d3b77f…/claude-opus-5
              (api=anthropic-messages, maxTokensField=max_tokens,
               adaptiveThinking=true, reasoning=true)
[agentbox-http] Prompt completed for session 5b0fde08…
```

That single line settles what the earlier rounds could not:

| claim | evidence |
|---|---|
| it ran on the model that failed | `claude-opus-5` |
| it took the **control-plane** path — the one a descriptor-only fix never touches | the provider is `sicore-custom-…`, minted by the control plane, not a Portal row |
| the fix is actually in effect | `adaptiveThinking=true` — the key did not exist on this path before |
| the turn succeeded | `Prompt completed`, and no `not supported` anywhere in the pod's log |

The pod was created after the deploy and runs the digest under test, so this is not the earlier trap of testing against a three-day-old pod.

## One addition since the last push

`runAttempt` now logs the model each attempt runs on, with the wire settings that decide the request's shape. This closes an observability regression **I introduced**: model identity used to be logged by the prompt path's own `setModel` call, that call moved into the runner, and for a while nothing logged it at all — leaving "which model was this turn on?" unanswerable from a box's stdout, which is the first question a provider-4xx triage asks. It is attempt-scoped, not turn-scoped, because a fallback runs on a different model than the one the turn started with.

It is amended into the same commit rather than added as its own: the change is what removed the line, so restoring it belongs to the same change rather than to a separate one a reviewer has to correlate.

That log is also what made this verification conclusive instead of inferential. The round before it, all I could honestly say was "a Chinese turn completed with no thinking error" — true, but silent on which model served it.

`npm test` — 251 files, 5251 passing, 2 skipped. `tsc --noEmit` clean. Artifacts verified in a digest-pinned pod before deploying.

jacoblee-io:
Both confirmed and fixed in `54848d44`. One correction to the provenance of P1 — it changes who owns the bug, not whether it is real.

## P2 — you are right, and my reasoning was scoped to Claude while the code was not

Reproduced exactly as reported: `minimax-m2.1` resolved to `forceAdaptiveThinking: true`.

The mistake is clean to state. My justification for "unknown → latest generation" was about *Claude's* release cadence — pi's table lags a Claude launch, the adaptive set is open and growing. All true, and all about Claude. But `anthropic-messages` is a **protocol other vendors implement**, and the code applied that Claude-shaped default to every unrecognised id on it. Telling a MiniMax model to use adaptive thinking is not a cautious guess; it is a guess about a different product, and it is exactly the payload such an endpoint rejects.

"Unknown id" now splits in two:

| id | resolution |
|---|---|
| unrecognised **Claude** id (`claude-neo-preview`) | latest generation — the original reasoning, now correctly scoped |
| not recognisably Claude (`minimax-m2.1`, `glm-4.6`) | **nothing** — pi's own defaults stand |

`withResolvedModelCompat` also had to stop writing keys the resolver did not answer: an explicit `undefined` still *adds* the key, which is enough for `modelNeedsRebind` to see a change and, worse, reads as a decision we never made.

The cost, stated rather than hidden: a Claude model renamed past recognition (`opus-5-fast`) is indistinguishable from a MiniMax id, so it needs the explicit override. At that point that is the honest answer, and it is what `compat_overrides` is for. If you would rather also match bare family names (`opus-*`, `sonnet-*`) I can — I left it out because a vendor naming a model "sonnet" is not impossible and a false positive there is a 400.

## P1 — the mechanism is exactly as you describe; the provenance is not

Verified every step:

- `runAttempt` unsubscribes in `finally`, i.e. when `brain.prompt()` returns;
- the live SSE subscription is gated on `_routeBrainEventsThroughExtra`;
- `onPromptFinish` *explicitly* anticipates the state — it re-subscribes to wait for `agent_end` / `auto_compaction_end` / `auto_retry_end`;
- `setIsCompacting(false)` in `usePilotChat` has **no other caller**, so a lost `auto_compaction_end` pins the UI in "compacting" for the life of that hook state.

So the window has no consumer at all: not the runner (unsubscribed), not the live path (gated). Fixed by lifting the diversion once the prompt resolves — it exists only while a fallback could still discard an attempt, and after resolution nothing can be discarded. The same flag gates `_eventBuffer`, so lifting it also restores replay for a late reconnect in that window.

**The correction:** this PR did not route ordinary prompts through the runner. `origin/main` already has `_routeBrainEventsThroughExtra = effectivePolicy !== undefined` (`http-server.ts:1085`) and the same `finally { unsubscribe() }` (`model-routing.ts:1043`) — the unified entry landed in 2026-06, before this branch. So the bug affects **every turn with a current model on main today**; this PR widens it only to the extra edge where a binding exists but the session has no current model yet. I fixed it here because the code is already open in front of me, not because this PR caused it — and it is worth knowing that reverting this PR would not remove the bug.

The new test asserts through the same gate the consumers use: a prompt resolving with compaction still in flight must leave the diversion off, and an event emitted in that window must land in `_eventBuffer`. No SSE harness exists in that file, and inventing one to assert a flag would have been more machinery than evidence.

## Verification

`npm test` — 251 files, **5254 passing**, 2 skipped. `portal-web` — 23 files, 229 passing. `tsc --noEmit` clean.

All three new tests confirmed failing on `b235d490`: the two P2 ones on the resolved value, the P1 one on `_routeBrainEventsThroughExtra` still being `true`.

Note the earlier end-to-end confirmation stands on `b235d490`, not on this commit — the P2 change alters what a **non-Claude** model resolves to and leaves `claude-opus-5` exactly as verified, but I have not re-run the live check on this build.

jacoblee-io:
Confirmed and fixed in `844f0993`. You were right that my previous fix only narrowed the window instead of closing it.

## The gap

I traced the timing rather than assuming: from `runAttempt` returning to `runPromptWithModelRouting` returning is **fully synchronous** on the success path (no awaits between the attempt's `finally` and the runner's `return`). So the gap is precisely one microtask boundary — the attempt unsubscribes, the runner resolves, and only then does the caller's `.then` run and lift the diversion. Anything the brain emits in between reaches neither owner. And as you note, `session.ts`'s own subscription is *not* gated, so `isCompacting` still flips to false, `onPromptFinish` sees an idle session, closes the SSE, and the frontend is left pinned with nothing that can ever clear it.

No amount of moving the flip earlier in the caller fixes that, because the caller has no code point inside the window.

## The fix

The handoff now happens **inside the attempt**, via a new `onEventCaptureChange` option the runner drives:

```
onEventCaptureChange?.(true);          // then subscribe
…
onEventCaptureChange?.(false);         // adjacent statements — no await between
unsubscribe();
```

Two adjacent synchronous statements on a single-threaded runtime leave no instant without a listener. That is the difference between closing this and narrowing it.

It is **paired, not one-shot**, deliberately: the next candidate has to capture again, or a fallback that is supposed to buffer would stream live and defeat the buffering the rollback machinery depends on.

## About the test — and a correction to my own first attempt at it

The new test asserts the **ordering** (`capture:on` → `capture:off` → `unsubscribe`), not the interleaving. That is not laziness; the loss window is genuinely unreachable from a prompt-scoped mock:

- an event queued from inside `brain.prompt` (e.g. `queueMicrotask`) is enqueued *before* the awaiting continuation, so it fires while the runner is **still subscribed** and travels the runner's own channel — never lost, in old code or new;
- anything scheduled as a macrotask (`setTimeout`) lands after `.then`, which the old code also survived.

I found this the hard way: my first version of this test asserted `_eventBuffer` for a `queueMicrotask` emit and **failed against the fix itself** — it was measuring the wrong channel, not a defect. The ordering invariant is what actually holds and it fails on `54848d44` with `['unsubscribe']` alone, since the hook did not exist there.

You said a timing test reproduces it reliably on your side. If your harness can reach that window directly, I would take that test over mine — the invariant test is the strongest thing I could write without one, not the best possible.

The `http-server` test is kept but re-scoped to what it can honestly show: the gate is already open when `brain.prompt` returns, the close is still deferred, and an event arriving then lands in the buffer.

## Verification

`npm test` — 251 files, **5256 passing**, 2 skipped. `tsc --noEmit` clean. `portal-web` unchanged since the last run (229 passing).

Both prior rounds' fixes are untouched by this: the compat resolution and the non-Claude guard are unchanged, and the earlier end-to-end confirmation of `claude-opus-5` remains valid for that path. I have not re-run the live check on this build — the change is confined to event-channel ownership and does not touch request construction.

jacoblee-io:
This finding is byte-identical to the previous round's, including the line numbers, and it does not match the current head — I think the run was against a stale revision. Evidence, so you can check rather than take my word:

`844f0993` (pushed before your comment, confirmed by `git ls-remote`) already moved the handoff inside the attempt:

```
src/core/model-routing.ts
1152:    onEventCaptureChange?.(false);
1153:    unsubscribe();
```

The cited `model-routing.ts:1126` now holds `await brain.prompt(text, media)`. The caller no longer waits for `.then` either — the runner owns the flag through `onEventCaptureChange`, and the two statements above are adjacent with no await between them, which is what closes the window rather than narrowing it. The comment's file path is the main checkout (`~/project/code/go/src/github.com/scitix/siclaw/src/...`), not a worktree on this branch, which may be why.

Please re-run against `f19e6d34`.

## That said, re-checking found one remaining instance of the same class — closed in `f19e6d34`

The pairing I added covered the prompt subscription, which left the **setup exits** — `model_not_found` and a setup throw — returning without handing delivery back. Those never captured, so there is nothing of the attempt's to give up; but the *caller* diverted delivery before the run began, so a quiet exit leaves it diverted across the next microtask. Same class, one level narrower.

Not theoretical: a primary that has already run a prompt can still be compacting while a **fallback** candidate fails setup, and that compaction's end event lands in exactly that window. Same consequence — `setIsCompacting(false)` has no other caller.

The invariant is now unconditional: **every** exit from `runAttempt` hands delivery back, with a test per exit.

## On the test, and a standing offer

You mention a temporary timing test reproduces the window reliably. I could not build one that does, and I want to be precise about why rather than leave it as "hard to test":

- an event queued from inside `brain.prompt` (`queueMicrotask`) is enqueued **before** the awaiting continuation, so it fires while the runner is still subscribed and travels the runner's own channel — not lost, in old code or new;
- a macrotask (`setTimeout`) lands after `.then`, which the old code also survived.

My first attempt at that test asserted the wrong channel and failed against the fix itself. So I pinned the **ordering** invariant instead (`capture:on` → `capture:off` → `unsubscribe`), which fails on `54848d44` with `['unsubscribe']` alone.

If your timing test does reach the window, please share it — I would rather ship your reproduction than my invariant test, and if it still fails on `f19e6d34` then there is a third instance of this I have not found, which is worth more than either test.

## Verification

`npm test` — 251 files, **5257 passing**, 2 skipped, across four consecutive runs. `tsc --noEmit` clean.

One thing worth flagging rather than hiding: an earlier run of the suite reported a single failure that I could not identify (the summary line showed it, my grep missed the name), and four subsequent full runs were clean. So there is a flake somewhere in the suite that I have not pinned down. It is not in the files this PR touches — those specific suites pass deterministically — but I am not going to claim a clean bill of health I did not actually establish.

nightmeng (approved):

PR #505, opened 2026-08-20
Require data pages for complete-set inventories

LikiosSedo:
## Summary
- Compile playbook now **requires** a data page (every source row) for catalogs Q&A will treat as a complete set.
- Ordinary ticket/log/inspection dumps stay optional two-page layering — the compiler still judges those.
- Sample rows plus a raw pointer is explicitly **not compiled**. Internal KBs must not hide catalog IDs / lookup tables under redaction, ROI, or "not the current decision set".
- English closer now matches Chinese: a complete inventory **must get** a data page (the previous sentence ended at "it must").
- Same rule mirrored in zh/en box_role and the starter constitution knobs.
- Examples stay domain-neutral: no product-family or current-topic special cases.

## Test plan
- [x] Playbook zh/en contain the new "sample + pointer ≠ compiled" language
- [x] English table cell finishes as "it must get one"; Chinese keeps "必须建"
- [x] Tests pin ordinary-collection judgment vs complete-set every-row must
- [ ] Live compile after `siclaw-kbc-box` rebuild + `SICLAW_COMPILE_BOX_IMAGE` bump


jacoblee-io (dismissed):
Approving. This is a well-constructed prompt fix, and I want to be specific about why, because the technique here is better than "we made the rule stronger".

## What makes this likely to actually work

**It removes the authority rather than arguing against it.** The old table handed me `which sources deserve layering` under "**I can** — settle it", and the old section closed with "*Which sources deserve this … are yours to judge — no rule decides it for you*". As a model reading that, the skip is not a violation, it is the delegation being exercised. Adding a competing rule elsewhere would have left the grant in place; deleting it from that column and appending `Whether a complete inventory gets a data page is **not** in this column — it must` closes the actual path.

**It names the excuses instead of restating the requirement.** "because it is long", "because transcription might drift", "because the rows are not the current GPU decision set", "because they are CPU/storage/management SKUs" — these read like rationales that were actually produced, and enumerating them is far more effective on a model than a stronger imperative. Same for `Do not stretch "avoid copying large tables" into hiding the shelf the user needs to filter`, which pre-empts borrowing an unrelated instruction as cover.

**It gives the mechanism, not just the rule.** `Q&A searches the wiki only; it will not follow a pointer into the source tree. "Does not exist" can be true of the excerpt and false of the source inventory.` That generalizes to cases the list doesn't enumerate, which a bare prohibition wouldn't.

**It closes the two bypasses in `constitution.md`, not only the playbook.** An owner could otherwise route around a stricter playbook via the `Sensitive` knob (declare machine IDs sensitive) or `Leave-blank` (mark the catalog not-to-be-compiled). Both are now explicitly out of scope for a complete-set inventory, with credentials still excluded. Fixing the playbook without those two would have left the rule optional in practice.

**The redaction boundary is drawn in the right place** — machine IDs, config lists and internal codes are not redaction targets while passwords / BMC / keys still stay off the page. That's the distinction that makes the rule safe to follow rather than a conflict for the model to resolve on its own.

I checked zh against en element by element — must-not-may, every-row coverage, sample-plus-pointer ≠ compiled, the wiki-only mechanism, the removed grant plus its negation in the table, the exclusion clause, and the internal-KB redaction boundary all appear in both, as does the `box_role` paragraph (condition, examples, full-data-page requirement, the "redaction / ROI / not-current-decision-set" refusals, credential carve-out). The line-count gap is Chinese being more compact, not missing content.

## Two small asymmetries

**1. zh drops two of the four named excuses.** en refuses both `not the current GPU decision set` *and* `they are CPU/storage/management SKUs` as separate rationales. zh compresses these into one — `非当前决策高频 / 纯 CPU 不像 GPU 选型` — which keeps the CPU case but loses storage and management. Since the whole power of this passage comes from enumerating excuses that were actually used, a zh compile facing a pure storage or management catalog has one fewer pre-refuted rationale than an en one. `纯 CPU / 存储 / 管理类 SKU` would restore parity.

**2. The zh assertion pins a whole sentence; en pins two fragments.**

```python
assert "摘若干「决策高频」样例 + 其余指路回 raw ≠ 编完" in zh_const
assert "high-frequency decision" in en_const and "is not compiled" in en_const
```

en survives rewording of everything between those two fragments. zh breaks on any edit to that line, including punctuation — and prompt copy is exactly the kind of text that gets tuned repeatedly (this PR is that). Two short zh fragments would give the same coverage with room to iterate.

Worth noting what these tests do and don't prove: they confirm the rules are packaged into the starter constitution, which is the right floor. Whether the rule changes behaviour is the third, unticked item in your test plan, and that can only be observed on a real compile.

## Deploy note

The prompt packs ship inside the box image, so this needs a `siclaw-kbc-box` rebuild and a `SICLAW_COMPILE_BOX_IMAGE` bump before any compile picks it up — and per the kbc contract, a rolled-out runtime won't pick up a new image for existing live sessions. Worth stating alongside that third test-plan item, since "next hardware incremental compile" only exercises the new rule after the image lands.

CI green across Test / Type Check / Portal Web Test / KBC Box Test.

jacoblee-io (approved):
Re-approving after the force-push dismissed my earlier review. Both new commits are improvements, and one of them fixes something I should have caught and didn't.

## The scope bug I missed

`d6224df9` is the important commit. The version I approved applied `must` to the whole *collections of records* category — and that category opens with ticket exports, on-call logs and inspection runs. So a 500k-row ticket export was also required to become a data page "covering every row of the source table", while the same section, a few lines further down, caps a page at 5MB / ~20,000 typical ticket records. The rule contradicted its own capacity limit.

Splitting it fixes that properly:

- ordinary collections (tickets, logs, inspection runs, a dump nobody will filter as "the complete list") → **can** become two pages, and *which* ones deserve it stays a judgement call;
- a **complete-set inventory** → **must**, with every row covered.

That also restores the discretion the earlier draft removed too broadly. The original problem was never "the model exercises judgement about ticket dumps" — it was "the model exercises judgement about a canonical inventory". Now the grant is removed exactly where it caused the miss and left intact everywhere else.

I read the previous version by following your description of it — canonical inventories must get a data page — and didn't check which noun `must` actually attached to. Worth saying plainly since I approved it.

## Domain-neutral examples resolve my first note, better than I proposed

I'd flagged that zh listed one excuse category (`纯 CPU 不像 GPU 选型`) where en listed three (`CPU/storage/management SKUs`), and suggested restoring the missing two. `78e465c5` instead abstracts both sides — `because they belong to a category the current brief did not highlight` / `某类行不像当前主题` — which covers every category rather than three, so the two locales are equivalent by construction instead of by matching enumerations. It also gets business-specific vocabulary out of a prompt pack that ships in this repo, which is the direction #499 established.

The same substitution runs through `constitution.md` and both `box_role` files (`entity ID overviews`, `Catalog IDs, lookup tables`, dropping `BMC` from the credential list), and the sensitivity rule survives it intact — IDs and lookup tables are still explicitly not redaction targets, credentials still stay off the page.

## The prose-pinning note, once more and lightly

Assertions went from two to six, and the four new ones are short paired fragments — `"ordinary collection"` + `"yours to judge"`, `"complete-set inventory"` + `"every row of the source table"`, and the zh equivalents. That's the shape I was hoping for: they pin the two-tier structure without pinning its wording.

The other two new assertions are full sentences, on both sides now:

```python
assert "Whether a complete inventory gets a data page is **not** in this column — it must get one" in en_const
```

This diff is its own evidence: that line changed from `it must` to `it must get one`, so the copy edit forced the test edit. Not worth another round on its own — just noting the pattern held true the first time the sentence was touched.

Everything from my previous review still applies: the grant is removed from the table's "I can settle it" column rather than argued against, the excuses are enumerated rather than the imperative strengthened, the wiki-only mechanism is stated so the rule generalizes, and the `Sensitive` / `Leave-blank` knobs in `constitution.md` are closed so the playbook rule can't be routed around.

Deploy note unchanged: prompt packs ship inside the box image, so this needs a `siclaw-kbc-box` rebuild plus a `SICLAW_COMPILE_BOX_IMAGE` bump before any compile sees it, and existing live sessions won't pick it up.

CI green across Test / Type Check / Portal Web Test / KBC Box Test.

PR #507, opened 2026-08-21
fix(tools): the plan and a hung command were eating 43% of triage time

jacoblee-io:
## What this changes

Two independent costs, both found by measuring rather than reading: the plan spent
a model round-trip per task, and a command with a timeout could hang for hours.
Together they account for **43% of executor wall time** in the measured month.

Measurement basis: the full history of the two executor agents (`srebot-国内`,
`srebot-海外`), 2026-07-09 → 08-21 — 1411 traces, 40321 tool calls, 69.2 hours,
every trace's message bodies read for per-call `durationMs` rather than sampled.
Time is attributed per **model round-trip**, defined as a batch of tool calls
whose start times fall within 2s of each other; the gap between batches is the
round-trip. Headline: **75% of the time is model round-trips, 25% is tools
executing.** The tools are not slow. The number of round-trips is the cost.

## 1. The plan cost a round-trip per task — 31% of wall time

- 9841 `task_*` calls (`task_update` 6068 / `task_create` 3434 / `task_list` 339)
- 8155 turns whose only tool call was `task_*`, **96% of them carrying exactly one**
- each paying a 5.5s median round-trip for a 0.1s ledger write
- **19.4h, 31% of total wall time**, against **0.15h** spent inside the tools
- in conversations that used a plan, it was a median **40%** of all their tool calls
- `c71561c4`: 604 seconds, 26 round-trips, 15 of them status-only, **1 second** of
  actual tool execution

Two causes, and the second is the bigger one:

**The schema took one task at a time**, so a five-step plan cost five round-trips.
Both tools now take an array. The single-item shape still lands — these tools have
no non-model caller, so it is kept not for compatibility but because bouncing a
call in the old shape costs exactly the round-trip the array form exists to save.

**The instruction said to do it.** `do not batch completions` was written against a
real failure (sitting on a finished task leaves the plan stale and anything
`blockedBy` it waiting), but it is about *timeliness* and was read as being about
*grouping*. A round-trip can carry several tool calls, so promptness never
required a separate turn: marking a task done and issuing the next `kubectl` is
one message. **92% of `task_*` calls were not co-emitted with any real tool call.**
Replaced, in both the tool description and the system prompt, with the rule that
actually holds.

Batch semantics are chosen so a partial result is never ambiguous: `task_create`
validates every entry before creating any (a half-written plan is worse than a
rejected call — the model cannot tell which half landed); `task_update` reports
per entry, and one unknown id among several good ones is not a failed call. One
`task_event` per task either way, so the plan panel still ticks task by task —
only the round-trips are shared.

## 2. A timeout that could not stop the command — 12% of wall time

Three conversations were held for **12280s, 12106s and 6465s** by a single kubectl
each — 8.6 hours, from three commands, each with a timeout configured the whole
time. Their tool-call counts were 33 / 24 / 38: not complex triage, a hung
command nobody could interrupt.

Three stacked reasons, and only the first was visible in the code:

1. `exec`'s own `timeout` sends SIGTERM to the **direct** child. In production
   that child is `sudo`, not the kubectl underneath — and a command can ignore
   SIGTERM outright.
2. The result settled on `close`, which fires when the stdio streams close, not
   when the process exits. A surviving descendant holding stdout keeps the pipe
   open, so `close` never fires and the call outlives its timeout without bound.
3. The escape hatch for both was `process.kill(-child.pid)` — kill the process
   group that `detached: true` was supposed to make the child leader of. **`exec`
   silently drops `detached`.** It is a spawn option; `exec` → `execFile` →
   `spawn` forwards only cwd/env/uid/gid/shell/signal/windowsHide. Measured
   before changing anything: the child's PGID equalled the **node process's**
   group, identically with `detached` true and false, so `-child.pid` was ESRCH
   on every call and always fell back to killing the direct child alone. The
   group kill had never worked, on any platform — and the real pgid could not be
   substituted, since signalling it would have killed the agent's own group.

The abort path used the same kill and had the same hole: a reproduction test on
that path hung for the full 60s cap before this change.

The exec call moves to `bounded-exec.ts`, which spawns (so `detached` is
honoured), runs its own timer, SIGKILLs the group, and **bounds the wait after the
kill** so the call settles even where the kill cannot land — `sudo` can place a
command in a session of its own. A fix that only works when its assumption holds
is what produced this bug in the first place.

`restricted_bash` was the only command tool still on `exec`: `local_script`,
`pod_script`, `bg-session` and `background-bash-runner` already spawn, and
`local_script` already had this exact timer-plus-group-kill shape. This brings the
highest-traffic one (10319 calls in the measured month) in line with them.

### Interaction with #493

Moving off `exec` stopped Node producing the two error shapes `classifyExit` reads,
and each would otherwise have been **silently reclassified**:

| case | would have become | why that is wrong |
|---|---|---|
| timeout | `channel_error` | "the command could not be started, so the target never ran it" — for a command that ran and was killed |
| output overflow | loses `output_truncated` | a cut-off prefix reads as a complete result; a search over it finding nothing then "proves" something — the exact false negative #493 cites |

So a timeout arrives as `code: null` / `signal: "SIGKILL"` and an overflow as
`ERR_CHILD_PROCESS_STDIO_MAXBUFFER`, both pinned by tests since these fields exist
only to be read by another module. The bespoke timeout branch this branch
originally carried is gone: #493's `interrupted` annotation is better — it names
what the tool *cannot* know, whether the time went to the apiserver proxy, an
unresponsive kubelet, or a log stream that never closed. The only thing added
rather than delegated is the cap in force, which `interrupted` advises raising
without being able to say what it currently is.

## Behaviour changes worth naming

- **stdin is now `/dev/null`** rather than an open pipe nobody writes to, so a
  command reading stdin gets EOF instead of blocking until the cap.
- **`maxBuffer` is explicit accounting**, since spawn has none. It counts BYTES via
  `Buffer.byteLength` (`.length` counts UTF-16 units and would let CJK output run
  ~3x past the cap) and stops collecting once passed, so a producer that outlives
  the kill cannot grow the string through the grace window.
- **An already-aborted signal no longer spawns the command** at all.
- An abort is classified `interrupted`, whose wording says "at the tool's
  timeout". #493 covers both under that branch deliberately ("a timeout kill, or
  the abort reap"), so it is left alone.

## How this was checked

`npx tsc --noEmit` clean; **5931 tests pass**, including #493's 268
`restricted-bash` tests unmodified — the move to `spawn` does not touch any
assumption they make.

Each guard was verified by reverting it:

- `detached: false` → the group-membership assertion and the
  grandchild-holds-stdout case both fail (3304ms against a 1500ms bound), so those
  test the **kill**, not the grace behind it
- removing the pre-spawn abort check → its test fails
- **one exception, marked as such in the test file**: the stop-collecting guard is
  not observable here. With a working group kill the process dies before another
  chunk arrives, so the test that looks like it covers that guard passes either
  way and says so rather than implying otherwise. It defends the survivor path,
  which needs `setsid` to reproduce and therefore only exercises on Linux.

## Deployment

Contains **agentbox-image** changes (`src/tools/`, `src/core/prompt.ts`), so
deploying needs an agentbox rebuild, an image-reference bump, and a pod recycle —
not a runtime-only roll.

`src/core/prompt.ts` is one clause of one bullet (+1/−1), approved separately
because that file is behaviour for every agent on every turn.

## How to tell whether it worked

The baseline for every number above is on hand, so re-running the same
measurement after deploy gives a clean A/B:

| metric | baseline | expected |
|---|---|---|
| turns whose only call is `task_*` | 8155 | down sharply |
| plan share of attributable time | 31% | down sharply |
| model round-trips per conversation | 17.1 | down |
| tool calls per round-trip | 1.52 | up |
| `details.timed_out` frequency | n/a | non-zero, and no single command past ~300s |

The two halves are separable in the data: `task_create` batching shows up as
fewer create calls per conversation, co-emission as fewer plan-only turns.

## Not in this PR

The same measurement found larger structural costs that need their own work — the
absence of a one-shot object-snapshot primitive (7 alert scenarios, 202
conversations, 13.5h, 56–75% repeated probes), cluster-identity resolution done as
a document the model reads 3.4 times per conversation (57% of all skill reads), and
diagnostic skills that carry prose but no scripts (1.3% of post-skill calls run a
script, 32.6% hand-assemble bash). Those are tracked separately.


LikiosSedo:
I reviewed the current head `79f8bc8` end to end, including task mutation persistence/replay and the production AgentBox UID boundary. I do not think this is ready to merge yet.

### Blocking findings

1. **Batched task events can be persisted out of order.**

   `task_create` / `task_update` synchronously emit multiple events, but `emitExtraEvent` starts an independent fire-and-forget `persistTaskEvent()` request for each one (`src/agentbox/session.ts:2605`). There is no per-`taskListId` queue or explicit mutation sequence. Portal sequence allocation can tie under concurrent appends, while Sicore can only sequence request arrival, not Siclaw emission order. Since plan replay is last-write-wins, a refresh can reorder newly created tasks, apply an older status after a newer one, or resurrect a task after a delete.

   Please serialize task-event persistence per task list, or attach a monotonic mutation sequence and replay by it. A regression test should delay/reorder the first append and verify that persisted replay matches the live ledger.

2. **A pending coalesced write can replace the durable ledger with an empty snapshot after session close.**

   The follow-up in `persistLedgerSnapshot` re-reads state through `getOrCreateLedger()` (`src/agentbox/session.ts:2036-2048`), while `close()` deletes that ledger at line 3282 without draining or cancelling the writer. If close happens while the first write is in flight, the pending follow-up recreates an empty ledger and renames `[]` over the durable snapshot that close explicitly intends to preserve. A minimal reproduction observed snapshots `[ ["1"], [] ]`.

   The coalescer needs per-key drain/cancel lifecycle ownership, and `close()` should await the final snapshot before deleting the in-memory ledger. Please cover batch -> blocked first write -> close with a regression test.

3. **The production timeout path can return while the actual sandbox command is still running.**

   `boundedExec` sends `SIGKILL` to the process group as the `agentbox` user (`src/tools/cmd-exec/bounded-exec.ts:128-188`), but production `restricted_bash` runs the real command through `sudo -u sandbox` (`src/tools/cmd-exec/restricted-bash.ts:540-543`), and the pod does not grant `CAP_KILL`. The group signal can kill the agentbox-owned shell/sudo while leaving the different-UID sandbox descendant alive. The catch fallback does not help when the group signal killed at least one eligible process. After the grace period, the implementation only destroys pipes and calls `unref()`; it does not verify or terminate the surviving command. The escaped-descendant test asserts only that the Promise settles, not that the descendant PID is gone.

   Please enforce the deadline through a sandbox-side supervisor/reaper or another retained killable boundary, and add a real AgentBox-image test for timeout, abort, and overflow that asserts the sandbox PID no longer exists after settlement.

### Additional correctness issue

An explicit abort is currently reported as a tool timeout. Both abort paths create `code=null, signal=SIGKILL` (`bounded-exec.ts:116-118,216`), and `exit-classification.ts:275-294` hard-codes that shape as "killed at the tool's timeout" and advises increasing `timeout_seconds`. Structured details do not set `timed_out`, so the text and result contract disagree. Please add a distinct aborted cause/error and test the complete `restricted_bash` result for live and already-aborted signals.

The exact-head CI is also still red: Type Check and Portal Web Test pass, but the full Test job has one failing mTLS assertion (`5943 passed, 1 failed, 2 skipped`). That failure looks independent/flaky because the focused mTLS test passes locally, but the current merge gate has not recovered.

nightmeng (approved):

PR #511, opened 2026-08-24
feat(knowledge): bind citations to exact evidence

LikiosSedo:
## Summary
- bind claim-level `evidence_refs` to stable source IDs; runtime re-derives the URL from the frozen manifest
- fail closed on malformed, missing, over-limit, or remount-raced evidence instead of substituting or pairing across generations
- accept `evidence_refs` and unmarked `pages` in one call; crossing 8 originals fails the whole call
- keep compiler and runtime on one `[space/tab]` marker grammar; one bad marker no longer wipes the page

## Binding
Read captures a mount receipt, then reads the page, then checks the receipt again. Register only those bytes plus that manifest. A raced Read drops earlier snapshots and invalidates the pin. A later consistent Read on a new mount re-pins so the model can re-read and cite in the same turn.

## Review follow-up
The original two blockers plus later remount/cap reviews:

1. TypeScript marker regexes match `selfcheck.py` (`[space/tab]`, not `\\s`). Shared cases live in `docs/design/okf-evidence-marker-fixtures.json`.
2. Mixed marked + unmarked answers are one `knowledge_cite` call. Marked pages cannot be cited through `pages`.
3. Mixed 8 evidence + 1 used legacy page fails closed (`cited: 0`); it does not truncate.
4. `noteRead` takes the exact Read bytes. Factory captures the mount before `fsReadFile` and passes that receipt in.
5. Receipt mismatch clears prior snapshots and the pin. Stable A → raced B → cite 0 → stable re-read B → cite B.
6. Contract is in `docs/design/okf-evidence-citations.md`. History is squashed to one commit.

## Deploy
- Agentbox (`SICLAW_AGENTBOX_IMAGE`): grammar, snapshots, mixed call.
- Gateway / Runtime image: the shared 3→8 render cap (Feishu / SSE).
- Compile box (`SICLAW_COMPILE_BOX_IMAGE`): `selfcheck.py` marker scan.
- Sicore apiserver: `okf_citations.go` import gate (companion MR to `develop`).

Live sessions keep the box they already have.

## Tests
- `npx vitest run src/core/knowledge-citation-tool.test.ts src/core/agent-factory.test.ts` (25 passed, 1 skipped)
- focused Python evidence selfcheck passed
- remote CI: Type Check, Test, Portal Web Test, AgentBox Build Graph, KBC Box Test

### DIFF
--- a/kbc/platform/pod/selfcheck.py
+++ b/kbc/platform/pod/selfcheck.py
@@ -60,6 +60,10 @@ MEDIA_ASSET_EXTS = {".png", ".jpg", ".jpeg", ".gif", ".svg", ".webp", ".bmp", ".
 EXCLUSIONS_PATH = "authoring/EXCLUSIONS.json"
 SELFCHECK_PATH = "authoring/SELFCHECK.json"
 SUPPORTED_OKF_VERSION = "0.2"
+_EVIDENCE_ID_RE = re.compile(r"^[A-Za-z0-9][A-Za-z0-9._:-]{0,127}$")
+_EVIDENCE_MARKER_RE = re.compile(r"<!--[ \t]*okf:evidence[ \t]+(\{[^\r\n]*\})[ \t]*-->")
+_EVIDENCE_MARKER_START_RE = re.compile(r"<!--[ \t]*okf:evidence\b")
+_MAX_EVIDENCE_SOURCES = 8
 
 # TEST_ROLE = the standing identity of a read-only knowledge CONSUMER over a
 # pinned wiki snapshot. Single-sourced in the locale prompt packs
@@ -826,6 +830,8 @@ def _okf_v02_metadata_violations(rel: str, fm: dict, body: str) -> list[dict]:
         violations.append({"page": rel, "kind": "okf_usage_window",
                            "detail": "usage_window 必须包含合法且有序的 YYYY-MM-DD from/to 日期"})
 
+    violations.extend(_okf_evidence_violations(rel, fm, body))
+
     if "generated" in fm:
         generated = fm.get("generated")
         if not isinstance(generated, dict) or not _valid_okf_actor(generated.get("by")):
@@ -883,6 +889,73 @@ def _okf_v02_metadata_violations(rel: str, fm: dict, body: str) -> list[dict]:
         if "attester" in fm and not _valid_resource_mapping(fm.get("attester")):
             violations.append({"page": rel, "kind": "okf_attested_computation",
                                "detail": "attester 必须包含非空 resource"})
+    return violations
+
+
+def _okf_evidence_violations(rel: str, fm: dict, body: str) -> list[dict]:
+    """Evidence markers bind one answerable section to declared source ids.
+
+    The marker is single-line JSON so both the compiler and Runtime can parse
+    the same deterministic shape without interpreting surrounding prose.
+    """
+    violations: list[dict] = []
+    declared = {
+        source.get("id").strip()
+        for source in fm.get("sources", [])
+        if isinstance(source, dict)
+        and isinstance(source.get("id"), str)
+        and source.get("id").strip()
+    }
+    scan = _markdown_prose(body)
+    matches = list(_EVIDENCE_MARKER_RE.finditer(scan))
+    if len(matches) != len(_EVIDENCE_MARKER_START_RE.findall(scan)):
+        violations.append({
+            "page": rel,
+            "kind": "okf_evidence",
+            "detail": "okf:evidence 必须是单行 JSON: {\"id\":...,\"sources\":[...]}",
+        })
+    seen: set[str] = set()
+    for match in matches:
+        try:
+            marker = json.loads(match.group(1))
+        except json.JSONDecodeError as exc:
+            violations.append({"page": rel, "kind": "okf_evidence",
+                               "detail": f"okf:evidence JSON 无效: {exc}"})
+            continue
+        if not isinstance(marker, dict) or set(marker) != {"id", "sources"}:
+            violations.append({"page": rel, "kind": "okf_evidence",
+                               "detail": "okf:evidence 只允许 id 与 sources"})
+            continue
+        evidence_id = marker.get("id")
+        if not isinstance(evidence_id, str) or not _EVIDENCE_ID_RE.fullmatch(evidence_id.strip()):
+            violations.append({"page": rel, "kind": "okf_evidence",
+                               "detail": f"evidence id 无效: {evidence_id!r}"})
+            continue
+        evidence_id = evidence_id.strip()
+        if evidence_id in seen:
+            violations.append({"page": rel, "kind": "okf_evidence",
+                               "detail": f"evidence id 重复: {evidence_id}"})
+        seen.add(evidence_id)
+        source_ids = marker.get("sources")
+        if (not isinstance(source_ids, list) or not source_ids
+                or len(source_ids) > _MAX_EVIDENCE_SOURCES):
+            violations.append({"page": rel, "kind": "okf_evidence",
+                               "detail": f"evidence {evidence_id} 必须声明 1-{_MAX_EVIDENCE_SOURCES} 个 sources"})
+            continue
+        marker_seen: set[str] = set()
+        for source_id in source_ids:
+            if not isinstance(source_id, str) or not _EVIDENCE_ID_RE.fullmatch(source_id.strip()):
+                violations.append({"page": rel, "kind": "okf_evidence",
+                                   "detail": f"evidence {evidence_id} source id 无效: {source_id!r}"})
+                continue
+            source_id = source_id.strip()
+            if source_id in marker_seen:
+                violations.append({"page": rel, "kind": "okf_evidence",
+                                   "detail": f"evidence {evidence_id} source 重复: {source_id}"})
+            marker_seen.add(source_id)
+            if source_id not in declared:
+                violations.append({"page": rel, "kind": "okf_evidence",
+                                   "detail": f"evidence {evidence_id} 引用了未在页面 sources 声明的 {source_id}"})
     return violations
 
 
--- a/src/core/knowledge-citation-tool.ts
+++ b/src/core/knowledge-citation-tool.ts
@@ -7,23 +7,111 @@ import { Text } from "@earendil-works/pi-tui";
 import { renderTextResult } from "../tools/infra/tool-render.js";
 import type { SessionEventEmitter } from "./tool-registry.js";
 import {
+  MAX_EVIDENCE_SOURCES_PER_MARKER,
   MAX_KNOWLEDGE_CITATIONS,
   type KnowledgeSourceCitation,
 } from "../shared/knowledge-citations.js";
 
 export const KNOWLEDGE_CITATION_MANIFEST = ".citation-manifest.json";
 
-interface CitationManifestSource { resource: string; title: string; url: string }
+interface CitationManifestSource { sourceId?: string; resource: string; title: string; url: string }
 interface CitationManifestRepo { id: string; root: string; sources: CitationManifestSource[] }
 interface CitationManifest { version: 1; repos: CitationManifestRepo[] }
 
-function result(text: string, cited: number) {
-  return { content: [{ type: "text" as const, text }], details: { cited } };
+interface PageSource { sourceId?: string; resource: string }
+interface PageEvidence { id: string; sourceIds: string[] }
+interface PageProvenance { sources: PageSource[]; evidence: Map<string, PageEvidence> }
+
+const EVIDENCE_ID = /^[A-Za-z0-9][A-Za-z0-9._:-]{0,127}$/;
+/** Canonical grammar: space/tab only. Must stay identical to selfcheck.py and Sicore. */
+export const EVIDENCE_MARKER_START = /<!--[ \t]*okf:evidence\b/g;
+export const EVIDENCE_MARKER = /<!--[ \t]*okf:evidence[ \t]+(\{[^\r\n]*\})[ \t]*-->/g;
+
+function result(text: string, cited: number, unresolved?: string[]) {
+  return {
+    content: [{ type: "text" as const, text }],
+    details: unresolved ? { cited, unresolved } : { cited },
+  };
 }
 
-function normalizedResource(value: string): string {
-  const clean = path.posix.normalize(value.replaceAll("\\", "/").replace(/^\.\//, "")).replace(/^\/+/, "");
-  return clean.startsWith("raw/") ? clean.slice(4) : clean;
+/** Match selfcheck._norm_source_entry: strip raw/ or drop/, keep a leading slash. */
+export function normalizedResource(value: string): string {
+  let entry = path.posix.normalize(value.trim().replaceAll("\\", "/"));
+  for (const prefix of ["raw/", "drop/"] as const) {
+    if (entry.startsWith(prefix)) {
+      entry = entry.slice(prefix.length);
+      break;
+    }
+  }
+  entry = path.posix.normalize(entry);
+  return entry === "." ? "" : entry;
+}
+
+function maskSpan(text: string): string {
+  return text.replace(/[^\n]/g, " ");
+}
+
+/**
+ * Mask fenced blocks and inline code the same way selfcheck._markdown_prose
+ * does, so an example marker in a code fence is not evidence.
+ */
+export function maskMarkdownCode(body: string): string {
+  const maskedLines: string[] = [];
+  let fenceChar: string | null = null;
+  let fenceLen = 0;
+  for (const line of body.match(/.*(?:\r?\n|$)/g) ?? []) {
+    if (!line) continue;
+    const raw = line.replace(/\r?\n$/, "");
+    if (fenceChar !== null) {
+      const close = /^[ ]{0,3}([`~]+)[ \t]*$/.exec(raw);
+      maskedLines.push(maskSpan(line));
+      if (close && close[1][0] === fenceChar && close[1].length >= fenceLen) {
+        fenceChar = null;
+        fenceLen = 0;
+      }
+      continue;
+    }
+    const opened = /^[ ]{0,3}(`{3,}|~{3,})(?:[^\r\n]*)$/.exec(raw);
+    if (opened) {
+      fenceChar = opened[1][0];
+      fenceLen = opened[1].length;
+      maskedLines.push(maskSpan(line));
+    } else {
+      maskedLines.push(line);
+    }
+  }
+  const text = maskedLines.join("");
+  const chars = text.split("");
+  let pos = 0;
+  while (true) {
+    const start = text.indexOf("`", pos);
+    if (start < 0) break;
+    let endRun = start;
+    while (endRun < text.length && text[endRun] === "`") endRun += 1;
+    const ticks = endRun - start;
+    let close = endRun;
+    let found = -1;
+    while (true) {
+      close = text.indexOf("`", close);
+      if (close < 0) break;
+      let closeEnd = close;
+      while (closeEnd < text.length && text[closeEnd] === "`") closeEnd += 1;
+      if (closeEnd - close === ticks) {
+        found = closeEnd;
+        break;
+      }
+      close = closeEnd;
+    }
+    if (found < 0) {
+      pos = endRun;
+      continue;
+    }
+    for (let i = start; i < found; i += 1) {
+      if (chars[i] !== "\n") chars[i] = " ";
+    }
+    pos = found;
+  }
+  return chars.join("");
 }
 
 function readManifest(knowledgeDir: string): CitationManifest | null {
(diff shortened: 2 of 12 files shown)


src/core/knowledge-citation-tool.ts
```diff
@@ -13,12 +13,23 @@ import {
...
 
-function result(text: string, cited: number) {
-  return { content: [{ type: "text" as const, text }], details: { cited } };
+interface PageSource { sourceId?: string; resource: string }
+interface PageEvidence { id: string; sourceIds: string[] }
+interface PageProvenance { sources: PageSource[]; evidence: Map<string, PageEvidence>; evidenceValid: boolean }
+
+const EVIDENCE_ID = /^[A-Za-z0-9][A-Za-z0-9._:-]{0,127}$/;
+const EVIDENCE_MARKER_START = /<!--\s*okf:evidence\b/g;
+const EVIDENCE_MARKER = /<!--\s*okf:evidence\s+(\{[^\r\n]*\})\s*-->/g;
```
chent1996: **Blocking 1.** `\s` here (and on line 25) includes newlines and U+00A0; `selfcheck.py:64/65` uses `[ \t]`. The two implementations therefore disagree about what counts as a marker, verified against both real implementations:

- A newline-form malformed marker (`<!--` newline `okf:evidence NOT-JSON -->`) -> python reports **0** `okf_evidence` violations (its start regex cannot see it), while here the start-count includes it and the full-match count does not, so `evidenceValid` goes false and **every** ref to that page returns unresolved. Measured: a page with one good marker plus this one returns `{"cited":0,"unresolved":["guide.md#good"]}`; the same page without it returns `cited: 1`. A compiler-green page silently loses all citations, with no diagnostic on either side.
- The pretty-printed form (`<!--` newline `okf:evidence {…}` newline `-->`) is invisible to python (0 violations, including the undeclared-source check) but matches here (1 start, 1 full) and is honored.

Suggest tightening this to `[ \t]` to match the compiler rather than loosening python: the single-line marker is the stated contract, the runtime already re-verifies the binding independently, and with both sides blind to the newline form a stray marker costs one unresolved ref instead of wiping the page. A shared fixture list driven from both `test_selfcheck.py` and `knowledge-citation-tool.test.ts` would keep them aligned.

kbc/platform/pod/selfcheck.py
```diff
@@ -60,6 +60,10 @@
 EXCLUSIONS_PATH = "authoring/EXCLUSIONS.json"
 SELFCHECK_PATH = "authoring/SELFCHECK.json"
 SUPPORTED_OKF_VERSION = "0.2"
+_EVIDENCE_ID_RE = re.compile(r"^[A-Za-z0-9][A-Za-z0-9._:-]{0,127}$")
+_EVIDENCE_MARKER_RE = re.compile(r"<!--[ \t]*okf:evidence[ \t]+(\{[^\r\n]*\})[ \t]*-->")
```
chent1996, line 64: Other half of Blocking 1: `[ \t]` here vs `\s` in `src/core/knowledge-citation-tool.ts:25-26`. Confirmed by running this function directly:

```
canonical-good        -> 0 violations   (correct)
canonical-malformed   -> 1 violation    (correct)
newline-malformed     -> 0 violations   (runtime wipes the whole page)
newline-undeclared    -> 0 violations   (the declared-source-ids check never runs)
```

The last line is the one that matters for this function's stated purpose — a marker referencing a source id absent from the page frontmatter passes selfcheck untouched as long as it is written with a newline after `<!--`. The runtime happens to reject it on its own, so this is not a forgery hole, but the compile gate is not gating. The new `test_okf_evidence_markers_bind_only_declared_source_ids` covers only the canonical single-line form, so nothing catches it.

src/core/knowledge-citation-tool.ts
```diff
@@ -35,30 +46,74 @@ function readManifest(knowledgeDir: string): CitationManifest | null {
...
+      const sourceId = typeof row.id === "string" && EVIDENCE_ID.test(row.id.trim())
+        ? row.id.trim()
+        : undefined;
+      return [{ sourceId, resource: normalizedResource(resource) }];
     });
+
+    const evidence = new Map<string, PageEvidence>();
+    const pageBody = body.slice(end + 4);
+    const markers = [...pageBody.matchAll(EVIDENCE_MARKER)];
+    let evidenceValid = (pageBody.match(EVIDENCE_MARKER_START)?.length ?? 0) === markers.length;
```
chent1996: Non-blocking, but it is what turns Blocking 1 into a page-wide outage: `evidenceValid` is a single page-global flag, so one marker this regex pair disagrees about invalidates every ref to the page, including refs whose own markers parsed perfectly.

Fail-closed is the right instinct, but the blast radius could be one ref rather than the page — a marker that fails to parse simply never enters the `evidence` map, so refs naming it already fail on lookup. Reserve the page-global flag for cases where the discrepancy means the page's provenance itself is untrustworthy, and consider logging when it trips, since today it is silent on both sides.

src/core/knowledge-citation-tool.ts
```diff
@@ -119,20 +180,116 @@ export function createKnowledgeCitationSupport(opts: {
...
       resetForTurn();
       const manifest = readManifest(opts.knowledgeDir);
       if (!manifest) return result("No trusted original-source metadata is available for this knowledge mount.", 0);
-      const rawPages = (rawParams as { pages?: unknown }).pages;
+      const params = rawParams as { evidence_refs?: unknown; pages?: unknown };
+      const rawEvidenceRefs = params.evidence_refs;
+      const rawPages = params.pages;
+      const evidenceMode = Array.isArray(rawEvidenceRefs);
+      if (evidenceMode === Array.isArray(rawPages)) {
+        return result("knowledge_cite requires exactly one of evidence_refs or legacy pages.", 0);
```
chent1996: **Blocking 2.** Strict XOR, combined with `hasEvidenceCitationManifest` being mount-wide (one manifest source with a `sourceId` anywhere flips the entire mount into evidence mode), means an answer drawing on one marked page and one unmarked page registers nothing. Verified:

```
knowledge_cite({ evidence_refs: ["guide.md#both"], pages: ["legacy.md"] })
  -> {"cited":0}  "knowledge_cite requires exactly one of evidence_refs or legacy pages."
```

Partial migration is the state this PR creates and it lasts until every page has markers. Throughout it, mixed-source answers lose citations they would have had before this change, and the system prompt gives the model no way to express the case. Accepting both parameters and merging the two resolutions, or deciding the mode per page instead of per mount, would close it.

src/core/knowledge-citation-tool.ts
```diff
@@ -119,20 +180,116 @@ export function createKnowledgeCitationSupport(opts: {
...
+          let refResolved = true;
+          const refCitations: KnowledgeSourceCitation[] = [];
+          for (const sourceId of evidence.sourceIds) {
+            const pageSource = pageSourceById.get(sourceId);
+            const source = manifestSourceById.get(sourceId);
+            if (!pageSource || !source || normalizedResource(source.resource) !== pageSource.resource) {
+              refResolved = false;
+              break;
+            }
+            if (!seenURLs.has(source.url)) {
```
chent1996: Non-blocking: `seenURLs` is read here but only written in the loop at line 266, so two distinct `sourceId`s inside one marker that resolve to the same URL both get pushed. Verified: `{"cited":2}` with the same URL twice. The legacy path adds to the set at push time and does not have this.

`normalizeKnowledgeSourceCitations` dedups by URL downstream so the rendered list is fine, but `details.cited` over-reports and the emitted event carries the duplicate. The two-phase structure is there so a ref that fails midway does not pollute `seenURLs`, so the minimal fix is a per-ref local set checked alongside `seenURLs`.

src/shared/knowledge-citations.ts
```diff
@@ -2,10 +2,15 @@ export interface KnowledgeSourceCitation {
...
+  sourceId?: string;
   page?: string;
+  evidence?: string;
 }
 
-export const MAX_KNOWLEDGE_CITATIONS = 3;
+// Evidence refs are already bounded and validated by knowledge_cite. Keep a
+// small transport safety ceiling, but do not silently discard the fourth
+// source of a multi-source answer as the old page-order resolver did.
+export const MAX_KNOWLEDGE_CITATIONS = 8;
```
chent1996, line 16: Non-blocking. The comment describes this as a transport ceiling, but the constant now also sets `evidence_refs` maxItems, `pages` maxItems, and the per-marker source cap in the runtime validator (`knowledge-citation-tool.ts:85`) — where it is shadowed by an independent `_MAX_EVIDENCE_SOURCES = 8` in `selfcheck.py:66` with no cross-check. Changing this number would silently change which markers are valid, on the TS side only.

Also worth calling out in the description: the raise from 3 applies to the legacy path too, so an answer that used to append at most 3 links to the references section can now append 8, including in Feishu cards.

chent1996 (dismissed):
Reviewed on a checkout of the branch. Base is `f106a858` (#509) rather than current main (`a42c237a`), but a trial merge against main is clean. All four CI checks green; locally `npx tsc --noEmit` is clean, `npm run build` produces no errors, the three targeted suites pass (76, matching the description), and the full suite passes (6031 passed / 2 skipped).

The core design is right, and I verified its central security property rather than taking the description's word for it: **the runtime re-derives the whole binding itself and does not trust the marker.** A marker's `sourceId` must appear in the page's own frontmatter `sources`, must appear in the frozen manifest, and `normalizedResource(manifest.resource)` must equal the page's declared resource — so a page cannot claim a source it does not declare, and a marker cannot re-point a manifest entry at a different original. Path handling for `evidence_refs` keeps the legacy guards intact (resolve → `.md` → containment under `knowledgeDir` → `readPages.has(page)`), and the additive `sourceId`/`page`/`evidence` fields flow through `sync-handlers.ts` as a pure passthrough (that file's change is type-only, so the immutable-session-handler row in the change-impact matrix does not bite).

Two findings block. Both are verified by execution against the real implementations, not read off the diff.

---

## Blocking 1: the compiler and the runtime do not implement the same marker grammar

`selfcheck.py` matches `<!--[ \t]*okf:evidence[ \t]+…[ \t]*-->`; `knowledge-citation-tool.ts` matches `<!--\s*okf:evidence\s+…\s*-->`. JavaScript `\s` includes newlines (and U+00A0); Python `[ \t]` does not. The two sides therefore disagree about what is even *a marker*, and both directions of the disagreement are damaging.

I ran the real `selfcheck.okf_v02_violations` and the real `knowledge_cite` tool against the same page bodies.

**Direction A — compiler-green page silently loses every citation.** Body containing a good marker plus a newline-form malformed one (`<!--` newline `okf:evidence NOT-JSON -->`):

```
python selfcheck  -> 0 okf_evidence violations   (marker invisible: start regex needs [ \t])
knowledge_cite    -> {"cited":0,"unresolved":["guide.md#good"]}
                    "Unresolved evidence refs: guide.md#good. No substitute citations were registered."
```

The TS start-count includes the newline form, the full-match count does not, so `evidenceValid` goes false and **every** ref to that page fails — including the perfectly well-formed one. The control (same page, no second marker) returns `cited: 1`. The author gets a green compile and zero citations at runtime, with no diagnostic on either side. That is the same silent-wrongness this PR exists to eliminate, moved from "substitutes the wrong source" to "drops every source on an approved page".

**Direction B — the compiler's core check is bypassable by whitespace.** The pretty-printed comment form an LLM would plausibly emit:

```
<!--
okf:evidence {"id":"e1","sources":["src-other"]}
-->
```

with `src-other` **not declared** in the page frontmatter:

```
python selfcheck  -> 0 okf_evidence violations
TS regexes        -> 1 start, 1 full match  (honored)
```

`_okf_evidence_violations`' entire stated purpose — "evidence markers bind one answerable section to declared source ids" — is skipped, along with the duplicate-id and >8-sources checks. This is not a citation-forgery hole (the runtime re-verifies against frontmatter + manifest, as above, so the undeclared id simply fails to resolve), but it means the compile gate does not gate what the runtime acts on. The new Python test only exercises the canonical single-line form, so nothing catches this.

The fix is to make one grammar canonical in both languages. I would tighten **TS to `[ \t]`** rather than loosen Python: the single-line marker is the stated contract, the runtime is already self-sufficient on validation, and with `[ \t]` on both sides a newline-form marker becomes invisible to both — so that one ref fails to resolve while the rest of the page still works, instead of a page-wide wipeout. Whichever direction you choose, a shared fixture list exercised by both `test_selfcheck.py` and `knowledge-citation-tool.test.ts` is what keeps them from drifting again.

## Blocking 2: an answer spanning a marked page and a legacy page registers zero citations

`hasEvidenceCitationManifest` is **mount-wide** — one manifest source with a `sourceId` anywhere flips the whole mount into evidence mode — while the tool enforces a strict XOR on its two parameters. Verified:

```
knowledge_cite({ evidence_refs: ["guide.md#both"], pages: ["legacy.md"] })
  -> {"cited":0}  "knowledge_cite requires exactly one of evidence_refs or legacy pages."
```

Partial migration is not an edge case here — it is the state this PR creates, and it persists for as long as it takes to add markers to every page. During it, any answer that draws on both a migrated and an unmigrated page loses **all** citations, where today it would have cited the legacy page. The system prompt tells the model to prefer `evidence_refs` and to use `pages` only for unmarked pages, so the model has no way to express the mixed case at all.

Either accept both parameters in one call and merge the two resolutions, or decide the mode per page instead of per mount. If the XOR is deliberate, the prompt needs to tell the model what to do when its answer spans both kinds of page.

---

## Non-blocking

- **Deploy note is missing, and the matrix requires one.** `kbc/platform/pod/*` changes need a `siclaw-kbc-box` image rebuild plus a `SICLAW_COMPILE_BOX_IMAGE` bump, and a rolled-out runtime does **not** pick up a new box image for existing live sessions. Nothing in the description says this. Worth stating the intended order explicitly, because the marker grammar now spans two independently-deployed artifacts.
- **Duplicate URL emitted within a single marker** — `seenURLs` is read inside the per-source loop but only written after it, so two distinct `sourceId`s resolving to the same URL both get pushed. Verified: `{"cited":2}` with two identical URLs. The legacy path adds to the set immediately and is correct. `normalizeKnowledgeSourceCitations` dedups on render so users do not see it, but `details.cited` over-reports.
- **One bad ref zeroes an otherwise-good call.** Two refs where only the second is bogus returns `cited: 0`. That is the documented fail-closed intent, but the system prompt says to call `knowledge_cite` **once**, so a model that obeys ships the answer with no citations even though one ref resolved cleanly. Either let the result text invite a corrective second call, or say in the prompt that a retry after an unresolved-ref error is expected.
- **`MAX_KNOWLEDGE_CITATIONS` now carries four unrelated meanings** (transport ceiling, `evidence_refs` maxItems, `pages` maxItems, per-marker source cap) and is mirrored by an independent `_MAX_EVIDENCE_SOURCES = 8` in Python with no cross-check. Changing what the comment calls a "transport safety ceiling" would silently change which markers are valid, on one side only. Worth splitting, and pinning the cross-language pair in a test.
- **3 -> 8 also raises the legacy path's cap.** The comment justifies the raise for evidence, but the same constant bounds legacy citations and the rendered references list — an answer that appended at most 3 links now appends up to 8, including in Feishu cards. Presumably intended; say so.
- **No design doc.** The repo convention is that contracts live in `docs/design/`. This PR introduces a cross-language wire contract (the marker grammar, the `sourceId` binding rules, the fail-closed matrix) recorded only in code comments and a prompt string. Given Blocking 1 is precisely "two implementations of one grammar drifted", a short doc naming the canonical grammar is the durable fix rather than a nicety.
- Minor: the description's "npm run build reaches unrelated baseline pi-coding-agent type mismatch" does not reproduce at `dd49ce0f` — both `tsc --noEmit` and `npm run build` are clean here, and CI Type Check passes.

### Verified clean

- Runtime binding re-verification (frontmatter AND manifest AND normalized-resource equality) holds independently of the compiler — this is the property that keeps Blocking 1 from being a forgery hole.
- `evidence_refs` path traversal guards match the legacy path exactly; `hash > 0` correctly rejects `#id` and bare page refs.
- Marker JSON shape is validated identically on both sides for the canonical form (exact key set, id charset, 1..8 sources, no duplicate source ids, no duplicate evidence ids); `JSON.parse` prototype keys are rejected by the exact-key-set check.
- The `g`-flagged regexes are only used with `.match()`/`.matchAll()`, so no `lastIndex` carry-over; `EVIDENCE_ID` is not `g`-flagged and is safe under `.test()`.
- Evidence mode is dormant until the upstream import service emits `sourceId` in `citationSources`, so the rollout is additive — but that also means no part of this has run against a real manifest end to end.

jacoblee-io (commented):
I read @chent1996's review first and am not going to restate it. I checked the load-bearing claims against the code and they hold: `EVIDENCE_MARKER_START` / `EVIDENCE_MARKER` do use `\s` against `[ \t]` on the Python side; `evidenceValid` really is one page-global flag derived from a start-count/full-match-count comparison; `MAX_KNOWLEDGE_CITATIONS` really does serve three unrelated ceilings at once; and the system prompt does tell the model to use `pages` for unmarked pages while the XOR forbids expressing the mixed case. I agree both blockers are blockers.

Three things I can add from having reviewed the PRs this one sits on top of — the citation renderer, the OKF sidecar exclusion, and the complete-set inventory rule.

## 1. Tightening TS to `[ \t]` is also the cheaper deployment, and this is the cheapest moment

Beyond the reasons already given for fixing the grammar on the TS side, there's a rollout argument for the same direction. The two halves of this contract ship in **different images**: `selfcheck.py` in the compile box (`SICLAW_COMPILE_BOX_IMAGE`), `knowledge-citation-tool.ts` in the agentbox (`SICLAW_AGENTBOX_IMAGE`). They have separate rebuild and recycle cycles, so any fix that touches both sides has a window where a new compile box runs against an old agentbox — producing a *third* grammar pairing that neither side was tested against. Fixing only the TS side needs one image and has no ordering constraint.

It also matters that this is the PR **introducing** markers. There are no published pages carrying `okf:evidence` yet, so the grammar can still be narrowed for free. Once pages exist in the wild, tightening it reclassifies already-compiled markers as invisible, and the page-global flag decides what that costs. Whichever grammar wins, it is much cheaper to settle here than in a follow-up.

## 2. The page-global flag's worst case is the one page type that is now mandatory

On the non-blocking note about `evidenceValid` having page-wide blast radius — there's a concrete worst case worth attaching to it. #505 made a **data page covering every row** mandatory for any inventory Q&A treats as a complete set, precisely so a reader can filter the full list. Those pages are the longest artifacts the compiler produces and will carry the most markers, so they're the most likely to contain one the two regexes disagree about.

The failure composes badly: a single malformed marker anywhere on the mandatory complete-set page drops citations for the entire inventory — the exact page whose completeness #505 exists to guarantee, and whose whole purpose is to be the thing Q&A can cite from. That seems like enough to move the blast-radius fix up from non-blocking; a per-ref failure would leave the rest of a 20,000-row data page citable.

## 3. `details` payload growth is now three PRs deep with nothing watching the total

The raise from 3 to 8 lands on a payload that has been growing in unrelated increments. #490 put `structuredContent`-style citation data into tool-result `details`; #502 added MCP `structuredContent` to the same field; this PR raises the citation cap and adds `sourceId` / `page` / `evidence` per entry. Each step is individually justified, and `tool-result-context-guard.ts` does charge `details` against the tool-result budget (`chars += estimateUnknownChars(details)`, with its own `TOOL_RESULT_CHARS_PER_TOKEN` weighting), so none of this escapes accounting.

The gap is that no single place bounds the sum. Worth a sanity check on what a full 8-ref result actually costs against that budget, since the cap moved by 2.7× and citation entries are now wider. Related, and already noted by @chent1996: the raise applies to the legacy path too, so a rendered references section can go from 3 links to 8 — worth confirming that reads acceptably in a Feishu card, which is the narrowest surface it lands on.

(One thing I did verify as *not* a problem: the Markdown-destination escaping added in #490 — percent-encoding `(` and `)` — is per-link, so eight links carry the same protection as three. No new injection surface from the cap alone.)

## Nothing to add on the security core

The property that matters here — the runtime re-derives the binding instead of trusting the marker — was verified in the other review by execution, and it's the same shape #496 established for the uploader-supplied sidecar: package content is input, the frozen server-side manifest is authority. `sourceId` having to appear in the page's own frontmatter *and* the manifest, with `normalizedResource` agreement between them, keeps a marker from re-pointing an entry at a different original. That part I'd have spent the review on, and it's already covered.

LikiosSedo:
Rebased onto current main and addressed the review blockers.

**Blocking 1.** Runtime regexes are now `[ \t]` only, same as `selfcheck.py`. Shared recognition cases are in `docs/design/okf-evidence-marker-fixtures.json` and are asserted by both `knowledge-citation-tool.test.ts` and `test_selfcheck.py`. Pretty-printed / newline forms are invisible to both sides.

**Blocking 2.** `knowledge_cite` accepts `evidence_refs` and `pages` together and merges the resolutions. The prompt says to do that in one call. A marked page cannot be cited through `pages`.

**Elevated blast radius.** `evidenceValid` is gone. A good marker on a page that also has a newline-broken comment still cites (`cited: 1`).

Also: same-URL source IDs inside one marker now count as one; unresolved-ref text asks for one retry; the grammar/binding contract is in `docs/design/okf-evidence-citations.md`.

Ready for re-review.

LikiosSedo:
Follow-up on the second review (`6fbc39c4`):

1. **Mixed cap.** Pages no longer append after evidence already filled 8. Test: 8 evidence sources + 1 legacy page → `cited: 8`.
2. **Code fences.** Compiler and runtime mask fenced/inline code the same way `selfcheck._markdown_prose` already did for link lint. A fenced example marker is not evidence.
3. **Hot reload.** `noteRead` snapshots the page body and pins the manifest. Cite uses that snapshot, so a mid-turn sync cannot retarget source A to source B.
4. **`drop/` and `/foo.md`.** `normalizedResource` now matches `selfcheck._norm_source_entry`.
5. **Deploy.** The 3→8 cap is in `src/shared/knowledge-citations.ts` and is used by Gateway (SSE + Lark) as well as AgentBox. Design doc updated.

Sicore import grammar is a different repo. Companion MR: http://gitlab.scitix-inner.ai/k8s/sicore/-/merge_requests/1041 (`okfEvidenceMarkerStart` now has `\b`, scan is post-frontmatter + code-masked).

LikiosSedo:
Head is now `d8f3d684` (squashed). Last same-scope blocker: a raced second Read of the same path no longer keeps the old A snapshot. Receipt mismatch clears `readPages` and the pin; the next consistent Read can re-pin and cite B in this turn.

Regression: stable A → raced B → `cited: 0` / no event → stable re-read B → cite B.

CI is green. The `CHANGES_REQUESTED` is still the 2026-08-25 review on `dd49ce0f` — please re-review this head.

jacoblee-io (commented):
Re-reviewed `d8f3d684`. The blockers from the earlier review are addressed, and the shared-fixture mechanism is a better fix than "make the two regexes match" — the fixture pins the recognition *decision*, so a future edit to either side has to justify itself against seven named cases rather than pass unnoticed. `nbsp-before-name` in particular is the case that would have survived any fix that only aligned the visible pattern text.

Scope note: at +1310 I concentrated on the recognition pipeline, since that's where the previous blocker lived and where a repeat would be hardest to see. I did not verify the read-snapshot / manifest-pin concurrency work described in your third comment.

## One finding: the fix covers the regex, but the pipeline now has a second cross-language algorithm

Blocking 1 was structural — one recognition rule implemented twice, in two languages, drifting. That's fixed for the pattern. But the pipeline gained a second such rule in the same round:

```js
/**
 * Mask fenced blocks and inline code the same way selfcheck._markdown_prose
 * does, so an example marker in a code fence is not evidence.
 */
export function maskMarkdownCode(body: string): string {
```

"the same way `selfcheck._markdown_prose` does" is a hand-port of an existing Python function, and `recognized` in the fixture describes the whole pipeline — mask, then match — so both halves have to agree for those verdicts to hold. The shared fixture carries exactly one fence case, the simplest form:

```json
{ "id": "fenced-code-example", "body": "```\n<!-- okf:evidence … -->\n```", "recognized": false }
```

The TS side tracks fence char and length and requires the closer to be the same char at >= the opening length, which is the CommonMark rule and more care than a quick port usually gets. What isn't pinned is everything either implementation might treat differently: a language-tagged opener, tilde fences, four-or-more backticks, a fence indented up to three spaces, an unclosed fence running to end of document, and multi-backtick inline spans. `_markdown_prose` was written for link linting, so its strictness was never chosen with marker recognition in mind.

The failure shape is the one you already fixed once, just moved: a compiler-green page whose marker the runtime masks (or doesn't) differently. Now that `evidenceValid` is gone the blast radius is one ref rather than the page — a real improvement — but the compile gate still stops gating what the runtime acts on for that ref.

The fix rides on the mechanism you built: add those forms to `cases` and let both suites drive them. If a form genuinely doesn't matter, saying so with `recognized` either way is cheaper than leaving it undefined.

Related and smaller: `pattern`, `start_pattern` and `max_sources` sit in the JSON, but both implementations hard-code their own (`_EVIDENCE_MARKER_RE`, `_MAX_EVIDENCE_SOURCES = 8`, and the TS equivalents). So those fields are a third copy that nothing forces to match unless a test asserts implementation-against-fixture. Either assert it (compare the compiled source strings to the JSON) or drop the fields and let `cases` carry the contract — as written, the file reads like a single source of truth that isn't one.

## Confirmed fixed

- **Grammar** is `[ \t]` on both sides with the newline and pretty-printed forms invisible to both, so a stray marker costs one unresolved ref instead of a page.
- **Mixed citation** accepts `evidence_refs` and `pages` in one call and merges, with a marked page unable to be cited through `pages` — which closes the partial-migration gap without reopening the "cite a marked page by name" path.
- **`evidenceValid` is gone.** This was the right call: a page-global flag turned any single disagreement into a whole-page outage, and the mandatory complete-set data pages from the inventory rule would have been the worst-hit — longest artifact, most markers, and the one whose citability is the point.
- **The 8-source cap** now lives in one place with `max_sources` echoed in the fixture, so the TS ceiling and the Python check can't part company silently.
- **Fenced example markers not counting as evidence** is a real catch that neither review raised — docs describing the marker format would otherwise have registered as evidence.

## Process note

`CHANGES_REQUESTED` from 2026-08-25 is against `dd49ce0f` and won't clear on its own; @chent1996 needs to re-review or dismiss it. My comment here doesn't unblock it either — and given that review was verified by running both implementations, its author is better placed than me to confirm the grammar fix behaves as measured.

jacoblee-io (approved):
Approving on my line of review — `c0040c50` addresses what I raised, and the fence coverage is better designed than what I asked for.

## The fence cases

All six forms I listed are now in the fixture, and the naming shows the part I didn't think to specify:

| form | case | recognized |
| --- | --- | --- |
| language-tagged opener | `language-tagged-fence-then-prose` | true |
| tilde fence | `tilde-fence-then-prose` | true |
| four-plus backticks | `four-backtick-short-run-not-close` | true |
| indented fence | `three-space-indented-fence-then-prose` | true |
| unclosed to EOF | `unclosed-fence-to-eof` | **false** |
| multi-backtick inline | `multi-backtick-inline-then-prose` | true |

The `-then-prose` suffix means each one asserts the *non*-over-masking direction: the fence closes correctly and a marker in the following prose still resolves. That's the failure mode I should have named — over-masking silently eats a legitimate marker, which is harder to notice than under-masking, since the page looks fine and one ref just never cites. Testing both directions per form is what makes this hold.

`start_pattern` being gone follows from dropping `evidenceValid` — with no start-count comparison there's nothing left to use it for. Good cleanup rather than an omission.

## The third implementation

This is the part I couldn't have found, and it makes my drift argument bigger than I framed it:

> Three implementations must agree on the grammar:
> - compile box: `selfcheck.py`
> - agentbox: `knowledge-citation-tool.ts`
> - the management-side import gate: `okf_citations.go`

I reviewed this as two implementations. The shared fixture is driven by `test_selfcheck.py` and `knowledge-citation-tool.test.ts` only, so the third one is outside the mechanism that keeps the other two aligned — and the doc states the resulting failure plainly:

> A compile-green page can still fail package import if Go's start regex or scan range differs.

Nothing in this PR can fix that, since the code lives elsewhere. Writing it down with the specific requirements the Go side must meet (same start regex with `\b`, post-frontmatter body only, ignore fenced/inline code) is the right move for a boundary this repo can't enforce. Worth having the other side load this same fixture file if it can reach it — that would close the loop instead of documenting it.

## The deploy section

This is the most complete version of a note I've raised on several PRs in this series, and it includes a consequence I missed:

> The 3→8 render cap lives in `src/shared/knowledge-citations.ts` and is used by AgentBox **and** Gateway (SSE consumer and Lark). Shipping only a new agentbox image leaves Feishu / SSE rendering on the old 3-link ceiling.

Four deploy units with what each one carries, plus the live-session caveat. That's the shape these notes should have had.

## Two small things

`pattern` and `max_sources` remain in the JSON while both implementations hard-code their own. Fine as documentation, but nothing forces them to match — if that's the intent, a test comparing the compiled pattern source to the fixture would make the file the authority it reads as.

The new doc names `okf_citations.go` and the management product directly. Given the ongoing effort to keep internal references out of this repository, that's worth a glance — though here the specificity is doing real work (a deployer needs to know which component gates the import), so a neutral phrasing that keeps the actionable part would be better than dropping it.

## Not verified

The read-snapshot / mount-receipt concurrency work. The contract in the doc is self-consistent — receipt before, read, receipt after, register only on match, a raced read drops snapshots and invalidates the pin, a later consistent read re-pins — but I checked the description, not the implementation.

`CHANGES_REQUESTED` from 2026-08-25 is against `dd49ce0f` and still stands; @chent1996 needs to clear it. That review verified its findings by executing both implementations, so its author is the right person to confirm the grammar fix measures as intended.

PR #512, opened 2026-08-24
fix(portal): treat subagent sessions as execution traces

jacoblee-io:
## Summary

`chat_sessions.origin = 'subagent'` shipped in 2026-05 (`7485b943`) but no Portal query ever learned about it: all eight hardcoded `NOT IN ('task', 'delegation')` predicates either predate it or were written without it. Sub-agent child sessions were therefore treated as top-level conversations everywhere. This centralizes the origin vocabulary in one registry so a future trace origin cannot repeat the failure.

### Problem

Three consequences, in descending severity:

| Site | Symptom |
|---|---|
| `adapter.ts` `task.prune` (×2 mirrors) | `origin='subagent'` rows were **unreachable by retention pruning** — unbounded growth since 2026-05 |
| `siclaw-api.ts` agent chat list | Sub-agent children appeared in the user's Chat list, which the surrounding comment explicitly says should be hidden |
| `metrics-entry.ts`, `adapter.ts` stats (×2 mirrors) | Session and prompt counts inflated by every sub-agent spawned; message-level parent attribution only recognized `delegation` |

The metrics axis (`2bd8e216`, 2026-06-25) was written a month *after* the origin existed, and its header comment asserts that `origin='delegation'` means "sub-agent execution traces" — conflating two different mechanisms. Its tests asserted the defective predicate verbatim (`expect(p).toContain("NOT IN ('task', 'delegation')")`), which is why this survived review.

### Solution

New `src/portal/session-origin.ts` owns the vocabulary: `TRACE_ORIGINS` (`task` / `delegation` / `subagent`) and `PARENT_ATTRIBUTED_ORIGINS` (`delegation` / `subagent`), plus the predicates built from them. All eight call sites now derive from it.

`delegation` and `subagent` stay separate origins because they are separate mechanisms:

- **Delegation** hands a request to *another* agent, which runs under its own model, prompt and capabilities. The coordinator configures nothing.
- **Sub-agent** is one agent's internal context isolation: same agent, parent's configuration, tool whitelist narrowed rather than widened.

Both are attributed to the parent session's entry form for message-level queries (both perform work on behalf of a parent turn). `task` is deliberately excluded from that set — a scheduled run is its own entry bucket with no parent turn to inherit from.

`docs/design/coordinator-routing.md` gains the ownership contract, which until now lived only in code comments (`delegate-api.ts:370` and `:701`). The overloaded `delegation_id` wire field — a sub-agent's spawn id travels under that name — is documented as historical transport naming, not a shared concept.

## ⚠️ Migration note

The first `task.prune` run after this merges will **delete the accumulated `origin='subagent'` backlog** (everything older than the configured retention window, going back to 2026-05). That is the intended behaviour — those rows were always supposed to be prunable — but the volume may be significant. Worth checking before deploying:

```sql
SELECT COUNT(*) FROM chat_sessions WHERE origin = 'subagent' AND last_active_at < <cutoff>;
```

## Questions for Upstream-mode review

`adapter.ts` is by its own docstring the Standalone implementation of "the same endpoints that Upstream provides". In Upstream mode these handlers do not run — the management plane's implementations do. Since `chat.ensureSession` writes `params.origin` through verbatim, the same `origin='subagent'` rows exist on that side, and this PR does not touch it.

Please confirm on the management-plane side:

1. **Session listing** — does the user-facing chat/session list exclude trace origins, and does its exclusion list include `subagent`?
2. **Stats / metrics** — are session and prompt counts filtered by origin, and is `subagent` in that filter?
3. **Retention pruning** — is there an equivalent of `task.prune`, and can it reach `origin='subagent'` rows? This is the one with unbounded-growth consequences.

I was unable to verify this myself: the checkout only exists on the build host and SSH was refused on both attempts. Every claim above about the management plane is inference from the RPC contract, not from reading its code.

## Test Plan

- [x] `npm test` — main repo green (45404 passed). The 5 failures are pre-existing, all inside historical worktrees (`portal-web` missing `node_modules`), unrelated to this change
- [x] `npx vitest run --dir src/portal` — 782 passed
- [x] Tests rewritten to assert against `TRACE_ORIGINS` rather than a pinned SQL string, plus a `'subagent'` regression case
- [ ] Not verified against a live DB: the count/list/prune deltas are read from the SQL predicates, not measured


jacoblee-io:
Addressed in `7609d44c` — the P1 is real, and it is worse than reported: **the inconsistency was introduced by this PR**, not pre-existing.

Before this branch, both endpoints (wrongly) counted sub-agent briefings as prompts. The first commit excluded trace origins from the session-level predicate but added `subagent` to the *parent-attributed* set, so `role='user'` rows kept counting through the parent branch — while the sibling `adapter.ts` summary, which I had switched to the plain non-trace predicate, stopped counting them. Two endpoints, one number, two answers.

Both affected sites were in `siclaw-api.ts`: `totalPrompts` and the daily prompts series. The tool/assistant queries share the same predicate and were correct to — they keep parent attribution.

Taking the suggested split as the contract:

- **`entryPromptPredicate`** (new) — `role='user'` counts. Trace sessions excluded, no parent join. Deliberately a thin alias of the session-level predicate: what matters is the difference from `entryMessagePredicate`, not the SQL it emits, so the distinction has a name instead of living in a comment.
- **`entryMessagePredicate`** — tool/assistant telemetry only, parent attribution retained. Its doc now says explicitly that a `role='user'` count must not use it.

Dropping the parent join also required dropping the `parent_s` argument to the channel/user filter — otherwise the query references an unjoined alias.

**One behaviour change worth flagging beyond the report:** this also changes the *pre-existing* treatment of `delegation`. A coordinator's delegated instruction no longer counts as a user prompt either. That is what makes this endpoint agree with the adapter summary, and it follows the same reasoning — a delegated instruction is text an agent wrote, not a request a person made. If the intended Upstream-mode semantics differ on that point, say so and I will narrow the change to `subagent` only.

Test plan delta: 3 new cases in `metrics-entry.test.ts`, including one that pins `entryPromptPredicate !== entryMessagePredicate` so the two cannot silently converge again. `npm test` 45407 passed; `npx tsc --noEmit` clean. Still not verified against a live DB — the count deltas are read from the predicates.

jacoblee-io:
Fixed in `f30aa4e7`. Confirmed, and it is a pre-existing bug this PR widened rather than one it introduced.

The mechanism, for the record: an unmatched LEFT JOIN NULLs every parent column, and two buckets *read* a NULL origin as a match — `web` is literally `origin IS NULL`, and `all` opens with the same disjunct. The specific-entry buckets were never affected (`parent_s.origin = 'api'` against NULL is not a match), which is exactly why this stayed invisible: only the default bucket and the one users open first were wrong.

Pre-existing because the parent branch has always been `s.origin = 'delegation' AND <entry on parent_s>`. Widened because attributing `subagent` to the parent multiplied the orphan population — sub-agent children vastly outnumber delegations.

On "legitimate": confirmed in code rather than assumed. `delegate-api.ts:466-468` persists a NULL parent **on purpose** —

```
// Link the parent ONLY when validated — never persist an unverified parent ref
const trustedParent = parentTrusted ? body.parentSessionId ?? null : null;
```

— so `origin='delegation'` with `parent_session_id IS NULL` is a designed state, not corruption. The second source is a parent removed by deletion or retention, which leaves children behind. The rule this encodes: a trace row whose parent is absent belongs to **no** entry bucket, and must not fall through to whichever bucket a NULL happens to satisfy.

Two new test cases pin the guard for both affected buckets and for a custom parent alias, and the three predicate-shape assertions were updated.

Test plan delta: `npx vitest run --dir src/portal` 787 passed; `npx tsc --noEmit` clean. `npm test` shows 3 failures, all inside historical worktree copies (`restricted-bash` skill-whitelist cases, `portal-knowledge-materializer`) — files this branch does not touch. Verified against the main repo directly: `--dir src/tools/cmd-exec` 555 passed, `portal-knowledge-materializer.test.ts` 8 passed. They also moved between worktrees across runs, so they track worktree environment, not this change.

Still not verified against a live DB. In particular I have not measured how many orphan trace rows exist — if that number is interesting for sizing the Overview correction, `SELECT COUNT(*) FROM chat_sessions s LEFT JOIN chat_sessions p ON s.parent_session_id = p.id WHERE s.origin IN ('delegation','subagent') AND p.id IS NULL` answers it.

nightmeng (approved):

PR #518, opened 2026-08-24
feat(subagent): route map children to a configured model tier

jacoblee-io:
## Summary

Lets an agent configure which models its sub-agents may use, and lets the lead pick one per `spawn_subagent` call. The speedup comes from the parallel map children — they already run concurrently, so a cheaper model per child makes the whole fan-out faster while the lead keeps a large model for orchestration.

Design: `docs/design/subagent-model-tiering.md` (added here).

### Problem

A sub-agent hard-inherits its parent's model — `session.ts` sets the parent turn's `{provider, modelId}` on the child with no override layer. `SubagentType.model` declared `"sonnet" | "opus" | "haiku" | "inherit"` but the repository contained only the two lines declaring it: a dead field, and those aliases cannot work where models are administered by a control plane and several providers coexist.

Execution-side measurement puts ~75% of wall-clock in model round-trips, so the headroom is real. **The actual gain is not claimed here — it has to be measured (see Test Plan).**

### Solution

**The two payloads travel separately, and that is the core of it.**

| | Content | Channel | Why |
|---|---|---|---|
| Menu | `tier` + `whenToUse`, **no credentials** | tools sync | Consumed while *building* a session's tool description. `http-server` creates the session before `setDelegationModel` runs, and `sync-handlers` states the tool-set is baked in at creation — a per-prompt field would never reach the schema |
| Candidates | `provider` + `modelId` + `modelConfig`, **credentials** | prompt binding | Per turn |

Beyond timing this is a confidentiality boundary: credentials never enter a tool description, so they never enter a prompt. The lead sees only a tier name and a sentence about when to use it — it cannot invent a model id, the same prompt works across deployments, and the model inventory stays out of the context.

**State scoping follows from that.** The menu is box-level but each session snapshots it at creation, because the menu that built a tool description is the menu that session must honour. Candidates are session/turn state, never box-level — a box holds many sessions, and box-level credentials would both share them across sessions and let one session resolve against another's binding.

**One immutable plan per dispatch.** `spawn_subagent` captures menu + candidates + the parent's effective candidate once, and every child of that call uses it: collapse path, each map child, every later wave, the detached background continuation, and reduce. Children start in waves, so reading live state per child let a 50-item batch straddle a config change and report both halves as one result. That hole predates tiering; the snapshot closes it.

**Failures divide by phase.** Everything before the child prompts falls back to the parent's *effective* model (not its configured primary — the parent may already have failed over) and runs anyway. Restoring the parent is an action, not a decision: the brain is already part-way onto the tier, so `setModel` and the params application both re-run. And it terminates — if the parent cannot be applied either, the child fails with `parent_fallback_failed` rather than looping.

**Reduce never downgrades.** Synthesis is what a cheaper model hurts most and it runs once.

Per-item results carry requested vs resolved tier, selection source, effective model, and a fallback reason — without those, "the report is weak", "the lead picked the wrong tier" and "the candidate never arrived" are indistinguishable, and one of the three stated mitigations for a mis-picked tier is that a weak report exposes it.

### Two design claims corrected while implementing

- **Forwarding sites are 8, not 13.** The design (and its review) counted every `modelRouting:` grep hit. Three pass those fields to `modelOptionsSupportImageInput`, a vision-capability check that must not carry tier state — adding it there is a type error, which is how the miscount surfaced. An invariant test pins the real count and asserts the check sites stay clean.
- **The TUI requirement rested on a false premise.** Earlier drafts had the CLI snapshot carry both payloads because the TUI has no tools channel. But the TUI registers no `spawnSubagentExecutor` at all — `spawn_subagent` does not exist there, so there is nothing to tier. Shipping tier state there would deliver *credentials* to a surface with no consumer for them. Design updated to say so explicitly.

### Drive-by fix

`agent-api`'s clone path omitted `tool_capabilities`, so a cloned agent silently lost its tool restriction. **A clone broader than its source is a privilege bug**, not a cosmetic one. Fixed alongside the new column; the INSERT assertions now resolve columns by name rather than position, which is what let three of them break on a new column.

## Not included, and not claimed

- **Endpoint-scoped concurrency limiting.** Every limiter counts *children*, not requests per model. Today those ~10 concurrent children hit the parent's provider — the endpoint the main path has already survived at that concurrency. After tiering they hit the cheap endpoint, which usually has a tighter quota, and a child session has no failover. §8.1 makes this a prerequisite for general rollout and lists three questions that must be frozen first (acquisition order, whether a waiter holds its session/pod slot, multi-replica quota). **This branch supports a bounded pilot, not general rollout.**
- **Child-side compaction and failover.** v1 adds a pure context-fit check that falls back pre-prompt; a post-prompt endpoint failure fails that child.
- **Portal UI.** Consistent with `model_routing`, which is API/DB-only in Standalone today. This does *not* generalize to Upstream mode — see below.

## Questions for Upstream-mode review

`resolveAgentModelBinding` exists twice and the two go different ways: `chat-gateway.ts` (local query, Standalone) and `agent-model-binding.ts` (`config.getModelBinding` RPC). **Under Upstream mode the binding is built entirely by the management plane and the Portal path does not run** — the Runtime returns `data.binding` verbatim with no validation. So this branch cannot make the feature work there. Please confirm on that side:

1. **Storage** — tier config in the management plane's own storage (a join table, not a column on the agents row), with a migration for existing data.
2. **Candidates** — `subagentTiers` in the `config.getModelBinding` response, hydrated with `modelConfig`, carrying a `revision`. Missing this means the feature does not exist there, **without any error** — it presents as "configured but has no effect".
3. **Menu** — `subagentTierMenu` on the resource-reload notification (`["tools"]` form). Missing this means the tool description carries no list, so the lead never sees the feature.
4. **The two revisions must be computed from the same canonical form**, or every spawn reports `revision_mismatch` and tiering never engages.
5. **Validation must match** (`tier` pattern and cap, uniqueness, `whenToUse` bounds). Divergence fails silently: the management plane stores something siclaw considers invalid, siclaw falls back, and the operator sees "configured but does nothing".
6. **A UI is needed there.** The "no UI in v1" decision rests on Standalone's `model_routing` having none either; the management plane's `model_routing` has a complete one.

**Deploy order**: management plane first (harmless extra fields to an older Runtime), then siclaw. Reversed, siclaw never receives the fields.

## Test Plan

- [x] `npm test` — 275 files / 6214 tests, all passing
- [x] New unit coverage: payload shapes and per-channel normalizers (33), resolution order / env override / config→menu projection (14), forwarding invariant (3), turn-state lifecycle (3), pure fit check (3)
- [ ] **Not verified against a live deployment.** No end-to-end run, and in particular the menu has not been observed reaching a real tool description — that is the easiest false green here: if candidates work but the menu does not, the lead never fills the parameter, the feature is silently absent, and every unit test still passes
- [ ] **Speedup not measured.** This decides whether to widen scope; §13 describes the comparison
- [ ] Failure drills from §13.10 (deleted model, live endpoint returning 429/5xx, quota exhaustion, menu reload mid-batch) not run

Ships in the **agentbox** image (`session.ts`, `src/tools/`, `sync-handlers.ts`) and the portal/runtime image (`agent-api`, `chat-gateway`, `internal-api`, `migrate`). Run the migration first.


jacoblee-io:
siclaw-side fixes pushed in `20d65a64`. The three management-plane-side items are **not** addressed here — different repository, they need to be handled there. Notes on each below.

## Fixed here

**[P1] Menu protocol mismatch.** Confirmed and fixed, and it is the one that mattered: the projection only handled the config array, so an already-projected `{revision, items}` returned null, no menu reached the tool description, `spawn_subagent` never exposed `model_tier`, and tiering was silently absent under Upstream mode — with this repo's entire suite green, because every test fed the shape this repo writes.

Both shapes are accepted now. A projected menu **keeps the producer's revision** rather than recomputing: re-hashing a projection that no longer carries `provider`/`modelId` could never reproduce the value shipped with the candidates, so recomputing would guarantee a permanent `revision_mismatch`.

On the field name — I did **not** ask the other side to change it. `config.getAgent` is snake_case throughout (`tool_capabilities`, `agent_type`, `model_provider`), so `when_to_use` follows that RPC's convention; the tools payload it feeds is camelCase like the rest of the AgentBox wire. Those are two different boundaries and each should keep its own convention. siclaw accepts both spellings and normalizes at the boundary. If that side renames to `whenToUse` it will still work, but nothing here requires it.

One further point for the design, now recorded: **projecting on the producer side is the better arrangement**, not a deviation. The menu channel has no business seeing `provider`/`modelId`, so a control plane that projects before answering discloses strictly less than one that hands over its whole config for the Gateway to trim. That shape is the one to prefer.

Contract test added from real upstream-shaped fixtures (`subagent-tier-wire-contract.test.ts`), including the end of the chain — that a description passage actually renders — since an empty passage is exactly how this failure presented.

**[P2] Tier fallback leaking the reasoning level.** Confirmed; my earlier comment claiming `setModel` restores the target's default was wrong. `_getThinkingLevelForModelSwitch` (pi 0.80.7) returns `this.thinkingLevel` whenever the new model supports thinking, and only falls back to the default when it does not — and `applyModelParams` cannot lower it, since an absent effort is a no-op rather than a reset. So an `xhigh` tier that failed its fit check left the *parent* running at `xhigh`. Added `captureModelParams` to the brain interface; the child's tunables are captured before any tier touches them and restored on the fallback path, before the candidate's own params so an explicit setting still wins.

**[P2] Standalone tier edits not reloading `tools`.** Confirmed and fixed. Comparison uses the same canonical form the revision does — not `normalizedToolCapabilities`, which dedupes with a `Set` and sorts, both meaningless for object arrays.

## management-plane-side, for that repo

I can't touch these, and each one blocks something real:

1. **UI dependency array** — a tier-only edit saving stale state means the config never reaches storage, so nothing downstream can work regardless of what siclaw does.
2. **`force=true` provider delete bypassing the tier guard** — worth noting siclaw degrades safely here: the resolver now verifies the *model* still exists (not just the provider), so a dangling reference ships as "no tiers" rather than as a menu whose candidates cannot resolve. That is containment, not a fix — the dangling config still needs preventing at the source.
3. **List response vs OAS** — agreed that `null` (not loaded) and `[]` (confirmed empty) are different states worth distinguishing, since they are exactly the clear-vs-absent distinction the wire contract depends on.

## Verification

`npm test` 276 files / 6227 passed, 2 skipped. `npx tsc --noEmit` clean.

Still unverified, unchanged from the PR description: no live end-to-end run, no speedup measurement, no failure drills. The contract test closes the specific gap that let the protocol mismatch through, but it is a fixture test — it does not prove the two deployments actually agree at runtime.

jacoblee-io:
Both siclaw-side items fixed in `90a21a15`. The two management-plane-side ones need that repo.

## [P1] Wrong restore baseline — confirmed, and it was my regression

This one I introduced in `4dd968a2` while fixing the level leak, and the fix was worse than the bug.

The baseline came from the freshly built **child** brain. That is not the parent's configuration, and it is not even a faithful reading of the child's: pi clamps the thinking level to the current model's capabilities, so a child starting on a non-reasoning model reads back `off` regardless of the deployment default. Meanwhile `_getThinkingLevelForModelSwitch` consults `supportsThinking()` on the model being switched **away from** and hands out the deployment default when that model is non-reasoning. So writing the captured value back overrode that default with a clamped artifact — your `off`-instead-of-`high` example exactly.

The part that makes it P1 rather than P2: **it ran on the plain inherit path too.** Every deployment that has never configured a tier goes through `applyParent(undefined, "inherit")`, so this was silently rewriting the thinking level for installations that are not using this feature at all — a direct breach of contract 1 ("no tiers configured ⇒ behaviour identical to today").

Fixed as you suggested: the parent publishes its own tunables through the same `onAttemptReady` hook, captured after the runner applied them, and they travel with the effective-parent snapshot. Restored **only** when a tier was actually attempted and rejected — `reason` is undefined on inherit, and on that path nothing has touched the child's level, so nothing is written and pi's own semantics stand. "Identical to today" has to mean not writing at all, not writing back a value that happens to match.

## [P2] One dangling tier disabling all of them — confirmed

And the asymmetry you identified is what made it bite: the menu is built from the raw config on a separate path, so collapsing the candidate side left the menu still advertising every tier. The revisions then disagreed and even the healthy tiers failed with `revision_mismatch` — so deleting one model of two broke both, which is strictly worse than the dangling tier alone.

Unresolvable entries are skipped now; the rest stay usable and the dangling one reports `candidate_missing`, which is the case §5 defines that reason for. Two things worth stating because they are easy to get wrong when touching this:

- **The revision is still computed over the full config, not the survivors.** The menu side hashes the same thing, so a revision over the surviving subset would mismatch on every spawn — reintroducing the failure from the other direction.
- **A list where nothing resolves still collapses to null.** A menu backed by no candidates would offer the lead choices that can never be honoured.

New `model-routing-config.test.ts` covers all-resolve, one-model-gone, one-provider-gone, revision-over-full-config, nothing-resolves, and empty.

## management-plane-side, for that repo

- **Gate flip not reloading warm sessions** — same shape as the Standalone bug I fixed last round (`agent-api.ts` now reloads `tools` when `subagent_models` changes). Worth noting the off-direction is the sharper half: a stale menu keeps advertising tiers the model can still select, and those selections then resolve against candidates that are no longer being sent.
- **Tier models not constrained to `model_type=llm`** — agreed. siclaw has no defence here worth relying on: an embedding model that resolves through `findModel` will be set as the child's model and fail at request time, and the fallback reason will say `model_not_found` only if the descriptor never made it into the registry. The write-side constraint is the real fix.

## Verification

`npm test` 277 files / 6233 passed, 2 skipped. `npx tsc --noEmit` clean.

Unchanged: no live end-to-end run, no speedup measurement, no failure drills.

jacoblee-io:
Both siclaw items fixed in `4bf055e3`. The the management plane gate-concurrency one needs that repo.

## [P1] Params applied with no tier configured

Confirmed. I verified the baseline against `30e2b4f9`: placing a child was exactly two calls, `registerProvider` and `setModel`, with **no params application at all**. So applying the parent candidate's `params` in `putBrainOnCandidate` meant a child would start honouring a `reasoning_effort` that every child has ignored — on the plain inherit path, i.e. for deployments that have never configured a tier. Contract 1 forbids that, and it was stated too abstractly to catch this; it now spells out the two calls.

Touching the tunables is gated on whether a tier was actually attempted:

| Path | Params |
|---|---|
| inherit | untouched — the neutral option is doing nothing |
| tier | the candidate's own apply; a tier *is* a request to run differently |
| fallback | restore the parent's real params, then the candidate's, so an explicit setting still wins |

Three tests pin it, including the ordering (restore before candidate, both after `setModel`, which is what re-clamps to the target).

## [P2] Null menu never taking effect

Confirmed, and it is the same root cause as the last round rather than a separate bug: **`config.getAgent` handed over the raw config while the candidates went through `resolveAgentSubagentTiers`, so the two channels answered "which tiers exist" independently.** That was latent until the previous commit made the candidate side drop unresolvable entries — at which point the divergence became visible in both directions (all-gone → empty candidates vs. full menu; partially-gone → filtered candidates vs. full menu).

Fixed by routing `config.getAgent` through the same resolver, as you suggested. Two things fall out of it:

- The Gateway stops receiving `provider`/`modelId` it never needed to build a menu — the projection now happens where the data already lives.
- "Which tiers exist" has one answer instead of two that have to be kept in sync by hand.

The raw-config path stays accepted for a producer that has not moved yet.

## management-plane-side gate concurrency

Not mine to fix, but the read-outside-the-transaction shape is worth flagging beyond the lost-notification symptom: two requests reading `false`, then writing `true` and `false`, can leave the *stored* value and the *last reload sent* describing different states. The suggestion to reload unconditionally on the gate key is the cheaper of the two fixes and does not depend on getting the locking right — a redundant `tools` reload costs a session rebuild, while a missed one leaves warm sessions advertising tiers that are switched off.

## Verification

`npm test` 277 files / 6236 passed, 2 skipped. `npx tsc --noEmit` clean.

Unchanged and worth repeating, since three rounds of review have all been static: **no live end-to-end run, no speedup measurement, no failure drills.** Every defect found so far was one that unit tests could not have caught on their own — the protocol mismatch, the level leak, the inherit-path regression. That pattern is itself the argument for running this against a real deployment before merge.

jacoblee-io:
Fixed in `a0f61325`. The the management plane notification-timeout item needs that repo.

## [P2] Malformed stored config crashing the config RPC

Confirmed, and the blast radius is the part worth naming: this defect only became serious in the *previous* commit. Routing `config.getAgent` through the same resolver was the right call for keeping the two channels consistent, but it also put a shared crash path under the model binding — so a malformed `subagent_models` row stopped being "no tiers for this agent" and became "this agent cannot start". Bad configuration for an optional feature turning into a total outage.

Root cause is the one you identified: the rules that would have caught `[null]` lived **only inside the throwing encoder**, where no read path could reach them. Every read path had to trust the column, and `safeParseJson`'s `as T` gave it nothing to trust.

Now there is a non-throwing `normalizeSubagentTierConfig` that the encoder wraps: the write path turns a rejection into a 400, read paths degrade to `{menu: null, candidates: null}` with a warning. Same rules everywhere by construction rather than by two implementations happening to agree — which is the same failure mode as the menu/candidate divergence from the last two rounds.

The menu projection's array branch now goes through it as well. That branch hand-checked fields and skipped the duplicate and cap rules entirely, so a config the write path would have refused could still produce a menu.

Tests cover `[null]`, `[undefined]`, `["fast"]`, `[42]`, `[[]]`, `[{}]`, missing fields, non-array, duplicates, over-cap, and snake_case keys — at both the normalizer and the resolver level, the latter asserting `.resolves` rather than a throw.

## Something I should flag about my own work

While fixing this I found the dedup key was built as `` `${provider}\x00${modelId}` `` with a **literal NUL byte** rather than an escape, which had made the source file register as binary to `grep`. Functionally fine; unmaintainable, and it silently broke tooling.

That is the third time in this branch I have written a raw control character into a file — twice of them while implementing the rule that rejects control characters. Replaced with `JSON.stringify([provider, modelId])`, which is unambiguous for the same reason and survives every tool that touches the file, and swept the whole tree to confirm nothing else is affected.

## Verification

`npm test` 277 files / 6241 passed, 2 skipped. `npx tsc --noEmit` clean. No stray control bytes anywhere under `src/`.

Still outstanding, unchanged across five rounds of review: **no live end-to-end run, no speedup measurement, no failure drills.** Every round so far has found something static review caught and the test suite did not — the protocol mismatch, the level leak, the inherit-path regression, this crash path. All four were shapes the tests fed themselves and therefore could not have discovered. I would not merge on green tests alone here.

jacoblee-io:
Fixed in `c0ba8ff0`. Both management-plane items need that repo.

## [P1] Corrupted config blocking its own repair

Confirmed, and the shape of it is worth naming: a PUT reads the current value to decide whether anything changed, so a comparison that throws on a corrupted row makes that row **permanently unfixable** — the one operation that repairs the problem, including simply clearing the field, is the one guaranteed to fail. I fixed the resolver last round and missed this helper, which is the same defect one layer over.

It goes through the non-throwing validator now, with an invalid stored value comparing as "no tiers". That is also the correct answer for the reload decision rather than a convenient one: an unusable config was already producing no tiers, so replacing it with a usable one *is* a change and must invalidate warm sessions. Two tests cover repair-by-overwrite and repair-by-clear.

## [P2] Background runs losing the tier outcome

Confirmed. Persisted on terminal delegation events now — per item inside `itemStatuses` for a group, and directly on the event for a single child. The single-child background path had the same gap, so both are covered.

Written through an explicit `persistableTierOutcome` projection rather than spreading the outcome object: this record is stored and replayed, so the safe-to-persist shape needs to be stated field by field rather than left to whatever the source type grows later.

## Shared golden fixture

Done — `src/core/__fixtures__/subagent-tier-wire.golden.json`. The fixture is now a file rather than inline literals, so the other implementation can assert against the same bytes; the header says explicitly to copy it verbatim rather than transcribe, since transcription is the step where a spelling drifts.

It also states the rules as **data** (tier pattern, cap, whenToUse bounds, accepted spellings, fields the menu must never carry, the clear signal), and a new test asserts those match what this side actually enforces. So a constant moving on either side fails a test rather than surfacing at runtime on a deployment nobody exercised.

## trim-then-reject: keeping it, documenting why

You flagged the inconsistency correctly, but I think the code is right and the doc was wrong, so I changed the doc rather than the behaviour.

Surrounding whitespace is an artifact of how a value was entered and carries no meaning; an interior control character changes what the model reads. Rejecting a pasted value over a trailing newline would fail perfectly good configurations for no gain, while silently rewriting the interior is the thing actually worth refusing. Both the doc and the function now say this explicitly, so the next reader does not have to re-derive it.

If you would rather have strict rejection including leading/trailing, say so — it is a one-line change, and I would rather it be a decision than a drift.

## Verification

`npm test` 277 files / 6254 passed, 2 skipped. `npx tsc --noEmit` clean.

Unchanged after six rounds: **no live end-to-end run, no speedup measurement, no failure drills.**

jacoblee-io:
Fixed in `ff507810`. The two management-plane items need that repo.

## [P1] Single-child tier outcome dropped at the Gateway

Confirmed, and this one is worth dwelling on because of *how* it hid.

I added the emission last round and never followed the field to storage. The group path copied its outcome (it rides inside `item_statuses`), so the group worked — which meant the feature looked delivered while the single-child case silently lost everything. And my test asserted that the AgentBox *sent* the field, which is precisely the shape of test that cannot catch a consumer dropping it. That is the fourth time in this PR a defect has come from a test feeding itself the thing it was checking.

Copied in **both** projections, because there are two: the RPC path (`internal-api`) and the direct-DB path (`chat-repo`). A field added to one is persisted on some deployments and silently dropped on others — the same class of bug as the menu/candidate divergence from earlier rounds. The new test asserts on the resulting `chat.appendMessage` metadata rather than on the emission, plus that no credential appears in it.

**One thing the audit turned up that I did not fix:** comparing both projections against `DelegationEventPayload` shows `traceId` is absent from the direct-DB path. Pre-existing, unrelated to tiering, and out of scope here — but someone should look at it, since it means delegation events lose their trace correlation on Standalone.

## [P2] Change-detection race

Confirmed. Took the unconditional-refresh option rather than the locked transaction, for the reason the asymmetry gives: a redundant reload costs one session rebuild, a missed one leaves warm sessions advertising a menu the binding no longer matches, producing `revision_mismatch` on every spawn until those sessions die. Same reasoning I gave for the equivalent gate on the other side, so the two now behave consistently.

The comparison helper is deleted rather than left unused. I also had to *invert* one of my own tests from last round — it asserted "unchanged means no reload", which is no longer true and should not be. Replaced with the two cases that matter: supplied-means-reload, and not-supplied-means-no-reload.

## [P2] Golden fixture disagreeing with the producer

You are right, and the irony is not lost on me: the file whose entire purpose is to be the shared byte-level contract was itself the thing disagreeing with the producer. I wrote it from the first round's observation that snake_case was emitted; that changed, and the fixture did not.

Canonical is camelCase now. The snake_case alias is still recorded, but explicitly as **consumer tolerance, not contract** — and the asymmetry is deliberate: a consumer that refuses a spelling turns a cosmetic difference into a silent outage, while a producer that emits only the canonical key keeps the contract crisp. So "the consumer must accept both, the producer should emit one, and the producer may reject the alias" is now stated in the fixture rather than left implicit.

The other half of that item — the fixture not yet being *loaded* on the other side — I cannot do from here. Until it is, this remains one repository asserting against a file the other has not adopted, which is better than two hand-written fixtures but is not the byte-level contract we described.

## Verification

`npm test` 277 files / 6257 passed, 2 skipped. `npx tsc --noEmit` clean.

Seven rounds, and unchanged: **no live end-to-end run, no speedup measurement, no failure drills.** Four of the defects found so far were consumer-side drops or shape mismatches that unit tests structurally could not surface. I would not merge this on green tests.

jacoblee-io:
Fixed in `ed274c7c`. The two management-plane items need that repo.

## [P1] No runtime projection at the persistence boundary

Confirmed, and this is the one I should have seen myself: I added a *compile-time* projection on the producer side and then treated the field as trusted everywhere downstream. The Gateway boundary is HTTP, an HTTP body is typed by assertion, and an assertion strips nothing — so `modelConfig`, a bare `apiKey`, an internal `detail` string all reach a durable record verbatim. The contents of a stored record cannot depend on a caller we do not control.

Both paths now share one runtime allow-list: `requestedTier` / `resolvedTier` / `source` / `provider` / `modelId` / `fallbackReason`, strings only. Two deliberate choices there — **allow-list rather than deny-list**, so a credential-bearing field added upstream is dropped by default instead of persisted until someone remembers to exclude it; and a nested object where a string belongs is **refused rather than coerced**, since that is either a mistake or an attempt to smuggle structure through.

Applied to a group's per-item outcomes as well. That path predates the single-child one and was persisting unsanitized data first — being older is not the same as being trusted.

**And you are right that my test proved nothing.** It asserted "no apiKey" against an input that never contained one. It now feeds `modelConfig: {apiKey}`, a bare `apiKey`, and a `detail` string, and asserts both the exact surviving key set and the absence of each planted value. Same for the group case.

## [P3] False-green test

Confirmed, and it was mine from last round. The mock included a current-state read that the request never performs — `description` is not in the needs-current-state set — which desynchronised every later mock and made the request throw. With no status assertion it passed anyway.

Mock corrected to the two queries that actually happen, and it asserts 200. That is what makes "no reload" mean *the request succeeded and chose not to reload* rather than *the request died before reaching the decision*.

## [P2] Two different fixtures

Agreed, and I want to be precise about what I did and did not do, because I cannot fix this one alone.

Done: the fixture is **byte-pinned by digest**. Changing it now fails a test, and that failure is the reminder that another copy exists and has to move in the same commit. The digest is published in the test and referenced from the fixture header — not stored in the file, since a hash over a whole file cannot live inside it.

Not done, and not doable from here: **choosing which copy is authoritative.** The header now says so explicitly — that the two repositories hold different fixtures with different top-level structure, that runtime only works because the consumer tolerates both spellings, and that this file is a candidate rather than a declaration. If you want mine as the source, the other side copies it and adds a `cmp`/digest check in CI; if you want theirs, I will adopt its structure. Either is fine, but it needs to be decided rather than left as two files that happen not to conflict yet.

## Verification

`npm test` 277 files / 6265 passed, 2 skipped. `npx tsc --noEmit` clean.

Eight rounds. Two of this round's three findings were **my own tests asserting things they could not fail on** — a filter tested with nothing to filter, and a status never checked. That is now the dominant failure mode in this PR, ahead of the code itself. Still no live end-to-end run, no speedup measurement, no failure drills.

jacoblee-io:
Fixed in `32dd6400`. Three of four resolved; the fourth needs a decision that is not mine to make — details below.

## [P2] Sanitizer producing data that does not match its declared type

Confirmed, and the honest description is that I wrote `as unknown as PersistedTierOutcome` — an assertion papering over the fact that the function did not actually produce that type. "Non-empty string" accepted any `source`, any `fallbackReason`, an invalid tier name, `status: "running"`, a negative or fractional `index`.

Now: enums checked against closed sets, tier names against the same pattern the config path enforces, `index` required to be a non-negative safe integer. Two different failure modes on purpose — an unrecognised `source` drops the **whole record** (without it there is nothing to report), while an unrecognised `fallbackReason` drops **just that field**. Both matter because this record is read back to render a UI: a status that is never terminal, or a `-1` index, produces a view nothing downstream can explain.

## [P3] Second persistence path untested

Added, with credentials and malformed items planted so the assertions exercise the filter rather than the input — the same mistake I made on the first path's test two rounds ago.

## [P3] Comment contradicting the implementation

Fixed. It still described the collapse-on-first-bad-tier behaviour that was replaced two commits earlier. Now documents skipping, plus why the revision covers the full config rather than the surviving subset.

## [P2] Golden fixture — partly fixed, and blocked on a decision

I read the canonical copy and agree with its reasoning: **the contract belongs with the sender, the receiver copies it.** So I adopted its structure — same `{config_getAgent, config_getModelBinding}` envelopes, same casing split, same already-projected menu — and the contract test now reads that structure, so a structural change on the producing side fails here.

**But I did not make the bytes identical, and I want to be explicit about why rather than quietly leaving it half-done.** The canonical copy's `provider` values are generated names carrying internal deployment identifiers. This repository is public and must not contain them. Byte-for-byte copying and that constraint are in direct conflict, and copying them in would trade a drift-detection gap for an information-disclosure one — a strictly worse trade.

Two ways out, both requiring a call on your side:

1. **Neutralise the provider names in the canonical copy.** They are sample values, not contract: `provider` is an opaque string to the consumer, which a new test now pins by feeding it `"!!! not a normal name @@@"` and asserting it is accepted. After that this file can be a byte-for-byte copy and `cmp` works.
2. **Keep the canonical copy internal and treat this file as the public projection**, with the producer asserting against both.

Until one is chosen: structure drift is caught, byte drift is not. That is stated in the fixture header rather than left for the next reviewer to rediscover.

One thing I changed about the previous approach: the rules the canonical copy states as prose are now asserted here as **behaviour** rather than mirrored as data in both files. Mirrored data meant each side agreed with its own copy, which is the same non-verification as two independently pinned hashes.

## Note on your `web/pnpm-lock.yaml` observation

Thanks for not reverting it. That file is in the main checkout, not this worktree, and nothing in this branch touches `web/` — so it is unrelated to this PR and I have left it alone too.

## Verification

`npm test` 277 files / 6268 passed, 2 skipped. `npx tsc --noEmit` clean. No internal identifiers under `src/` for this branch's files.

Nine rounds. Still no live end-to-end run, no speedup measurement, no failure drills.

jacoblee-io:
This round's report appears to be against `ed274c7c`, not the current head. Three of its four items were fixed in `32dd6400` (pushed before this review). Evidence, so it can be checked rather than taken on my word:

| Item | Status | Where |
|---|---|---|
| Sanitizer accepting any `source`/`fallbackReason`, bad tier names, `status: "running"`, negative/fractional `index` | Fixed in `32dd6400` | `subagent-models.ts` — `TIER_SELECTION_SOURCES.find`, `TIER_FALLBACK_REASONS.find`, `GROUP_ITEM_STATUSES.find`, `Number.isSafeInteger(entry.index) && entry.index >= 0` |
| `chat-repo` second persistence path untested | Fixed in `32dd6400` | `chat-repo.test.ts:212` — "sanitizes a tier outcome before persisting it (direct-DB path)", with `apiKey`/`modelConfig`/`detail` planted |
| Resolver comment contradicting the implementation | Fixed in `32dd6400` | `model-routing-config.ts:146` — "A tier that cannot be resolved is SKIPPED…" |

The quoted SiClaw digest `8cc4d0…` also predates that commit; it was `e5c5036…` at `32dd6400`. Worth checking the worktree is on the current head — I would rather you re-review the real state than my summary of it.

## The fixture item was real, and measuring it dissolved the conflict

I had left this as "blocked on a decision". That was me not doing enough work. Fixed in `4da568a8`.

**Whole-file hashing could never have detected drift**, and that is the actual defect — not that we pinned the wrong values. Both repositories carry the same payloads under different prose, because each side documents the contract for its own readers, in its own language. So the files differ for reasons that are not the contract, each side pins a digest only it can satisfy, and both CIs pass while nothing is compared. That is two independent constants wearing the costume of a shared contract.

The digest now covers the **canonicalized payload**: commentary dropped, keys sorted recursively, array order preserved (menu order is what the lead reads), no insignificant whitespace, sha256 of the UTF-8 bytes. In Go it is nearly free — unmarshal into `map[string]any` and re-marshal, since `encoding/json` sorts map keys. Two tests pin the properties that make it worth having: **differing prose must not change the digest, and a renamed tier must.**

Then I measured the two payloads instead of continuing to describe the standoff:

> **They differ in exactly one field — the `provider` prefix.** Neutralise it in the canonical copy and both sides compute `de8aa805a00228e637f47d7480940f89361ba35de255ae9e8f76d6c2c31b7e4e`.

That value is recorded in the fixture header, so you can verify it without running anything here.

And it is a one-line change to a **sample value**, not to the contract: `provider` is opaque to the consumer, which the contract test now pins by feeding it `"!!! not a normal name @@@"` and asserting acceptance. So the public-repository constraint and byte-comparability were never actually in conflict — they only looked that way while the comparison was whole-file. I should have measured that two rounds ago instead of writing up the conflict twice.

## Verification

`npm test` 277 files / 6270 passed, 2 skipped. `npx tsc --noEmit` clean. Note the prose edit to the fixture in this commit did **not** move the digest, which is the mechanism demonstrating itself.

Ten rounds. Still no live end-to-end run, no speedup measurement, no failure drills — and those remain the only findings I would expect from here that the suite cannot produce on its own.

jacoblee-io:
The siclaw half of item 3 is fixed in `29029ba9`. Items 1 and 2 are on the other side; notes below.

## [P2] Tier outcome recorded but not displayed — the siclaw half

This is the one that stings, and your framing is right: the runtime records requested vs resolved tier, selection source and fallback reason **precisely so a mis-picked tier is diagnosable**. The design names "a weak child report exposes a bad choice" as one of three mitigations for a wrong tier, and that mitigation needs the tier visible per item. The fields were reaching the payload and being dropped when the card model was built — so every earlier round of this work added plumbing behind a UI that discarded it, and the diagnostic entry point had no exit.

Both cards restore it now, from the two differently-shaped sources: the foreground result flattens the outcome onto `item_results` in snake_case, while a background run only has the persisted terminal event with it nested under `tier` — its tool call returned `launched` long before the outcome existed, so that event is the only record it ever has.

Three deliberate choices in the rendering:

- A fallback shows `requested → actual`, not just the actual model. The difference *is* the diagnostic; showing only the result hides the fact that a choice was overridden.
- An operator-pinned tier (`selection_source: "env"`) is marked, because it explains a choice the agent did not make.
- An unrecognised fallback reason renders **verbatim** rather than vanishing — a reason we have no wording for is still information, and swallowing it would leave a fallback with no explanation at all.

Nothing renders when no tier was involved; a badge on every item of every batch would be noise, and "no tier" is not a diagnostic.

**I also covered the single-spawn card, which your report did not ask for.** A single task is the common case, so leaving it out would have meant the most-used path was the one that could not show which model ran.

Logic extracted into `subagent-tier-view.ts` rather than written inline twice — that is exactly where a shape mismatch silently yields an empty badge, so it is the part worth testing, and `PilotArea` is too large to test through. 11 tests cover both shapes, foreground precedence, the no-tier case, unknown reasons, and hostile input.

## Items 1 and 2

**1 (golden payload):** everything on this side is done — canonical payload algorithm, the pinned value, and the measured fact that the two payloads differ in exactly one sample field. Neutralise the `provider` prefix in the canonical copy and both sides compute `de8aa805a00228e637f47d7480940f89361ba35de255ae9e8f76d6c2c31b7e4e`. Also worth fixing the completion doc claiming byte-level sharing already exists — it does not yet, and a doc asserting it removes the pressure to finish it.

**2 (skipped items showing as running after reload):** this side already merges `item_statuses` for exactly that case, so the recovery logic is there to copy.

## Verification

Backend: `npm test` 277 files / 6270 passed, 2 skipped. `npx tsc --noEmit` clean.
Frontend: `npx vitest run` in `portal-web` — 24 files / 240 passed, `tsc --noEmit` clean.

On the frontend suite not starting for you: a fresh worktree has no `portal-web/node_modules`, so it needs its own `npm ci` in that directory — the root install does not cover it. That is what I did here.

Eleven rounds. Still no live end-to-end run, no speedup measurement, no failure drills — and note this round's finding was invisible to every test on both sides, because it lived in the gap between "data is persisted" and "a human can see it".

src/agentbox/session.ts
```diff
@@ -419,6 +490,237 @@ export class AgentBoxSessionManager {
...
+  private effectiveParentCandidate(
+    parentSessionId: string,
+  ): (ModelRouteCandidate & { params?: BrainModelParams }) | null {
+    const parent = this.sessions.get(parentSessionId);
+    const live = parent?.effectiveModelCandidate;
+    // Carry the parent's ACTUAL tunables alongside its model. Only meaningful
+    // together: restoring one without the other puts the child on the right model
+    // at the wrong level.
+    if (live) return { ...live, params: parent?.effectiveModelParams ?? undefined };
+    if (this.delegationModelProvider && this.delegationModelId) {
```
LikiosSedo: [P1] This box-global fallback can bind a synthetic turn to another session's model credentials. Normal HTTP turns publish `ManagedSession.effectiveModelCandidate` through `onAttemptReady`, but `processSyntheticPrompt` invokes `runPromptWithModelRouting` without that callback. If session B most recently called `setDelegationModel`, a synthetic turn in session A reaches this branch and can spawn with B's provider/model/`modelConfig`. That violates the session-isolation contract documented for credential-bearing candidates. Please publish/clear the effective candidate around synthetic routing too, and remove the box-global fallback in favor of binding state owned by the specific `ManagedSession`. A two-session regression should switch B's binding while A's synthetic turn spawns and assert A keeps its own candidate.

src/agentbox/session.ts
```diff
@@ -2468,6 +2821,16 @@ export class AgentBoxSessionManager {
       // no-op with no active run — so DON'T start the prompt at all, or it would run a fresh,
       // un-aborted turn. Throw straight into the stopRequested branch below.
       if (stopRequested) throw new Error("stopped before sub-agent prompt started");
+      // Not even the parent's effective model could be applied (§7.4). Throw rather
+      // than returning early: this path has to reach the same `finally` — which
+      // shuts down an MCP manager this child opened — and the terminal-event
+      // persistence below, or the card sticks "Running" on reload and the child
+      // leaks its connections.
+      if (tierOutcome.failed) {
```
LikiosSedo, line 2887: [P1] `tierOutcome.detail` contains the raw `registerProvider`/`setModel` exception text. This line puts that text into the thrown error; the catch then copies it into `finalText`, logs it, and builds the persisted/returned terminal summary from it without applying the union redactor created above. A provider setup error that includes an API key or base URL can therefore reach logs, delegation persistence, UI details, and the parent-model-visible tool result. Please redact the setup error before logging it and derive every terminal summary from a sanitized `finalText`; add a regression that plants a tier API key/base URL in the setup error and asserts it is absent from all outputs.

LikiosSedo (commented):
Reviewed the current head `bbef5473`. The three CI jobs are green, but the paths called out inline are not covered by that signal. The first three comments are merge blockers; the final two break the tier-observability acceptance path.

src/agentbox/session.ts
```diff
@@ -2880,6 +3248,9 @@ export class AgentBoxSessionManager {
       // spawn_subagent is available in normal chat (top-level sessions only — child
       // sessions above omit this executor, so sub-agents cannot recurse).
       spawnSubagentExecutor: this.createSpawnSubagentExecutor(),
+      // Snapshot the box's menu at creation — this is the value the session will
+      // hold for its lifetime, and the tool schema built below derives from it.
+      subagentTierMenu: this.subagentTierMenuState,
```
LikiosSedo, line 3329: [P1] This snapshot is always `null` on a cold LocalSpawner. `local-spawner.ts` bootstraps only `tool_capabilities` and `agent_type`, and creating the tools handler merely registers the later reload callback; it never resolves/installs `subagentTierMenuState`. K8s avoids this because `agentbox-main.ts` awaits an initial tools sync before listening. A preconfigured Standalone agent's first turn consequently has candidates but no `model_tier` in the tool schema, so children silently inherit until an admin-triggered tools reload happens. Please bootstrap the resolved menu before LocalSpawner accepts prompts and cover the cold first-turn path.

portal-web/src/components/chat/PilotArea.tsx
```diff
@@ -1461,6 +1468,11 @@ function agentWorkSummary(message: PilotMessage): {
       stringValue(metadata.status) ??
       message.toolStatus ??
       "ready",
+    // Foreground: the collapse envelope's item_results[0] (already resolved above
+    // as firstItem). Background: only the persisted terminal event exists, which
+    // nests the outcome under metadata.tier — the launch returned before it was
+    // known.
+    tier: extractTierOutcome(firstItem, metadata),
```
LikiosSedo: [P2] The foreground single-spawn badge cannot receive a tier outcome here. `result` prefers parsed model-visible content over `details`; for the collapsed single form, `content.item_results[0]` has only `item/status/summary`, while the tier fields exist only in `details.item_results[0]`. Therefore `firstItem` has no tier fields and the common single-task path renders no badge. Please prefer the details item for this UI projection (as the group path already does) and add an integration-level card-model test, not only tests of `extractTierOutcome` in isolation.

portal-web/src/components/chat/PilotArea.tsx
```diff
@@ -2451,6 +2506,7 @@ function groupWorkSummary(message: PilotMessage): {
       childSessionId:
         stringValue(inline?.child_session_id) ??
         stringValue(fold?.childSessionId),
+      tier: extractTierOutcome(inline, fold),
```
LikiosSedo, line 2521: [P2] The background fold never carries the tier data this extraction expects. `annotateSubagentCompletions` reduces terminal events to `{status, summary}`, and `annotateGroupCompletions` reduces both child events and `item_statuses` to index/status/summary, dropping `metadata.tier` / `item_statuses[].tier`. As a result, background single and the default background batch path lose the outcome, including after refresh. Please preserve the tier object through both folds and test the full persisted-event -> folded-message -> card-model path.

LikiosSedo:
I cannot approve the current head `9f0d067e` yet because it is no longer the green head that was re-reviewed.

I refreshed the exact refs:

- PR head: `9f0d067eceee4d703181257b78af7cadb7386042`
- current `main`: `25e7047f162b43ab8f5687ade5454f833f6b605d`

The only merge conflict is `Dockerfile.agentbox`. The new head adds `COPY src/knowledge/ ./src/knowledge/`, but current `main` already contains that same build-graph fix via `c922ccf2` (`ci(agentbox): verify isolated build graph`). Consequently this PR is currently `CONFLICTING / DIRTY`, and GitHub has not created CI checks for `9f0d067e`.

Please rebase onto current `main` and drop the redundant `9f0d067e` change (or resolve it by retaining main's version), then let CI run on the resulting exact head. The five earlier review findings remain fixed in the preceding commits; this is only a fresh-head/merge-readiness blocker.

LikiosSedo (dismissed):
Approved at exact head `3d40b8708f8073381466f3df5cd39a772b573579`.

Re-review evidence:

- rebased onto current `main` (`25e7047f162b43ab8f5687ade5454f833f6b605d`)
- GitHub reports the PR mergeable
- the rebased tree is byte-for-byte identical to the clean merge of current `main` with the previously reviewed and locally validated head `36904460`
- all five prior review findings remain fixed
- Type Check, AgentBox Build Graph, Test, and Portal Web Test all pass on this exact head

The synthetic-turn regression could still be strengthened to drive the full interleaving rather than seed session state directly, and live E2E/performance/failure-drill validation remains outside this PR's static acceptance. Neither is a remaining merge blocker for this bounded pilot.

LikiosSedo (dismissed):
Approved at exact head b28229ab6d4a1ccaa067857be492cb42be6020eb after re-reviewing the new tier-observation commits. Tier state now participates in replica identity, inconsistent or missing menu/candidate revisions fail consensus, and the aggregate exposes tier revisions only when every running box agrees. The previously reviewed patch series remains preserved, the merge tree is clean, focused sync-status coverage passes locally, and all four required CI checks are green.

nightmeng (approved):

PR #520, opened 2026-08-25
fix(runtime): apply release models at turn boundaries

LikiosSedo:
## Summary

Apply Agent Type Release model changes to existing AgentBoxes at a turn boundary, and expose evidence only after the intended release model completes a successful turn.

### Problem

The management server could publish a new Release while warm AgentBox sessions kept the previous model. The reload protocol also had no release identity, so an acknowledgement could not distinguish preparation from actual model execution.

### Solution

- Add identity-bound `model` reloads that invalidate warm sessions without transporting credentials.
- Reject stale release/fingerprint notifications before touching AgentBoxes.
- Name reload response identities `preparedReleaseId` / `preparedModelFingerprint` so an ACK cannot be mistaken for runtime evidence.
- Carry release identity on every prompt path and record it only when the intended model, not a fallback, successfully completes the turn.
- Fail closed when a completed turn does not report which model candidate actually ran.
- Return that observed identity through `/api/sync-status` for the management server's promotion gate.
- Document the turn-boundary and prepared-vs-verified contract in ADR-015.

## Test Plan

- [x] `npm run build` passes with lockfile dependencies
- [x] Full Vitest suite passes: 273 files / 6176 tests passed, 2 skipped
- [x] Focused reload/observation suites pass: 4 files / 151 tests
- [x] Model observation tests cover successful primary turns and reject fallback or unidentified turns as verification
- [x] Reload tests cover online invalidation, stale delivery rejection, and zero-box cold start

## Architecture Checklist

- [x] **Deployment mode**: The reload descriptor/handler is shared by local and K8s AgentBox HTTP servers; both keep model configuration on the established prompt channel.
- [x] **Security model**: Reload notifications contain release identity only, never provider credentials.

## General Checklist

### DIFF
--- a/docs/design/decisions.md
+++ b/docs/design/decisions.md
@@ -559,3 +559,50 @@ than profile) must not be evicted while it is still running.
   the old (richer) payload by ignoring the extra fields.
 
 **Files**: `src/gateway/metrics-aggregator.ts`, `src/shared/metrics-types.ts`, `src/agentbox/http-server.ts`, `src/gateway/server.ts`, `src/portal/siclaw-api.ts`, `portal-web/src/hooks/useMetrics.ts`
+
+---
+
+## ADR-015: Agent Type Releases Apply at Turn Boundaries
+
+**Status**: Active
+
+**Context**:
+An external management server can move an Agent Type's production Release while
+existing AgentBoxes and sessions are warm. Skills, MCP, knowledge, prompt, and
+model configuration belong to one immutable Release, but replacing a model in
+the middle of an in-flight turn would make one turn span two configurations.
+Treating a generic reload ACK as proof that a particular model was prepared also
+allows stale or unrelated Release notifications to report success.
+
+**Decision**:
+Release application is identity-bound and happens at a turn boundary:
+
+- A release reload carries the expected `releaseId` and `modelFingerprint`.
+- Runtime resolves the current binding from the management server and rejects a
+  notification whose observed identity does not match the expected identity.
+- `model` is a reload type whose AgentBox action invalidates warm sessions. The
+  existing invalidation fence lets an in-flight turn finish and rebuilds the
+  session before its next prompt.
+- The next prompt remains the only carrier of provider credentials and model
+  configuration. Reload notifications carry identity, never secrets.
+- Runtime reports the observed Release identity and number of reached boxes.
+  Zero boxes means cold-start convergence, not runtime model verification.
+- A reload ACK proves preparation only. AgentBox records an observed release
+  and model fingerprint after a real turn completes successfully; only that
+  observed inventory can satisfy management-server runtime verification.
+
+**Consequences**:
+
+- ✅ Existing Agents switch models without a Pod restart on their next turn.
+- ✅ A turn never mixes model configurations.
+- ✅ Reload events are safe to retry and stale Release ACKs are detectable.
+- ✅ Provider credentials stay on the established prompt/config channel.
+- ⚠️ "Published" and "applied to an online box" are distinct states.
+- ⚠️ "Prepared for next turn" and "successfully ran a turn" are distinct
+  states; promotion gates consume only the latter.
+- ⚠️ Management servers predating this contract must keep model deliveries
+  pending while an older Runtime is deployed; rolling upgrades should deploy
+  Runtime support before relying on verified model delivery.
+
+**Files**: `src/shared/gateway-sync.ts`, `src/gateway/server.ts`,
+`src/gateway/agent-model-binding.ts`, `src/agentbox/sync-handlers.ts`
--- a/docs/design/decisions.md
+++ b/docs/design/decisions.md
@@ -577,15 +577,18 @@ allows stale or unrelated Release notifications to report success.
 **Decision**:
 Release application is identity-bound and happens at a turn boundary:
 
-- A release reload carries the expected `releaseId` and `modelFingerprint`.
-- Runtime resolves the current binding from the management server and rejects a
-  notification whose observed identity does not match the expected identity.
+- Identity-bound callers send the expected `releaseId` and `modelFingerprint`.
+  Runtime resolves the current binding from the management server and rejects a
+  notification whose prepared identity does not match either supplied value.
+  Legacy callers may omit the expected identity; that compatibility path reports
+  what was prepared but cannot reject a stale notification.
 - `model` is a reload type whose AgentBox action invalidates warm sessions. The
   existing invalidation fence lets an in-flight turn finish and rebuilds the
   session before its next prompt.
 - The next prompt remains the only carrier of provider credentials and model
   configuration. Reload notifications carry identity, never secrets.
-- Runtime reports the observed Release identity and number of reached boxes.
+- The reload response reports the prepared Release identity and number of
+  reached boxes as `preparedReleaseId` and `preparedModelFingerprint`.
   Zero boxes means cold-start convergence, not runtime model verification.
 - A reload ACK proves preparation only. AgentBox records an observed release
   and model fingerprint after a real turn completes successfully; only that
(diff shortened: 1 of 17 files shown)


jacoblee-io (commented):
The mechanism is well chosen: `model` joins the existing immutable-session reload family and reuses `invalidateSessions`, so the turn-boundary guarantee comes from a fence that already exists rather than new machinery. `gatewayPath: ""` with `requiresGatewayClient: false` makes it structurally impossible for this reload to carry provider config, which is a stronger statement than a comment saying it doesn't. ADR-015 is the right place for the prepared-vs-verified distinction, and the consequences section states the rule a promotion gate needs.

Three things below. The first is a naming change I'd want before this merges, because it works against the exact distinction the PR exists to establish — and it is the field a management-side implementation is most likely to read.

## 1. The reload RPC returns `prepared` state under the name `observed`

```js
let preparedReleaseId = "";
let preparedModelFingerprint = "";
…
return {
  ok: true, reloaded, failed, boxes: targets.length,
  observedReleaseId: preparedReleaseId, observedModelFingerprint: preparedModelFingerprint,
};
```

The local variables are named accurately — these come from `resolveAgentModelBinding`, i.e. what the Runtime *resolved and is about to apply*. Nothing in that value reflects a turn having run. The genuinely observed identity is the separate `model` field on `/api/sync-status`, recorded by the box after a successful turn.

ADR-015 has the same slip, in adjacent bullets:

> - Runtime reports the **observed** Release identity and number of reached boxes.
> - A reload ACK proves **preparation only**.

Both sentences describe the same return value, and they disagree. The consequences section then gets it right (`"Prepared for next turn" and "successfully ran a turn" are distinct states; promotion gates consume only the latter`), but a reader implementing the gate is most likely to reach for the field literally called `observedReleaseId` on the reload response — which would reproduce precisely the failure this PR fixes, while appearing to follow the ADR.

Renaming the response fields to `preparedReleaseId` / `preparedModelFingerprint` and aligning that first ADR bullet makes the wrong integration hard to write. This is worth doing even though the current names are internally consistent with nothing, because it is a cross-repository contract and the name is the API.

## 2. `ranIntendedModel` defaults open, and it gates the evidence

```js
const intendedCandidate = body.modelProvider && body.modelId
  ? candidateKey({ provider: body.modelProvider, modelId: body.modelId })
  : undefined;
const ranIntendedModel = !result?.activeCandidateKey || result.activeCandidateKey === intendedCandidate;
```

The two unknowns are treated oppositely:

- `intendedCandidate` missing → the equality fails → **no** evidence recorded (fail-closed, correct).
- `activeCandidateKey` missing → short-circuits true → evidence **is** recorded, without anything having confirmed which model ran (fail-open).

For a value whose only purpose is to satisfy a promotion gate, both unknowns should fail closed. As written, "we could not tell which candidate ran" and "the intended candidate ran" produce the same output.

I did not fully establish how a successful turn reaches this line with the key unset, so I'm flagging the asymmetry rather than claiming a specific trigger. What I did confirm: the field is optional throughout `model-routing.ts`, it is explicitly cleared in `clearModelRouteUserSelectionIfDifferent`, normalization at the state boundary maps a non-string to `undefined`, and CLAUDE.md documents a routing edge — "the no-current-model edge where the runner does a bare `brain.prompt`" — where no candidate participates at all. If that edge is genuinely unreachable when `releaseId` and `modelFingerprint` are both present, a comment saying so would be enough. Otherwise requiring `result?.activeCandidateKey === intendedCandidate` outright costs nothing: an unverified box simply stays unverified until a turn proves otherwise, which is the conservative direction for a gate.

## 3. Stale rejection is opt-in — worth stating as such

```js
if (expectedReleaseId && preparedReleaseId !== expectedReleaseId) throw …
if (expectedModelFingerprint && preparedModelFingerprint !== expectedModelFingerprint) throw …
```

A caller that omits the expected identity gets no staleness check. That reads as deliberate backward compatibility, and given the caller is the management plane over the internal mesh rather than an untrusted client, the threat model is a management-side bug rather than an attacker — so opt-in is defensible. It's a different shape from the negotiated handshake used elsewhere in the codebase (where the *box* only enables new behaviour when the runtime advertises support), and the ADR's summary line "stale Release ACKs are **detectable**" is accurate in a way "rejected" would not be. Worth keeping that wording precise, since a management-side author might otherwise assume the Runtime enforces this unconditionally.

## Deploy note not covered by the ADR

The ADR's rollout bullet covers Runtime-before-management, but `observedModel` is recorded in `src/agentbox/http-server.ts`, which ships in the **agentbox** image, not the runtime. So verified-model delivery additionally needs an agentbox rebuild plus a `SICLAW_AGENTBOX_IMAGE` bump and pod recycle — and existing boxes won't pick it up. Without that, a Runtime and management server that both support this contract will wait indefinitely for evidence that no box is able to produce, which fails as a hang rather than an error. Worth adding to the ADR's rollout consequence next to the Runtime ordering.

## On reviewing this together with the management-side change

I can't reach the internal GitLab MR from here, so I've reviewed only this side. The parts I'd specifically want checked against it, since they're where a mismatch would be invisible in this repo's tests:

1. Which field the promotion gate consumes — `/api/sync-status` → `model.{releaseId,modelFingerprint}` (correct) versus the reload response's `observedReleaseId` (prepared only). Item 1 above exists to make this hard to get wrong.
2. Whether the management server sends `releaseId` **and** `modelFingerprint` **and** `modelProvider`/`modelId` on the prompt path. All four matter: the first two gate whether evidence is recorded at all, the last two are what `intendedCandidate` is built from — and per item 2, omitting the provider/model pair currently means no evidence rather than false evidence, which is the safe direction but silent.
3. Whether it treats `boxes: 0` as cold-start convergence rather than success — the ADR says it must, and this response now carries `preparedReleaseId` even in the zero-box branch.
4. Whether it keeps model deliveries pending against an older Runtime, per the ADR's rollout bullet, and whether its rollout accounts for the agentbox image above.

Happy to review the pairing properly if you can paste the MR's diff or description.

LikiosSedo:
Thanks for the careful review. Addressed in `9fc77eda`:

1. The reload RPC now returns `preparedReleaseId` / `preparedModelFingerprint`; the old `observed*` fields are absent, including the zero-box path.
2. Model evidence now fails closed: a successful turn must report an `activeCandidateKey` that exactly matches the requested provider/model candidate. Missing identity and fallback execution both leave `/api/sync-status.model` unverified.
3. ADR-015 now states that stale rejection applies when the caller supplies expected identity; the legacy compatibility path reports preparation without claiming unconditional rejection.
4. The rollout consequence now calls out the AgentBox image update and one-time recycle of existing Pods before verified delivery is enabled. Subsequent Release model changes still apply at the turn boundary without a Pod restart.

Verification with lockfile dependencies:

- `npm run build`
- Full Vitest suite: 273 files, 6176 passed, 2 skipped
- Focused reload/observation suites: 4 files, 151 passed

jacoblee-io (approved):
Approving. All three items are addressed, and two of the fixes go past what I asked for.

## The test settles what I couldn't

My second item came with an admission — I flagged the fail-open asymmetry but said I hadn't established how a successful turn reaches that line with `activeCandidateKey` unset. `9fc77eda` answers it by construction:

```js
it("does not verify the release model when the successful turn has no observed candidate", async () => {
  session.brain.getModel.mockReturnValue(undefined);
  … releaseId: "release-2", modelFingerprint: "fingerprint-2"
  expect(prompt.status).toBe(200);
  expect((await getJson(port, "/api/sync-status")).data.model).toBeNull();
});
```

That's exactly the combination that worried me — turn succeeds, both identity fields present, no observed candidate — and it now asserts the evidence stays `null`. So the path is reachable, which means the old `!result?.activeCandidateKey ||` short-circuit really would have recorded false verification, and the behaviour is pinned rather than reasoned about. Better than a comment explaining why it can't happen.

The predicate itself is now closed on both unknowns, and explicitly rather than by truthiness:

```js
const ranIntendedModel = intendedCandidate !== undefined
  && result?.activeCandidateKey === intendedCandidate;
```

## The rename is complete, and guarded against coming back

Both return sites carry `preparedReleaseId` / `preparedModelFingerprint`, including the zero-box branch, and the tests add:

```js
expect(result).not.toHaveProperty("observedReleaseId");
```

Asserting the *absence* of the old name is the part I wouldn't have thought to ask for. It stops someone re-adding `observedReleaseId` as a compatibility alias later, which would quietly restore the ambiguity this PR exists to remove.

ADR-015 now matches the code — the reload response is described as reporting *prepared* identity, and the staleness bullet states plainly that identity-bound callers get rejection while legacy callers "cannot reject a stale notification". That last clause is the honest version; "detectable" was accurate but easy to read as enforced.

## The rollout consequence is more complete than my note

I only said the agentbox image needed rebuilding. The ADR now gives the whole ordering — Runtime first, then an AgentBox image with observed-model reporting, then `SICLAW_AGENTBOX_IMAGE`, then recycle existing Pods, *then* enable verified delivery — and adds the distinction I didn't make:

> Subsequent Release model changes use the turn-boundary session fence and do not require another Pod restart.

That separates the one-time cost of turning the contract on from the steady-state behaviour, which is what an operator reading this actually needs to know. `src/agentbox/http-server.ts` in the Files list is right too, since that's where the evidence is recorded.

## On the management-side pairing

I withheld approval last round partly because I can't reach the paired management-side change and the contract spans both. That concern is much smaller now: with `preparedReleaseId` no longer wearing the word "observed", the field a gate must *not* consume is named accurately, and the ADR bullet describing it agrees. The wrong integration is now hard to write by accident rather than the default reading — which is the property I actually wanted, and it holds whether or not I can see the other side.

Still worth confirming there, since a mismatch wouldn't show up in this repo's tests: that the gate reads `/api/sync-status` → `model.{releaseId,modelFingerprint}`; that the prompt path sends `modelProvider`/`modelId` alongside `releaseId`/`modelFingerprint` (without the first pair, `intendedCandidate` is `undefined` and evidence is now correctly withheld — safe, but silent); and that `boxes: 0` is treated as cold-start convergence.

Thanks for the itemized reply — knowing which commit addressed what, and that item 4 was folded into the ADR rather than left implicit, made this pass quick.

PR #521, opened 2026-08-26
feat(core): compile agent prompts with capability harness

LikiosSedo:
## Summary

Compile every Agent session from one type-aware context so role prompt, built-in capabilities, configured resources, memory, and skill roots no longer drift apart. Knowledge QA now cold-starts without SRE cluster-binding, shell, memory, workflow, or ambient operational-skill guidance, and has a complete source-backed retrieval stack.

### Problem

The shared default prompt was SRE-oriented and every Agent type appended its identity after it. A read-only Knowledge QA Agent therefore still saw cluster and command guidance. Its knowledge prompt also said there was no search tool even though Grep/Find existed, so misses often recovered only after a user named the document. Initial capability lookup failures shared the same null value as legacy unrestricted Custom, and LocalSpawner could expose process-global MCP descriptions while dropping explicitly bound Skills.

Built-in capability groups and configured resources are different axes. Dynamic MCP tools cannot be safely classified from their names, and the current MCP payload has no trustworthy read/write or binding-source metadata.

### Solution

- Add a shared Agent Context compiler for the role-neutral platform prompt, built-in tool policy, configured MCP exposure, memory, and skill roots.
- Give `read_files` a first-class Knowledge QA retrieval stack: Agent-scoped hybrid `knowledge_search`, Grep/Find fallback, full-page Read, and manifest-backed `knowledge_cite`.
- Lock Knowledge QA and Coordinator built-in capabilities in Portal-backed TUI and AgentBox sessions while retaining explicitly configured skills, knowledge, and MCP.
- Keep Coordinator resource discovery and authorization distinct: an attached locator skill/MCP may resolve a concrete Pod/Job/Node/ID/IP to one binding; `list_delegates` remains the coverage check; Coordinator does not receive its own `cluster_list`/`host_list`.
- Require explicit, valid Agent type and capability provenance at permission boundaries. Missing rows/types, malformed capability values, failed sync, and built-in-plus-null payloads fail closed; only explicitly resolved Custom-plus-null stays unrestricted.
- Apply domain-neutral state-change authorization, blast-radius, and confirmation guidance to every harness; keep infrastructure examples conditional.
- Isolate LocalSpawner Knowledge, Skills, MCP config, clients, initial sync, and reload handlers per Agent. It never writes Agent MCP bindings into process-global settings or materializes Skills into a shared tree.
- Keep visual MCP usage generic: channel guidance marks render tools as optional/availability-dependent; binding `create-chart` remains valid for QA.
- Emit non-sensitive session and final provider-envelope manifests with prompt length/hash, tool-schema hash, tool names, configured MCP server names, and known-section markers.
- Document the lifecycle, retrieval stack, resource axes, and decision in ADR-016.

### Boundary

Siclaw does not guess whether a configured MCP is “knowledge” or “SRE” from server/tool names. Agent-type-safe MCP selection remains a control-plane binding responsibility until effect classification and binding provenance exist end to end. Domain-neutral side-effect safety applies even to unclassified configured MCP, and the runtime manifest makes the actual model-visible set auditable.

### Compatibility

Custom Agents with an explicit, successfully resolved null capability selection keep legacy unrestricted built-ins. Scoped QA/Coordinator Agents no longer inherit repo-bundled or host-global operational skills; required Skills and MCP must be explicitly bound. LocalSpawner now waits for concurrent Knowledge/Skill/MCP initial sync before accepting its first prompt.

## Test Plan

- [x] `npx tsc --noEmit`
- [x] Full root suite: 276 files, 6215 passed, 2 skipped
- [x] Knowledge cold-start regressions: title-free hybrid recall, exact version fallback, empty result
- [x] Local two-Agent Knowledge/Skill/MCP isolation and authoritative unbind tests
- [x] Strict Gateway/Local/TUI capability provenance and built-in-null rejection tests
- [x] Direct prompt assembly safety tests
- [x] Portal Web suite: 23 files, 229 passed
- [x] Portal Web production build


jacoblee-io (commented):
Scope note first: at 32 files / +1097 I did not read this exhaustively. I focused on four things — whether the platform safety layer survived the refactor, whether the capability fail-closed is complete, whether the new envelope manifest is actually non-sensitive, and the blast radius of the breaking change. Everything below comes from those; the Agent Context compiler's structure and the doc updates I read but did not audit line by line.

## The capability fix is the right shape, and better than fail-closed alone

Three separate fail-open paths are gone: `data ?? { allowedTools: null }` no longer manufactures an unrestricted payload from a missing fetch, a non-array or non-string-bearing `allowedTools` now throws instead of degrading to `null`, and `agentType` must be in `VALID_AGENT_TYPES` rather than defaulting to `custom`.

What makes it more than a fail-closed patch is `harnessResolvedState` / `resolution`. "Resolved, and the operator really chose unrestricted" and "we never established what this Agent may do" were the same `null` before; they're now distinct states, so legacy Custom keeps working without that compatibility path doubling as a swallow-all for lookup failures. Every `include*` in `resolveAgentHarness` then leads with `resolution === "resolved"`, so an unproven Agent gets no guidance at all rather than SRE defaults.

`mcpExposure` failing closed on unresolved and delegated-read-only contexts is consistent with that, and the comment explaining why MCP can't be narrowed by name is the honest version of the boundary.

## The one asymmetry I'd want changed

Five of the six harness booleans gate on the tools actually present:

```js
includePlanningGuidance:   … hasAnyTool(allowedTools, ["task_create", …]),
includeSubagentGuidance:   … hasAnyTool(allowedTools, ["spawn_subagent"]),
includeInfrastructureGuidance: … hasAnyTool(allowedTools, ["cluster_list", "host_list"]),
```

`includeOperationalSafety` gates on the type instead:

```js
includeOperationalSafety:
  resolution === "resolved" && !delegatedReadOnly &&
  (agentType === "sre" || (agentType === "custom" && canOperate)),
```

So `knowledge_qa` and `coordinator` can never receive the operational-safety section, by type rather than by absence of operational capability. That's fine while their capability sets stay read-only — but this PR also states, correctly, that Siclaw cannot classify a configured MCP's effects:

> Siclaw does not guess whether a configured MCP is "knowledge" or "SRE" from server/tool names.

and `mcpExposure` is `"configured"` for any resolved, non-delegated Agent regardless of type. A `knowledge_qa` Agent with an MCP server that performs actions therefore has operational reach with the operational guidance structurally unreachable. The old code's deleted comment — `// Append hardcoded safety section — NOT overridable by agent templates` — is what makes this worth raising: that line was the marker for the invariant CLAUDE.md states for this file ("never replace platform safety/mode/context assembly"), and the replacement makes the section conditional on something other than capability.

To be clear about what this is and isn't: `resolveAgentHarness`'s own comment says *"Prompt wording is never used as a permission gate"*, and that's right — omitting the section grants nothing, since reach comes from tool registration and operator-configured MCP. So this is not a privilege hole. It's an Agent that can act without being told the rules for acting, in exactly the configuration the PR says the control plane cannot yet prevent.

Aligning it with the other five — `hasAnyTool` over the exec/write tools, or `|| mcpExposure === "configured"` — costs nothing and removes the exception. If the type gate is deliberate, the reasoning belongs next to it, because the surrounding five lines establish a pattern that this one silently breaks.

## Envelope manifest: non-sensitive as claimed

`ModelEnvelopeManifest` emits `{chars, sha256}` for instructions, tool names plus a schema hash, and six booleans. `instructionMessages` collects only `system` / `developer` roles, so user turns never enter the hash — which matters twice: content stays out, and the hash stays stable enough to be worth comparing. Tool names are already model-visible, so exposing them costs nothing. I don't see a leak here.

## Two process points

**`prompt.test.ts` wasn't touched.** `prompt.ts` changed by +111/-33, including `buildSreSystemPrompt` → `buildSystemPrompt` with a positional-to-object signature change and the safety section becoming conditional. CLAUDE.md names `prompt.test.ts` as the verification for this file. Coverage does exist — `agent-context.test.ts` (+136) tests the harness derivation and `agent-factory.test.ts` picks up the wiring — so I'm not claiming a hole, but the conditional-safety-section behaviour is now asserted only indirectly, one layer away from the function that assembles it. A case in `prompt.test.ts` pinning "safety section present when `includeOperationalSafety`, absent when not" would put the assertion where the matrix says to look.

**The PR does several things at once.** The Solution section lists eight items spanning a new compiler, a security fix, a Coordinator capability change, a breaking removal of ambient skills, a new audit surface, and an ADR — while the General Checklist ticks "Changes are focused on a single logical change." The capability fail-closed is a fix you'd want to land quickly; "Scoped QA/Coordinator Agents no longer inherit repo-bundled or host-global operational skills" is a change existing Agents feel silently and that you'd want to stage. Bundled, they inherit each other's pace. Not asking for a split at this point — but the breaking half deserves its own callout for whoever operates the upgrade, since an existing QA Agent loses skills with no error, and the compatibility note is currently a sentence inside a long description.

## Noted as correct

Dropping `inspect_infra` from Coordinator with the reasoning inline (`cluster_list` describes the coordinator's *own* bindings and primes the wrong route) strengthens the existing `list_delegates` contract — the tool most likely to be mistaken for coverage evidence is now simply absent. The rewritten routing prose keeps discovery and authorization separate in the same terms that contract uses ("the helper discovers identity but NEVER authorizes delegation"), and `portal-web/src/lib/agentTypes.ts` moved with the registry, which is the pairing that file requires.

LikiosSedo:
Addressed the review and the related QA cold-start boundary in `89c03b84`:

- moved domain-neutral state-change authorization / blast-radius / confirmation rules into the always-present common Safety section; infrastructure examples remain conditional;
- added direct `prompt.test.ts` coverage for common safety and conditional operational guidance;
- restored `knowledge_search` as a real Agent-scoped hybrid index and made Knowledge QA use search → Grep/Find → Read → `knowledge_cite`;
- made missing/unknown Agent type, malformed capabilities, missing Local row, and built-in + null payloads fail closed; only explicit Custom + null remains unrestricted;
- isolated LocalSpawner Knowledge, Skills, MCP config, per-box clients, initial sync, and reload handlers, including two-Agent/unbind tests;
- made optional Skill/visual prompt claims availability-aware and updated the compatibility/upgrade callout in the PR description.

Verification: `npx tsc --noEmit`; root suite 6215 passed / 2 skipped; Portal Web 229 passed + production build; all three GitHub checks pass.

LikiosSedo:
Follow-up review fixes landed in `1e7753df`:

- Agent/Portal-scoped Skill roots are now authoritative even when the directory does not exist yet; Local cold-start sync failure no longer falls back to process-shared `skills/resolved/`.
- Skill-root selection moved behind one tested policy used by `agent-factory`, with coverage for missing scoped roots, scoped-vs-shared precedence, and the unscoped K8s path.
- The compiler now expands `null` through each built-in Agent type’s locked capability groups. Only explicit Custom + null remains unrestricted, and unknown Agent types are rejected rather than normalized to Custom.

Verification on this commit: `npm test` (277 files, 6220 passed / 2 skipped), `npx tsc --noEmit`, and `npm run build`. GitHub Type Check, Test, and Portal Web Test are all green.

jacoblee-io (commented):
Both of my points are addressed, and the safety fix is a better cut than the one I proposed. I'm still not approving, for a reason that isn't about a defect — see the last section.

## The safety split is the right boundary

I suggested gating `includeOperationalSafety` on tool presence. Splitting by *content* is better: the domain-neutral rules (state change requires an explicit ask, state the target/impact/blast radius, require explicit confirmation) are now unconditional, and only the infrastructure examples stay behind the harness flag. That means the case I raised — a `knowledge_qa` Agent whose configured MCP can act — now receives the rules that actually govern acting, without inheriting `kubectl` vocabulary that would be noise for it.

The test pins exactly that boundary, including the direction that's easy to lose later:

```js
expect(prompt).toContain("change external state only when the user explicitly asks");
expect(prompt).not.toContain("delete/evict/cordon");
```

Asserting the neutral section *doesn't* contain the domain example is what stops someone folding the examples back into the common text. And it's in `prompt.test.ts`, which is where the change-impact matrix says to look.

## Two fail-opens you found that I missed

`allowedTools === null` no longer means unrestricted for a built-in type — it expands through that type's locked capability groups, with only explicit Custom + null staying unrestricted. A caller that hadn't already expanded capabilities would previously have handed a `knowledge_qa` Agent every tool. `normalizeAgentType` → `requireAgentType` closes the sibling case, so an unknown type is rejected rather than quietly becoming Custom.

Same shape in `1e7753df`: a Local cold-start sync failure no longer falls back to the process-shared `skills/resolved/`. That's the same class of bug as the knowledge-directory isolation work — a failure path silently reaching shared state in the one mode where every box shares a filesystem. Good catch, and not something I flagged.

## Why I'm not approving: I can't claim to have reviewed this

Numbers, because they're the argument:

- when I reviewed it: 32 files, +1097/-196 — and I opened that review by saying I had *not* read it exhaustively, and named the four things I did check
- now: 47 files, +2067/-392
- the delta alone: 38 files, +1002/-228 — larger than half the PR I already couldn't fully cover

This round I verified two files closely (`prompt.test.ts`, `agent-context.ts`) and read the reply. I have not meaningfully reviewed `src/knowledge/indexer.ts`, `src/core/skill-directories.ts`, the reworked LocalSpawner per-box isolation, the `sync-handlers.ts` changes (+93/-39 on top of the earlier +14/-7), or the four doc updates. Approving would put my name on roughly a third of a 2000-line change to prompt assembly, capability resolution, and resource isolation — three areas where the failure mode is silent.

This isn't a complaint about the fixes. Each round has answered the review precisely, and several of the additions are genuinely better than what was asked for. But the loop has a shape worth naming: a review finds a boundary problem, the fix lands in the same PR, the PR grows, and the next review covers a smaller fraction of it. The unreviewed remainder is where the next boundary problem will be.

Three of the additions look separable to me, and each would get more scrutiny alone than it will here:

1. **`knowledge_search` / `src/knowledge/indexer.ts`** — a new retrieval surface with its own index. It's a feature, not a fix to the harness.
2. **`src/core/skill-directories.ts` + the scoped-root precedence policy** — a standalone refactor with its own tests; it happens to fix a fail-open but doesn't depend on the context compiler.
3. **LocalSpawner per-box Knowledge/Skills/MCP isolation** — continues an earlier isolation effort and stands on its own, including the two-Agent/unbind tests.

What's left after those — the context compiler, the capability fail-closed, the prompt split, the envelope manifest, ADR-016 — is still a large PR but a coherent one, and the security-relevant half would stop being gated on a new feature's review.

If you'd rather land it whole, that's a legitimate call and I won't re-litigate it. In that case I'd ask for a second reviewer on the parts I've listed as uncovered, rather than an approval from me that overstates what was checked. Happy to review any of the three split-outs properly if you go that way.

jacoblee-io (approved):
Approving. My previous comment withheld approval on coverage, not on any defect — so I went and read the parts I'd listed as uncovered. They hold up, and one of them answers a question only someone who reviewed the earlier isolation work would think to ask.

## The lesson from the knowledge-handler work was generalized correctly

This is the check I most wanted to run. That effort ended with a subtle bug: one local box created *two* handler instances — one in LocalSpawner for the initial sync, another inside `createHttpServer` for reloads — so per-handler state split in half. The fix was to build it once and pass it in.

This PR adds two more per-box handlers, which is exactly where that bug would recur. It doesn't:

```js
if (options.mcpHandler) { perServerHandlers.mcp = options.mcpHandler; }
else if (sessionManager.mcpServersState !== undefined) { … }

createHttpServer(sessionManager, { disableIdleShutdown: true, knowledgeHandler, skillsHandler, mcpHandler });
```

Built once in LocalSpawner, passed in, with the self-construct branch reserved for the path that doesn't supply one — and the fallback conditions (`mcpServersState !== undefined`, `skillsDir`) leave K8s on the process-global handlers, matching how knowledge already behaves. That's the pattern applied deliberately rather than copied loosely.

## The three previously-unread modules

**`skill-directories.ts`** — `scopedSkillsDir !== undefined` short-circuits the whole fallback chain, so a scoped root that doesn't exist yet yields *no* skill dirs rather than the shared `resolved/` tree. The comment names the exact scenario. Checking `!== undefined` rather than truthiness is the right call here, since the question is "was a scope supplied", not "does it exist".

**`knowledge/indexer.ts`** — two things I'd have flagged if they were wrong, and both are right. The index db is keyed by a hash of the resolved knowledge path, so mounts can't share an index; and it lives in a separate `indexRoot`, deliberately outside the mount. That second one matters because the knowledge directory is replaced atomically via the staging swap — an index stored inside it would be deleted by its own content update. Disabling temporal decay for knowledge (unlike memory) is also the right default.

**LocalSpawner per-box isolation** — same shape as the knowledge work, applied to skills and MCP: `resolveUnderDir(skillsRoot, agentId)` for the path, `mcpServersState = {}` set explicitly so an unanswered Gateway pull can't fall through to process-global `settings.json`, `preserveExistingOnEmpty: false` matching the empty-bundle semantics, and the agent row query tightened from `rows.length > 0 ? … : null` to `rows.length !== 1 → throw`. The comment "a failed axis stays empty/fail-safe" describes what the code does.

Together with the earlier round — the safety split with its negative assertion in `prompt.test.ts`, `null` no longer meaning unrestricted for a built-in type, `requireAgentType` rejecting unknown types, the envelope manifest emitting only hashes/lengths/names/booleans — I've now covered every path where I could construct a failure.

## What I still haven't read closely

Two things, both of which I'm comfortable accepting rather than claiming to have verified:

- the bodies of `createSkillsHandler` / `createMcpHandler` in `sync-handlers.ts` (I read their *call sites* and the capability-parsing changes, and they're structural copies of the knowledge factory that's been reviewed several times);
- the four design-doc updates, which don't affect runtime behaviour.

Stating that rather than implying full coverage.

## On my splitting suggestion

That still seems right to me as a process point — a review loop where each round's fix enlarges the PR does shrink the next round's coverage, and `knowledge_search`, the skill-root policy, and the LocalSpawner isolation would each get more scrutiny alone. But it was never a merge condition, and I don't want it read as one. Having now read those parts, I don't have a reason to hold the PR for them.

CI green across Test / Type Check / Portal Web Test; 6220 passing locally per your last comment.

PR #522, opened 2026-08-26
feat(runtime): observe personal preview harness

LikiosSedo:
## Summary
- version the AgentBox sync-status contract and query every running box
- materialize personal Preview Skill overlays and explicit image Built-in masks
- expose configured Agent prompt, loaded Skill names/digests and actual tool functions after a successful turn
- refresh Skill evidence after hot reload and require replica consensus for aggregate Harness/model evidence
- preserve per-box evidence while redacting transport failures
- bind Skill script lookup to each Agent session so LocalSpawner previews cannot share resolved resources or Built-in policy
- normalize older/partial AgentBox status payloads during Runtime/AgentBox rolling upgrades
- include the knowledge indexer in the isolated AgentBox image build and verify the builder stage in CI

## Contract boundary
Sicore sends the effective non-image Skill bundle after Agent Type inheritance, masks, and personal overlay resolution. Siclaw applies the accompanying policy only to image Built-ins. The observed prompt field is the configured Agent prompt/template, not the fully compiled platform prompt.

## Validation
- npm test: 278 files passed; 6233 passed, 2 skipped
- npx tsc --noEmit
- npx tsc -p tsconfig.agentbox.json --noEmit
- focused script resolver/tool tests: 69 passed
- gateway sync-status tests: 9 passed
- AgentBox builder-stage Docker build runs in CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,6 +19,13 @@ jobs:
       - run: npm ci
       - run: npx tsc --noEmit
 
+  agentbox-build-graph:
+    name: AgentBox Build Graph
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - run: docker build --target builder -f Dockerfile.agentbox .
+
   test:
     name: Test
     runs-on: ubuntu-latest
--- a/Dockerfile.agentbox
+++ b/Dockerfile.agentbox
@@ -22,7 +22,7 @@ WORKDIR /app
 COPY package*.json ./
 RUN npm ci
 
-# Copy backend source only (AgentBox runtime graph: agentbox-main + core + tools).
+# Copy backend source only (AgentBox runtime graph: agentbox-main + core + tools + knowledge).
 # Gateway code (spawners, security, db) is intentionally NOT built here — the
 # agentbox image runs only agentbox-main.ts, which never imports src/gateway/*.
 # Those modules are compiled in the runtime/portal images instead.
@@ -31,6 +31,7 @@ COPY src/agentbox-main.ts ./src/
 COPY src/agentbox/ ./src/agentbox/
 COPY src/core/ ./src/core/
 COPY src/cron/ ./src/cron/
+COPY src/knowledge/ ./src/knowledge/
 COPY src/memory/ ./src/memory/
 COPY src/shared/ ./src/shared/
 COPY src/tools/ ./src/tools/
--- a/docs/design/decisions.md
+++ b/docs/design/decisions.md
@@ -83,7 +83,9 @@ Bundles contain **only global + skillset (dev only) + personal skills**. Core sk
 - ✅ Core skills are versioned with the code, not the database
 - ⚠️ `skillsHandler.materialize()` does NOT restore core skills — it only writes what's in the bundle
 - ⚠️ In local mode, only the scoped `createSkillsHandler()` may materialize a bundle, under `.siclaw/skills/agents/<agentId>/`; the process-global handler remains unsafe
-- ⚠️ If a user disables a core skill, the `disabledBuiltins` list in the bundle tells AgentBox which core skills to skip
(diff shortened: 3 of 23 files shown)


jacoblee-io (approved):
Approving. Scope first: at 967 lines I concentrated on three places where this line of work has produced bugs before — per-agent isolation entry points, cross-version payload parsing, and the direction of the evidence gate. Uncovered parts listed at the end.

## What I verified

**No isolation entry point was missed.** `script-resolver.ts` moves from process-global path derivation to injected options, which is the same shape as the earlier knowledge/skills isolation work. The risk with that refactor is a call site that keeps the implicit fallback and silently shares state under LocalSpawner — so I enumerated the consumers on `main`: `node-script`, `host-script`, `local-script`, `pod-script`, and this PR updates all four, plus `agent-factory` constructing the resolver and `tool-registry` carrying it. Nothing left reaching for the global path.

**The evidence gate fails closed, on both conditions.**

```js
const consistent = observed.length === targets.length &&
  observed.every((item) => identity(item.status) === firstIdentity);
```

An unreachable replica is as disqualifying as a disagreeing one, which is the right pairing — "we couldn't see one of them" must not read as consensus. Comparing a normalized identity while retaining the content/version fields keeps the agreement test from being sensitive to fields that legitimately differ. And `observations.some((item) => … item.reason === "unsupported")` preserves the actionable reason when every replica fails, which is the same property worth keeping from the earlier single-box work.

**`normalizeBoxSyncStatus` is written for the problem it actually has.** Per-field type checks, malformed repos dropped individually rather than failing the batch, `fileCount` preserving the null-vs-number distinction, a missing `schemaVersion` normalizing to 0, and `throw` reserved for a non-object root. Degrading instead of throwing is the correct choice for rolling upgrades — an older box's partial payload shouldn't take down the whole query.

**On `harness.systemPromptTemplate` being full text.** I checked this against the envelope manifest added earlier, which deliberately emits only `{chars, sha256}` and states that prompt content stays private — the two look contradictory at first glance. They aren't: that manifest inspects the *final provider payload*, which carries platform instructions and user content, while this field is the configured Agent prompt/template with platform instructions excluded. Its content is either an operator-authored prompt or a built-in default that lives in this repository, and the reader is the control plane that supplied it. Consistent, and the right split.

## Two notes

**1. `.inherit-builtins.json` parses loosely for a policy file.**

```js
if (fs.existsSync(filePath)) return JSON.parse(fs.readFileSync(filePath, "utf-8")) !== false;
```

Any JSON that isn't literally `false` means "inherit" — including `null`, `0`, and the string `"false"`. Default-on is the right default for compatibility, but a policy file that governs whether image Built-ins reach a preview session is the kind of input where a typo should be loud. Accepting only a boolean and treating anything else as invalid-and-logged would keep a malformed file from silently choosing the permissive branch.

**2. The reasoning for full-text `systemPromptTemplate` lives only in the PR description.** Given the other manifest in the same system reduces prompt content to a hash, the next person reading `systemPromptTemplate: string | null` will reasonably wonder whether it should be hashed too. One line next to the field — configured layer, platform instructions excluded, reader is the supplier — would settle that at the point of doubt instead of in a merged PR body.

Incidentally, the `Revert` / `Revert "Revert"` pair in the history suggests the image-build change went back and forth. The way it ended is good: the uncertainty is now pinned by the new `AgentBox Build Graph` CI job rather than by whoever remembers it.

## Not covered

Stating this rather than implying full coverage: I did not read `skill-overlay.ts` (+67), the `agent-factory.ts` changes (+73/-15), `http-server.ts` (+25), `session.ts` (+12), or the Dockerfile/CI edits beyond noting the new job exists. The three areas I did check are the ones where this series has previously hidden problems; the rest is either a new module or a continuation of patterns already reviewed. Worth a second pair of eyes on `skill-overlay.ts` in particular, since preview overlays decide which Built-ins a session can see.

PR #523, opened 2026-08-26
feat(lark): add topic-scoped group conversations

LikiosSedo:
## Summary
- add a Topic context mode that shares one claimed Feishu Topic across authorized participants
- keep the main group and unrelated Topics silent and isolated
- expose Topic mode in the Feishu mode card and standalone Portal
- document the claim, authorization, and reset contracts

## Validation
- 493 focused Vitest tests passed
- TypeScript build passed
- standalone Portal production build passed
- full Vitest suite: 6225 passed, 2 skipped; one unrelated timing test timed out under parallel load and passed alone

### DIFF
--- a/docs/design/2026-08-26-feishu-topic-closed-loop.md
+++ b/docs/design/2026-08-26-feishu-topic-closed-loop.md
@@ -1,0 +1,59 @@
+# Feishu Topic Closed Loop
+
+## Goal
+
+Let a Feishu group use one Siclaw Agent as a natural, shared conversation inside
+one Topic: the first message mentions the bot, then authorized participants can
+continue in that Topic without mentioning it again. The main group and other
+Topics must remain unaffected.
+
+This is a third context mode. It does not change the existing Team or Personal
+contracts.
+
+## Modes
+
+| Mode | Session key | Follow-up without `@bot` | Reply surface |
+| --- | --- | --- | --- |
+| Team (`shared`) | `chat:<chat_id>` | Main-group chatter may be buffered; Topic follow-ups do not run the Agent | Main group |
+| Personal (`per_user`) | `<participant>:lark_thread:<root_id>` | Only the same participant can reuse their Topic session | Topic |
+| Topic (`topic`) | `lark_thread:<root_id>` | Any authorized participant can reuse a claimed Topic | Topic |
+
+## Claim and isolation rules
+
+1. An explicit `@bot` message is allowed to create the Topic session.
+2. An unmentioned Topic follow-up is resolved with `conversation_existing_only`.
+   It can reuse an existing Topic session, but cannot create one.
+3. The Topic root ID is the complete session scope in Topic mode. Sender IDs are
+   deliberately excluded so authorized participants share one context.
+4. A different Topic root produces a different key. If it has not been claimed,
+   the runtime receives no binding and stays silent.
+5. Main-group messages without `@bot` do not enter Topic sessions.
+6. `/new` resets only the current Topic session. It never resets another Topic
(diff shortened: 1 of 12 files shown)


jacoblee-io (approved):
Approving. Topic mode introduces something this channel didn't have before — a space that *feels* private but is shared by several people — so I spent the review on whether the existing personal-only protections still land on the right side of that line. They do.

## What I verified

**`/apikey` and `/webchat` still drop in a Topic.** Both gates sit at the top of the group-message handler with no condition of their own — the function *is* the group path — and ahead of the @-gate, which the following comment confirms ("Computed here rather than at the @-gate below: /mode needs both"). Every change in this PR lands after that point, so a Topic message reaches those gates exactly as a main-group message does. That matters more than usual here: a Topic reads as a side channel, so it's the surface where someone would most plausibly try `/apikey` expecting privacy, and the drop has to happen anyway.

**One-time links can't reach a Topic.** `deliverSingleUseLink` is confined to the personal branch; the group path resolves denials through `isChannelAccessDenied` and its shareable console URL. This PR doesn't touch that split — it only changes session keying and reply surface *inside* the group path — so the type distinction that keeps a redeemable token out of a room where anyone could take it stays intact.

**The claim boundary is enforced where it has to be.** `conversation_existing_only` for unmentioned follow-ups means an unauthorized sender can reuse but never create, an unclaimed Topic yields no binding and silence, and `modeExistingOnly` extends the same rule to no-mention commands so `/mode` can't establish state in an unrelated Topic either. Deliberately excluding sender IDs from the Topic session key is what makes sharing work, and pairing it with per-turn sender-identity authorization is what keeps that from becoming a bypass — the design doc states both halves.

`normalizeGroupContextMode` keeps the historical fail-closed behaviour with an allow-list, so an unknown or future mode value lands on `per_user` rather than being inferred as shared. That's the right direction for the one decision where guessing wrong puts unrelated group chatter into a shared session.

## Three notes, none blocking

**1. `im:message.group_msg` widens what reaches the process, not just what's answered.** The doc correctly names the permission as a prerequisite. Worth stating the consequence too: with it granted, every group message arrives at Siclaw, including main-group chatter and unrelated Topics that the logic then correctly declines to act on. The handling is right — but "not acted on" and "not observed" are different, and anything that logs message text at debug level now sees the whole room. Worth a look at the logging on that path before this is enabled on a busy group.

**2. Two adjacent decisions use opposite defaults.** `normalizeGroupContextMode` is an allow-list (unknown → `per_user`), while the behaviour switch is a negation:

```js
const topicScopedMode = isGroupMessage && contextMode !== "shared";
```

Today those agree, because normalization only ever emits the three known values. But they'd diverge the moment a fourth mode is added to the allow-list: it would inherit Topic keying and Topic reply surface implicitly, without anyone deciding that. `contextMode === "per_user" || contextMode === "topic"` would make the next mode a compile-or-review decision instead of a default.

**3. Shared history is the part users will misjudge.** Team mode's shared context is self-evident — the replies are in the main group. A Topic looks like a side room, so a participant may reasonably assume it's theirs, while any authorized participant who joins later can read the whole exchange with the Agent. Per-turn authorization is the correct control and it's in place; the gap is expectation, not enforcement. The `/mode` switch confirmation is a good place for it — the zh string already says 已授权成员在该话题内无需再次 @, and adding that they also see the existing context would set the expectation at the moment someone opts in.

## Also good

The mode-card and Portal selector moved with the runtime, the reset contract is scoped to the current Topic only (`/new` never reaching another Topic or the room), and the design doc separates what standalone supports from what remains a platform capability rather than blurring them.

PR #524, opened 2026-08-27
feat(tools): persist invocation toolset labels

caihongwei2006:
## Summary
- stamp registry tools with their configured category, restricted file tools as `filesystem`, and dynamic MCP tools as `mcp:<server>`
- preserve invocation-scoped toolset labels across parallel/out-of-order start/end events and abort finalization
- carry the nullable label through chat, delegation, AgentBox, internal API, SSE, and Lark persistence paths
- leave unknown tools unlabelled instead of inferring a value

## Verification
- 14 focused test files, 689 tests passed
- `tsc --noEmit`
- `git diff --check`

## Rollout
A companion downstream integration change should be deployed first or together; older integrations safely ignore the nullable field.


LikiosSedo (changes requested):
Requesting changes for three gaps that are not covered by the green focused tests:

1. **[P1] Standalone Portal drops every persisted `toolset`.** `src/gateway/chat-repo.ts` now forwards the field, but Siclaw's bundled Portal has no matching column in `src/portal/migrate.ts`, and its `chat.appendMessage`, `chat.updateMessage`, and `chat.getMessages` implementations in `src/portal/adapter.ts` neither write nor return it. In the supported Gateway + LocalSpawner shape, every label is therefore silently lost and reads back as `null`. Please add the nullable schema/upgrade migration, propagate the field through both standalone persistence paths, and cover MySQL/SQLite plus adapter persistence.

2. **[P1] The public PR description exposes a private downstream repository/MR reference.** Please remove the internal project name, domain/path, and MR identifier rather than copying them into another public comment. The rollout note can be stated neutrally, for example: "A companion downstream integration change should be deployed first or together; older integrations safely ignore the nullable field."

3. **[P2] Synthetic background-completion turns still persist known tools with `toolset = null`.** `PiAgentBrain.enrichToolEvent()` enriches only lifecycle start/end events, while `AgentBoxSessionManager.runSyntheticPrompt()` buffers `message_end.message` objects and `persistSyntheticMessage()` persists those tool results without correlating the lifecycle label. Please associate them by `toolCallId` (or enrich the tool-result message) and add a focused `notify-parent` regression test.

The core start/end propagation, out-of-order invocation pairing, abort finalization, unknown-tool behavior, and downstream nullable rolling contract otherwise look sound. On exact head `f5bd93a8`, the 7 focused files pass 423/423 tests, `npx tsc --noEmit` passes with lockfile dependencies, and `git diff --check` is clean.

LikiosSedo (changes requested):
Independently confirmed the existing change request on `f5bd93a8`. CI is green and the start/end stamping path looks sound; do not merge until the persistence holes are closed.

## Must fix

1. **Standalone Portal drops every persisted `toolset`.** `chat-repo` now forwards the field, but Siclaw's bundled Portal has no `chat_messages.toolset` column (`src/portal/migrate.ts`) and `chat.appendMessage` / `chat.updateMessage` / `chat.getMessages` in `src/portal/adapter.ts` neither write nor return it. Extra RPC params are ignored, so Gateway + LocalSpawner / `siclaw local` silently reads `null`. Add the nullable `safeAlterTable` (same pattern as `trace_id` / `seq`), thread the field through both standalone persistence paths, and cover a write/read round-trip. Payload-only `chat-repo` tests are not enough.

2. **The public PR description exposes a private downstream repository/MR reference.** Remove the internal project name, domain/path, and MR identifier. Neutral rollout note is enough: a companion downstream integration change should be deployed first or together; older integrations safely ignore the nullable field.

3. **Synthetic background-completion turns still persist known tools with `toolset = null`.** `PiAgentBrain.enrichToolEvent()` stamps lifecycle start/end only. `runSyntheticPrompt()` buffers `message_end.message` objects and `persistSyntheticMessage()` writes `toolName` without correlating the lifecycle label. Associate them by `toolCallId` and add a focused notify-parent / synthetic-turn regression test.

## Checked, not bugs

- `extraListeners` / `emit()` only fan out `auto_retry_*`; they do not bypass enrich for tool events.
- `tool_end` is an alias of `tool_execution_end`, not a second persist.
- `knowledge_cite` is a registry `query` tool, so the primary path already gets a label.

Registry / filesystem / `mcp:<server>` stamping, invocation-id pairing for out-of-order start/end, abort finalization, and unknown-tool non-inference can stay as-is.

caihongwei2006:
Thanks for the careful review and independent confirmation. I agree with all three findings.

- **Standalone Portal persistence:** the RPC payload coverage is insufficient if the bundled Portal schema and adapter paths drop the field. I’ll add the nullable migration/upgrade path, propagate `toolset` through append, update, and read operations for the supported standalone stores, and add persistence round-trip coverage.
- **Public rollout wording:** fixed. The description now uses a neutral downstream-integration note and contains no private repository, domain/path, or MR reference.
- **Synthetic background completion:** agreed. Lifecycle enrichment alone does not cover buffered `message_end.message` tool results. I’ll correlate the invocation label by `toolCallId` through synthetic persistence and add a focused notify-parent regression test.

I’ll keep unknown tools nullable and preserve the rolling compatibility contract. I’ll push the implementation and test updates before requesting another review.

caihongwei2006:
Addressed the three requested gaps in b64d0dd2.

- Added nullable `chat_messages.toolset` to fresh and upgrade migrations, with historical rows left `NULL`.
- Propagated `toolset` through the standalone HTTP and RPC append paths, RPC update, and reads.
- Correlated synthetic lifecycle labels to tool-result messages by `toolCallId` before persistence.
- Added MySQL DDL coverage, SQLite upgrade coverage, a real standalone adapter round-trip, HTTP/RPC path assertions, and a notify-parent regression.
- Updated the public rollout wording to remain downstream-neutral.

Verification: 14 focused files / 689 tests passed, `npx tsc --noEmit` passed, and `git diff --check` is clean. Please take another look when convenient.

LikiosSedo (approved):
Re-reviewed on b64d0dd2. The standalone Portal migration and append/update/read paths now preserve toolset, synthetic background completions correlate labels by toolCallId, and the public rollout wording is clean. Independently verified 14 focused test files (689/689), TypeScript typecheck, diff check, mergeability, and green CI. No remaining blocking findings.

PR #526, opened 2026-08-27
fix(lark): allow no-mention turns in claimed two-party Topics

LikiosSedo:
## Summary
- allow an unmentioned human follow-up only in Topic mode and only inside an already-claimed Feishu Topic
- verify the Topic participant set from its root plus every paginated reply; allow only exactly one human and this Siclaw app, regardless of the parent group roster
- require an explicit bot mention when another human or app participates, sender identity is unknown, history is malformed or unavailable, pagination is invalid, or the Topic exceeds 100 pages of 50 replies
- keep Personal and Team Agent turns mention-gated, apply the same Topic gate to `/mode`, and treat `@all` as an unmentioned human follow-up inside an otherwise eligible Topic
- preserve rich-post Markdown and code-block content while parsing Feishu messages

## Validation
- `npx vitest run src/gateway/channels/lark.test.ts` (238 passed)
- `npm run build`
- `git diff --check`
- GitHub CI

### DIFF
--- a/docs/design/2026-07-11-lark-group-context-mode.md
+++ b/docs/design/2026-07-11-lark-group-context-mode.md
@@ -71,6 +71,18 @@ no user-facing choice, and non-@ group messages are dropped unconditionally
    messages never arrive; `shared` degrades to "shared @-turns only" (multiple
    senders share one session but the agent doesn't hear the chatter). This is
    functional, not an error.
+8. **No-mention activation is a two-party Topic exception**: the parent group
+   roster does not define Topic participation. Only `topic` mode may invoke the
+   agent without `@bot`, and only inside an already-claimed Topic whose root
+   message plus every paginated reply contain exactly one human sender and this
+   Siclaw app. The current event sender is added explicitly to cover Feishu
+   history-list eventual consistency. Two humans or another app in that Topic
+   require an explicit mention on every Agent turn, including turns from the
+   original asker. Personal and Team modes never run the Agent from an
+   unmentioned message. Missing/malformed root/history data, incomplete or
+   cyclic pagination, unknown sender identities, and lookup failures all fail
+   closed to "mention required". The positive result is revalidated before
+   every unmentioned turn so a new participant takes effect immediately.
 
 ## Feishu permission matrix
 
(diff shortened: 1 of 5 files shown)


jacoblee-io (approved):
Approving. This tightens the Topic contract in the direction I'd raised on the previous PR — and further than I suggested there.

My note then was that a Topic *looks* like a private side room while any authorized participant who joins can read the whole exchange, and I proposed setting expectations in the `/mode` confirmation copy. Restricting no-mention continuation to a genuinely two-party Topic solves the underlying thing instead: the room now behaves the way it looks, and the moment a second human or another app participates, every turn goes back to requiring a mention. That's a better answer than better wording.

## What I verified

**Every unverifiable path requires the mention.** Seven distinct failure modes — sender identity unavailable, unknown `sender_type`, non-zero Feishu `code`, root message unavailable, `data.items` not an array, invalid pagination state, exceeding the page ceiling — all throw into one handler that returns `false` and logs why. For a gate whose job is deciding when authorization can be skipped, "couldn't tell" landing on "require @bot" is the only correct default, and there's no path that reaches the permissive branch without a positive answer.

**The participant set is the Topic's, not the group's.** Seeding `humanIds` with the current sender and verifying the root separately closes the case the comment names — replies alone would read as a one-human Topic. `appIds.every((id) => id === siclawAppId)` means any other app disqualifies it, not just other humans.

**The positive result is deliberately uncached, and that reasoning is right.** Caching it would let a Topic keep skipping mentions after a second participant arrives, which is exactly the window this gate exists to close.

**Mention detection can't be forged by the new parsing.** Since this PR also preserves rich-post Markdown and code-block content, I checked whether a mention could be faked through text: `isBotMentioned` reads Feishu's structured `mentions` array and matches `id.open_id`, so `@bot` inside a code block produces no entry. `@all` is excluded via `key !== "@_all"`, which is what makes treating it as an unmentioned follow-up consistent rather than a special case.

## Two notes, neither blocking

**1. Negative results are cacheable, and that would cut the per-message cost a lot.** Right now an unmentioned follow-up costs one root fetch plus up to 100 paginated reply fetches, on every message, with no caching. A Topic sitting at 500 replies pays ten API calls per turn; near the ceiling it pays a hundred.

The asymmetry worth exploiting: a Topic's participant set only grows — someone who has posted cannot un-post. So once the gate answers "more than two parties", that answer can never flip back, and caching it is caching the *stricter* conclusion. That keeps the safety property you deliberately preserved for positive results while removing the repeated pagination for the case that will dominate in an active group.

**2. Crossing the ceiling silently changes behaviour for good.** A Topic past 100 pages of 50 replies permanently returns to requiring a mention, since the throw fires every time. That's a defensible DoS bound, but from a user's seat it reads as "this Topic used to work without @ and now doesn't". Worth checking that `channels.mdx` states the threshold in user terms rather than only the mechanism — the PR description has the condition, the docs are where someone will look for it.

**3. Minor, pre-existing:** `isBotMentioned`'s fallback when `botOpenId` is absent treats *any* non-`@all` mention as mentioning the bot. That's fail-open, and this PR gives `botMentioned` more to decide than before (it now selects between the mention gate and the two-party path). Low impact — a missing bot open_id means the integration is misconfigured anyway, and authorization still happens at binding resolution — but it's the one input to this gate that doesn't fail closed.

## Shape

176 lines of implementation against 427 of tests, on a change whose whole value is the boundary being exactly right. Three docs updated alongside. That ratio is the reason this one was quick to review.

PR #527, opened 2026-08-28
build: remove legacy KBC image alias

LikiosSedo:
## Summary

- stop republishing the KBC image under the retired `kbc-compile-box` repository name
- remove the legacy image from `make info`
- record that the compatibility window has ended

## Verification

- `make -n push-kbc REGISTRY=registry.example/team TAG=test-tag`
- `make -n push REGISTRY=registry.example/team TAG=test-tag`
- `make info REGISTRY=registry.example/team TAG=test-tag`
- `git diff --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ Renamed the KB compile box so operators can tell it apart from chat agentboxes a
 - **Image rename**: `kbc-compile-box` → `siclaw-kbc-box` (Makefile build/push targets, helm `siclaw.compileBoxImage` derivation, Dockerfile/README/docs). `agentbox.compileBoxImage` explicit-override semantics are unchanged.
 - **Pod-name prefix**: compile boxes now spawn as `kbc-box-<id>` instead of `agentbox-<id>`, declared on the BoxProfile (`podNamePrefix`) for the `kb-compile` / `kb-compile-codex` profiles. Derived resources (cert Secret, hostname) follow the prefix. Chat agentboxes and the read-only `kb-test` box keep the `agentbox-` prefix.
 - **Upgrade behavior**: on the first compile spawn after upgrade, the spawner reaps any pod left under the old `agentbox-<id>` name for that agent (guarded to compile boxes only — a chat box under the same name is never touched), so the old and new pods do not coexist.
-- **Transition double-push**: `make push-kbc` republishes the same image digest under the legacy name `kbc-compile-box` for one release cycle so runtimes still pinned to the old name keep resolving. Remove the legacy tag/push next release.
+- **Legacy alias removed**: the one-release compatibility window has ended, so `make push-kbc` now publishes only `siclaw-kbc-box`. Runtimes and deployment overrides must no longer reference `kbc-compile-box`.
 - **DevOps action required**: the internal registry replication rule must gain an entry for `siclaw-kbc-box` (the old `kbc-compile-box` was not in the rule and had to be pushed manually). Until the rule is added, push `siclaw-kbc-box` to the production registry manually.
 
 ### Added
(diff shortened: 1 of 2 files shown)


jacoblee-io (commented):
The Makefile change is clean and I verified there's nothing left pointing at the old name: `git grep kbc-compile-box` over `main` returns only the Makefile and CHANGELOG lines this PR touches. helm's `_helpers.tpl` derives `siclaw-kbc-box` in both branches, and the docs and CLAUDE.md reference the new name only. So removing the variable, the double-push, and the `make info` line leaves no dangling reference.

One wording change I'd like before approving, and one thing worth confirming.

## The CHANGELOG entry describes a window that never opened

Both bullets live under `## [Unreleased]` — the rename and the double-push have not shipped. So rewriting that bullet is right (nothing published yet to contradict), but the replacement text describes history that didn't happen:

> **Legacy alias removed**: the one-release compatibility window has ended, so `make push-kbc` now publishes only `siclaw-kbc-box`.

Since the rename and the alias removal will go out in the *same* release, no release ever double-pushed. An operator reading "the compatibility window has ended" reasonably concludes one was provided and has expired — and therefore that anything still pinned to `kbc-compile-box` had a release to migrate in. It didn't. A deployment upgrading from any version before this one, with `kbc-compile-box` pinned in an override, gets no transition at all.

That's the opposite of what the sentence sets up, and the consequence is a skipped check: nobody goes looking for pinned overrides because the text implies that migration already had its window.

Stating it as what it is would set the right expectation — the rename lands in this release with no transition alias, so any runtime or deployment override pinning `kbc-compile-box` must be updated as part of the same upgrade. If a real window is wanted instead, the double-push stays and this PR waits a release; that's a product call, not a wording one, but the two options shouldn't be describable by the same paragraph.

## Worth confirming: the replication rule

The bullet directly below the one you changed is untouched and still says:

> the internal registry replication rule must gain an entry for `siclaw-kbc-box` (the old `kbc-compile-box` was not in the rule and had to be pushed manually). **Until the rule is added, push `siclaw-kbc-box` to the production registry manually.**

If that's still true, then after this PR the production registry has no automatic path to any KBC image: the old name is no longer pushed, and the new name isn't replicated. The manual push becomes the only route, and it's now the only route — previously a runtime pinned to the old name at least resolved to something.

Not introduced here, but this PR removes the fallback that made it survivable. Worth checking the rule actually has `siclaw-kbc-box` before this merges; if it does, that bullet is stale and could go in the same commit.

Fix the first point and I'll approve — the build change itself is correct and verified.

jacoblee-io (approved):
Approving — and correcting the premise my last comment rested on.

I argued that since both bullets sit under `## [Unreleased]`, the double-push had never shipped and therefore no compatibility window existed. That inference was wrong: `v0.2.8` through `v0.3.10` published under both names, so the window was real and lasted a good while. I treated a CHANGELOG section heading as authoritative evidence about release state, when the section is a cumulative account of the rename spanning several releases. The release history was something to ask about, not derive.

The replacement text is better than what I asked for. I wanted the transition state stated accurately; it now gives the version range, the exact point the behaviour changes, and — most usefully — the two override keys someone actually has to grep for:

> update any explicit `agentbox.compileBoxImage` or `SICLAW_COMPILE_BOX_IMAGE` override before upgrading

That's the sentence an operator can act on. "Any runtime or deployment override", which is roughly what I suggested, would have left them to work out where those live.

Standing by the rest of the earlier check: `git grep kbc-compile-box` over `main` finds nothing outside the two files this PR touches, helm's `_helpers.tpl` derives `siclaw-kbc-box` on both branches, and the Makefile edit removes the variable, the double-push, and the `info` line without leaving a dangling reference.

The **DevOps action required** bullet is unchanged, so I'll assume the replication rule still lacks `siclaw-kbc-box` and the manual push remains necessary. That's pre-existing and documented rather than something this PR should carry — noting it only because after this merges, the manual push is the *only* route to the production registry, where before a runtime pinned to the old name would still have resolved to something. Worth having that rule entry land before the next release actually goes out.

PR #529, opened 2026-08-28
feat(delegation): stamp delegated legs with their own trace id and link them from the coordinator's tool row

glow-Kev1n:
## Summary

Delegated legs were invisible to every trace-keyed read: on the local route the peer's rows were persisted with `trace_id NULL`, and on both routes nothing recorded which trace a delegation produced. This PR stamps every delegated leg with the peer turn's own root trace id (a delegated turn is its own trace — the peer runs under its own authority) and records the cross-trace link as `child_trace_id` on the coordinator's `delegate_to_agent` tool row, next to the existing `child_session_id`.

### Problem

`delegate-api`'s local route ran `consumeAgentSse` without a trace id, so an entire delegated investigation (often the bulk of a coordinator turn) persisted as `trace_id NULL` — while the remote route already stamped the peer's own id. Either way there was no machine-readable edge from the coordinator's trace to the leg, so audit and analysis reads keyed by trace id stopped dead at the `delegate_to_agent` boundary.

### Solution

- **Local route parity**: consume stamps every persisted peer row with the trace id from the box's prompt ack; the opening user row (appended before the trace exists) is back-bound afterwards — the same append-then-bind pattern `chat.send` uses. The two routes now leave identical rows.
- **Terminal carries the id**: the target Runtime rides the leg's trace id on the delegation terminal event — the one channel the source can learn a REMOTE leg's trace from. Both terminal producers (consumer paths and the shutdown/box-roll supervisor) read it from the delegated turn's ledger entry, recorded at prompt ack, so they cannot disagree.
- **The tool row records the link**: `delegate_to_agent` result details gain `child_trace_id` on every outcome (done / failed / input_required / stopped). An early `delegate_trace` frame announces the id at prompt ack, so a stopped leg's tool row keeps the link even though the final result frame dies with the destroyed socket.
- **One ingestion gate**: every trace id crossing a process boundary goes through `validTraceId` (32 lowercase hex, the same contract the bind RPC enforces server-side). Absent or malformed ids degrade to "no link" — an older peer Runtime simply produces the pre-feature behaviour.
- **Late-terminal salvage**: a terminal that outlives its subscriber (Stop, relay idle-timeout, source restart) still binds the opening row, memoized per delegation id.
- Contract documented in `docs/design/coordinator-routing.md`.

### Compatibility

No breaking changes: the new SSE frame is ignored by older consumers (if/else-if dispatch), `DelegateResponse.peerTraceId` is an optional addition, no new RPC methods (`chat.bindMessageTraceId` pre-exists), and persistence changes are additive (leg rows gain a real `trace_id`; tool-row metadata gains one key). Rolling back simply stops writing them.

**Known gap** (also noted in code): on the REMOTE route, a coordinator Stop or relay idle-timeout tears the subscription down before the terminal arrives, so those tool rows carry no `child_trace_id`. The opening-row bind is still salvaged via the retried terminal, so the leg stays reachable through `child_session_id`.

## Test Plan

- [x] 12 new tests: local-route stamp/bind/frames, tracing-off degradation, remote terminal extraction, malformed-id gate, late-terminal salvage + memo, stop-path link retention, tool-row link fields on every outcome
- [x] Full vitest suite and `tsc --noEmit` pass

### DIFF
--- a/docs/design/coordinator-routing.md
+++ b/docs/design/coordinator-routing.md
@@ -273,6 +273,43 @@ And a re-delivered terminal must be acknowledged, not rejected: its consumer is
 gone precisely because it consumed the original, and refusing it keeps a sender
 retrying — which keeps supervision alive over a turn that already ended.
 
+### The delegated leg is its own trace, linked — not folded — into the coordinator's
+
+A delegated turn runs under the peer's own authority (its persona, model,
+permissions), so its rows are persisted under the peer turn's OWN root trace id —
+the same rule every other cross-agent boundary follows — never under the
+coordinator's. What connects the two traces is an explicit link, written at three
+places:
+
+- **Row stamping.** The local route's consume stamps every persisted peer row with
+  the trace id from the box's prompt ack, exactly as the target Runtime's ordinary
+  `chat.send` path does for a remote peer — the two routes leave identical rows.
+  The opening user row is appended by the source before the trace exists and is
+  back-bound afterwards (`chat.bindMessageTraceId`, append-then-bind, the same
+  pattern `chat.send` uses for its prompt row).
+- **The terminal carries the id.** `chat.send` acks in milliseconds — before the
+  trace exists — and `chat.getMessages` does not project `trace_id`, so the
+  terminal event is the one channel the source can learn which trace a REMOTE
+  leg's rows were persisted under. Both terminal producers (the consumer paths and
+  the shutdown/box-roll supervisor) read the id from the delegated turn's ledger
+  entry, recorded at prompt ack, so they cannot disagree. A terminal that arrives
+  after its delegation settled (coordinator Stop, idle-timeout, source restart)
+  still salvages the opening-row bind — best-effort, memoized per delegation id.
+- **The tool row records the link.** `delegate_to_agent` returns
+  `child_trace_id` next to `child_session_id` in its result details, which persist
+  into the coordinator tool row's metadata: that is the machine-readable edge an
+  audit or analysis read walks to get from the coordinator's trace into the leg.
+  An early `delegate_trace` frame announces the id the moment the local ack names
+  it, so a stopped leg's tool row keeps the link even though the final
+  `delegate_result` frame dies with the destroyed socket.
+
+Trace ids cross process boundaries only through one gate (`validTraceId`,
+32 lowercase hex — the same contract the bind RPC enforces server-side): accepting
+a malformed id at one boundary and rejecting it at another is how rows end up
+persisted under an id no link references. Absent or malformed ids degrade to "no
+link", never to an error — an older peer Runtime simply produces the pre-feature
+behaviour.
+
 After the terminal event, the source always derives remote `finalText` from
 assistant rows after the current delegation boundary in durable session history.
 Recovery widens a window over the newest rows rather than walking a timestamp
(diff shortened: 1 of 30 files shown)


jacoblee-io (approved):
Approving. Scope: 30 files, but the substance is seven — I focused on the trust boundary for an id arriving from another process, the lifetime of the structures that carry it, and whether the two terminal producers can disagree. I did not read the docs change or the append-then-bind implementation line by line.

## What I verified

**One gate, and it fails closed.** `validTraceId` is anchored (`^[0-9a-f]{32}$`, lowercase) and returns `undefined` rather than throwing, so a malformed or absent id degrades to "no link" instead of persisting a bogus one. It's applied at each place an id crosses in — the delegation event, the relay envelope, and the prompt ack. For a value that becomes a persisted foreign key into audit reads, the anchored-whitelist-plus-degrade shape is the right one; a substring match or a truthiness check would both have let a peer write something unqueryable.

**The two producers read the same source, and the comment says why.**

> It lives HERE — on the turn's ledger entry — rather than in the `chat.send` closure, so both terminal producers (the consumer paths in the handler and the shutdown/box-roll supervisor above) report the same trace

That's the failure this repo has hit before in delegation — two paths reporting the same event with different content. Putting the id where both already read from makes divergence structurally impossible instead of relying on both being updated together.

**The structures are bounded.** `salvagedDelegations` has an explicit cap with oldest-first eviction, and the comment is honest about what the memo is for ("keeps a redelivery storm from re-walking the history") — since it's a performance guard rather than a correctness dependency, evicting an old entry costs at most one repeated history walk. `unsupportedUpstreamMethodsReported` is naturally bounded by the method namespace, and de-duping per method rather than globally is the detail worth having: "an absent method must not silence the next one."

`delegatedTurns` gaining `traceId?` doesn't change its lifetime — the existing `delete(turnId)` still runs, and the read precedes it. I checked because a per-turn map on a long-lived Runtime is where a leak would hide; this isn't one.

**The degradation layering is sound.** Opening row appended before the trace exists, back-bound after, `warnTraceBindFailure` on failure, and the late-terminal salvage as a third chance. Each layer's failure lands on "row keeps `trace_id NULL`" rather than on a wrong id, which is the correct direction for a link that audit reads follow.

## Two notes

**1. The known gap is well handled — say where it lives in code.** Declaring in the description that a REMOTE-route Stop or idle-timeout leaves the tool row without `child_trace_id`, and that the leg stays reachable via `child_session_id`, is exactly the disclosure a reviewer needs. The description says it's also noted in code; worth making sure that comment names the same fallback, since the next person to see a `child_trace_id`-less row will be reading the code, not this PR.

**2. Fourteen of the thirty files are `+2/-0` test boilerplate.** Every file that mocks `chat-repo.js` had to gain the same `validTraceId` line. That's mechanical and correct here, but it means adding one export to `chat-repo` costs a touch in fourteen files, and the next one costs fifteen — while also inflating a seven-file change into a thirty-file diff, which is its own review tax. A shared mock factory for `chat-repo` would make that one edit. Not for this PR.

## Worth flagging for whoever merges

The change-impact rules for `delegate-api.ts` and delegated persistence in `server.ts` call for deploying the management plane before the Runtime. This PR is additive on the wire in both directions — the new SSE frame lands in an if/else-if dispatch that older consumers skip, `peerTraceId` is optional, and an older peer simply produces the pre-feature behaviour — so I don't think the ordering bites here. Stating it because that row exists and a reader shouldn't have to re-derive whether it applies.

CI green across Test / Type Check / Portal Web Test / AgentBox Build Graph, and the 12 new tests cover the paths I'd have asked for: local-route stamping, tracing-off degradation, remote terminal extraction, the malformed-id gate, late-terminal salvage plus its memo, and the tool-row link on every outcome.

PR #530, opened 2026-08-28
Honor authoritative empty Skill bundles

LikiosSedo:
## Summary
- Treat `skillsAuthoritative: true` as a successful full Skill resolve, including `skills: []`.
- Clear `resolved/` when that marker arrives so an archived last Skill actually stops executing without a pod restart.
- Keep unmarked empty payloads on the live K8s and Local reload constructors; a failed fetch never reaches materialize.

## Test plan
- [x] `npx vitest run src/agentbox/sync-handlers.test.ts src/agentbox/http-server.test.ts src/gateway/agentbox/local-spawner.test.ts src/agentbox/resource-sync.test.ts` (198 passed)
- [ ] Confirm the control plane emits `skillsAuthoritative` on a successful `config.getSkillBundle`, including an intentional empty set
- [ ] Withdraw the last Skill, reload, and check sync status no longer lists it

### DIFF
--- a/src/agentbox/http-server.test.ts
+++ b/src/agentbox/http-server.test.ts
@@ -1951,3 +1951,11 @@ describe("resolveDelegation (worker autonomy — readOnly is explicit opt-in)", 
     expect(resolveDelegation({ delegationId: "d1", readOnly: true }, "api")?.readOnly).toBe(true);
   });
 });
+
+describe("http-server — Skill handler wiring", () => {
+  it("does not disable empty-bundle preservation on the K8s/hot-reload path", () => {
+    const src = fs.readFileSync(path.resolve(__dirname, "http-server.ts"), "utf8");
+    expect(src).toContain("createSkillsHandler");
+    expect(src).not.toContain("preserveExistingOnEmpty");
+  });
+});
--- a/src/agentbox/http-server.ts
+++ b/src/agentbox/http-server.ts
@@ -661,7 +661,6 @@ export function createHttpServer(
   } else if (sessionManager.skillsDir) {
     perServerHandlers.skills = createSkillsHandler({
       skillsDir: sessionManager.skillsDir,
-      preserveExistingOnEmpty: false,
       boxClient: sessionManager.gatewayClient ? sessionManager.gatewayClient.toClientLike() : null,
     });
   }
--- a/src/agentbox/sync-handlers.test.ts
+++ b/src/agentbox/sync-handlers.test.ts
@@ -21,6 +21,7 @@ import { knowledgeRepoDirName } from "../shared/knowledge-package.js";
 import type { GatewaySyncClientLike } from "../shared/gateway-sync.js";
 import { CredentialBroker } from "./credential-broker.js";
 import { resolveSkillDirectories } from "../core/skill-directories.js";
+import { syncResource } from "./resource-sync.js";
 import type {
   CredentialTransport,
   ClusterMeta,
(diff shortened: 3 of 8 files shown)


jacoblee-io (approved):
Approving. This is the right shape for a problem this repo has argued about twice, and it retires the argument rather than winning it again.

## Why the explicit marker is the fix

The knowledge-side version of this landed as "a successful empty fetch is authoritative", which then required proving that a failed fetch can never reach `materialize` — a claim that held on the HTTP reload path and not on the retry path, and which had to be re-litigated per call site. Skills couldn't take that route at all, since an empty Skill bundle is a legitimate steady state (core Skills ship in the image, the bundle carries only global + personal).

`skillsAuthoritative` sidesteps both. The predicate moved from "is the list empty" to "did the control plane say it resolved" — and a failed fetch doesn't set the flag, so the wire distinguishes the two cases that the emptiness test could not. I checked where it's set: both `config.getSkillBundle` success returns in the adapter, the empty one and the populated one, with nothing on the failure paths. That's the correct placement.

## Verified

- **Clearing is scoped to `resolvedDir`.** The guard and the wipe both key off `resolvedDir`, so image-provided `core/` is untouched — which is the invariant that made the knowledge approach unsafe to copy here verbatim. Post-isolation that directory is per-agent, so Local mode clears one Agent's tree rather than a shared one.
- **The unmarked path keeps the old behaviour.** An older control plane emits no flag, so its empty payloads preserve the last copy exactly as before. Degradation direction is "Skill stays" rather than "Skill vanishes", which is the safe half for a mistake to land on.
- **`hasExplicitBuiltinPolicy` / `hasExplicitBuiltinMask` counting as authoritative** is consistent with the preview-overlay work: those markers are also the control plane stating intent explicitly, so treating them the same way avoids a second notion of authority.
- **Removing `preserveExistingOnEmpty` tightens LocalSpawner.** That option previously made Local mode clear on *any* empty payload; it now waits for the marker like everything else. Worth noting as a deliberate narrowing of earlier behaviour rather than a refactor — it's a behaviour change in the safe direction, and the right one given empty-is-normal for Skills.

## Two notes

**1. The "failed fetches never reach materialize" sentence is a leftover from the old argument.** It appears in both the description and the new doc comment:

> Failed fetches never reach materialize, so they keep the last copy.

The design no longer rests on that. What keeps the last copy is the *absence of the marker* — which is true whether or not a failed fetch reaches materialize. And the claim itself is only true on one of the two call paths (the HTTP reload route returns before materialize; the retry loop in `resource-sync` retries on a throw but hands a successful-but-empty payload straight through). Dropping the sentence and saying "a failed fetch does not carry the marker, so an unmarked empty payload keeps the last copy" makes the guarantee accurate and independent of that path analysis.

**2. The unticked first test-plan item is the deployment matrix.** "Confirm the control plane emits `skillsAuthoritative` on a successful `config.getSkillBundle`" is the whole feature's precondition, and this PR only changes one producer of that contract — the standalone adapter. On a deployment whose control plane is the other implementation, empty payloads stay unmarked, so archiving the last Skill still won't take effect until that side emits the flag too.

That's the safe failure (nothing gets wrongly wiped), but it means "withdraw the last Skill and it stops executing" is fixed for standalone and pending elsewhere. Worth stating in the description alongside the second unticked item, since someone verifying the fix on the wrong deployment would conclude it doesn't work.

CI green across Test / Type Check / Portal Web Test / AgentBox Build Graph; 198 tests in the focused run cover both the marked-empty clear and the unmarked-empty preserve.

PR #531, opened 2026-08-28
feat(runtime): add chat dispatch observability

LikiosSedo:
## Summary

- log the AgentBox selected for each `chat.send` using `agentId`, `sessionId`, and `turnId`
- log `/api/prompt` acceptance or rejection status, error code, retriable flag, duration, and a bounded single-line error summary without prompt text or configuration
- preserve structured AgentBox HTTP status, code, and retriable fields in asynchronous `stream_error` events
- add a post-rollout acceptance gate: at least five consecutive first turns in fresh sessions must succeed with correlated Runtime and AgentBox evidence

This change has a paired Sicore control-plane MR.

## Verification

- `npx tsc --noEmit`
- 34 targeted Vitest tests passed across dispatch summary, Runtime chat send/abort/system-prompt paths, and AgentBox `/api/prompt` acceptance

No production configuration or deployment is included.

### DIFF
--- a/docs/install/kubernetes.mdx
+++ b/docs/install/kubernetes.mdx
@@ -159,3 +159,22 @@ runtime:
   re-adopt live capability runs. Cluster-wide box deletion remains an explicit
   cleanup operation, not a process-shutdown side effect.
 - If you deploy monitoring resources from the chart, do not also apply duplicate monitor manifests manually.
+
+## Post-rollout Chat Acceptance
+
+After a production Runtime or AgentBox rollout reports Ready, verify cold-session
+dispatch before declaring the release complete:
+
+1. Create at least five new chat sessions. Do not reuse an existing session or
+   retry a failed first message.
+2. Send one first-turn prompt in each session, sequentially. All five must reach
+   assistant output and `prompt_done` without a `stream_error` or manual retry.
+3. For every turn, correlate `sessionId` and `turnId` across the control-plane,
+   Runtime, and AgentBox logs. Confirm the selected Runtime and AgentBox, and a
+   `/api/prompt` result with `status=200` and `outcome=accepted`.
+4. Treat any first-turn failure, missing terminal event, or unexplained
+   Runtime/AgentBox selection as a failed release acceptance. Preserve the
+   correlated log lines before retrying or rolling again.
+
+This check validates observed post-rollout behavior. Pod readiness and a
+successful image rollout alone do not satisfy chat acceptance.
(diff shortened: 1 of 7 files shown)


jacoblee-io (approved):
Approving. Logging is a data egress, so I reviewed it the same way as the envelope manifest work: what can reach the sink, and is the boundary structural or by convention. I did not review the acceptance-gate text in `kubernetes.mdx`.

## What the sanitizer gets right

```js
const message = value instanceof Error ? value.message : String(value ?? "Unknown error");
const compact = message.replace(/\s+/g, " ").trim();
```

Three separate things, each load-bearing:

- **`.message`, not the error.** No stack, so file paths and internal structure stay out of the log.
- **Whitespace collapse.** Beyond formatting, this removes the ability to forge a log record — an error message containing `\n[agentbox-http] …` can't become its own line. Worth a word in the comment, since a future edit that switches to trimming only the ends would silently drop that property.
- **512 with an explicit `…`.** Truncation is visible rather than making a clipped message look complete.

On the AgentBox side, `JSON.stringify(compactDispatchLogMessage(message))` adds quoting on top, so the value stays one unambiguous field even if it contains delimiters.

The dependency split does what the second commit says: `src/shared/` holds the zero-dependency compaction and AgentBox imports only that, while `wrapRpcError` stays behind the Gateway-side wrapper. AgentBox reaching into Gateway code would have been an image-boundary violation, and it doesn't.

## The claim is narrower than it looks, and that's mostly fine

"without prompt text or configuration" is a property of the *inputs*, not something the function enforces — it only shortens and single-lines whatever it receives. So I traced what actually arrives. Of the six call sites, five pass internal literals or `undefined`:

```js
logPromptResponse(400, "rejected", "Missing 'text' field");
logPromptResponse(409, "busy", "Session is already running", managed.id);
logPromptResponse(200, "accepted", undefined, managed.id);
```

Only `setup_failed` forwards an external error:

```js
logPromptResponse(500, "setup_failed", err, managed.id);
```

That's a genuinely small surface, and setup failure happens before the prompt reaches a provider, so prompt text isn't a realistic passenger. Configuration is the one that gives me pause: setup failures are typically model-binding or provider-config problems, and a provider error message can carry the endpoint URL — which is operator-supplied and, in the shape this codebase already handles elsewhere, can carry credentials in the query string.

Not a finding I can demonstrate — it depends on what a given provider SDK puts in `.message` — but it's the one input where the description's guarantee rests on an upstream that isn't under this file's control. Either strip known-sensitive patterns at that call site, or state in the comment that content safety depends on the upstream message, so nobody later reads "without prompt text or configuration" as something the sanitizer enforces. The distinction matters because the compaction function is the obvious place a future author would look for that guarantee.

## Also

Preserving structured `status` / `code` / `retriable` on async `stream_error` events is the useful half of this for anyone debugging a failed first turn — a retriable flag that survives into the event means a consumer can distinguish "try again" from "this will never work" without parsing prose.

PR #532, opened 2026-08-28
feat(kbc): ingest repository snapshots and harden compile image

LikiosSedo:
## Summary

- install Source Snapshot v3 as an immutable repository Raw tree while preserving the exact Git revision, paths, MD/MDX documents, `docs.json`, and standalone media
- keep Source Snapshot v2 compatibility and the existing Agent `read` / `search` / `slice` / `compile` workflow; v3 adds provenance, not a product-specific parser
- retain repository media as first-class Raw evidence instead of deleting or auto-excluding files that are not linked by ordinary Markdown syntax
- normalize `/app` permissions in the KBC image before switching to the non-root `kbc` user, and add an image-level regression smoke from a deliberately restrictive build context (integrates #533)

## Deliberate boundary

This PR does not add a Mintlify/Docusaurus parser, execute JSX or JavaScript, create a prepared-view tree, or make the control plane infer document/media relationships. Sicore owns repository snapshot intake; Siclaw receives the unchanged Raw tree and exposes MD/MDX through the existing source tools.

## Real repository acceptance

Validated with `siflow-ai/mintlify-docs` commit `c85062168f772b348f93aa75de4856237bc922db` after the matching Sicore archive intake:

- 363 tracked files
- 361 accepted immutable Raw files; only `logo/favicon.ico` and `siverse/llm/.ds-store` were excluded
- Agent inventory and batch planner exposed 360 sources: all 134 MDX documents and all 219 images
- `docs.json` was readable and searchable through SourceInspector
- `.mintignore` remained preserved in Raw but stayed outside the existing hidden-file planning allowlist

## Verification

- Source Snapshot: 15 tests passed, 1 environment-dependent test skipped
- the pre-integration Snapshot v3 head passed all five GitHub checks
- PR #533 exact commit `721d18f3845d0156c028269608819db46e317fcd` passed all five GitHub checks and a dev02 restrictive-context image build/start smoke as uid 1000
- `sh -n kbc/platform/pod/test_image_permissions.sh`
- `git diff --check`

The combined exact-head CI is the final merge gate.

## Related delivery and rollout

- Sicore archive intake: http://gitlab.scitix-inner.ai/k8s/sicore/-/merge_requests/1052
- deploy this Siclaw revision to every compile box before deploying Sicore Snapshot v3 producers

No production deployment is included in this PR.


jacoblee-io (approved):
Approving. Scope: I focused on the image-permission change and on what the accounting rewrite does to the coverage ledger, since that's where the deleted code carried explicit fail-open reasoning. I did not read `test_image_permissions.sh`, the `materialize.ts` contract change, or the fixtures in detail.

## The Dockerfile change is right, including the detail that's easy to get wrong

```dockerfile
RUN chmod -R u=rwX,go=rX /app
```

Capital `X` rather than `x` is the part worth calling out: it adds the execute bit to directories and to files that already had one, instead of making every `.py` and `.md` in the tree executable. Combined with root ownership and `go=rX`, the runtime user gets read-only access to a tree it doesn't own, which is what the comment claims. The root cause is also stated rather than implied — Docker preserves build-context modes, so a checkout under a restrictive umask produces an image the non-root user can't read — and that's exactly the kind of environment-dependent breakage that needs the regression script this PR adds rather than a comment.

## v3 provenance validation

`_validate_source_revision` is a whitelist projection: dict required, `origin` and `revision` bounded at 64 and 256 characters, both normalized through `str(... or "").strip()`, and only those two keys returned. Nothing else from the payload survives into the committed snapshot. Version acceptance widens to `{2, 3}` without touching the v2 shape.

## One thing the description doesn't say: this changes accounting for v2 libraries too

The summary frames v3 as "provenance, not a product-specific parser", which is accurate for the snapshot format. But the ledger rewrite isn't v3-scoped.

`is_media_asset` previously required an `assets/` path segment *and* an image extension, with the narrowness deliberate:

> Deliberately NARROW: sheet placeholders at `assets/sheets/*.md` and every other `.md`/`.json` are content files … Auto-attaching a sheet placeholder would launder 'the table data was never compiled' into 'covered'

It now keys on extension alone. The extension half of that guard survives — `.md` and `.json` still can't be auto-attached, so the laundering case the old comment names is still closed. What relaxes is location: an image anywhere in the tree is now auto-accountable, where before only images under `assets/` were.

The second change is the bigger one. Orphans used to stay unaccounted on purpose:

> an ORPHAN asset embedded by no accounted document (upload residue — it stays unaccounted and must be excluded with a reason, **so a human still sees it**)

They're now auto-accounted with a `<source-tree>` receipt. For a repository snapshot that's the honest answer — a git-tracked image is something an author committed, not upload residue, and JSX or a product config can reference it in ways no Markdown parser will see. But the same code path serves v2 wiki libraries, where "upload residue" was a real category and forcing a human to look at it was the point.

Both directions stay monotonic, which the new comment says explicitly and which matters: nothing that was green turns red. The flip side is that things which were *red* can now go green — an existing v2 library recompiled after this ships may close a ledger that previously required someone to write an exclusion reason. That's a defensible trade for the repository case; it's just not something a reader would infer from "v3 adds provenance".

Worth a line in the description, and ideally a note in the ledger docstring, saying that orphan media no longer surfaces for human review in any library — so the next person debugging "why did this library go green" finds the answer where they look.

## Minor

The description links an internal GitLab MR and names an internal repository and commit. PR bodies are public repository metadata, and there's been an effort to keep internal references out of this repo — same note I've raised before, so treat it as a standing preference rather than a new finding.

CI green across all five checks, including KBC Box Test.

PR #533, opened 2026-08-29
fix(kbc): make image permissions independent of build context

LikiosSedo:
## Summary

- normalize the root-owned KBC application tree before switching to the non-root `kbc` user
- add an image-level regression smoke that builds from a deliberately restrictive `0600`/`0700` context
- verify the image default user can read every `/app` file, import `compile_box`, start the real entrypoint, and pass `/health`

## Root cause

Docker preserves build-context modes on `COPY`. The production image was built from KBC source files with mode `0600`, then switched to `USER kbc` (uid 1000), so Python could not read `/app/compile_box.py` and every compile box exited before compilation began. Source-level Python CI did not build or start the final image as its runtime user.

## Verification

- `sh -n kbc/platform/pod/test_image_permissions.sh`
- `git diff --check`
- dev02 (`si-dev02`), exact commit `721d18f3845d0156c028269608819db46e317fcd`: restrictive-context image build, uid 1000 full-tree readability, import, entrypoint startup, and `/health` passed
- negative control using the pre-fix Dockerfile failed on the expected unreadable `/app` files

## Deployment

No production change is included. Recovery requires publishing a new immutable KBC image digest and updating the production compile-box image after approval.

### DIFF
--- a/.github/workflows/kbc-ci.yml
+++ b/.github/workflows/kbc-ci.yml
@@ -61,3 +61,5 @@ jobs:
         working-directory: kbc/platform/pod
       - run: python test_emit.py
         working-directory: kbc/tools
+      - name: Build and start the KBC image as its default non-root user
+        run: sh kbc/platform/pod/test_image_permissions.sh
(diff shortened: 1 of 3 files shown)


LikiosSedo:
Integrated into #532 as commit e5c37194015e84e0f9f926379d0671e5615d652f, preserving the original authorship. PR #532 is now the single KBC release candidate covering both repository Snapshot v3 intake and the non-root image-permission fix; its combined exact-head CI is green. Closing this PR as superseded.

PR #535, opened 2026-08-31
fix(knowledge): preserve Agent-led Wiki exploration

LikiosSedo:
## Outcome

Keep the Karpathy-style LLM Wiki as the Knowledge QA reasoning and discovery system. `knowledge_search` is an optional accelerator for a unique, high-confidence, complete single-page lookup; it is not a mandatory router or knowledge authority.

## Product contract

- the mounted Wiki pages, catalog, and links remain the factual source and discovery graph
- the Agent decides whether a question is concrete and likely single-page before optionally calling `knowledge_search` once
- broad, novel, ambiguous, comparative, weak-match, oversized-page, and cross-page questions use Agent-led Find/Grep/Read exploration
- search candidates are hypotheses, not evidence or proof of applicability
- only one unique identity-and-passage match whose complete page fits the evidence budget can return `direct_hit`
- `explore` exposes bounded unverified leads with no page body or registered evidence
- `explore` and `unavailable` never mean that the Wiki lacks an answer
- embeddings are optional candidate generation; FTS and Wiki exploration remain available without them

## Changes

- split editable Knowledge QA identity from platform-owned runtime retrieval policy; exact historical built-in defaults are upgraded while administrator-edited prompts remain authoritative
- inject the actual runtime Wiki root and `indexPath`, including LocalSpawner Agent-scoped mounts and custom Portal CLI locations; results return directly usable `readPath` values
- preserve negation and numeric/version qualifiers at the direct-hit gate so opposite actions and wrong versions cannot pass after recall-oriented FTS tokenization
- return `explore` for large pages instead of isolated matched sections that may omit applicability, deprecation, conflict, or version context
- normalize hybrid weights per candidate so a strong FTS-only exact match is not suppressed when vectors find a different page
- start FTS immediately; hydrate knowledge vectors in the background with the durable retry policy, keep query embedding short and fail-fast, and preserve blocking durable indexing for investigation memory
- keep answering through FTS/Wiki exploration when an optional release embedding descriptor is malformed
- retain exact mounted-snapshot citation validation for pages actually read and used
- compute direct-hit citation capability against the selected page and its own repository manifest; uncitable pages now return `citationMode: none` without a doomed citation call
- update architecture and ADR documentation around the accelerator boundary

## Regression coverage

- catcher versus tangential sichek
- weak and competing candidates
- opposite-action queries
- requested year/version and attached product versions such as `CUDA12.4`
- GSP semantic aliases
- navigation pages
- oversized pages
- invalid paths and index failures
- vector-only and FTS-only candidates in the same result set
- unavailable and malformed embedding configuration
- LocalSpawner Agent-scoped Wiki paths
- materialized default Prompt compatibility
- mixed citable and uncitable knowledge repositories, including unresolved evidence mappings

## Verification

- `npm test` — 284 files, 6345 passed, 2 skipped
- `npx tsc --noEmit`
- `npm run build`
- `git diff --check`

## Acceptance boundary

This PR deliberately prefers extra Wiki exploration over a false direct hit. It does not add a programmatic completion gate that forces model behavior after `explore`; the Agent retains control of Wiki discovery and synthesis. Further threshold or embedding changes require a fixed E2E evaluation set and must prove that worst-case answer quality is no worse than Agent-led Wiki exploration without embeddings.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -1,0 +1,25 @@
+# Siclaw Knowledge Q&A
+
+Siclaw Knowledge Q&A helps an Agent understand a question, discover applicable material in a mounted LLM Wiki, and synthesize an answer from sources it has verified.
+
+## Language
+
+**Retrieval Accelerator**:
+An optional fast path that proposes a direct page for a concrete, high-confidence question. It is a navigation aid, not the authority for what the question means or which facts answer it.
+_Avoid_: Retrieval authority, answer router
+
+**Direct Hit**:
+A single page whose identity, matched passage, and applicability strongly cover the question without unresolved competing pages.
+_Avoid_: Top result, nearest page
+
+**Exploration**:
+Agent-led discovery through the Wiki's catalog, paths, links, and page contents when a question is novel, ambiguous, broad, or distributed across pages.
+_Avoid_: Search failure, fallback answer
+
+**Candidate**:
+A page proposed for inspection by retrieval or navigation. A Candidate is a hypothesis until the Agent verifies its subject, task, version, environment, and scope.
+_Avoid_: Evidence, answer source
+
+**Evidence**:
+Exact mounted page content that the Agent has read, judged applicable, and materially used in its answer.
+_Avoid_: Search result, citation candidate
--- a/README.md
+++ b/README.md
@@ -266,7 +266,7 @@ All configuration happens through the web UI:
 | Language | TypeScript 5.9 |
 | Agent | [pi-coding-agent](https://github.com/badlogic/pi-mono) |
 | Database (portal) | MySQL (prod) or SQLite (local, via [node:sqlite](https://nodejs.org/api/sqlite.html)) — single DDL, driver chosen by `DATABASE_URL` scheme |
-| Database (memory) | node:sqlite + FTS5 + bge-m3 embeddings |
+| Database (memory) | node:sqlite + FTS5 + optional OpenAI-compatible embeddings (for example BGE-M3) |
 | Frontend | React + Vite + Tailwind CSS |
 | K8s Client | @kubernetes/client-node |
 | MCP | @modelcontextprotocol/sdk |
--- a/docs/design/decisions.md
+++ b/docs/design/decisions.md
@@ -125,14 +125,22 @@ Whitelist-only model with 4-pass validation: binary allowlist → per-command va
 Memory search needs to handle both semantic similarity ("find memories about OOMKilled pods") and exact keyword matching ("find memories mentioning node-23"). Pure vector search misses exact terms; pure BM25 misses semantic similarity.
 
 **Decision**:
-Hybrid scoring: `score = vectorWeight × cosineSimilarity + ftsWeight × BM25`
+Generate candidates independently with FTS5 and, when configured, vector
+similarity. Fuse each candidate using only the channels that found that
+candidate:
 
-Default: `vectorWeight = 0.70`, `ftsWeight = 0.30` (source: `src/memory/indexer.ts:14-15`). Configurable via `MemorySearchConfig`.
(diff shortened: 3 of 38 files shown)


jacoblee-io (approved):
Approving. Scope: 2371 lines is more than I can claim to have reviewed. I concentrated on the `direct_hit` gate, since it is the only path that produces an authoritative answer and the entire premise of this PR is that a false one is worse than more exploration. I did not review the hybrid weighting, the background vector hydration, or the citation-mode changes in depth.

## The product judgement is the strongest part

Demoting `knowledge_search` from router to accelerator, and stating that candidates are hypotheses rather than evidence, is the right call for a retrieval layer sitting under an agent that can already read the corpus. So is the acceptance boundary — preferring extra exploration over a false hit, and requiring a fixed evaluation set plus a no-worse-than-baseline proof before thresholds move again. Good to see that last one landed in the design doc rather than only in the PR description, since the person who eventually tunes a threshold will read the doc, not this thread.

## The message copy does real work

Worth calling out because it's easy to treat as cosmetic. Every non-hit outcome says some version of:

> This is not evidence that the Wiki lacks an answer; explore its catalog, links, and pages with Find/Grep/Read.

> The retrieval accelerator is unavailable (…). Continue with Wiki Find/Grep/Read; this does not mean the knowledge is absent.

Reading these as the consumer: an empty or low-confidence retrieval result is exactly the kind of output a model will treat as "I checked, there's nothing there" and stop. Saying otherwise in the payload itself — rather than only in the system prompt — is what makes `explore` behave as a redirect instead of a dead end. Same for "Treat the hints as unverified leads": without it, a returned candidate list reads as an answer set.

The `__negated__` sentinel is the other detail I'd have flagged if it were missing. Recall-oriented tokenization dropping "not" means *do not restart* and *restart* collapse to the same tokens, and a `direct_hit` on the opposite action is the worst output this system can produce. Encoding the negation as a token that survives tokenization, then requiring it on the page side, is the right shape.

## One thing I'd tighten

```js
directCandidate.qualifiersMatch !== false &&
```

`routingQualifiersMatch` itself returns a plain boolean — `true` when the query carries no `__`-prefixed qualifiers, otherwise every required token must appear on the page. So the function never yields `undefined`. But the field is declared `qualifiersMatch?: boolean`, and the gate is written to accept `undefined`.

That means "verified to match" and "never evaluated" are the same answer at the gate. If every candidate that can reach `direct_hit` passes through the assignment that computes it, then `=== true` is behaviourally identical and costs nothing — while making the invariant structural instead of dependent on all construction paths remembering to set the field. If some candidates *can* arrive without it — the description mentions FTS-only and vector-only candidates being normalized into one result set — then a vector-recalled page that is semantically close but describes the opposite action would currently clear the qualifier check by not having been checked.

I couldn't establish which of those is true without reading the full candidate construction, so I'm raising the asymmetry rather than claiming a reachable bug. For a gate whose stated job is that "opposite actions and wrong versions cannot pass", the unevaluated case should fail closed like the mismatch case does.

## Also noticed

`citationMode: none` for uncitable pages — computing citation capability against the selected page's own manifest, so the model isn't sent into a citation call that cannot succeed — is the kind of thing that only shows up after watching real traces. Failing early with an explicit mode beats a doomed call whose failure the model then has to interpret.

CI green across Test / Type Check / Portal Web Test / AgentBox Build Graph, with 6345 tests passing.

LikiosSedo:
Closing as superseded.

This branch rewrites `knowledge_search` as an FTS/`direct_hit` accelerator. `main` after #542 / #546 / #549 / #551 is labels-only and does not search page bodies. The nine commits are not in `main` (`git cherry`), and the PR now conflicts (70 behind). Merging it would restore a retrieval model we already replaced.

If a later retrieval change is still needed, start from current `main`, not this branch.

PR #537, opened 2026-09-01
fix(runtime): support strict product support sessions

LikiosSedo:
## Summary

- add `product_support` as a read-only built-in Agent type with a minimal runtime-owned contract while leaving the business prompt and result schema to the control plane
- let machine callers require durable session/message persistence and exactly one successful structured result-tool output per turn
- run one bounded, provider-neutral repair pass when the main response omits the required result, without allowing compaction or concurrent steers to consume the repair contract
- reject strict concurrent sends instead of degrading them into the active turn, and surface missing/inactive result tools as explicit failed outcomes
- advertise strict-result protocol version `1` only when the complete durable-session/result contract is active, so the control plane can reject older runtimes during rolling upgrades
- validate the strict result-tool input at the AgentBox boundary and clean up the in-memory session registry on strict persistence failure

Deploy this runtime support before accepting strict product-support traffic from a control plane.

## Contract boundary

- Siclaw owns the generic Agent type, strict turn mechanics, and runtime failure semantics.
- The control plane owns the released product-support prompt, result-tool schema, session API, and channel-facing streaming contract.
- A text-only tool completion is not a valid strict result; success requires the configured tool's non-error `structuredContent` object.

## Verification

- `npm run build`
- `npm test` (296 files; 6,604 passed, 2 skipped)
- `cd portal-web && npm test` (25 files; 246 passed)
- `cd portal-web && npm run build`

### DIFF
--- a/src/agentbox/http-server.ts
+++ b/src/agentbox/http-server.ts
@@ -113,6 +113,8 @@ interface PromptRequestBody {
    * lets the box mint nothing can still correlate.
    */
   turnId?: string;
+  /** Trusted control-plane contract for one required structured result tool. */
+  requiredResultToolName?: string;
 }
 
 /**
@@ -876,6 +878,13 @@ export function createHttpServer(
       console.log(`[agentbox-http] /api/prompt result boxId=${process.env.SICLAW_POD_NAME ?? "unknown"} sessionId=${sessionId ?? "pending"} turnId=${body.turnId ?? "unknown"} status=${status} outcome=${outcome} durationMs=${Date.now() - promptStartedAt}${suffix}`);
     };
 
+    if (body.requiredResultToolName !== undefined && typeof body.requiredResultToolName !== "string") {
+      logPromptResponse(400, "rejected", "'requiredResultToolName' must be a string");
+      sendJson(res, 400, { error: "'requiredResultToolName' must be a string" });
+      return;
+    }
+    const requiredResultToolName = body.requiredResultToolName?.trim() || undefined;
+
     const promptMediaValidation = validatePromptMedia(body.images, body.files);
     if (promptMediaValidation.error) {
       logPromptResponse(400, "rejected", promptMediaValidation.error);
@@ -1294,6 +1303,9 @@ export function createHttpServer(
         },
       },
       promptMedia,
+      requiredResultToolName
+        ? { requiredResultToolName }
+        : undefined,
     );
 
     promptPromise.then((result) => {
--- a/src/core/agent-types.ts
+++ b/src/core/agent-types.ts
@@ -15,6 +15,9 @@
  *                   No skills by default.
  *   - knowledge_qa — researches bound knowledge bases and synthesizes sourced
  *                    answers. Read-only, no skills by default, and no delegation.
+ *   - product_support — managed front-door customer support. Its persisted
+ *                       prompt and bound MCP define the intake/result contract;
+ *                       built-in filesystem access stays read-only.
  *   - custom      — the legacy free-form agent. Standalone Portal may persist
  *                   an operator's tool_capabilities selection; integrations
  *                   that omit it intentionally retain unrestricted built-ins.
@@ -24,7 +27,7 @@
  * is the built-in type contract; Custom has no built-in contract.
  */
 
-export type AgentType = "sre" | "coordinator" | "knowledge_qa" | "custom";
+export type AgentType = "sre" | "coordinator" | "knowledge_qa" | "product_support" | "custom";
 
 export interface AgentTypeDef {
   label: string;
@@ -236,10 +239,21 @@ const REPLACED_KNOWLEDGE_QA_DEFAULT_PROMPTS = new Set([
   COMPLETE_CATALOG_KNOWLEDGE_QA_DEFAULT_PROMPT,
 ]);
 
+/**
+ * Public runtime invariant for Product Support. The control plane still owns
+ * the released business prompt and result schema; this layer only keeps the
+ * built-in type safe and meaningful when its managed addendum is absent.
+ */
+export const PRODUCT_SUPPORT_DEFAULT_PROMPT =
+  "You are a product-support agent. Answer product questions using the configured tools and resources. " +
+  "When the current turn provides a result-submission tool, call it exactly once with a machine-readable outcome that follows its declared schema. " +
+  "Do not claim that downstream actions such as ticket creation or human handoff succeeded unless the caller confirms them.";
+
 const MATERIALIZED_TYPE_PROMPTS: Record<Exclude<AgentType, "custom">, ReadonlySet<string>> = {
   sre: new Set([SRE_DEFAULT_PROMPT]),
   coordinator: new Set([COORDINATOR_DEFAULT_PROMPT, PREVIOUS_COORDINATOR_DEFAULT_PROMPT]),
   knowledge_qa: REPLACED_KNOWLEDGE_QA_DEFAULT_PROMPTS,
+  product_support: new Set([PRODUCT_SUPPORT_DEFAULT_PROMPT]),
 };
 
 export interface AgentPromptLayers {
@@ -277,6 +291,16 @@ export const AGENT_TYPES: Record<AgentType, AgentTypeDef> = {
     defaultPrompt: KNOWLEDGE_QA_DEFAULT_PROMPT,
     defaultNoSkills: true,
   },
+  product_support: {
+    label: "Product Support Agent",
+    description: "Answers product questions and prepares structured customer-support handoffs through its bound result tool.",
+    capabilities: ["read_files"],
+    // The control plane owns the managed business prompt and result schema.
+    // This generic runtime contract is deliberately schema-free, so those stay
+    // single-source.
+    defaultPrompt: PRODUCT_SUPPORT_DEFAULT_PROMPT,
+    defaultNoSkills: true,
+  },
   custom: {
     label: "Custom Agent",
     description: "Free-form built-in capabilities; explicitly resolved Custom agents with no selection retain legacy unrestricted compatibility.",
@@ -288,7 +312,7 @@ export const AGENT_TYPES: Record<AgentType, AgentTypeDef> = {
 
 /** Normalize an unknown stored value to a valid AgentType (default custom). */
 export function normalizeAgentType(v: unknown): AgentType {
-  return v === "sre" || v === "coordinator" || v === "knowledge_qa" ? v : "custom";
+  return v === "sre" || v === "coordinator" || v === "knowledge_qa" || v === "product_support" ? v : "custom";
 }
 
 /**
@@ -299,7 +323,7 @@ export function normalizeAgentType(v: unknown): AgentType {
  * tools enter a model session must fail closed when provenance is absent.
  */
 export function requireAgentType(v: unknown): AgentType {
-  if (v === "sre" || v === "coordinator" || v === "knowledge_qa" || v === "custom") {
+  if (v === "sre" || v === "coordinator" || v === "knowledge_qa" || v === "product_support" || v === "custom") {
     return v;
   }
   throw new Error(`Invalid or missing agent_type: ${String(v)}`);
--- a/src/core/brains/pi-agent-brain.ts
+++ b/src/core/brains/pi-agent-brain.ts
@@ -16,6 +16,7 @@ import type {
   BrainProviderResponse,
   BrainContextPreflightResult,
   PromptMedia,
+  PromptRequirements,
 } from "../brain-session.js";
 import { estimateMessagesTokens } from "../compaction.js";
 import { rememberPromptFiles } from "../openai-file-payload.js";
@@ -48,6 +49,46 @@ function readCompatBoolean(compat: unknown, key: string): boolean | undefined {
   return typeof value === "boolean" ? value : undefined;
 }
 
+/**
+ * Provider-neutral "call this tool" choice. OpenAI chat-completions providers
+ * can name the exact function; providers without a named-tool form still get a
+ * deterministic choice because the repair exposes exactly one tool.
+ */
+export function requiredToolChoice(
+  api: unknown,
+  toolName: string,
+): "any" | "required" | { type: "function"; function: { name: string } } {
+  if (api === "openai-completions") {
+    return { type: "function", function: { name: toolName } };
+  }
+  return api === "anthropic-messages" || api === "bedrock-converse-stream" ||
+    api === "google-generative-ai" || api === "google-vertex"
+    ? "any"
+    : "required";
+}
+
+function requiredResultRepairOptions(
+  model: unknown,
+  options: Record<string, any> | undefined,
+  requiredToolName: string,
+): Record<string, any> {
+  return {
+    ...options,
+    toolChoice: requiredToolChoice(
+      model && typeof model === "object" ? (model as { api?: unknown }).api : undefined,
+      requiredToolName,
+    ),
+  };
+}
+
+function contextHasTool(context: unknown, toolName: string): boolean {
+  if (!context || typeof context !== "object") return false;
+  const tools = (context as { tools?: unknown }).tools;
+  return Array.isArray(tools) && tools.some((tool) =>
+    tool != null && typeof tool === "object" && (tool as { name?: unknown }).name === toolName
+  );
+}
+
 export class PiAgentBrain implements BrainSession {
   readonly brainType = "pi-agent" as const;
 
@@ -61,6 +102,9 @@ export class PiAgentBrain implements BrainSession {
   /** True from abort() until the next prompt() starts. Stops the empty-response retry loop from
    *  firing a fresh (un-aborted) re-prompt when Stop lands during the backoff sleep. */
   private aborted = false;
+
+  /** A strict repair owns the session loop; steers must become separate turns. */
+  private requiredResultRepairActive = false;
 
   constructor(
     readonly session: AgentSession,
@@ -101,6 +145,11 @@ export class PiAgentBrain implements BrainSession {
   }
   private static readonly RETRY_DELAY_MS = 2000;
   private static readonly PROMPT_PREFLIGHT_SAFETY_TOKENS = 2048;
+  private static readonly REQUIRED_RESULT_REPAIR_PROMPT =
+    "[System: required-result repair]\n" +
+    "The preceding attempt did not submit the required structured turn result. " +
+    "Call the only available result-submission tool exactly once now, using the conversation and any preceding draft to fill its schema. " +
+    "If validation rejects the call, correct it. After one successful call, give the user the final answer without mentioning this repair, tools, or schemas.";
 
   private emit(event: any): void {
     for (const listener of this.extraListeners) {
@@ -116,10 +165,11 @@ export class PiAgentBrain implements BrainSession {
     });
   }
 
-  async prompt(text: string, media?: PromptMedia): Promise<void> {
+  async prompt(text: string, media?: PromptMedia, requirements?: PromptRequirements): Promise<void> {
     this.aborted = false;
     let lastAssistantHadContent = false;
     let lastAssistantMessage: any = null;
+    const successfulTools = new Set<string>();
 
     const promptOptions = this.promptOptionsForMedia(media);
 
@@ -134,6 +184,17 @@ export class PiAgentBrain implements BrainSession {
         const hasText = content.some((c: any) => c.type === "text" && c.text?.trim());
         const hasToolCalls = content.some((c: any) => c.type === "toolCall");
         if (hasText || hasToolCalls) lastAssistantHadContent = true;
+      }
+      if (
+        event.type === "tool_execution_end" &&
+        typeof event.toolName === "string" &&
+        event.isError !== true &&
+        event.result?.details?.error === undefined &&
+        event.result?.details?.structuredContent !== null &&
+        typeof event.result?.details?.structuredContent === "object" &&
+        !Array.isArray(event.result?.details?.structuredContent)
+      ) {
+        successfulTools.add(event.toolName);
       }
     });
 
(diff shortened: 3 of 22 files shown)


src/core/brains/pi-agent-brain.ts
```diff
@@ -207,11 +289,69 @@ export class PiAgentBrain implements BrainSession {
...
+
+      const requiredResultToolName = requirements?.requiredResultToolName?.trim();
+      if (
+        requiredResultToolName &&
+        !this.aborted &&
+        lastAssistantMessage?.stopReason !== "aborted" &&
+        lastAssistantMessage?.stopReason !== "error" &&
+        !successfulTools.has(requiredResultToolName)
+      ) {
+        await this.repairMissingRequiredResult(requiredResultToolName, successfulTools);
```
jacoblee-io, line 280: high: The repair pass is not failure-isolated — this `await` has no `catch`, and inside `repairMissingRequiredResult` the `await session.prompt(REPAIR_PROMPT)` is wrapped only in `try`/`finally`.

Scenario: the provider rejects the forced `toolChoice` (a gateway with no named-function-choice support, Ollama, a non-thinking Kimi variant), or the repair round hits a rate limit or a context overflow. The rejection propagates out of `brain.prompt`, `runAttempt` converts it into a candidate failure, and a turn whose main pass already streamed a complete answer to the user is reported as a failed turn (`stopReason:"error"` plus an error bubble). A bounded best-effort repair must never be able to destroy an answer that already succeeded.

Fix: catch around the repair (or around `session.prompt` inside it), log, and let `required_result_repair_end` report `success:false` instead of rethrowing.

src/core/brains/pi-agent-brain.ts
```diff
@@ -135,6 +206,17 @@ export class PiAgentBrain implements BrainSession {
         const hasToolCalls = content.some((c: any) => c.type === "toolCall");
         if (hasText || hasToolCalls) lastAssistantHadContent = true;
       }
+      if (
+        event.type === "tool_execution_end" &&
+        typeof event.toolName === "string" &&
+        event.isError !== true &&
+        event.result?.details?.error === undefined &&
+        event.result?.details?.structuredContent !== null &&
```
jacoblee-io, line 193: high: The success criterion requires `details.structuredContent` to be a non-null, non-array object, but that key is optional in practice: `src/core/mcp-client.ts:358` sets it only when the server actually returned one, and an MCP server emits `structuredContent` only when the tool declares an `outputSchema`.

Scenario: a fully compliant `mcp__result__submit` that returns text content only. It is judged "result missing", so every strict turn runs the repair and invokes the side-effecting submit tool a second time — a duplicate customer-support handoff — and `required_result_repair_end.success` still reports `false` even though both calls succeeded.

Note: this PR's test `"repairs a result-tool call that succeeded at transport level without structuredContent"` pins exactly this behaviour, so a passing suite does not clear the finding. The stated contract is "the result tool must succeed", not "must return structuredContent".

Fix: treat a non-error `tool_execution_end` for the contracted tool as success, and require `structuredContent` only if the contract also guarantees the bound server declares an `outputSchema`.

src/core/brains/pi-agent-brain.ts
```diff
@@ -207,11 +289,69 @@ export class PiAgentBrain implements BrainSession {
...
+      agent: AgentSession["agent"] & { streamFn: (...args: any[]) => any };
+    };
+    const activeToolNames = session.getActiveToolNames?.() ?? [];
+    if (!activeToolNames.includes(requiredToolName) || !session.setActiveToolsByName) {
+      console.error(`[pi-agent-brain] Required result tool is not active: ${requiredToolName}`);
+      return;
+    }
+
+    const originalStreamFn = session.agent.streamFn;
+    let forceNextProviderRequest = true;
```
jacoblee-io, line 317: high: The one-shot `forceNextProviderRequest` is not guaranteed to be spent on the repair's own model request. `AgentSession.prompt()` runs `_checkCompaction(lastAssistant, false)` before the agent loop (`agent-session.js:844`), and `_runAutoCompaction` calls `compact()` with `this.agent.streamFn` (`agent-session.js:1635`) — the very function patched two lines below.

Scenario: a long support conversation past the compaction threshold. The first call through the patched `streamFn` is the compaction request: it receives the injected `toolChoice` (e.g. `{type:"function",function:{name:"mcp__result__submit"}}`) while carrying no tools, which OpenAI-compatible and Anthropic endpoints reject with a 400. pi swallows that (emits `compaction_end` with an `errorMessage`, returns false), and the force is already consumed — so the actual repair request goes out unforced and the model typically answers with text and no tool call. The repair silently no-ops in exactly the long-context case it exists for, plus a spurious compaction failure surfaces to the UI.

Fix: gate the force on the request really being the agent turn (e.g. only when the outgoing context carries the repair's tools), or arm the flag after the pre-prompt compaction check rather than before `session.prompt`.

src/core/brains/pi-agent-brain.ts
```diff
@@ -207,11 +289,69 @@ export class PiAgentBrain implements BrainSession {
...
+    requiredToolName: string,
+    successfulTools: ReadonlySet<string>,
+  ): Promise<void> {
+    const session = this.session as AgentSession & {
+      getActiveToolNames?: () => string[];
+      setActiveToolsByName?: (names: string[]) => void;
+      agent: AgentSession["agent"] & { streamFn: (...args: any[]) => any };
+    };
+    const activeToolNames = session.getActiveToolNames?.() ?? [];
+    if (!activeToolNames.includes(requiredToolName) || !session.setActiveToolsByName) {
```
jacoblee-io, line 304: medium: When the contracted tool is not in the active set this returns after a `console.error`, emitting neither `required_result_repair_start` nor `required_result_repair_end` — so the turn reports ordinary success and the strict caller receives no signal at all.

Scenario: the MCP client swallows a connect failure at startup (`src/core/mcp-client.ts:270` catches and continues), so the result tool never registers. Every strict turn then completes "successfully" with no structured result, and the control plane cannot distinguish "result submitted" from "the contracted tool was never bound".

Fix: emit `required_result_repair_end` with `success:false` and a reason on this path (or a distinct event), so an unbound contract tool is observable rather than silent.

src/core/brains/pi-agent-brain.ts
```diff
@@ -207,11 +289,69 @@ export class PiAgentBrain implements BrainSession {
...
+    session.agent.streamFn = ((model: any, context: any, options?: Record<string, unknown>) => {
+      const forceThisRequest = forceNextProviderRequest;
+      forceNextProviderRequest = false;
+      return originalStreamFn(model, context, forceThisRequest
+        ? requiredResultRepairOptions(model, options, requiredToolName)
+        : options);
+    }) as typeof session.agent.streamFn;
+
+    this.emit({ type: "required_result_repair_start", toolName: requiredToolName });
+    session.setActiveToolsByName([requiredToolName]);
```
jacoblee-io, line 331: medium: `setActiveToolsByName([requiredToolName])` narrows the tool set session-wide for the whole repair window, not just for the repair's own reasoning.

Scenario: a user steer queued during the main pass is delivered into the repair run (pi hands queued steering messages to the next loop). That steer — "also check X" — is then answered by an agent that has only the submit tool: no read, grep, or knowledge search. The same applies to any steer arriving while the repair is in flight.

Fix: drain or defer queued steering before narrowing and re-deliver after the restore, or skip the repair when the queue is non-empty and report `success:false`.

src/gateway/server.ts
```diff
@@ -797,6 +804,7 @@ export async function startRuntime(opts: StartRuntimeOptions): Promise<RuntimeSe
     const promptOpts: PromptOptions = {
       sessionId,
       turnId,
+      requiredResultToolName,
```
jacoblee-io, line 807: medium: This comment is about `src/gateway/server.ts:909` and `src/gateway/server.ts:1020` — the two steer fallbacks — anchored here because those lines are outside the diff.

Both call `client.steerSession(sessionId, text, { images, files })`, which has no field for the requirement being threaded on this line, so a degraded send loses `requiredResultToolName` entirely.

Scenario: a strict caller re-sends on a session whose previous turn is still running, or retries after a lost ack. The send degrades to a steer, no repair pass ever runs, and the caller gets an accepted ack while "one result tool must succeed per turn" is silently broken for that request.

Fix: carry the requirement on the steer path as well (extend `steerSession`), or refuse rather than degrade when `requiredResultToolName` is set.

src/core/agent-types.ts
```diff
@@ -277,6 +283,15 @@ export const AGENT_TYPES: Record<AgentType, AgentTypeDef> = {
     defaultPrompt: KNOWLEDGE_QA_DEFAULT_PROMPT,
     defaultNoSkills: true,
   },
+  product_support: {
+    label: "Product Support Agent",
+    description: "Answers product questions and prepares structured customer-support handoffs through its bound result tool.",
+    capabilities: ["read_files"],
+    // Sicore owns this managed type's released business prompt. Keeping no
+    // Runtime fallback avoids a second copy of its result schema drifting here.
+    defaultPrompt: null,
```
jacoblee-io: medium: `product_support` is the only built-in type with `defaultPrompt: null` **and** an empty `MATERIALIZED_TYPE_PROMPTS` entry, so `resolveAgentPromptLayers` returns `typeContract: undefined` and surfaces the entire prompt supplied by the upstream control plane as the editable Agent Addendum. That breaks the invariant stated in this file's header and mirrored in `portal-web/src/lib/agentTypes.ts` — that a built-in type provides an immutable contract the Portal never replaces.

Scenario: `src/portal/agent-api.ts:338` writes `agentPromptAddendum(nextAgentType, body.system_prompt) ?? null` on PUT. An operator opens a provisioned Product Support agent, clears the Addendum field — which is where the whole released prompt landed — and saves. The agent now has no prompt at all: no intake instructions and no result-schema guidance, while the strict path still forces the submit tool every turn, so the forced call just fails validation repeatedly. Every other built-in type falls back to its type contract in the same situation.

Fix: give the type a minimal Runtime-side contract layer, or mark it as control-plane-owned so the Portal renders its prompt read-only instead of as an editable addendum.

src/agentbox/http-server.ts
```diff
@@ -1294,6 +1296,9 @@ export function createHttpServer(
         },
       },
       promptMedia,
+      body.requiredResultToolName
```
jacoblee-io: low: `body.requiredResultToolName` is forwarded on truthiness alone, with no type check — unlike the gateway, which gates it with `typeof params.requiredResultToolName === "string"` (`src/gateway/server.ts:736`).

Scenario: an older or buggy caller posts a non-string (a number, or an object). The value reaches `requirements.requiredResultToolName?.trim()` at `src/core/brains/pi-agent-brain.ts:293` and throws a TypeError inside `brain.prompt`, which `runAttempt` classifies as a candidate failure — so the turn fails outright instead of degrading to an unconstrained run.

Fix: mirror the gateway's `typeof === "string"` check here.

src/gateway/server.ts
```diff
@@ -860,6 +868,16 @@ export async function startRuntime(opts: StartRuntimeOptions): Promise<RuntimeSe
         promptMessageId = await appendMessage({ sessionId, role: "user", content: text, deferSequence: true });
         await incrementMessageCount(sessionId);
       } catch (persistErr) {
+        if (requireSessionPersistence) {
```
jacoblee-io, line 871: low: The strict-persistence rollback is incomplete. It undoes `registerPendingStart`, `addLiveTurn` and `delegatedTurns`, but not `sessionRegistry.remember(sessionId, userId, agentId)` at line 762, and it does not compensate the half-committed case where `ensureChatSession` succeeded and `appendMessage` failed.

Scenario: a database wobble that fails `appendMessage` repeatedly. Each rejected `chat.send` leaves an in-memory registry entry plus an orphan `chat_sessions` row for a session no turn ever ran; over a sustained outage those accumulate.

Fix: forget the registry entry on this path, and either wrap the two writes in one transaction or delete the session row when this call is the one that created it.

jacoblee-io (commented):
## Review scope and method

A single careful pass over the 21-file diff between merge-base `55452f25` and head `d4505972`, plus targeted verification against the surrounding source and the `pi-coding-agent` package internals. **The test suite was NOT run for this PR** — every finding below is a source-level argument, not a test failure.

### Findings 1-3 should block merge

1. **`repairMissingRequiredResult` has no `catch`**, so a repair-pass failure (provider rejecting the forced `toolChoice`, a rate limit, a context overflow) is thrown out of `brain.prompt` and fails a turn whose main pass had already streamed a complete answer.
2. **The required-result success criterion demands `details.structuredContent`**, which MCP servers only emit when the tool declares an `outputSchema` — so a compliant text-only result tool is judged "missing" and the side-effecting submit tool is called a second time on every strict turn.
3. **The one-shot `forceNextProviderRequest` can be consumed by pi's pre-prompt auto-compaction request**, which travels through the same `agent.streamFn` — the compaction request gets a `toolChoice` while carrying no tools, and the actual repair request then goes out unforced.

**On finding 2 specifically:** this PR adds a test that pins the disputed behaviour — `"repairs a result-tool call that succeeded at transport level without structuredContent"` — so a green suite does not clear it. The test locks in the criterion the finding argues is wrong.

### Remaining findings

- **Medium (4-7):** silent no-op with no event when the contracted tool is not bound; session-wide tool narrowing that catches queued steers; both steer fallbacks dropping `requiredResultToolName`; `product_support` being the only built-in type with no immutable prompt contract.
- **Low (8-10):** missing type check on the box-side `requiredResultToolName`; incomplete strict-persistence rollback; substring-based Kimi match with an unconditional payload rewrite.

Two comments are anchored to the nearest changed line in their file rather than the line they discuss (GitHub only accepts anchors inside the diff); each of those names its real target on the first line.

Verified and explicitly *not* reported: the `setActiveToolsByName` prompt rebuild does not clobber the compiled system prompt (the resource loader's `systemPromptOverride` covers it); every `requiredToolChoice` return value is type-valid for its `Api`; MCP tools bypass the read-only `allowedTools` whitelist so a `product_support` agent can still reach its result tool; the strict-persistence throw does precede any AgentBox dispatch; the repair's synthetic user echo neither mis-claims a pending user row nor renders as a chat bubble; and the new positional argument order into `runPromptWithModelRouting` is correct.

src/core/brains/pi-agent-brain.ts
```diff
@@ -48,6 +49,70 @@ function readCompatBoolean(compat: unknown, key: string): boolean | undefined {
...
+  if (api === "openai-completions") {
+    return { type: "function", function: { name: toolName } };
+  }
+  return api === "anthropic-messages" || api === "bedrock-converse-stream" ||
+    api === "google-generative-ai" || api === "google-vertex"
+    ? "any"
+    : "required";
+}
+
+function isKimiThinkingModel(model: unknown): boolean {
```
jacoblee-io: low: The Kimi match is a substring test and the payload rewrite it gates is unconditional. `identity.includes("kimi-k2")` also matches non-thinking K2 variants (e.g. `kimi-k2-0905-preview`), and line 111 then injects `thinking: {type:"disabled"}` into an OpenAI-compatible payload for all of them.

Scenario: a deployment whose K2 variant does not accept that top-level field — OpenAI-compatible endpoints commonly reject unknown params with a 400. Combined with the missing failure isolation (finding 1 on line 301), every strict turn on that provider fails.

Fix: narrow the match to the model ids that genuinely need thinking disabled, and make the injection tolerant of rejection rather than a hard prerequisite of the repair.

src/core/brains/pi-agent-brain.ts
```diff
@@ -207,11 +268,81 @@ export class PiAgentBrain implements BrainSession {
...
+
+    const originalStreamFn = session.agent.streamFn;
+    let forceNextProviderRequest = true;
+    session.agent.streamFn = ((model: any, context: any, options?: Record<string, unknown>) => {
+      // AgentSession can auto-compact before starting the repair turn. Compaction
+      // uses the same streamFn but carries no result tool; do not spend the
+      // one-shot force on that unrelated provider request.
+      const forceThisRequest = forceNextProviderRequest && contextHasTool(context, requiredToolName);
+      if (forceThisRequest) forceNextProviderRequest = false;
+      return originalStreamFn(model, context, forceThisRequest
```
jacoblee-io, line 324: The forced `toolChoice` injected here is silently dropped for every provider API except openai-completions: pi-ai's per-API `streamSimple` (anthropic-messages, google-generative-ai, google-vertex, openai-responses) rebuilds options via `buildBaseOptions()`, whose fixed key list does NOT include `toolChoice` — only `api/openai-completions.js` forwards it. So a product_support agent bound to a Claude or Gemini model gets a bare re-prompt with a restricted toolset instead of a forced call, and a model that answers in text again yields `required_result_repair_end success:false` on exactly the providers this was written for. Unit tests can't see it because they stub `streamFn`.

src/gateway/server.ts
```diff
@@ -886,6 +905,10 @@ export async function startRuntime(opts: StartRuntimeOptions): Promise<RuntimeSe
           releaseTurn = await sessionTurnLocks.acquire(sessionId);
         } catch (busyErr) {
           throwIfStoppedBeforePrompt();
+          // A strict machine request is one distinct turn with one distinct
+          // result. Folding it into the running turn as a steer would lose its
+          // required-result contract, so surface busy and let the caller retry.
+          if (requiredResultToolName) throw busyErr;
```
jacoblee-io, line 911: This strict busy guard (and the `throw err` at ~1025) executes inside the detached async IIFE AFTER `chat.send` already resolved `{ok:true, sessionId, turnId, messageId}` — so "busy, retry" can never reach the strict caller as an RPC failure. Concretely: `appendMessage` + `incrementMessageCount` succeed, the RPC returns ok, then the lock acquire throws and the IIFE's outer catch emits `stream_error` + `prompt_done`. chat-gateway's SSE watcher filters by sessionId only and closes on `prompt_done` (`res.end()`), so the RUNNING turn's stream is torn down mid-flight — the exact hazard the comment above warns about. The machine caller holding ok:true + a messageId for a never-dispatched turn retries, appending a duplicate unanswered user row per attempt. The busy check needs to happen (or be surfaced) before the RPC resolves.

src/core/brains/pi-agent-brain.ts
```diff
@@ -135,6 +185,17 @@ export class PiAgentBrain implements BrainSession {
         const hasToolCalls = content.some((c: any) => c.type === "toolCall");
         if (hasText || hasToolCalls) lastAssistantHadContent = true;
       }
+      if (
+        event.type === "tool_execution_end" &&
+        typeof event.toolName === "string" &&
+        event.isError !== true &&
+        event.result?.details?.error === undefined &&
+        event.result?.details?.structuredContent !== null &&
```
jacoblee-io, line 193: Strict success requires `details.structuredContent`, which mcp-client sets only when the MCP server returns it (i.e. only for tools declaring an `outputSchema`). A side-effecting result tool that succeeds but returns text-only content is classified as missing, so the repair force-calls it a SECOND time every strict turn — e.g. a ticket/handoff tool duplicates the ticket per turn, and the repair still ends `success:false`. Nothing validates `outputSchema` at bind time and there's no distinct reason like `tool_succeeded_without_structured_content`, so "ran but text-only" and "never called" are indistinguishable to operators.

src/core/brains/pi-agent-brain.ts
```diff
@@ -207,11 +268,81 @@ export class PiAgentBrain implements BrainSession {
...
+
+      const requiredResultToolName = requirements?.requiredResultToolName?.trim();
+      if (
+        requiredResultToolName &&
+        !this.aborted &&
+        lastAssistantMessage?.stopReason !== "aborted" &&
+        lastAssistantMessage?.stopReason !== "error" &&
+        !successfulTools.has(requiredResultToolName)
+      ) {
+        await this.repairMissingRequiredResult(requiredResultToolName, successfulTools);
```
jacoblee-io, line 280: A transient provider failure during the repair leg propagates out of `brain.prompt()` (or arrives as a `stopReason:'error'` message_end) and is classified by `runAttempt` as the CANDIDATE's failure. If the main run answered in text without tool executions (`fallbackBlockedReason` unset), a repair-leg 5xx/timeout → checkpoint restore wipes the already-streamed answer, `model_route_rollback` tells consumers to drop what the user watched stream, and the entire original prompt re-runs on the fallback model. If the main run DID execute tools, the same blip instead rethrows as `prompt_error` — a failed turn over a materially delivered answer. Nothing distinguishes a repair-leg failure from a main-run failure at the routing boundary.

src/core/brains/pi-agent-brain.ts
```diff
@@ -207,11 +268,81 @@ export class PiAgentBrain implements BrainSession {
...
+        type: "required_result_repair_end",
+        toolName: requiredToolName,
+        success: false,
+        reason: "required_tool_inactive",
+      });
+      return;
+    }
+
+    const originalStreamFn = session.agent.streamFn;
+    let forceNextProviderRequest = true;
```
jacoblee-io, line 317: `forceNextProviderRequest` is consumed at request-issue time, so pi's in-turn transport retry (429/529/5xx re-invokes streamFn via `agent.continue`) and the follow-up request after a schema-rejected first tool call both go out UNFORCED. A forced repair request that dies on a provider 529 gets retried with the original (absent) toolChoice → model answers in plain text → `required_result_repair_end success:false`, despite the mechanism existing for exactly this case. Suggest: keep forcing while `!successfulTools.has(requiredToolName)` (the set is already passed in), and stop after success.

src/gateway/server.ts
```diff
@@ -731,6 +731,13 @@ export async function startRuntime(opts: StartRuntimeOptions): Promise<RuntimeSe
     // Runtime already persisted and sequenced the user row, so the target must
     // not bind or update a second local copy of it.
     const skipInitialPersistence = params.skipInitialPersistence === true && Boolean(delegation?.delegationId);
+    // Machine-facing strict callers cannot safely publish a reusable sessionId
+    // until its session and first user message are durable. Interactive chat
+    // keeps the legacy best-effort behavior; strict /run opts into this ACK.
+    const requireSessionPersistence = params.requireSessionPersistence === true;
+    const requiredResultToolName = typeof params.requiredResultToolName === "string"
```
jacoblee-io, line 738: Normalization disagrees with the agentbox side: http-server.ts trims and folds whitespace to `undefined`, while this keeps any raw string and silently coerces non-strings to `undefined`. A whitespace-only value (template bug) engages strict busy-rejection semantics at the gateway with ZERO required-result enforcement at the brain — all the strict costs, none of the guarantee. And a non-string (e.g. `{name:'...'}`) is silently dropped here, so the agentbox 400 never fires and the caller believes the contract is enforced while the turn runs unrestricted. Suggest applying the same trim-or-reject normalization here.

src/gateway/server.ts
```diff
@@ -886,6 +905,10 @@ export async function startRuntime(opts: StartRuntimeOptions): Promise<RuntimeSe
           releaseTurn = await sessionTurnLocks.acquire(sessionId);
         } catch (busyErr) {
           throwIfStoppedBeforePrompt();
+          // A strict machine request is one distinct turn with one distinct
+          // result. Folding it into the running turn as a steer would lose its
+          // required-result contract, so surface busy and let the caller retry.
+          if (requiredResultToolName) throw busyErr;
           const running = sessionTurnLocks.busyOn(sessionId);
           const steered = running ? await new AgentBoxClient(running.endpoint, 10000, agentBoxTlsOptions)
```
jacoblee-io, line 913: The strict/steer exclusion is one-directional: an ordinary concurrent `chat.send` (or `chat.steer`, which has no strict check at all) on the same sessionId is steered INTO a running strict turn's main run, since neither `sessionTurnLocks` nor the live-turn registry records that the running turn is strict. Session ids are one flat caller-chosen namespace (chat-gateway accepts `body.session_id`, and 'api'-origin sessions aren't trace-hidden), so a web/channel message to the same session folds foreign input into the strict turn — the end-of-turn check still forces the result tool, so the machine-consumed structured result is shaped by input from another principal, and the steering user's message is swallowed by a machine turn.

src/core/brains/pi-agent-brain.ts
```diff
@@ -207,11 +268,81 @@ export class PiAgentBrain implements BrainSession {
...
+    this.emit({ type: "required_result_repair_start", toolName: requiredToolName });
+    this.requiredResultRepairActive = true;
+    session.setActiveToolsByName([requiredToolName]);
+    try {
+      await session.prompt(PiAgentBrain.REQUIRED_RESULT_REPAIR_PROMPT);
+    } finally {
+      session.agent.streamFn = originalStreamFn;
+      session.setActiveToolsByName(activeToolNames);
+      this.requiredResultRepairActive = false;
+      this.emit({
```
jacoblee-io, line 338: A failed repair (including `required_tool_inactive` when the result MCP tool is unbound/down) leaves no durable or actionable failed outcome: `required_result_repair_end{success:false}` reaches only the live SSE stream, has no consumers, and is never persisted by sse-consumer — the turn still resolves with `/api/prompt` outcome=success, `model_route_success`, and a normal `prompt_done`. A control-plane caller that misses the one transient chat.event frame (reconnect, reload, or replaying from DB) records a successful support handoff for a turn whose required structured result was never submitted — despite the PR description's "surface missing/inactive result tools as explicit failed outcomes".

jacoblee-io (approved):
Reviewed with focus on the strict required-result contract. Approving to unblock, but there are several confirmed correctness issues worth follow-ups — 8 posted as inline comments, plus two that fall outside this PR's diff:

**1. `src/gateway/sse-consumer.ts:740` — repair's synthetic user echo flushes commit-gated rows mid-attempt.** The repair pass's synthetic `session.prompt()` emits a user-role `message_start` mid-attempt, and sse-consumer treats any user `message_start` as a turn boundary and calls `flushTerminalError()`, flushing `pendingAssistantOps`/`pendingErrorOps` to the DB while the routed attempt is still discardable. On a strict turn with a fallback candidate: the live primary produces a text-only draft (deferred assistant row — no tool_start commit point, which is precisely the repair trigger); the repair's user echo flushes the draft row; the repair then fails transiently → `model_route_rollback`, but `discardRoutedAttempt()` only clears in-memory arrays — the row is already written. Reload shows the rolled-back primary's draft stacked above the fallback's answer, defeating exactly what routed-turn commit gating exists to prevent.

**2. `src/agentbox/http-server.ts` `/api/steer` (~1564) — `RequiredResultRepairInProgress` flattened to a retriable 500.** `steer()` can now reject with `err.name='RequiredResultRepairInProgress'`, but the `/api/steer` route maps every rejection to HTTP 500 `{code:'INTERNAL_ERROR', retriable:true}`, and `chat.steer` persists the steer user row + `incrementMessageCount` BEFORE the box call with no rollback. A caller keying on the structured fields retries a steer that can never succeed, and each retry durably appends another `kind:'steer'` row that never reaches the model — ghost steer bubbles on reload while the input is silently lost.

Minor (no inline anchor): `supervisorEndedTurns`/`boxAbortAsked` map entries leak on the strict persistence-failure early-return path in server.ts (~872-875); `VALID_AGENT_TYPES` in sync-handlers.ts is hand-enumerated with no drift test against `AGENT_TYPES`; the strict contract has no design doc yet.

PR #541, opened 2026-09-01
feat(agent): redesign and inspect effective prompts

LikiosSedo:
## Summary

Redesign Siclaw prompts as an explicit, capability-aligned assembly instead of a long shared SRE prompt plus mutable role text. The same compiled Agent Context now produces the runtime harness, exact prompt layers, and an administrator-only resident-session inspection of the final prompt and actual model-visible tools.

### Problem

Knowledge QA and Coordinator sessions inherited irrelevant SRE, infrastructure, memory, and workflow guidance; a progress-only update could satisfy the common prompt; persisted `system_prompt` text could replace a built-in type's behavior; and normal manifests could not show the exact prompt/tool payload a provider received.

### Solution

- Introduce a lean role-neutral Platform Kernel with explicit completion semantics.
- Compile capability/mode guidance, immutable Agent Type contracts, optional Agent Addenda, and trailing platform safety as owned prompt layers.
- Keep exact historical built-in defaults as migration data while preserving real administrator-authored Addenda.
- Add `agent-context/v2` layer hashes and an explicit `prompt-inspection/v1` path from Portal to Runtime to the resident AgentBox.
- Show exact prompt layers, provider-visible tools, Skills, replica consistency, and deterministic `siclaw-prompt-design/v1` checks in the Portal Prompt tab.
- Keep exact prompt text on demand behind admin authentication and Runtime/AgentBox mTLS; routine status, telemetry, and logs remain hash-only.

Retrieval ranking, Embeddings, and Wiki search behavior remain in #535. The inspector intentionally reports `retrieval_below_reasoning: fail` while a deployed Knowledge QA tool contract mandates search before every answer; it turns `pass` when the runtime retrieval contract advertises search as an optional accelerator with Wiki exploration fallback.

## Test Plan

- [x] `npx tsc --noEmit` passes
- [x] `npm test` passes — 282 files, 6326 passed, 2 skipped
- [x] `npm --prefix portal-web test` passes — 23 files, 229 passed
- [x] `npm run build` passes
- [x] `npm --prefix portal-web run build` passes
- [ ] Manual verification in a deployed Portal/Runtime/AgentBox environment

## General Checklist

- [x] Changes are focused on a single logical change
- [x] Commit messages follow Conventional Commits
- [x] No unrelated changes included

PR #539, opened 2026-09-01
feat(knowledge): add complete-catalog label routing

LikiosSedo:
## Product outcome

Make compiled page labels a first-class, embedding-independent routing signal while keeping every Wiki route visible on the first Knowledge QA turn.

## Runtime contract

- inject the complete root `index.md`; remove the silent 8K head-only truncation
- use the complete catalog (title, path, one-line description) as the primary route
- keep `KnowledgeResolver` labels-only and in-memory: no FTS, vector DB, embedding client, or hydration wait
- use `knowledge_search` only to resolve typed labels and aliases when catalog routing is ambiguous
- return page metadata and match reasons, never body snippets or answer evidence
- require the Agent to read the complete supporting leaf pages before answering and citing

## Compiler contract

- add typed page-level `labels` to OKF frontmatter (`entity`, `topic`, `task`, `component`, `environment`, `version`)
- guide Siclaw KBC to emit 4–12 high-signal labels with alternate names in `aliases`
- validate label shape for generated pages and preserve labels through the existing OKF package
- require each generated `index.md` link entry to include the page's one-line description

## Compatibility

- existing and third-party OKF packages without labels remain usable through the complete root index
- labels remain navigation metadata; canonical facts and citations still come from leaf pages
- no DB schema, UI, model configuration, or embedding quota is required

## B300 regression

- a 185-line, 21K+ character catalog keeps its final B300/LSTM/cudagraph route in the system prompt
- B300 + LSTM + cudagraph labels rank the specific benchmark page ahead of a generic B300 page
- label results do not leak the measured `29.71 ms` body value; the Agent must read the leaf page

## Validation

- `npm test`: 281 files passed; 6311 tests passed, 2 skipped
- `npm run build`
- `npx tsc --noEmit`
- affected routing/session regression suite: 114 passed
- targeted KBC OKF, label, emitter, index-description, and locale-prompt assertions passed
- `git diff --check`

The full local asynchronous `compile_box` suite could not start because the local Python environment lacks `aiohttp`; KBC Box CI remains the authoritative full-suite check.

PR #542, opened 2026-09-01
feat(knowledge): unify label routing and prompt contracts

LikiosSedo:
## Product outcome

Keep the Karpathy-style LLM Wiki as the Knowledge QA reasoning and discovery system while making first-turn routing and the effective Agent prompt explicit and inspectable. This is the single `main`-target delivery for complete-catalog labels-only routing, Prompt/Harness redesign, usable Wiki paths, and leaf-only citations.

## Knowledge compiler and routing contract

- emit typed page labels (`entity`, `topic`, `task`, `component`, `environment`, and `version`) and aliases as compact navigation metadata
- inject the complete root `index.md`; do not silently truncate the catalog
- use titles, paths, descriptions, labels, and aliases as navigation signals
- keep `KnowledgeResolver` labels-only and in memory
- let `knowledge_search` resolve labels and aliases only when catalog routing is ambiguous
- never search page bodies or return answer evidence from the resolver
- read complete relevant leaf pages before answering and cite only leaf content actually used
- expose the actual mounted Wiki root and index paths to the model

## Prompt and harness contract

- use a lean, role-neutral Platform Kernel with explicit completion semantics
- add capability and mode guidance only when the corresponding tools exist
- compose immutable built-in Agent Type contracts with optional Agent Addenda
- keep trailing platform safety outside editable Addenda
- expose exact resident-session prompt and tool inspection through Portal → Runtime → AgentBox
- report layer ownership, provenance, hashes, provider-wire tools and schemas, Skills, and replica consistency
- mark exact prompt responses `Cache-Control: no-store`
- keep deterministic design checks aligned with complete-catalog, labels-only routing and leaf evidence

## Citation and compatibility

- reject root, index, catalog, and navigation pages as factual citation evidence
- preserve exact original-source citation mapping for leaf pages
- migrate exact historical materialized defaults, including the complete-catalog default, without treating them as administrator Addenda
- let existing and third-party packages without labels continue to route through the complete root catalog

## Explicit non-goals

This PR does not introduce a content-ranking retrieval path:

- no BM25, FTS, vector index, or embedding model
- no embedding hydration or per-turn embedding request
- no body-snippet or page-body retrieval
- no `direct_hit`
- no catalog truncation

Historical prompt constants may contain retired retrieval wording only so already-materialized rows can be recognized and migrated; they are not current runtime policy.

## Validation

- [x] `npm test` — 284 files, 6339 passed, 2 skipped
- [x] `npx tsc --noEmit`
- [x] `npm run build`
- [x] `npm --prefix portal-web test` — 23 files, 229 passed
- [x] `npm --prefix portal-web run build`
- [x] `git diff --check origin/main...HEAD`

## Delivery

- targets `main`
- supersedes #539 and #541
- #535 remains outside this delivery


jacoblee-io (approved):
Approving. Scope, stated up front: 53 files / +2533 is past what I can claim to have reviewed. I concentrated on `prompt-inspection.ts`, because it is the largest new file, a new data egress, and carried an "add it, then immediately fix its caching" commit pair — which is usually where something was noticed late. I did not review the labels resolver, the KBC emitter changes, the Portal UI, or the prompt-lifecycle doc rewrite in depth.

## Removing the retrieval path is a consistent ending, not a reversal

The non-goals section is the part I'd want a future reader to see first:

> no BM25, FTS, vector index, or embedding model … no `direct_hit` … no catalog truncation

The previous round's argument was that a retrieval accelerator must not become an authority, that candidates are hypotheses, and that a false direct hit is worse than more exploration. Taking that to its conclusion — drop content ranking entirely, keep compiler-emitted labels as navigation metadata, and let the Agent read leaf pages — is the same position followed through rather than a change of mind. Keeping the retired wording in historical constants purely so materialized rows can be recognized and migrated, with an explicit note that it is not current policy, is the right way to avoid a migration hazard without leaving misleading text as live documentation.

## The two exposure surfaces are deliberately different, and the code says so

This was my main question, since the envelope manifest reduces prompt content to `{chars, sha256}` on the stated grounds that prompt content stays private, and this file returns full text. The answer is in the comment:

> Agent Context manifest continues to contain hashes and layer metadata only.

So they're separate outputs with separate rules rather than a contradiction — the manifest stays hash-only, and inspection is a distinct admin-facing surface. Worth having that sentence there; it's the first thing a reader will ask.

The controls on the new surface fit it:

```js
// Explicit admin-only access to the exact resident-session prompt and tool
const auth = requireAdmin(req, res, jwtSecret);
…
res.setHeader("Cache-Control", "no-store");
```

`no-store` matters more here than the commit message suggests — the response body is exactly the kind of thing an intermediary or browser cache should never retain, and it arrived as a follow-up commit rather than with the feature.

## One note: the payload is broader than "the prompt"

The layer list includes:

```js
layer("runtime.composed_context", …, "pi ResourceLoader (profile, Wiki, citations, context files, Skills, cwd)", …)
```

That layer carries whatever ResourceLoader assembled — Wiki catalog content, context files, Skill text. Those are tenant data, not platform constants, so the response is not only "the effective prompt" in the configuration sense; it's a window onto the knowledge and skills mounted for that session.

Admin-only is the correct gate for that, and the code's own layer description is accurate. What's narrower than reality is the summary wording ("expose exact resident-session prompt and tool inspection"), which reads as prompt text plus tool schemas. That matters for one future decision in particular: if anyone later proposes relaxing this to an agent owner or a non-admin role, they'll evaluate it against the description, not against the layer list. A sentence at the route saying the response includes mounted knowledge and Skill content assembled into the prompt would put that fact where the next permission change gets reasoned about.

## Also worth noting

Keeping trailing platform safety outside editable Addenda preserves the invariant that matters most in that file — the composition model moved from "editable full prompt" to "immutable built-in contract plus optional addendum", which narrows what an administrator can displace rather than widening it. And rejecting root, index, catalog and navigation pages as citation evidence is right on its own terms: a catalog page contains links and descriptions, so citing one asserts provenance without carrying any fact.

CI green across all five checks.

PR #546, opened 2026-09-02
feat(knowledge): prove label routes through catalogs

LikiosSedo:
## Outcome

`knowledge_search` returns only root-catalog-reachable labeled leaves and includes a deterministic shortest `routeProof` from `index.md` to each result. Label routing stays optional: the complete Wiki catalog remains the fallback, and answer evidence still comes from reading the final leaf.

## Changes

- parse standard Markdown links and legacy wikilinks from navigation pages
- build one canonical shortest root-to-leaf catalog route during label-index sync
- exclude orphan labeled pages from search and label-catalog results, with an explicit count
- rank label matches by query coverage, unique matched terms, facet diversity, and matched term frequency across reachable pages
- preserve full confidence for rare exact labels while downranking broad shared aliases
- return compact search metadata by default; expose full labels only with `includeLabels=true`
- return `matchedPages` and `hasMore` before and after top-K truncation, plus a non-blocking weak or ambiguous match hint
- report invalid label declarations separately from unlabeled pages so an empty label index is diagnosable
- return a compact paginated label catalog by default; expose page paths only with `includePages=true`
- use conservative defaults (`topK=3`, label-catalog page size 20) while preserving explicit expansion up to existing limits
- add a fail-closed, manifest-driven tool for backfilling typed labels into frozen Wiki snapshots without changing page bodies
- document the reachability and evidence contract

## Verification

- `npm test` — 296 files, 6582 passed, 2 skipped
- focused knowledge tests — 4 files, 20 passed
- `npm run build`
- `git diff --check`
- synthetic multi-constraint queries rank the intended leaf above generic candidates
- shared high-frequency aliases are downranked and report result truncation
- invalid, unlabeled, and unreachable page counters are covered by regression tests

### DIFF
--- a/docs/design/agent-prompt-lifecycle.md
+++ b/docs/design/agent-prompt-lifecycle.md
@@ -81,10 +81,14 @@ responsibility.
 For Knowledge QA, the complete root `index.md` is the primary navigation map.
 `knowledge_search` resolves typed page labels and aliases only when catalog
 routing is ambiguous; it never searches page bodies or returns answer evidence.
-The Agent must Read the complete relevant leaf pages before answering and may
-cite only leaf content it actually used. This preserves semantic reasoning and
-the Wiki's linked navigation model without adding an embedding or content-index
-cold-start dependency.
+Each candidate includes one deterministic shortest `routeProof` from the root
+catalog to the leaf. The proof establishes that the candidate is published in
+the Wiki's navigation graph; it is not a claim that the Agent already traversed
+or read those pages. Intermediate catalogs do not need to be reread. The Agent
+must Read the complete relevant leaf pages before answering and may cite only
+leaf content it actually used. This preserves semantic reasoning and the Wiki's
+linked navigation model without adding an embedding or content-index cold-start
+dependency.
 
 ## Stored Addendum compatibility
 
--- a/docs/design/decisions.md
+++ b/docs/design/decisions.md
@@ -705,9 +705,13 @@ The complete root `index.md` is injected into the model prompt as the common
 navigation baseline. AgentBox also derives an in-memory, paginated Label Catalog
 from page frontmatter before the first turn. `knowledge_search` resolves typed
 labels and aliases only; it never searches page bodies, creates a database, or
-opens an embedding provider. It returns page metadata, matched labels, and alias
-reasons, never snippets. Labels and catalog entries never count as answer
-evidence; the Agent must still read and cite supporting pages.
+opens an embedding provider. It returns page metadata, matched labels, alias
+reasons, and one deterministic shortest route from the root catalog to each
+leaf, never snippets. A labeled page outside the root catalog graph is excluded
+and reported as unreachable. The route proves publication and reachability; it
+does not claim that the Agent read every intermediate catalog. Labels and
+catalog entries never count as answer evidence; the Agent reads and cites only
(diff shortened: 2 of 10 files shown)


jacoblee-io (approved):
Approving. I focused on the two places where this design can lose information silently: whether `routeProof` is actually deterministic, and what the backfill tool is allowed to touch. Both hold up. I did not review the ranking weights in depth — the reported separations (0.992 vs 0.459, 0.855 vs 0.517) are convincing as evidence, but ranking quality isn't something I can verify by reading.

## `routeProof` determinism has implementation backing

Two properties, each with a mechanism rather than an intention:

```js
return links.sort((a, b) => a.file.localeCompare(b.file) || a.text.localeCompare(b.text));
…
const queue = [rootCatalog];
for (let cursor = 0; cursor < queue.length; cursor++) {
```

An array-plus-cursor queue is BFS, so first arrival is shortest; the two-key sort fixes the tie-break when several equal-length routes exist. Without the sort, "shortest" would still be true and "canonical" would quietly depend on directory iteration order — the kind of thing that produces a different proof on a different filesystem and looks like a flake.

## The backfill tool refuses the dangerous shapes

Seven separate throws, and the two that matter most aren't about labels at all:

```js
if (…) throw new Error("Output directory must be outside the source directory");
if (fs.existsSync(output)) throw new Error(`Output already exists: ${output}`);
```

It's a source→output transform that cannot write in place and cannot overwrite, so "frozen snapshot" stays frozen by construction rather than by care. Insertion is at the frontmatter end offset with `markdown.slice(endOffset)` carried through verbatim, which is what makes "without changing page bodies" a byte-level claim rather than an intent.

`Page already declares labels; refusing to overwrite` gives it idempotence, and the round-trip check is the one I'd have asked for if it were missing:

```js
if (!parsed || parsed.labels.length !== labels.length) {
  throw new Error(`Generated labels failed runtime validation: ${file}`);
}
```

Emitting YAML and then reading it back with the *runtime* parser closes the failure this line of work has hit before — a producer and a consumer disagreeing about the same on-page format, where the producer looks successful and the consumer sees nothing. Validating against the real parser rather than the generator's own assumptions is what makes that impossible here.

Also worth noting: reporting invalid label declarations separately from unlabeled pages. "No labels" has two very different causes, and collapsing them is what makes an empty label index undiagnosable.

## One thing to confirm: how long unreachability stays visible

Reachability exclusion is the only mechanism here that makes a page vanish from retrieval entirely, and its trigger is mundane — a link in `index.md` whose case doesn't match the file (case-sensitive filesystem in the container), or a link form the parser doesn't cover. The migration report showing zero unreachable pages is good evidence for today's corpus, and excluding them with an explicit count is the right behaviour rather than hiding it.

What I couldn't establish is whether that count rides on every search and label-catalog response, or only appears at sync/migration time. It matters because the trigger recurs: every Wiki recompile, and every hand edit to a catalog page, can introduce a broken link. If the count is in the response, an operator notices "five labeled pages are unreachable" while debugging a missing answer. If it only surfaces at migration, a link that breaks afterwards produces a page that is simply never returned, with the labels still present and valid — the hardest shape to diagnose, because every layer reports success.

If it's already per-response, disregard this. If not, surfacing it alongside `matchedPages` / `hasMore` would put it where someone looking for a missing page will actually be looking.

PR #548, opened 2026-09-02
fix(portal/metrics): count prompts by kind, not by matching the metadata column

chent1996:
## Why

The prompt figure excluded only `delegation_event`, so `task_event` — the agent's own to-do ledger, and by far the largest category — was counted as questions people asked. Over 30 days of production: **19,502 ledger rows against 7,512 real questions**.

Those rows carry `role='user'` only because that is the channel the model reads from: the ledger it maintains while working, status reports from delegated children, the notification that wakes it when a background job finishes. None appear in a transcript, and none are questions.

## What changed

- **`src/shared/message-kinds.ts`** — the vocabulary: every kind this repo persists to `metadata.kind`, and the subset that is plumbing. Follows the pattern `session-origin.ts` established after `'subagent'` shipped and eight hardcoded origin predicates failed to learn about it; the same shape had formed here, with four call sites each naming a single kind by hand.
- **`src/portal/human-prompt.ts`** — the predicate. All four call sites go through it: the v1 summary, its daily series, the internal summary, and the `metrics.summary` RPC.
- **Reads `metadata.kind` through JSON.** The old predicate was `metadata NOT LIKE '%"kind":"delegation_event"%'`, which depended on the exact serialized shape (no space after the colon) and matched anywhere in the column, so a row merely mentioning a kind name in an unrelated field was dropped.
- **Restated the basis in both places it is spelled out** — the Prompts card hint and the daily `TrendChart` subtitle, which both read "user messages".

## The JSON read has to be dialect-aware

Not a refinement — either half alone is broken:

```
MySQL :  CASE WHEN JSON_VALID(col) THEN JSON_UNQUOTE(JSON_EXTRACT(col, '$.kind')) END
SQLite:  CASE WHEN JSON_VALID(col) THEN json_extract(col, '$.kind')               END
```

- MySQL's `JSON_EXTRACT` returns a **quoted JSON string** (`"task_event"`), so without `JSON_UNQUOTE` the comparison never matches and the filter silently degrades to a no-op.
- SQLite has **no `JSON_UNQUOTE` at all**. A MySQL-only predicate 500s every metrics endpoint under `siclaw local`.
- The `JSON_VALID` guard is load-bearing on **both** sides: given a column that is not valid JSON, MySQL raises `ER_INVALID_JSON_TEXT` and SQLite raises `malformed JSON` — neither returns NULL, so one bad row would fail the whole query.

It goes through a new `jsonScalarOrNull()` in `dialect-helpers.ts`, the repo's single outlet for dialect differences. That is why the predicate lives in `portal/` (it needs `db.driver`) while the vocabulary stays in `shared/`, which the AgentBox image bundles and which must not reach into the gateway.

**Asserting on the SQL string is not a test for this.** `human-prompt.test.ts` builds a real `node:sqlite` table and executes the predicate against it.

## Registering a new kind is a compile error, with one known hole

`ChatMessageMetadata` narrows `metadata.kind` to the registry, and the six write-path payloads carry that type. A comment is exactly what `session-origin.ts` had when `subagent` shipped past it. Narrowing surfaced **three kinds no author had enumerated**: `steer`, `error_response`, `model_route_notice`.

`steer` is deliberately **not** plumbing — a person typing while a turn runs. The tag says *when* the message was sent, not who wrote it; counting it as synthetic would undercount exactly the users who interact most.

The hole is worth knowing because it is not where you would guess: TypeScript rejects an unregistered kind in an **inline literal**, but a value arriving as `Record<string, unknown>` is assignable to the narrowed type with **no error** — a source index signature need not satisfy a target's declared optional property. So a helper returning an open record slips past, which is how `model_route_notice` ran a release unregistered while the type was in place and the suite was green. No shape of the type fixes it (dropping the index signature rejects every literal carrying a payload key, i.e. all of them), so the two root metadata builders in `sse-consumer.ts` are typed at their own declarations. `message-kind-invariants.test.ts` pins those and the six interfaces, since widening any of them back breaks nothing else.

## No cross-repo parity is claimed

Earlier revisions of this PR asserted this list was literally equal to sicore's `chatfields.syntheticUserKinds`. **The file, the symbol and the reciprocal comment all do not exist** — verified against sicore's current main. What sicore actually has:

| Location | Question it answers | State |
|---|---|---|
| `metrics/handler.go` totalPrompts | prompt total | **no kind filter at all** |
| `adapter/rpc.go` metrics.summary | same | same |
| `metrics/trace_kinds.go` `TraceNonPromptUserKinds` | which user row may title a trace / feed analysis | 8 kinds, **includes `steer`** |
| `chat/service.go` `previousUserMessageSnapshot` | which prompt an assistant answered (feedback) | 3-kind LIKE, missing `exec_job_event` |

So sicore's Prompts figure is inflated by *more* than this one was — a separate problem, recorded in `docs/design/prompt-basis.md`, not addressed here. `trace_kinds.go` answers an adjacent question, which is why `steer` belongs in that set and must stay out of this one. Copying either list into the other would be wrong in both directions.

## Verification

- `npx tsc --noEmit` — clean.
- `npm test` — 298 files, **6601 passed**, 2 skipped.
- `portal-web` — 25 files, **246 passed**.
- Both enforcement mechanisms were checked by breaking them: unregistering `model_route_notice` fails the build at both `sse-consumer.ts` sites; widening a builder back fails `message-kind-invariants.test.ts`.

## ⚠️ This makes the Prompts figure drop substantially

The number is becoming correct, not the usage falling. Please say so when releasing: a counting basis that changes silently leaves readers to discover it by sensing the figure looks wrong, and by then they have been trusting a wrong number for a while.


src/shared/message-kinds.ts
```diff
@@ -0,0 +1,95 @@
...
+ * their own copy — the values are written by this runtime and read by both — so
+ * a change on either side needs the other updated in step. There is no shared
+ * contract to lean on, and inventing one that only this side can see would be
+ * worse than the honest duplication.
+ *
+ * The SQL that consumes this list lives in `portal/human-prompt.ts`: it needs
+ * the database driver to pick a dialect, and `src/shared` is bundled into the
+ * AgentBox image and must not reach into the gateway.
+ */
+export const SYNTHETIC_USER_KINDS = [
```
jacoblee-io, line 107: medium: the cross-repo parity claim is false at merge time.

This comment and `prompt-basis.md` both assert that this list is literally equal to the counterpart registry in the other repository, and `message-kinds.test.ts` pins it as "the agreed set". It is not equal: the counterpart list carries three kinds and does not include `exec_job_event`.

Scenario: the two dashboards report different Prompts totals over the same rows, and the next person told to "change both sides in step" reads a comment saying they already match and skips the diff.

Fix: add the missing kind on the other side in step, or drop the equality claim here and state which side is authoritative and by which kinds they differ.

src/shared/message-kinds.ts
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * Metadata persisted alongside a `chat_messages` row.
+ *
+ * Open on every other key — metadata carries per-kind payloads, model-routing
+ * detail and channel bookkeeping, and typing all of that here would only move
+ * the guesswork. `kind` is the one key with a cross-cutting meaning: it decides
+ * whether the row counts as a human prompt, so it is narrowed to the registry
+ * above and a literal outside it is a compile error at the write site.
+ */
+export type ChatMessageMetadata = {
```
jacoblee-io, line 125: medium: the "adding a kind is a compile error" mechanism does not hold for the write pattern already in tree.

TypeScript does not use a source type's string index signature to satisfy a target's declared optional property, so `Record<string, unknown>` is assignable to `ChatMessageMetadata` with no error (verified against this repo's tsc). `src/gateway/sse-consumer.ts:193` and `:260` build metadata in a helper typed `Record<string, unknown> | null` and pass it to `appendRow` — which is exactly why `kind: "model_route_notice"` compiles today despite never being registered. Only inline object literals are actually checked, which is why `error_response` was caught and this was not.

Scenario: the next notice row is added the same way — a helper returning `Record<string, unknown>`, this time on a `role: "user"` append. It compiles, no test fails, and the Prompts figure silently over-counts again: the exact regression this PR exists to fix.

Fix: type the metadata builders in `sse-consumer.ts` as `ChatMessageMetadata` (then `model_route_notice` becomes the compile error the design intends), and pin that in `message-kind-invariants.test.ts` alongside the four interfaces.

src/shared/message-kinds.ts
```diff
@@ -0,0 +1,95 @@
...
+ */
+
+/**
+ * Every value that may appear as `chat_messages.metadata.kind`.
+ *
+ * Not every kind is about prompts — `error_response` rides on an assistant row
+ * — so this list is broader than {@link SYNTHETIC_USER_KINDS} by design. Its
+ * job is to be the place a new kind has to pass through.
+ */
+export const CHAT_MESSAGE_KINDS = [
```
jacoblee-io, line 55: low: `CHAT_MESSAGE_KINDS` is documented as exhaustive but omits a kind that is live in production data.

The header says "Every value that may appear as `chat_messages.metadata.kind`", yet `model_route_notice` is written to that column at `src/gateway/sse-consumer.ts:204` and `:271` (history confirms those and `error_response` are the only kinds ever added there). `message-kinds.test.ts` then pins the incomplete list as "the registered set".

Scenario: a maintainer asking "is this kind a human prompt?" consults the registry that is meant to be the single answer and finds no entry for a kind that exists in the DB — the same discovery-by-absence failure this file was written to prevent. It is harmless for counting today only because the row is `role='assistant'`.

Fix: register `model_route_notice` with a comment noting it rides an assistant row (like `error_response`), and update the pinned list in `message-kinds.test.ts`.

src/portal/message-kind-invariants.test.ts
```diff
@@ -0,0 +1,56 @@
...
+import { describe, it, expect } from "vitest";
+import fs from "node:fs";
+import path from "node:path";
+import { fileURLToPath } from "node:url";
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+const SRC_ROOT = path.resolve(__dirname, "..");
+
+/** Every interface carrying the metadata of a row headed for `chat_messages`. */
+const WRITE_PATH = [
```
jacoblee-io, line 32: low: `WRITE_PATH` omits two interfaces this PR itself narrowed.

The comment claims "Every interface carrying the metadata of a row headed for `chat_messages`", but `UpdateDelegationToolMessageInput` (`src/gateway/chat-repo.ts:70`) and `DelegationToolUpdatePayload` (`src/shared/delegation-persistence.ts:46`) are both narrowed to `ChatMessageMetadata` in this diff and are not listed.

Scenario: either one is widened back to `Record<string, unknown>` and the invariants test — the file whose whole purpose is to catch that widening — stays green.

Fix: add both entries to `WRITE_PATH`.

portal-web/src/components/metrics/KpiCards.tsx
```diff
@@ -39,7 +39,7 @@ export function KpiCards(p: Props) {
         <div className="grid grid-cols-2 sm:grid-cols-3 lg:grid-cols-5 gap-4">
           <KpiCard label="Users" value={fmt(p.distinctUsers)} hint="distinct" icon={<Users className="h-3.5 w-3.5" style={{ color: "#60a5fa" }} />} accent="#60a5fa" />
           <KpiCard label="Sessions" value={fmt(p.totalSessions)} hint="conversations" icon={<Activity className="h-3.5 w-3.5" style={{ color: "#fbbf24" }} />} accent="#fbbf24" />
-          <KpiCard label="Prompts" value={fmt(p.totalPrompts)} hint="user messages" icon={<MessageSquare className="h-3.5 w-3.5" style={{ color: "#34d399" }} />} accent="#34d399" />
+          <KpiCard label="Prompts" value={fmt(p.totalPrompts)} hint="human prompts" icon={<MessageSquare className="h-3.5 w-3.5" style={{ color: "#34d399" }} />} accent="#34d399" />
```
jacoblee-io, line 42: low: the daily-trend chart's subtitle was missed — the line I mean is `portal-web/src/pages/Metrics.tsx:173`, which is not in this diff (anchored here because this is the sibling label that WAS updated).

`Metrics.tsx:173` still reads `subtitle="user messages per day"` for the Prompts `TrendChart`, which is fed by the same `dailyPrompts` query this PR changed.

Scenario: the chart plots human prompts only while claiming to plot user messages, so a reader reconciling it against a raw `role='user'` count over `chat_messages` sees a large unexplained gap — the silent basis change the PR's own release note warns about, in the one place the basis is still spelled out the old way.

Fix: change that subtitle to "human prompts per day" to match this card's hint.

jacoblee-io (commented):
Reviewed the diff, ran `tsc --noEmit` (clean) and the full `vitest` suite on a clean checkout of this branch: 298 files / 6597 tests, all passing. Five findings inline — no functional SQL bug found (the dialect split, the `JSON_VALID` guard, the `COALESCE` NULL handling and the parameter-free interpolation all check out, and the SQLite branch is genuinely executed). The findings are about the enforcement mechanism built around the fix, plus one parity claim that is false as written.

One note on the description: it says `adapter-rpc.test.ts` and `siclaw-api.misc.test.ts` cannot run locally because `@earendil-works/pi-ai` is missing its `./providers/anthropic.models` export. That did not reproduce — both suites ran and passed here (176 and 60 tests) on a clean checkout of this PR, as did the whole suite. Worth dropping that caveat so the next reader does not skip running them.

jacoblee-io (commented):
## Re-review after the force-push (`f373eef0` → `3f7effb8`)

Verified on a clean checkout of `3f7effb8`: `npx tsc --noEmit` clean, `npm test` green (298 files, 6601 passed, 2 skipped). No functional SQL bug — the dialect split, the `JSON_VALID` guard on both sides, `COALESCE` NULL handling, MySQL collation coercibility of `JSON_UNQUOTE` against a literal, and the parameter-free interpolation all check out; the SQLite branch is genuinely executed by `human-prompt.test.ts`; all four prompt call sites are converted and there is no fifth.

### Status of my five earlier findings

| # | Topic | Status |
|---|---|---|
| 2 | metadata builders typed so the compile-error contract holds | **fixed** |
| 3 | `model_route_notice` registered | **fixed** |
| 4 | `WRITE_PATH` completeness | **fixed** |
| 5 | `TrendChart` subtitle | **fixed** |
| 1 | cross-repo parity claim | **not fixed — now false in the opposite direction** |

The four mechanism findings are fixed well: both root builders declare `ChatMessageMetadata | null` with inline literals, so unregistering `model_route_notice` is now a compile error at both sites, and `ROOT_METADATA_BUILDERS` pins that by name. The `Record<string, unknown>` hole is now documented accurately in the header, which is the right outcome — the mechanism's real boundary is stated rather than overclaimed.

**Finding 1 is the one still open, and it moved rather than closed.** The earlier revision claimed this list was identical to its counterpart; this revision claims the counterpart does not exist at all ("**There is NO sicore counterpart to `SYNTHETIC_USER_KINDS`**", and specifically that `chatfields/message_kind.go` is a file that does not exist). I checked the other repository's current `main` again: that file **does exist** — it was added on 2026-08-24 — and it defines a synthetic-kind set together with a human-prompt predicate and a metadata-kind expression, using the same dialect split and the same `JSON_VALID` guard as this PR. It is applied in the prompt-count query, the daily series, and three call sites in the platform-usage module. The directory the doc describes as holding one file holds four. Separately, `metrics.summary` has been removed from that repo's RPC surface, so the doc's table row points at a call site that is no longer there.

The actual divergence is what my original finding said: **that set carries three kinds and does not include `exec_job_event`**, so the two dashboards report different Prompts totals over the same rows. What makes this worse than the first version is reach — the claim is now asserted in four places (this file's header, `docs/design/prompt-basis.md`, `CLAUDE.md`, and a test comment), and `CLAUDE.md` makes the doc mandatory reading before touching these files. A future maintainer is now told in four places that there is nothing to sync.

### Verdict

Not approving on that basis alone — everything else here is good, and the counting fix itself is correct. What is needed is to replace the "no counterpart" claim with the real divergence and say which side is authoritative; the code needs no change. Five smaller findings inline, all low severity.

src/shared/message-kinds.ts
```diff
@@ -0,0 +1,128 @@
...
+  "task_notification",
+] as const;
+
+export type ChatMessageKind = (typeof CHAT_MESSAGE_KINDS)[number];
+
+/**
+ * The `role='user'` rows that are workflow plumbing rather than a person's
+ * question — i.e. the rows a prompt count must exclude.
+ *
+ * **This list has no counterpart in sicore, and earlier revisions of this
```
jacoblee-io, line 86: medium: the "no counterpart" claim is not correct, and it is now asserted in four places.

The counterpart file this paragraph names as non-existent does exist on the other repository's current `main` (added 2026-08-24), and it defines a synthetic-kind set plus a human-prompt predicate and metadata-kind expression — same dialect split, same `JSON_VALID` guard as here. It is applied in the prompt-count query, the daily series, and three call sites in the platform-usage module; the directory described as holding a single file holds four. `metrics.summary` has also been removed from that repo's RPC surface, so the doc's table row names a call site that no longer exists.

The real divergence is the one from the earlier review: that set carries three kinds and does not include `exec_job_event`, so the two dashboards report different Prompts totals over the same rows.

Scenario: the divergence becomes unreachable by reading. The same claim appears here, in `docs/design/prompt-basis.md:149`, in `CLAUDE.md:94`, and in `message-kinds.test.ts` — and `CLAUDE.md` now makes that doc mandatory reading before touching this file, so the next maintainer is told four times that there is nothing to sync. `prompt-basis.md:158` also contradicts its own section by citing "sicore 在 2026-08-24 经历的是同一件事" — that date is the commit that added the file the section says does not exist.

Fix: state the real divergence (the missing `exec_job_event`) and which side is authoritative, in all four places. No code change needed.

src/shared/message-kinds.ts
```diff
@@ -0,0 +1,128 @@
...
+ * hand, all four naming a single kind, so the largest category by far —
+ * `task_event` — was counted as a prompt everywhere.
+ *
+ * **Adding a kind means adding it HERE**, and the metadata type below is what
+ * makes that more than a request: it narrows `kind` to
+ * {@link CHAT_MESSAGE_KINDS}, so an unregistered value fails to compile, and
+ * registering it means arriving at this file — where the next question, whether
+ * a human asked it, is the one being asked. The rule was a comment first, and a
+ * comment is exactly what the origin predicates had; it did not survive
+ * `subagent` there. Three of the seven values below were found by the compiler
```
jacoblee-io, line 24: low: this header and the design doc disagree on what the compiler actually caught.

Here: "Three of the seven values below were found by the compiler". `docs/design/prompt-basis.md:101`: two (`steer`, `error_response`). Two is right — `model_route_notice` went through the `Record<string, unknown>` hole documented three lines below and was found by review, not by `tsc`.

Scenario: a reader concludes the compiler catches helper-built kinds, which is exactly the belief the paragraph below exists to correct.

Fix: say two, and name `model_route_notice` as the one the hole let through.

src/portal/message-kind-invariants.test.ts
```diff
@@ -0,0 +1,91 @@
...
+  { file: "gateway/chat-repo.ts", iface: "AppendMessageInput" },
+  { file: "gateway/chat-repo.ts", iface: "UpdateMessageInput" },
+  { file: "gateway/chat-repo.ts", iface: "UpdateDelegationToolMessageInput" },
+  { file: "shared/delegation-persistence.ts", iface: "DelegationAppendMessagePayload" },
+  { file: "shared/delegation-persistence.ts", iface: "DelegationUpdateMessagePayload" },
+  { file: "shared/delegation-persistence.ts", iface: "DelegationToolUpdatePayload" },
+];
+
+/**
+ * Functions whose return value is passed as the `metadata:` ROOT of an append.
```
jacoblee-io, line 42: low: `ROOT_METADATA_BUILDERS`'s stated rule is broader than its contents.

The rule is "functions whose return value is passed as the `metadata:` ROOT of an append". `attachModelRouteMetadata` (`src/gateway/sse-consumer.ts:298`) matches it exactly — its result is the `metadata:` root at `:683` and `:824` — and returns `Record<string, unknown> | null`. So do `extractPersistableDetails`, which feeds it, and the `stoppedMeta` local at `:1074`.

Scenario: a future notice kind is stamped inside `attachModelRouteMetadata`. It compiles, no test fails, and the kind never reaches the registry — the hole this mechanism exists to close. Blast radius is smaller today because those roots land on `role='tool'` rows, so the Prompts figure is unaffected, but the invariant is weaker than its comment claims.

Fix: either type those builders `ChatMessageMetadata | null` and list them, or narrow the comment to say the list covers only builders on `role='user'`/`role='assistant'` paths and why.

src/gateway/dialect-helpers.ts
```diff
@@ -122,6 +123,38 @@ export function jsonArrayFlattenSql(
...
+ * optimisation: handed a column that is not valid JSON, MySQL raises
+ * ER_INVALID_JSON_TEXT and SQLite raises `malformed JSON`. Neither returns
+ * NULL, so without the guard one unparseable row fails the whole query.
+ *
+ * A missing column, unparseable content, and an absent path all collapse to
+ * NULL — the caller decides what that means.
+ *
+ * `column` and `path` are interpolated: pass literals, never request input.
+ */
+export function jsonScalarOrNull(db: Db, column: string, path: string): string {
```
jacoblee-io, line 150: low: the new fourth dialect outlet has no case in `dialect-helpers.test.ts`, and `CLAUDE.md` was not updated for it.

`CLAUDE.md:98` — the row the impact matrix tells you to read before touching this file — still enumerates only "All 4 dialect differences (upsert, INSERT IGNORE, JSON ops, `safeParseJson`)", so it now disagrees with this file's own updated header. And the MySQL branch of `jsonScalarOrNull` is asserted only as a substring inside `human-prompt.test.ts` — the class of test this PR's own design doc argues is insufficient ("A string assertion is not a test here").

Scenario: the next person needing a JSON scalar read hand-rolls a fifth `JSON_UNQUOTE`, because the documented registry of dialect outlets does not list this one.

Fix: add a `jsonScalarOrNull` case to `dialect-helpers.test.ts` covering both drivers, and update the `CLAUDE.md` row to name the outlet.

src/portal/adapter-rpc.test.ts
```diff
@@ -15,7 +17,10 @@ function mockQuery(...results: any[][]) {
   for (const rows of results) {
     query.mockResolvedValueOnce([rows, []]);
   }
-  (getDb as any).mockReturnValue({ query, getConnection: vi.fn() });
+  // driver is explicit: adapter.ts builds dialect-aware SQL, and a mock
```
jacoblee-io, line 20: low: pinning `driver: "mysql"` on the shared `mockQuery` drops SQLite-branch coverage for this whole file, and the stated reason does not hold.

155 `mockQuery` call sites previously ran with `driver` undefined — i.e. through the SQLite branch of every helper they touch. They all take the MySQL branch now. The comment says these mirrors "only ever run against MySQL in production", but `registerAdapterRoutes` / `buildAdapterRpcHandlers` are exactly what `siclaw local` serves over SQLite; `insertIgnorePrefix(db)` at `src/portal/adapter.ts:210` and `:528` runs there.

Scenario: nothing exercises `adapter.ts`'s SQLite SQL any more — in the very mode whose 500s this PR cites as the motivating regression, and which the design doc says a shape-only assertion failed to catch the first time.

Fix: scope the `mysql` driver to the metrics describe block that needs it, or parameterise the file over both drivers.

docs/design/prompt-basis.md
```diff
@@ -0,0 +1,159 @@
+# 提问口径：哪些 user 行算是人在提问
```
jacoblee-io, line 1: low: this is the only Chinese-language document in `docs/design/`.

Every other doc there is English, as is `CLAUDE.md` — which this PR updates to make this file mandatory reading before touching `message-kinds.ts` / `human-prompt.ts`.

Scenario: a contributor who reads the English impact-matrix row is sent to a document they cannot read, for a contract the matrix says they must understand first.

Fix: translate to English to match the rest of the directory, or note the language in the `CLAUDE.md` row.

src/portal/message-kind-invariants.test.ts
```diff
@@ -0,0 +1,91 @@
...
+ *
+ * It exists because narrowing the interfaces above is not enough on its own.
+ * TypeScript checks an unregistered kind in an inline literal, but a value
+ * arriving as `Record<string, unknown>` is assignable to `ChatMessageMetadata`
+ * with no error — the source's index signature need not satisfy the target's
+ * declared optional property. Both builders below were written that way, which
+ * is how `model_route_notice` reached production unregistered while the type
+ * was in place and every test was green.
+ */
+const ROOT_METADATA_BUILDERS = [
```
jacoblee-io, line 58: `attachModelRouteMetadata` (sse-consumer.ts:298) belongs on this list by the list's own inclusion rule: it returns `Record<string, unknown> | null` and its result IS the metadata root of persisted rows (via `metadataWithRoute` ~683 and `startMetadataWithRoute` ~824). Worse, its input comes from `extractPersistableDetails` (~138), which copies every tool-authored `details` key except blocked/error — so a tool emitting `details.kind = "delegation_event"` (plausible; delegation tools already emit `details.delegation_id`, read a few lines below) persists an unregistered/colliding kind to `chat_messages` through the exact Record-assignability hole this PR documents, while this test stays green. Same failure mode that shipped `model_route_notice`, one function over from the two builders being pinned.

src/portal/message-kind-invariants.test.ts
```diff
@@ -0,0 +1,91 @@
...
+  { file: "gateway/sse-consumer.ts", fn: "modelRouteSwitchMetadata" },
+  { file: "gateway/sse-consumer.ts", fn: "modelRouteRecoveryMetadata" },
+];
+
+function readInterfaceBody(file: string, iface: string): string {
```
jacoblee-io, line 63: The hand-rolled source parsing is fragile in three ways: (a) `readInterfaceBody` terminates at the first `"\n}"` after the opener — a column-0 close brace in a comment/template inside the interface region slices wrong, and the exact `export interface X {` opener breaks on an `extends` clause; (b) the builder regex matches the FIRST `function <name>(` occurrence anywhere, including comment text — a docstring quoting the signature could keep the test green after the real declaration is widened; (c) the negative `not.toContain("Record<string, unknown>")` on the captured return type is unreachable while the positive passes. Net effect: it can fail open (JSDoc example satisfies it) or fail closed on an intact contract (arrow-function reformat reads as "was it renamed?"), which trains people to edit the parser instead of the contract. Simpler: anchored multiline `^function <name>` and an exact property-line assertion.

src/portal/adapter.ts
```diff
@@ -1969,7 +1970,7 @@ export function registerAdapterRoutes(router: RestRouter, internalSecret: string
       JOIN chat_sessions s ON m.session_id = s.id
       WHERE m.role = 'user' AND m.created_at >= ?
         AND ${nonTraceOriginPredicate("s")}
-        AND (m.metadata IS NULL OR m.metadata NOT LIKE '%"kind":"delegation_event"%')`;
+        AND ${humanPromptPredicate(db, "m")}`;
```
jacoblee-io, line 1973: Only 2 of the 4 migrated call sites are pinned by tests: this HTTP mirror (`GET /api/internal/siclaw/metrics/summary`) and siclaw-api's dailyPrompts (~3838) have no assertion that `humanPromptPredicate` is in their SQL — adapter.routes.test.ts exercises only the invalid-period 400 path, and its `getDb` mock still lacks the `driver` field the two sibling test files just added (with a comment explaining exactly this blindness). A merge conflict or partial revert dropping `AND ${humanPromptPredicate(db, "m")}` from the mirror or the daily series alone stays green, and the HTTP summary silently disagrees with RPC `metrics.summary` (or the trend chart with the KPI card) — the exact mirror-drift mode the session-origin history documents.

src/gateway/dialect-helpers.ts
jacoblee-io, line 150: `jsonScalarOrNull` — the new 4th dialect difference — has zero coverage in dialect-helpers.test.ts, which is the file the CLAUDE.md matrix row names as the mandatory verify step for this module; and that row wasn't updated (it still enumerates "All 4 dialect differences (upsert, INSERT IGNORE, JSON ops, safeParseJson)" while the PR's new message-kinds row calls this helper "the 4th") — two rows in the same file now enumerate the set differently. A future editor of this function runs the row's named verification and it proves nothing about the helper; a reader of the stale row concludes JSON scalar reads aren't routed through this file and hand-rolls a fifth spelling elsewhere.

src/gateway/dialect-helpers.ts
```diff
@@ -122,6 +123,38 @@ export function jsonArrayFlattenSql(
...
+  const extract =
+    db.driver === "mysql"
+      ? `JSON_UNQUOTE(JSON_EXTRACT(${column}, '${path}'))`
```
jacoblee-io, line 153: Hardening nit: `column` and `path` are interpolated bare / inside single quotes with no escaping or validation, guarded only by the "pass literals, never request input" comment — but unlike the registry-fed sibling helpers, this is an exported general-purpose `(db, column, path)` helper, so nothing machine-checks the guard. A derived path containing an apostrophe (a legal JSON key: `$."o'brien"`) is a syntax error on both dialects; request-derived input is direct injection. All current call sites are literals, so not a live bug — but a one-line throw on quote/backslash characters would fail closed, matching the repo's stated preference for machine-readable refusals over prose.

src/portal/human-prompt.ts
```diff
@@ -0,0 +1,45 @@
...
+ * Rows with no metadata, unparseable metadata, or no `kind` at all are KEPT:
+ * a plain question carries no kind, and that is the common case, not an edge
+ * one. They reach the comparison as NULL, and COALESCE — rather than a second
+ * copy of the extraction under an `IS NULL` arm — is what lets the JSON be
+ * parsed once per row instead of twice.
+ */
+export function humanPromptPredicate(db: Db, alias: "m" | "msg" | "" = "m"): string {
+  const column = alias ? `${alias}.metadata` : "metadata";
+  const kind = jsonScalarOrNull(db, column, "$.kind");
+  const list = SYNTHETIC_USER_KINDS.map((k) => `'${k}'`).join(", ");
```
jacoblee-io, line 43: Nit: this re-implements `sqlList()` from `session-origin.ts:51` — the very file this module's header names as its companion. `sqlList` is module-private; exporting (or lifting) and reusing it avoids two hand-rolled SQL-literal quoters one directory apart, where any future hardening (escaping, empty-list producing an invalid `IN ()`) applied to one copy silently misses the other — the one-rule-two-spellings drift both registries were created to end.

src/shared/message-kinds.ts
jacoblee-io, line 24: Factual contradiction across this PR's own authoritative sources: this comment and the new CLAUDE.md row say the compiler surfaced THREE unregistered kinds, while `message-kinds.test.ts:12` says two and `docs/design/prompt-basis.md` says the compiler found two (`steer`, `error_response`) with `model_route_notice` explicitly slipping PAST the type. Whichever number is right, two of the four copies are wrong — and a reader following the Development Protocol hits the disagreement in the row's own "Must read" doc, undermining exactly the history-of-the-defect rationale these registries lean on.

jacoblee-io (approved):
Reviewed the kind-registry mechanism, the dialect predicate, and all four migrated call sites. The predicate's NULL/dialect semantics were checked hard from several angles and hold on both engines (including legacy native-JSON MySQL columns, where the old LIKE was silently broken), and no legacy LIKE site was missed. Approving, with inline comments plus three findings that can't be anchored because their files aren't in this diff:

1. **`src/gateway/sse-consumer.ts` ~807/1065/667/490 — open-record locals defeat the narrowing.** Metadata-root locals in the primary persistence consumer keep explicit `Record<string, unknown>` annotations: `startMetadata` (~807), `stoppedMeta` (~1065, passed to `updateMessage`), the tool-end metadata local (~667), and `lastAssistantMetadata` (~490, spread at ~549). All are assignable to `ChatMessageMetadata` via the index signature, so literal-kind checking never fires at these write sites — anyone adding `kind: "stopped"` to one of them compiles clean and ships an unregistered kind, the PR's own definition of the rot mode. Fix is annotation-only: declare them `ChatMessageMetadata`.

2. **`src/agentbox/session.ts:2374` — the prompt-count contract rests on an unpinned string sniff.** `persistSyntheticMessage` tags the injected background-turn user row `kind:"task_notification"` only when `content.trimStart().startsWith("<task_notification>")`. A non-matching injection persists with `metadata: null` in the parent (origin-NULL) session, where a kind-less row is counted as a human prompt *by design* — invisible to both the origin axis and the new kind axis. No test pins the builder-output-starts-with coupling, so if `buildNotificationBatch` ever leads with a header/digest line (or a second `runSyntheticPrompt` caller uses a different wrapper), the Prompts figure silently re-inflates at background-job volume, unfixable at the query layer. Deeper fix: tag by call-site identity (`runSyntheticPrompt` only ever injects notification text), not by string prefix.

3. **`portal-web` — the registry is invisible to its largest consumer.** The frontend dispatches on bare kind string literals in ~18 places across ≥6 files against `metadata?: Record<string, unknown>`, with no mirror of `CHAT_MESSAGE_KINDS`. The repo already has the exact pattern for this: `portal-web/src/lib/agentTypes.ts`, asserted against the backend registry. Without it, the next registered kind reaches the frontend with no list to fail and renders the runtime's raw injected text as an ordinary user bubble — the same regression the `<task_notification>` XML bubble had — and a typo in any of the 18 literal comparisons is a silently never-matching branch today.

PR #549, opened 2026-09-02
fix(knowledge): label index correctness — hot-update re-sync and navigation-page exclusion

LikiosSedo:
Two label-index correctness fixes, no agent-visible contract change. Together they make `knowledge_search` serve the index the mounted knowledge actually implies — after hot updates, and only for content pages.

## 1. Re-sync the label index on hot update in K8s

In K8s mode, a knowledge hot update (publish/takeover) lands on disk and the prompt catalog refreshes through `brain.reload()`, but `knowledge_search` keeps serving the label index built at pod start until the pod recycles; new sessions in the pod inherit the stale index (resolver is shared per process).

Observed live as `knowledge_search(listLabels=true)` returning `totalLabels: 0, totalPages: 0` on an agent whose repos had just been updated with labeled packages.

Cause: the per-server knowledge handler — the only one carrying the `afterMaterialize` hook that calls `sessionManager.syncKnowledgeIndex()` — was gated on `sessionManager.knowledgeDir`, which only `LocalSpawner` assigns. Fix: create it unconditionally; `materialize` already resolves its target dir with the same config fallback the module-level handler uses, so the only behavioral change is that the hook now fires. Local mode unchanged.

## 2. Keep labeled navigation pages out of the label index

The label index excluded navigation pages by filename only (`index.md`, `log.md`), while citation validation and the catalog graph classify with `isKnowledgeNavigationPage` (path or frontmatter `type: index`). A page named anything else carrying `type: index` plus labels was searchable, readable — and then hard-failed the entire `knowledge_cite` call as a navigation ref. This shape is about to become load-bearing (generated route/FAQ navigation pages).

Fix: classify with the same `isKnowledgeNavigationPage` so the three layers cannot disagree. Navigation pages are routing surfaces, not content pages missing labels — they no longer inflate the `unlabeledPages` diagnostic (also fixes `_index.md` sub-catalogs being counted as unlabeled).

## Validation

- `npx vitest run src/agentbox/http-server.test.ts src/tools/query/knowledge-search.test.ts` — 112 passed, including two new regression tests (K8s shape: reload → `syncKnowledgeIndex` called once; labeled `type: index` page and labeled `sub/_index.md` absent from search results, label catalog, and diagnostic counters)
- `npx tsc --noEmit`

## Deploy note

Agentbox image: rebuild + repoint `SICLAW_AGENTBOX_IMAGE` + recycle pods. After deploy, worth re-verifying the earlier `totalLabels: 0` incident against fix 1.

### DIFF
--- a/src/agentbox/http-server.test.ts
+++ b/src/agentbox/http-server.test.ts
@@ -26,6 +26,12 @@ const mockConfigState = vi.hoisted(() => ({
   mcpServers: {} as Record<string, unknown>,
 }));
 
+// Captures the options createHttpServer passes when building its per-server
+// knowledge handler, so tests can pin the afterMaterialize wiring.
+const knowledgeHandlerState = vi.hoisted(() => ({
+  lastOptions: null as null | { knowledgeDir?: string; afterMaterialize?: () => void | Promise<void> },
+}));
+
 // Silence metrics auth side effects.
 vi.mock("../shared/metrics.js", () => ({
   checkMetricsAuth: () => true,
@@ -71,12 +77,18 @@ vi.mock("./sync-handlers.js", async (importOriginal) => ({
   getSyncHandler: () => undefined,
   createClusterHandler: () => ({ type: "cluster", fetch: async () => 0, materialize: async (n: number) => n }),
   createHostHandler: () => ({ type: "host", fetch: async () => 0, materialize: async (n: number) => n }),
-  createKnowledgeHandler: () => ({
-    type: "knowledge",
-    fetch: async () => ({ repos: [] }),
-    materialize: async () => 0,
-    getLastKnowledgeSyncStatus: () => null,
-  }),
+  createKnowledgeHandler: (options: { knowledgeDir?: string; afterMaterialize?: () => void | Promise<void> } = {}) => {
+    knowledgeHandlerState.lastOptions = options;
+    return {
+      type: "knowledge",
+      fetch: async () => ({ repos: [] }),
+      // The real handler awaits afterMaterialize on EVERY materialize path
+      // (empty-wipe and normal) — the fake must model that, or the reload
+      // route test cannot see whether the wiring passed the hook at all.
+      materialize: async () => { await options.afterMaterialize?.(); return 0; },
+      getLastKnowledgeSyncStatus: () => null,
+    };
+  },
   createToolsHandler: (target: { allowedToolsState: string[] | null }) => ({
     type: "tools",
     fetch: async () => ({ allowedTools: null }),
@@ -262,6 +274,8 @@ function makeFakeSessionManager() {
     onSessionRelease: undefined as undefined | (() => void),
     credentialBroker: undefined,
     credentialsDir: undefined,
+    // K8s shape: knowledgeDir stays unset (only LocalSpawner assigns it).
+    syncKnowledgeIndex: vi.fn(async () => {}),
   };
 }
 
@@ -1965,6 +1979,20 @@ describe("http-server — reload routes delegate to handlers", () => {
     // short-circuits to 200 count:0 before ever calling the per-box handler.
     expect(r.status).toBe(200);
     expect(r.data).toMatchObject({ ok: true, count: 0, type: "tools" });
+  });
+
+  it("POST /api/reload-knowledge re-syncs the label index even when sessionManager.knowledgeDir is unset (K8s shape)", async () => {
+    // K8s boxes never set sessionManager.knowledgeDir (only LocalSpawner does),
+    // and the per-server knowledge handler used to be gated on it — so a hot
+    // update materialized new files while knowledge_search kept serving the
+    // label index built at pod start. Pin the wiring: the handler exists
+    // without the dir, and its materialize drives syncKnowledgeIndex.
+    process.env.SICLAW_GATEWAY_URL = "http://gateway.test";
+    const r = await getJson(port, "/api/reload-knowledge", "POST");
+    expect(r.status).toBe(200);
+    expect(r.data).toMatchObject({ ok: true, type: "knowledge" });
+    expect(knowledgeHandlerState.lastOptions?.knowledgeDir).toBeUndefined();
+    expect(sm.syncKnowledgeIndex).toHaveBeenCalledTimes(1);
   });
 
   it("POST /api/reload-tracing is wired standalone and is a clean no-op without a gateway URL", async () => {
--- a/src/agentbox/http-server.ts
+++ b/src/agentbox/http-server.ts
@@ -655,16 +655,19 @@ export function createHttpServer(
     perServerHandlers.cluster = createClusterHandler(sessionManager.credentialBroker);
     perServerHandlers.host = createHostHandler(sessionManager.credentialBroker);
   }
-  const perServerKnowledgeHandler = sessionManager.knowledgeDir
-    ? options.knowledgeHandler
-      ?? createKnowledgeHandler({
-        knowledgeDir: sessionManager.knowledgeDir,
-        afterMaterialize: () => sessionManager.syncKnowledgeIndex(),
-      })
-    : undefined;
-  if (perServerKnowledgeHandler) {
-    perServerHandlers.knowledge = perServerKnowledgeHandler;
-  }
+  // The knowledge handler is per-server UNCONDITIONALLY — never gated on
+  // sessionManager.knowledgeDir. K8s boxes leave the dir unset (materialize
+  // falls back to config.paths.knowledgeDir, identical to the module-level
+  // handler's resolution), but only this closure carries the afterMaterialize
+  // hook that re-syncs the shared label index. Behind the gate, a K8s
+  // knowledge hot update landed on disk while knowledge_search kept serving
+  // the index built at pod start until the pod recycled.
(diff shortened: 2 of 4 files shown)


src/knowledge/labels.ts
```diff
@@ -270,13 +271,21 @@ export class KnowledgeLabelIndex {
...
         try { markdown = fs.readFileSync(absolute, "utf8"); } catch { continue; }
+        const file = path.relative(this.knowledgeDir, absolute);
+        // Navigation pages (`_index.md` by path, `type: index` by frontmatter;
+        // plain `index.md` never gets here) must not enter the label index:
+        // citation validation rejects them as evidence, so a labeled navigation
+        // page would hand knowledge_search a candidate whose cite call then
+        // hard-fails the whole turn's citations. Same classifier as citation
+        // validation and the catalog graph, so the layers cannot disagree.
+        // They are routing surfaces, not content pages — no counter increments.
+        if (isKnowledgeNavigationPage(file, markdown)) continue;
```
jacoblee-io, line 282: The comment says "same classifier … so the layers cannot disagree", but the two frontmatter grammars still disagree on the same file: `isKnowledgeNavigationPage` requires `body.startsWith('---\n')` exactly (page-kind.ts:17), while this file's `frontmatterSource` normalizes `\r\n` and trims a BOM (labels.ts:88). A CRLF-encoded or BOM-prefixed navigation page (`type: index`) fails classification here but parses fine below — so it enters the label index, search results, and `totalPages`, and the exclusion never fires. Files land on disk byte-exact from the sync pipeline; nothing normalizes upstream.

Related cost: every scanned page's frontmatter is now `yaml.load`ed 2–3× per sync (the classifier's own parse, then `parseKnowledgeLabels` → `parseFrontmatter`, plus `declaresKnowledgeLabels` on the unparsed branch) — and with the re-sync fix, `sync()` now runs on every hot update inside the reload request path, not just at pod start.

Both have the same root fix: parse frontmatter once per file (export this file's parser for page-kind.ts to consume, or classify from the already-parsed object via path check + `meta.type === 'index'`), so the classifier and the label parser cannot read one file two ways.

src/knowledge/labels.ts
```diff
@@ -270,13 +271,21 @@ export class KnowledgeLabelIndex {
         if (lowerName === "index.md" || lowerName === "log.md") continue;
```
jacoblee-io, line 271: This retained filename skip half-duplicates the classifier called 11 lines later: `isKnowledgeNavigationPath` already matches basename `index.md` and `_index.md`, so the `index.md` branch is redundant — and a future navigation-definition change in page-kind.ts must be mirrored here or the index silently diverges again. Meanwhile `log.md` remains a labels-layer-only exclusion the shared classifier does not know: a labeled `log.md` is excluded from the label index but citation validation (gated on `isKnowledgeNavigationPage` only) accepts it as evidence, and the catalog graph treats it as content — the same cross-layer drift this PR exists to eliminate. Suggest folding `log.md` into page-kind (or a named predicate beside it) and keeping only a pre-read short-circuit that delegates to `isKnowledgeNavigationPath`.

src/agentbox/http-server.ts
```diff
@@ -655,16 +655,19 @@ export function createHttpServer(
...
+  // handler's resolution), but only this closure carries the afterMaterialize
+  // hook that re-syncs the shared label index. Behind the gate, a K8s
+  // knowledge hot update landed on disk while knowledge_search kept serving
+  // the index built at pod start until the pod recycled.
+  const perServerKnowledgeHandler = options.knowledgeHandler
+    ?? createKnowledgeHandler({
+      knowledgeDir: sessionManager.knowledgeDir,
+      afterMaterialize: () => sessionManager.syncKnowledgeIndex(),
+    });
+  perServerHandlers.knowledge = perServerKnowledgeHandler;
```
jacoblee-io, line 670: Unconditionally registering the per-server handler regresses K8s `GET /api/sync-status`: this handler's in-memory `lastKnowledgeSyncStatus` is `null` until the first hot reload — startup `syncAllResources` (agentbox-main.ts:73) populates only the now-shadowed module-level handler — so status silently falls back to the disk manifest. And on the empty-bundle startup path (agent with zero bound knowledge repos), the wipe writes no `.sync-manifest.json` at all: `readBoxSyncStatus` → `readKnowledgeInventoryFromDisk` finds no manifest and reports `syncedAt: null / repos: []` ("never synced") for a successfully synced box, until the first `/api/reload-knowledge`. Pre-PR the same request returned the module handler's real `syncedAt`. A console verifying that a publish landed misreads the box as unsynced.

src/agentbox/http-server.ts
```diff
@@ -655,16 +655,19 @@ export function createHttpServer(
...
-    perServerHandlers.knowledge = perServerKnowledgeHandler;
-  }
+  // The knowledge handler is per-server UNCONDITIONALLY — never gated on
+  // sessionManager.knowledgeDir. K8s boxes leave the dir unset (materialize
+  // falls back to config.paths.knowledgeDir, identical to the module-level
+  // handler's resolution), but only this closure carries the afterMaterialize
+  // hook that re-syncs the shared label index. Behind the gate, a K8s
+  // knowledge hot update landed on disk while knowledge_search kept serving
+  // the index built at pod start until the pod recycled.
+  const perServerKnowledgeHandler = options.knowledgeHandler
```
jacoblee-io, line 665: This fixes the reload path but leaves TWO knowledge-handler instances per K8s box with asymmetric behavior: startup `syncAllResources` still materializes through the module-level handler (sync-handlers.ts:900), which has NO `afterMaterialize` hook. That's benign today only because the sessionManager is constructed after `syncAllResources` and the shared indexer initializes lazily on first session — an ordering nothing pins. Any retry path, startup reorder, or future caller resolving `getSyncHandler('knowledge')` re-creates exactly the stale-index bug this PR fixes. Deeper fix: pass this hooked handler into the startup sync (what LocalSpawner already does, local-spawner.ts:195) — or fold the re-sync into `createKnowledgeHandler` itself — then delete the shadowed module-level singleton and its now importer-less `getLastKnowledgeSyncStatus` accessor. That would also resolve the `/api/sync-status` regression noted below.

src/agentbox/http-server.ts
```diff
@@ -655,16 +655,19 @@ export function createHttpServer(
...
+    ?? createKnowledgeHandler({
+      knowledgeDir: sessionManager.knowledgeDir,
+      afterMaterialize: () => sessionManager.syncKnowledgeIndex(),
```
jacoblee-io, line 668: Nit: this `afterMaterialize` → `syncKnowledgeIndex` wiring is now hand-copied at two construction sites, and the copies already drift — this one calls `sessionManager.syncKnowledgeIndex()` bare while local-spawner.ts:119 uses `syncKnowledgeIndex?.()` (a shape lacking the method throws on one path, no-ops on the other). The next spawner/handler construction site has to remember the same closure or the stale-index bug recurs there. Worth owning the re-sync in one place, e.g. `createKnowledgeHandler` given a sessionManager.

src/agentbox/http-server.test.ts
```diff
@@ -71,12 +77,18 @@ vi.mock("./sync-handlers.js", async (importOriginal) => ({
   getSyncHandler: () => undefined,
   createClusterHandler: () => ({ type: "cluster", fetch: async () => 0, materialize: async (n: number) => n }),
   createHostHandler: () => ({ type: "host", fetch: async () => 0, materialize: async (n: number) => n }),
-  createKnowledgeHandler: () => ({
-    type: "knowledge",
-    fetch: async () => ({ repos: [] }),
-    materialize: async () => 0,
-    getLastKnowledgeSyncStatus: () => null,
-  }),
+  createKnowledgeHandler: (options: { knowledgeDir?: string; afterMaterialize?: () => void | Promise<void> } = {}) => {
```
jacoblee-io, line 80: The fake's own comment admits it hand-models the real handler's contract ("awaits afterMaterialize on EVERY materialize path") — so this regression test pins the wiring but cannot catch the real `createKnowledgeHandler` dropping the hook (e.g. on the empty-wipe path). The `totalLabels:0` incident is then guarded only by mock fidelity. Suggest a companion sync-handlers.test.ts case asserting `afterMaterialize` fires on both materialize paths (empty-wipe and normal) against the real handler.

jacoblee-io (approved):
Reviewed with a focus on the two fixes (hot-update re-sync wiring, navigation-page exclusion). The core direction is right and the tests pin the wiring well. Approving, with inline comments worth addressing — the `/api/sync-status` regression in particular.

One finding that can't be anchored inline because the file isn't in this diff — `src/agentbox/session.ts:794` (`ensureSharedComponents`): the hot-update re-sync this PR wires in is a permanent silent no-op whenever the shared knowledge indexer failed to initialize. The once-latch is set *before* the init attempt, the failure is only `console.warn`ed, and `_sharedKnowledgeIndexer` stays pinned to `null` with no retry — so `syncKnowledgeIndex()`'s `?.sync()` never fires for the pod's lifetime while the reload route keeps reporting success. That reproduces the exact `totalLabels:0` incident this PR fixes, with no error surfaced. Suggest not setting (or resetting) the latch on init failure, or at minimum logging the no-op on the reload path.

PR #550, opened 2026-09-02
fix(knowledge): keep labeled navigation pages out of the label index

LikiosSedo:
## Problem

The label index (`KnowledgeLabelIndex.sync`) excludes navigation pages by **filename only** (`index.md`, `log.md`), while citation validation and the catalog graph classify navigation with `isKnowledgeNavigationPage` — path (`index.md`/`_index.md`) **or** frontmatter `type: index`.

A page named anything else that carries `type: index` plus a `labels` array is therefore label-indexed and returned by `knowledge_search`; the agent reads it and cites it, and the cite call **hard-fails in full** (`Cannot cite navigation pages` registers zero citations for the whole call, valid refs included). This trap is about to become load-bearing: generated route/FAQ navigation pages are exactly this shape.

## Fix

Classify pages in the label index with the same `isKnowledgeNavigationPage` the citation validator and catalog graph use, so the three layers cannot disagree. Navigation pages are routing surfaces, not content pages missing labels — they no longer inflate the `unlabeledPages` diagnostic either (this also stops `_index.md` sub-catalogs being counted as unlabeled).

## Validation

- `npx vitest run src/tools/query/knowledge-search.test.ts src/knowledge/label-backfill.test.ts` — 14 passed, including a new regression test (labeled `type: index` page and labeled `sub/_index.md`: absent from search results and label catalog, zero diagnostic-counter noise)
- `npx tsc --noEmit`

## Deploy note

Agentbox image (label index runs in the box).

### DIFF
--- a/src/knowledge/labels.ts
+++ b/src/knowledge/labels.ts
@@ -4,6 +4,7 @@ import path from "node:path";
 import yaml from "js-yaml";
 
 import { buildKnowledgeCatalogRoutes, type KnowledgeRouteProof } from "./catalog-graph.js";
+import { isKnowledgeNavigationPage } from "./page-kind.js";
 
 export const KNOWLEDGE_LABEL_FACETS = [
   "entity", "topic", "task", "component", "environment", "version",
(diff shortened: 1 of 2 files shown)


LikiosSedo:
Folded into #549 (same label-index correctness scope, one review context) — commit 76a69853 there is identical to this branch's 073236e2.

PR #552, opened 2026-09-02
fix(exec): stop the tool telling the agent things that are not true

chent1996:
## Summary

The `bash` tool has been reporting four different things to the agent that were not true: a filter
that matched nothing as an upstream failure, a successful early-exit pipeline as an incomplete read,
a legitimate regex as a path, and a missing metric as a missing object. Each one makes the agent
abandon a correct conclusion and retry, or route around a command that was never wrong.

### Problem

From the review backlog on `builtin_tool/bash` — 126 findings, 190 mentions, high severity, first
reported 2026-08-20 and still growing on 2026-09-01. `pipeline_upstream_failed` alone accounts for 25
of them. Every defect below was reproduced locally against the production trace that reported it
before anything was changed.

The sharpest example, from trace `cc0c857…`: five consecutive log-filter commands were refused with
`grep file operand "seed brokers|Operation timed out|..." names a path`, and the agent gave up on
that line of evidence. The `-m 8` value had spent grep's pattern quota, so the regex was screened as
a filename.

### Solution

Stage attribution turned out to be the root of two of the four. `PIPESTATUS` describes **one**
pipeline — the last bash executed — while the command text may hold several joined by `;`/`&&`/`||`.
Splitting text on `|` and indexing it with a status subscript is how a trace came to read
`stage 1/2 (echo) exited 141` for a pipeline whose stages were `kubectl` and `grep`.

- **Segment with the splitter we already have.** `extractPipeline` (quote-, paren- and
  escape-aware) replaces the regex split, so a `|` inside `'a|b'` is no longer a stage boundary. One
  splitter, not two.
- **Refuse to name a stage unless exactly one candidate pipeline fits the status count.** Returning
  nothing is a correct answer: for `a || b` with one status, "a succeeded and b never ran" fits as
  well as "a failed and b succeeded".
- **Judge SIGPIPE from the statuses alone.** A writer gets SIGPIPE only when the read end closed, so
  the first status below it that is not itself 141 settles it — no stage text needed. This matters
  because SIGPIPE *cascades*: `seq … | cat | head -1` reports `[141, 141, 0]`, and judging the first
  141 against only its neighbour made `kubectl logs | grep -v noise | head -20` a failure.
- **Relax a mid-pipeline `grep` exit 1 to `no_match`, but only when the stage can be attributed.**
  That condition is a safety property: with the old split, `echo "a | grep -v x" | kubectl get pods |
  jq .items` put `grep` at subscript 1, so relaxing unaligned would downgrade kubectl's real failure
  to "nothing found" — the false success this mechanism exists to remove.
- **Scope metrics answers to `kubectl top`.** `ServiceUnavailable` sits in the channel markers with
  no context guard, so `kubectl top` on a cluster without metrics-server was reported as "the target
  never ran the command". Matched by the `metrics.k8s.io` group rather than by resource kind, since
  the string appears in singular and plural forms. `Forbidden` is excluded and both possible causes
  of a NotFound are named — RBAC is a definite cause, and one message covers both "does not exist"
  and "not scraped yet".
- **Implement `TextExpressionArgs.valueFlags`**, documented since the layer was written but never
  built. Consuming a flag's value frees the pattern quota; screening it still refuses
  `grep -m /etc/shadow`. Both halves are load-bearing.
- **Ask the expression slot whether the shell will rewrite it**, per character and per quote kind.
  The slot was exempt from *everything*, so an unquoted glob reached the command as expanded paths —
  the first becoming the pattern, the rest becoming files it reads. The operand rule is untouched:
  its character class is a strict superset of the expansion metacharacters and it matches after
  quotes are stripped, so zero regression there is structural rather than tested.
- **Refuse a blocked capability by its other spellings.** `-d recurse` is `-r`, and `getopt_long`
  takes any unambiguous prefix — real GNU grep 3.8 runs `--recursi` and `--di=recurse` recursively,
  past both denylists.

Contracts and rationale: `docs/design/2026-09-02-exec-classification-and-operand-quoting.md`.

## ⚠️ Agent-visible contract change

Annotation wording and tool outcome change for the shapes below. Consumers reading tool outcomes
should expect **fewer** failures, not different ones.

| Shape | Before | After |
|---|---|---|
| `… \| grep X \| tail` (no match) | `pipeline_upstream_failed`, error | `no_match`, not an error |
| `… \| grep -v X \| head -N` (cascaded SIGPIPE) | `pipeline_upstream_failed`, error | `success` |
| Upstream failure in an unalignable chain | named the wrong stage | reported without naming a stage |
| `kubectl top` with no metrics data | `no_match` / `channel_error` | `target_reported_failure`, both causes named |
| `grep -A 5 '<regex>'`, `yq -o json '<filter>'`, `jq --arg k v '<filter>'` | refused as a file operand | runs |
| Unquoted glob in an expression slot | ran, reading files | refused, with the corrected command in `hint` |
| Unquoted `jq`/`yq` filter with brackets (`jq .items[]`) | ran | refused; quote the filter |
| `grep -d recurse`, `--di=recurse`, `--recursi` | ran | refused |

`kubectl top` moving from `no_match` (not an error) to `target_reported_failure` (an error) changes
what counts toward failure rates for that shape.

## ⚠️ Deployment

All three modified source files ship inside the **agentbox image**. A runtime rollout does not pick
them up: rebuild the image, repoint `SICLAW_AGENTBOX_IMAGE`, and recycle pods. `imagePullPolicy:
Always` only pulls on pod **create**, so a live pod keeps the old code and verifying against an
existing session proves nothing.

## Test Plan

- [x] `npm test` — 295 files / 6609 tests, 0 failures; `npx tsc --noEmit` clean
- [x] Real-bash differential test for stage alignment (`pipeline-status.test.ts`): 10 shapes covering
      `|`, `&&`, `||`, `;`, subshells, `pipefail` and a quoted `|`, each pinning the stage count bash
      actually reports. This is how the `&&`/`||` short-circuit defect was found — it is not visible
      from reading the code.
- [x] Security baseline pinned in `command-sets.test.ts`: every pre-existing payload still refused,
      plus 15 that were **allowed before this change** (`grep .siclaw/*/*/*`, `grep -c "$HOME/x"`,
      `grep -e <glob>`, `jq <glob>`, `--di=recurse`, …), and the regex idioms that must keep working
      (`grep "error$"`, `"^$"`, `"\$HOME"`).
- [x] Deployed to `siclaw-tc` and verified against the **deployed agentbox image** with strings
      captured from a real cluster, not reconstructed:
      - `error: Metrics API not available` (this cluster has no metrics-server) → correct apiDown
        wording
      - `apiservices.apiregistration.k8s.io "v1beta1.metrics.k8s.io" not found` from
        `kubectl get apiservice` → still `no_match`, not hijacked by the metrics branch
      - `PIPESTATUS=[141,141,0]` → `success`; `[0,1,0]` → `no_match`; `[1,0]` → still
        `pipeline_upstream_failed`
- [ ] **Not verified:** the `nodemetrics.metrics.k8s.io "X" not found` form (metrics-server online
      but no sample for that name). No cluster available with metrics-server running; that marker
      rests on metrics-server issue reports, not a captured string.
- [ ] **Not verified:** that the class reaches the model as TEXT rather than only in `details`.
      `siclaw-tc` has no registered clusters, so the `bash` tool cannot be driven through an agent
      there; the deployed image was exercised directly instead. Unit tests cover the contract.


src/tools/infra/command-sets.ts
```diff
@@ -384,6 +466,30 @@ function validateByRule(
       }
     }
 
+    // blockedFlagValues: the flag is fine, this value is not. Handles all three spellings the shell
+    // allows for a value — `--directories=recurse`, `-d recurse` and `-drecurse`.
+    if (rule.blockedFlagValues && arg.startsWith("-")) {
```
jacoblee-io, line 514: **HIGH — `blockedFlagValues` is one keystroke deep: a short-option cluster bypasses it**

`extractFlag` only splits on `=`, so it never decomposes a short-option cluster. For `-id recurse`, `extractFlag("-id")` returns `"-id"`, which is not a key in the map; `blockedLongAbbrev` bails immediately because the token does not start with `--`; `flag` falls back to `"-id"`, `.get()` returns `undefined`, and the whole check is skipped. GNU grep parses `-id recurse` as `-i -d recurse` (short `d:` takes its argument from the next argv), so recursion is enabled — the exact capability this check was added to block.

Verified against this branch (head `3254182e`):

```js
import { validateCommand } from "./src/tools/infra/command-validator.js";
validateCommand("echo x | grep -id recurse -l cred .", { context: "local" }) // => null  (ALLOWED)
validateCommand("echo x | grep -d recurse -l cred .",  { context: "local" }) // => refused
validateCommand("echo x | grep -r -l cred .",          { context: "local" }) // => refused
```

and at the restriction layer directly:

```js
import { validateCommandRestrictions } from "./src/tools/infra/command-sets.js";
const local = { context: "local", piped: true };
validateCommandRestrictions("grep -id recurse -l cred .", local)     // => null  (ALLOWED)
validateCommandRestrictions("grep -in -d recurse -l cred .", local)  // => refused
```

`applyContextPolicy` already solves this exact problem one screen down (line 1417: it walks the cluster looking for the first letter in `spec.shortValueFlags` and rewrites `flag`/`inlineValue`). `validateByRule` needs the same resolution — or, more cheaply here, check every letter of a short cluster against the map's single-letter keys.

Separately: the comment above this block claims it handles `-drecurse`, but `extractFlag("-drecurse")` returns the whole token, so this code does not see that spelling either. It is refused only by accident — the `blockedFlags` per-letter scan a few lines up hits the `r` in `"recurse"`. A banned value with no `r` in it would sail straight through, and the comment would still claim otherwise.

src/tools/infra/command-sets.ts
```diff
@@ -1805,9 +2036,12 @@ export interface CommandDef {
  */
 export const COMMANDS: Record<string, CommandDef> = {
   // ── text processing ──
-  grep:   { category: "text", blockedFlags: ["-r", "-R", "--recursive"] },
-  egrep:  { category: "text", blockedFlags: ["-r", "-R", "--recursive"] },
-  fgrep:  { category: "text", blockedFlags: ["-r", "-R", "--recursive"] },
+  grep:   { category: "text", blockedFlags: ["-r", "-R", "--recursive"],
```
jacoblee-io: **MEDIUM — `--dereference-recursive` is the long form of the already-blocked `-R` and is still allowed**

`blockedFlags` remains `["-r", "-R", "--recursive"]`. GNU grep also spells recursion `--dereference-recursive` (the long form of `-R`), and that string is in neither list, so `blockedLongAbbrev` cannot reach it — resolution is one-sided, mapping an abbreviation *onto* something already blocked, and this spelling never is.

Verified against this branch (head `3254182e`):

```js
import { validateCommand } from "./src/tools/infra/command-validator.js";
validateCommand("echo x | grep --dereference-recursive -l cred .", { context: "local" }) // => null  (ALLOWED)
validateCommand("echo x | grep -R -l cred .",                      { context: "local" }) // => refused
```

```js
import { validateCommandRestrictions } from "./src/tools/infra/command-sets.js";
const local = { context: "local", piped: true };
validateCommandRestrictions("grep --dereference-recursive pat .", local) // => null  (ALLOWED)
validateCommandRestrictions("grep --deref pat .", local)                 // => null  (ALLOWED)
```

A recursive grep started at `.` walks the working directory while naming no path at all, which is the vector `-d recurse` was added for. Suggested fix: add `--dereference-recursive` (and `-R`'s companion spellings) to `blockedFlags` on `grep`/`egrep`/`fgrep`; the new prefix resolution then covers its abbreviations for free.

src/tools/infra/command-sets.ts
```diff
@@ -367,9 +444,14 @@ function validateByRule(
     // blockedFlags: explicitly forbidden flags (checked before allowedFlags)
     if (rule.blockedFlags) {
       if (arg.startsWith("--")) {
-        if (rule.blockedFlags.includes(extractFlag(arg))) {
+        const named = extractFlag(arg);
+        const blocked = rule.blockedFlags.includes(named)
```
jacoblee-io, line 484: **MEDIUM (regression) — prefix resolution refuses `dmidecode --dump`, which the code says must keep working, and explains it with something untrue**

`blockedLongAbbrev` (line 1325) accepts any written flag that is a *prefix* of a blocked one, without first asking whether the written flag is itself a real option of that command. `dmidecode`'s only blocked flag is `--dump-bin`, and `"--dump-bin".startsWith("--dump")` is true, so the legitimate `--dump` (dump SMBIOS to **stdout**) is now refused. The `COMMANDS` entry's own comment states the opposite intent: `// --dump-bin <file> writes SMBIOS binary to an arbitrary file → blocked; -u/--dump (stdout) kept`.

Verified against this branch (head `3254182e`) vs `main`:

```js
import { validateCommandRestrictions } from "./src/tools/infra/command-sets.js";

// on main
validateCommandRestrictions("dmidecode --dump")  // => null  (ALLOWED)

// on 3254182e
validateCommandRestrictions("dmidecode --dump")
// => { "error": "dmidecode \"--dump\" is not allowed — the shell's option parser reads it as \"--dump-bin\"." }
validateCommandRestrictions("dmidecode --dump -t 1")  // => same refusal
validateCommandRestrictions("dmidecode -u")           // => null (still fine)
```

Also confirmed through `validateCommand("dmidecode --dump", { context: "local" })`, so it is refused on the real path, not just the restriction helper.

The error text is independently wrong: `getopt_long` prefers an **exact** match over an abbreviation, so real `dmidecode --dump` is `--dump`, never `--dump-bin`. The message asserts a shell behaviour that does not exist — the same class of untruth this PR set out to remove.

Suggested fix: skip abbreviation resolution when the written flag exactly matches a known option of the command (e.g. it appears in `def.allowedFlags`, or short-circuit when `flag` equals any full option spelling the command declares). Failing that, at minimum do not claim the option parser reads it as the blocked flag.

src/tools/infra/command-sets.ts
```diff
@@ -1190,15 +1366,43 @@ function applyContextPolicy(
...
+        expressionValueFollows = false;
+        const err = rejectUnquotedExpansion(baseName, arg, expansion);
+        if (err) return err;
+        continue;
+      }
+
+      // A value-flag's value is neither an expression nor a positional. CONSUMING it is what stops
+      // `grep -m 5 PATTERN` from spending the pattern quota on `5`; SCREENING it is what stops
+      // `grep -m /etc/shadow` from hiding a path there. Both halves are load-bearing.
+      if (plainValuesLeft > 0) {
```
jacoblee-io, line 1466: **LOW — the new arity consumption skips the `fileFlags` check, so a file-reading flag parked in a value slot is no longer seen**

The consumed token is screened with `rejectPathishOperand` only. It is never tested against `spec.fileFlags`, which is the check that used to catch it back when the same token was read as a flag. So `-f` hidden in `-m`'s value position is now invisible, and the following positional then spends the expression quota and is exempt from the operand rule.

Verified against this branch (head `3254182e`) vs `main`:

```js
import { validateCommandRestrictions } from "./src/tools/infra/command-sets.js";
const local = { context: "local", piped: true };

// on main: refused with
//   grep "-f" reads a file, which is not allowed here — grep may only process piped input.
// on 3254182e:
validateCommandRestrictions("grep -m -f /etc/shadow", local)   // => null  (ALLOWED)

// the arity-satisfied forms are still caught, which is why this is narrow:
validateCommandRestrictions("grep -A 5 -f /etc/shadow", local) // => refused ("-f" reads a file)
validateCommandRestrictions("grep -m 5 -f /etc/shadow", local) // => refused
```

Not exploitable with today's grep — `-m -f` makes grep exit 2 with `invalid max count`, and the same holds for `-C -f`, `-d -f` and `jq --arg -f x` — so this is a weakened guard rather than a live hole. But it is weaker than what it replaced, and the weakening is invisible from the tests. Screening the consumed value against `spec.fileFlags` in addition to `rejectPathishOperand` restores the old floor at no cost.

src/tools/infra/command-sets.ts
```diff
@@ -24,23 +24,80 @@ export function shellEscape(arg: string): string {
...
+   * "Active" is a per-character property, not a per-argument one, because quoting is applied per
+   * character: in `'$CRED'/clusters/*` the `$` is inert and the `*` is not. See
+   * `docs/design/2026-09-02-exec-classification-and-operand-quoting.md` for why this is the right
+   * question to ask of a stdin-only command's expression.
+   */
+  expansion?: string;
+}
+
+/** Metacharacters the shell acts on only when they are UNQUOTED. Double quotes suppress all of these. */
+const GLOB_METACHARS = new Set(["*", "?", "[", "]", "{", "}", "~"]);
```
jacoblee-io: **LOW — `~`, `}` and `]` are treated as live at positions where bash never expands them, producing refusals whose stated reason is false**

`GLOB_METACHARS` is consulted purely per character, with no regard for position or for a matching opener. But bash tilde-expands only at the **start of a word**, and brace-expands only when a matching `{` with a `,` or `..` is present. So an unquoted `~`, `}` or `]` in the middle of a word is a literal, and the refusal tells the agent something untrue about its own command.

Verified against this branch (head `3254182e`):

```js
import { validateCommandRestrictions } from "./src/tools/infra/command-sets.js";
const local = { context: "local", piped: true };
validateCommandRestrictions("grep 'a~b'", local)  // => null (allowed, as expected)
validateCommandRestrictions("grep a~b", local)
// => grep expression "a~b" contains an unquoted "~", which the shell expands into file names
//    before grep runs — so this would be a list of paths, not a pattern.
```

`echo a~b` prints `a~b`; there is no expansion to warn about.

The `[`/`{` halves of the class are load-bearing (`jq .items[]` and `grep -E a{2,3}` really do expand, and the PR documents refusing them), so the fix is positional rather than removing characters: treat `~` as live only at index 0 of the argument, and `}`/`]` only when a matching `{`/`[` appeared earlier in the same unquoted run.

src/tools/infra/exit-classification.ts
```diff
@@ -203,59 +203,90 @@ export function classifyExit(opts: {
...
       return {
         exitClass: "pipeline_upstream_failed",
         isError: true,
         annotation:
           `[pipeline_upstream_failed: ${who}, while the last stage exited 0 — so the pipeline's exit code `
           + "says success and the output above is NOT a complete answer. An empty result here means the "
           + "EARLIER stage failed, not that nothing matched. Fix the failing stage; see STDERR for its "
-          + "error.]",
+          + "error."
+          + (aligned ? "" : " (This command chains pipelines with `&&`/`||`, so which command the status "
```
jacoblee-io: **LOW — the degraded annotation asserts `&&`/`||` chaining, but `alignPipelineStages` returns null for other reasons too**

The appended sentence states as fact "This command chains pipelines with `&&`/`||`". That is only one of the ways `alignPipelineStages` returns null. The other is step (4): *no* candidate group's length matches the status count — reachable without any `&&`/`||` in the command, e.g. a brace group (`{ a; b; } | c`, which `extractPipeline` splits on the inner `;`), a `|&` pipe (the `&` is read as a separate separator, so the two stages land in different groups), or any construct whose segment count disagrees with what bash actually forked.

In those cases the agent is told something about its own command that is not true, and the remedy offered ("run the failing pipeline on its own") is aimed at a chain that does not exist. Given the annotation's whole purpose is to stop asserting things the tool cannot know, the safe wording is the one that does not name a cause — e.g. "which command this status belongs to could not be determined; run the pipeline on its own to see it" — or gate the `&&`/`||` clause on actually having observed such a separator in `extractPipeline`'s output.

src/tools/infra/exit-classification.ts
```diff
@@ -203,59 +203,90 @@ export function classifyExit(opts: {
...
-      const at = label(expectedSigpipes[0].i);
+    // (2) Nothing failed; a filter simply matched nothing. Reported as the RESULT it is, so the
+    //     agent stops re-running a query whose answer it already has.
+    if (last === 0 && noMatchAt.size > 0) {
+      const who = [...noMatchAt].map((i) => label(i)).join(", ");
+      return {
+        exitClass: "no_match",
+        isError: false,
+        annotation:
+          `[no_match: ${who} ran and matched nothing, and every other stage exited 0 — so the empty or `
```
jacoblee-io: **LOW — "every other stage exited 0" / "the source command succeeded" is emitted even when upstream stages were SIGPIPE'd**

Branch (2) is reached before branch (3), and `realFailures` has already excluded benign SIGPIPEs — so a pipeline that has both a cascaded 141 *and* a no-match filter takes this branch, and the annotation then makes two claims that are false for it.

Reproduced with real bash on this branch (head `3254182e`):

```js
const cmd = "seq 1 200000 | cat | head -3 | grep zzz | wc -l";
// bash reports:
//   PIPESTATUS = [141, 141, 0, 1, 0]
classifyExit({ command: cmd, exitCode: 0, stdout, stderr: "", context: "local", pipeStatuses: statuses });
// => exitClass "no_match", annotation:
//    [no_match: stage 4/5 (grep) ran and matched nothing, and every other stage exited 0 — so the
//     empty or zero result above IS the answer, not a failure. The source command succeeded; ...]
```

Stages 1 and 2 did not exit 0, and the source (`seq`) did not run to completion — it was killed by SIGPIPE. The verdict itself (`no_match`, not an error) is right, because the `head -3` closure was deliberate; only the two supporting clauses are untrue. The realistic shape is `kubectl logs pod | cat | head -N | grep X | wc -l`.

Suggested fix: compute the benign set before composing the message and either soften the clause ("no other stage failed") or name the SIGPIPE alongside the no-match, so the annotation describes the statuses it actually saw.

src/tools/infra/command-sets.ts
```diff
@@ -1155,6 +1287,49 @@ const TEXT_EXPRESSION_ARGS = new Map<string, TextExpressionArgs>([
...
+ * working directory silently replaces the pattern.
+ */
+function rejectUnquotedExpansion(baseName: string, arg: string, expansion: string | undefined): string | null {
+  if (!expansion) return null;
+  return JSON.stringify({
+    error: `${baseName} expression "${arg}" contains an unquoted "${expansion}", which the shell expands `
+      + `into file names before ${baseName} runs — so this would be a list of paths, not a pattern.`,
+    matched: expansion,
+    rejected_by: "unquoted_expansion",
+    hint: `Quote the expression so the shell passes it through literally: ${baseName} ... '${arg}'`,
```
jacoblee-io: **LOW — the "copy-pasteable" hint single-quotes the argument without escaping embedded `'`**

The argument is interpolated straight into `'…'`. `shellEscape` at the top of this same file exists precisely for this (`"'" + arg.replace(/'/g, "'\\''") + "'"`), and it is not used here. For any expression containing a single quote, the hint hands the agent a command with an unterminated quote — which the test in `command-sets.test.ts` asserts is "copy-pasteable".

Note also that `arg` is the **parsed** value, not the text as written, so for a partly-quoted argument the hint changes the command's meaning: `grep '$CRED_DIR'/clusters/*` yields the hint `grep ... '$CRED_DIR/clusters/*'`, in which `$CRED_DIR` is now a literal rather than an expansion. Same for `grep "$HOME/x"` → `'$HOME/x'`, a pattern that cannot match what the caller was looking for.

Suggested fix: `hint: ... ${shellEscape(arg)}`, and consider phrasing the hint as "quote the expression" plus the escaped form rather than presenting it as the corrected command.

src/tools/infra/command-sets.ts
chent1996: Confirmed and fixed in a7edc5c7 — you are right on both halves.

The cluster is now decomposed in the `blockedFlagValues` check: the long/`=` reading is tried first, then each letter of a short cluster is looked up against the single-letter keys, with the value taken from the rest of the token if non-empty and from the next argv otherwise. `grep -id recurse`, `grep -in -d recurse` and `grep -idrecurse` are all refused now, and `grep -id skip` still runs.

The point about the comment is the more useful half and I have said so in the code: `-drecurse` was refused **by accident**, through the `r` inside the word `recurse` hitting the per-letter `blockedFlags` scan, so a banned value with no `r` in it would have gone straight through while the comment claimed coverage. That is a comment asserting something untrue about its own code, which is the same failure this PR is about. The replacement comment says which spellings are reached and why.

I did not reuse `applyContextPolicy`'s resolver: it keys on `spec.shortValueFlags` (the stdin-only table) while this check keys on `rule.blockedFlagValues`, so sharing it would mean threading one table into the other layer. Regression tests added in `command-sets.test.ts` ("decomposes a short cluster, which is where the value actually hides").

src/tools/infra/command-sets.ts
chent1996: Confirmed and fixed in a7edc5c7 — added `--dereference-recursive` to `blockedFlags` on grep/egrep/fgrep, and the prefix resolution then covers `--deref` and friends for free. Both are now refused; verified.

The generalisation is worth stating and I have put it in the design doc and the CLAUDE.md row: prefix resolution being **one-sided** has a corollary I did not draw at the time — a spelling that is never listed is never reached. So "block the capability" means naming every spelling of it, not just the one that happens to be short.

src/tools/infra/command-sets.ts
chent1996: You are right, and this one is mine — a regression this branch introduced, with an error message that asserts a `getopt_long` behaviour that does not exist. That is exactly the failure mode the PR set out to remove, turned inward. Fixed in a7edc5c7.

`dmidecode` has no `allowedFlags` (it is not default-deny), so there was nothing for the resolver to consult. Added a small `exactFlags` field for the case where a legitimate option is a prefix of a blocked one, and declared `exactFlags: ["'--dump"']` on `dmidecode` — the entry's own comment already promised `-u/--dump` was kept, so this makes the promise enforceable rather than aspirational. Resolution now short-circuits on an exact match, as getopt does.

The message is reworded too: it no longer claims the parser "reads it as" the blocked flag; it says the option parser accepts it *as an abbreviation of* it, which is only emitted now for genuine abbreviations. `dmidecode --dump`, `--dump -t 1` and `-u` all run; `--dump-bin` and `--dump-b` are refused. Pinned in tests.

src/tools/infra/command-sets.ts
chent1996: Confirmed and fixed in a7edc5c7 — the consumed token is now checked against `spec.fileFlags` before the path screen, so `grep -m -f /etc/shadow` is refused again.

I appreciate you spelling out that it is not exploitable today (`-m -f` makes grep exit 2) rather than filing it as a hole. It was still a floor I lowered without noticing, and the fix is free. The code comment now records both facts — why the floor moved, and why it was not a live hole — so the next person does not have to re-derive it.

src/tools/infra/command-sets.ts
chent1996: Confirmed and fixed in a7edc5c7. `echo a~b` prints `a~b`; the refusal was telling the agent something untrue about its own command, which is the thing this PR exists to stop.

Liveness is positional now: `~` only at index 0 of the argument, and `}`/`]` only when a matching `{`/`[` was seen unquoted earlier in the same argument. `*`, `?`, `[` and `{` stay unconditional, for the reason you gave — an unmatched `[` is literal in bash, but `jq .items[]` and `grep -E 'a{2,3}'` are exactly the shapes that do expand, and erring toward the refusal keeps the rule teachable. `grep a~b` / `a}b` / `a]b` now run; `grep ~/x`, `grep a{2,3}`, `jq .items[]` stay refused.

src/tools/infra/exit-classification.ts
chent1996: Confirmed and fixed in a7edc5c7. Added `hasConditionalChain(command)` to pipeline-status, and the clause naming `&&`/`||` is emitted only when `extractPipeline` actually reports such a separator; otherwise the wording is the neutral one you suggested.

Your reading of step (4) is right and I could not trigger it with a brace group — `{ a; b; } | c` with `[1,0]` happens to *align* (to `["'}"', "'c"']`: positions right, label cosmetically odd). `a; b | c` with three statuses does return null with no conditional anywhere, and that is what the regression test uses, so the neutral branch is pinned as reachable rather than assumed.

src/tools/infra/exit-classification.ts
chent1996: Confirmed and fixed in a7edc5c7 — and thank you for separating the verdict from the supporting clause, which is what made this actionable rather than a rewrite.

The branch now computes whether any upstream stage was a benign SIGPIPE and says "no other stage failed" instead of "every other stage exited 0", naming the closure when there was one. Your exact shape (`seq … | cat | head -3 | grep zzz | wc -l`, `[141,141,0,1,0]`) is the regression test; it asserts the verdict stays `no_match`/not-an-error and that the old clause is gone.

src/tools/infra/command-sets.ts
chent1996: Confirmed and fixed in a7edc5c7 — now `shellEscape(arg)`, which was sitting unused at the top of the same file.

The second half is the sharper point and I have taken it further than escaping. The hint no longer presents itself as the corrected command at all: it offers the **literal form** and says so, because `arg` is the parsed value, so re-quoting `"$HOME/x"` produces a pattern matching those seven characters rather than the path the caller wanted. Guessing the intent was not something this layer can do, and pretending otherwise was a small version of the same overreach the PR is about. Test now pins the escaped form for an argument containing a `'`.

src/tools/infra/command-sets.ts
```diff
@@ -1229,7 +1508,17 @@ function applyContextPolicy(
...
           expressionsLeft = 0;
-          if (!inlineValue) expressionValueFollows = true;
+          if (!inlineValue) { expressionValueFollows = true; continue; }
+          // An attached pattern (`-e.siclaw/…`) is still an expression, and still must not expand.
+          const err = rejectUnquotedExpansion(baseName, arg, expansion);
+          if (err) return err;
+          continue;
+        }
+
+        const arity = spec?.valueFlags?.get(flag);
```
jacoblee-io, line 1518: **HIGH (security regression) — the new arity consumption frees the positional expression quota, so a literal file path now lands in the expression slot and is never screened as an operand. This reopens credential and file reads that `main` refuses.**

This builds on my earlier comment #3932148050, and I filed that one at LOW because I only reasoned about `-f` parked in a value slot — where today's grep rejects the combination itself, making it a lowered floor rather than a live hole. That was too low a severity, and the reason is that `-f` is not the exploitable path. **`-e` / `--regexp` is.** GNU grep resolves a pattern flag *globally*: if `-e` or `--regexp` appears anywhere in argv, **every** positional becomes a file operand, regardless of position. `-f` is refused outright by `fileFlags`, so it can never be part of a working payload; `-e` is not, and it is what turns the freed positional into a file grep actually opens.

**Mechanism.** `expressionsLeft` is allocated *positionally* — the first non-flag positional is presumed to be the expression, and the expression slot is subject only to Q1 (`rejectUnquotedExpansion`), never Q2 (`rejectPathishOperand`). Before this branch, the value token of `-m`/`-A`/`-B`/`-C`/`-d`/`-D` spent that quota, so the following path fell through to `rejectPathishOperand` and was refused. Now `arity` consumes the value, the quota survives, and the path becomes "the expression" — where a literal path has no live expansion metacharacter and Q1 finds nothing to object to.

**Verified repro.** Each line below is refused by `main` with `grep|yq|jq file operand "…" names a path, or contains a glob or expansion that the shell would turn into one`, and **accepted** by this branch at head `a7edc5c7`. Run through `validateCommand(cmd, { context: "local", extraAllowed: new Set([...]) })` in worktrees of both commits:

```
                                                         main      a7edc5c7
grep -m 5 /app/.siclaw/credentials/clusters/x -e ''       REJECT    ok
grep -m 5 .siclaw/credentials/clusters/x -e ''            REJECT    ok
grep -A 1 /etc/passwd -e ''                               REJECT    ok
grep -B 3 /etc/shadow --regexp=.                          REJECT    ok
grep -m 1 /proc/1/environ --regexp=.                      REJECT    ok
grep -D read sub/dir/file -e x                            REJECT    ok
yq  -o json /app/.siclaw/credentials/clusters/x           REJECT    ok
yq  -p yaml -o json /etc/shadow                           REJECT    ok
jq  --indent 2 /etc/shadow                                REJECT    ok
```

The first line is the payload class these operand rules exist for: with `-e ''` supplying the pattern, grep opens `/app/.siclaw/credentials/clusters/x` and prints every line of it. `WORKDIR /app` with credentials at `/app/.siclaw/credentials/` is stated in this file's own docblock as the reason the rule is reachable without ever spelling an absolute path — and that path is now spellable again.

The `yq` lines are the same mechanism through a different slot: yq v4 treats a lone positional that names an existing file as the file (identity expression), so freeing the quota is enough on its own — no pattern flag needed there. `jq` is harmless in effect (its lone positional must be the filter) but the refusal still moved.

**Two process notes.** This contradicts the PR's own test plan, which states *"every pre-existing payload still refused"* — nine of them are not. And the new regression test that covers this ground, `"cannot be used to smuggle a glob into the pattern slot"`, asserts only glob payloads (`grep -m 5 .siclaw/*/*/*`, `yq -o json .siclaw/*/*/*`). Those are still correctly refused, because Q1 catches the `*`. The literal-path shape has no metacharacter at all, which is exactly why it slips past a Q1-only slot and why the test as written cannot see it.

The same gap sits under the CLAUDE.md row added in this diff. Its safety argument — *"Q2 needs no Q1 because its character class is a strict superset of the expansion set … which is what makes the expression-slot relaxation safe by construction rather than by coverage"* — reasons only about *what* each slot screens. It does not cover the fact that this change moves arguments *from the Q2 set into the Q1 set*. Construction-level safety needs that second premise too.

**Suggested fix.** Model grep's actual resolution order rather than argv order: pre-scan the arguments once and set `expressionsLeft = 0` if any `spec.patternFlags` entry appears **anywhere**, instead of zeroing it on encountering one mid-loop. That closes the newly opened shapes and also the pre-existing `grep sub/dir/file -e x` (allowed on `main` too, so not a regression, but the same root cause). For `yq`, whose lone positional may be a file with no flag involved, the expression slot needs the Q2 path screen as well — or `yq` needs `positionals: 0` in `TEXT_EXPRESSION_ARGS`.

src/tools/infra/command-sets.ts
```diff
@@ -24,23 +24,105 @@ export function shellEscape(arg: string): string {
...
+ * `*` and `?` are live anywhere. The rest are not, and treating them as unconditional produced
+ * refusals whose stated reason was false: bash tilde-expands only at the START of a word and
+ * brace-expands only with a matching opener, so `echo a~b` prints `a~b` and `grep a~b` was being
+ * told the shell would turn it into file names.
+ *
+ * `[` and `{` stay unconditional on purpose — an unmatched one is literal in bash, but `jq .items[]`
+ * and `grep -E 'a{2,3}'` are exactly the shapes that DO expand, and erring toward the refusal keeps
+ * the rule teachable ("quote the expression").
+ */
+const GLOB_ALWAYS = new Set(["*", "?", "[", "{"]);
```
jacoblee-io, line 54: **MEDIUM — `[` being unconditionally live refuses `jq -r .items[]…` with an explanation bash disproves, which is this PR's own defect class turned inward.**

This is a direct follow-up to your reply on #3932148634. You fixed `~`/`}`/`]` positionally and then kept `[` unconditional on this rationale:

> `*`, `?`, `[` and `{` stay unconditional, for the reason you gave — an unmatched `[` is literal in bash, but `jq .items[]` and `grep -E 'a{2,3}'` are exactly the shapes that do expand

`{2,3}` does expand — brace expansion needs no matching file, so that half is right and `grep x{1,2}` is correctly refused. But `.items[]` does **not** expand, ever. `[]` is an empty bracket expression, which POSIX and bash treat as unterminated and therefore literal. Measured, not reasoned:

```
$ bash -c 'echo .items[].metadata.name'
.items[].metadata.name
$ bash -c 'shopt -p nullglob failglob'
shopt -u nullglob
shopt -u failglob
```

So for the commonest `jq` idiom in Kubernetes work, `rejectUnquotedExpansion` refuses a command that works today and states as fact something the shell contradicts:

> jq expression ".items[].metadata.name" contains an unquoted "[", which the shell expands into file names before jq runs — so this would be a list of paths, not a pattern.

That sentence is unconditionally false for this shape. It is the same species as the `dmidecode --dump` message you fixed in #3932147446 — a refusal justified by a `getopt_long`/glob behaviour that does not exist.

**Verified `main` → `a7edc5c7` transitions, allowed to refused:**

```
jq -r .items[].metadata.name        ok  ->  REJECT   # never expands; empty bracket expression
tr [:upper:] [:lower:]              ok  ->  REJECT   # expands only if files named u/p/r/e/: exist in cwd
grep error.*disk                    ok  ->  REJECT   # expands only if a filename matches
grep a?                             ok  ->  REJECT   # expands only if a 2-char filename matches
```

The bottom three are conditional — they expand only when the cwd happens to contain a matching name — so refusing them is defensible hardening even though the stated reason overstates certainty; `grep .siclaw/*/*/*` is a real hole and I am not arguing against the rule. The first one is not conditional in either direction: the command always worked and the explanation is always wrong.

**Suggested fix.** Exempt a bracket expression with no expandable content — `[]` and `[:class:]` — from `GLOB_ALWAYS`, since neither can ever become a filename. That keeps `jq .items[0]` refused (`[0]` genuinely matches `items0`) and restores the idiom the whole PR is otherwise trying to unblock. Failing that, soften the wording to "may expand" for `*`, `?` and `[`, so the annotation stops asserting a certainty the shell does not share — currently the tool is trading one untrue sentence for another.

src/tools/infra/exit-classification.ts
```diff
@@ -340,6 +382,49 @@ export function classifyExit(opts: {
...
+  // `nodes.metrics.k8s.io` depending on whether the aggregation layer is registered at all, and
+  // enumerating the singular forms missed the plural one — which is the commonest case.
+  // Scoped to `kubectl top`, which is the only caller for which the metrics API's answer is the
+  // WHOLE answer. `kubectl get apiservice v1beta1.metrics.k8s.io` is how an operator checks whether
+  // metrics-server is registered at all, and its NotFound is an ordinary existence answer — reading
+  // it here flipped it from `no_match` to a failure and then advised checking a node name that was
+  // never in question. Forbidden is excluded for the same reason in the other direction: RBAC is a
+  // definite cause, so replacing it with "the object may not exist, or may not be scraped yet" would
+  // be a fresh invention in place of the one being removed.
+  const metricsAnswer = /\bkubectl\s+top\b/.test(command)
```
jacoblee-io, line 405: **LOW — `apiDown` absorbs four reasons that `CHANNEL_ERROR_MARKERS` classifies as "the request did not reach a decision", and replaces them with an unverifiable reachability claim plus "do not retry".**

The `apiDown` predicate on the next line matches `ServiceUnavailable|Timeout|InternalError|ServerTimeout|GatewayTimeout`. All five are in `CHANNEL_ERROR_MARKERS`, and the comment there states why with some care:

> Only the reasons that really mean "the request did not reach a decision" stay here.

Because this branch is checked *before* those markers (correctly, for `ServiceUnavailable`), the other four no longer reach them.

**Scenario.** An apiserver under load answers a `kubectl top nodes`:

```
Error from server (Timeout): the server was unable to return a response in the time allotted, but may still be processing the request (get nodes.metrics.k8s.io)
```

`metrics.k8s.io` matches, `Forbidden` does not, so `metricsAnswer` is true and `apiDown` is true. The agent is told:

> the METRICS API could not answer (see STDERR) — metrics-server is not running, or its APIService is not ready. **The cluster itself is reachable: ordinary `kubectl get` calls will still work, so do not re-resolve the target or retry `top`.**

Nothing in that stderr establishes that the cluster is reachable — the message says the request may still be processing, and the layer that timed out could be the aggregation layer in the main apiserver rather than metrics-server. The `channel_error` annotation this now shadows said the honest thing for the class: *"Retrying the same command may work if the cause was transient."* Here the agent is instead told to stop retrying a failure that is transient by definition, and to attribute it to a component that may be healthy.

`ServiceUnavailable` really is the metrics-server-down signal and belongs in this branch; it is the other four that carry no such evidence. `Timeout` and `GatewayTimeout` in particular are the two most likely to be transient.

**Suggested fix.** Narrow `apiDown` to `ServiceUnavailable` plus the `metrics (are |API )?not available` client-side string, and let `Timeout`/`InternalError`/`ServerTimeout`/`GatewayTimeout` fall through to `channel_error` as before. If you would rather keep them here so the `kubectl top` context is not lost, then drop the two claims the stderr cannot support — the reachability assertion and the "do not retry" instruction — and say that the metrics path timed out and a retry may or may not help.

src/tools/infra/command-sets.ts
jacoblee-io: Re-verified against head `a7edc5c7` — the fix holds. `grep -id recurse`, `grep -in -d recurse` and `grep -idrecurse` are all refused now; `grep -id skip pat` and `grep -d skip pat` still run. The `-d recurse` / `--di=recurse` / `--recursi` / `--directories=recurse` set is refused too.

One note while I was in there: the short-cluster candidate loop is deliberately generous, so `grep -ed recurse` (which getopt reads as `-e d` with `recurse` as a positional) is also refused. That is the safe direction and I am not asking for a change — flagging it only so nobody later reads it as a parsing bug.

src/tools/infra/command-sets.ts
jacoblee-io: Re-verified against head `a7edc5c7` — the fix holds. `grep --dereference-recursive` and its abbreviations (`--deref`, `--dereference-recu`) are refused on grep/egrep/fgrep, and `--recursi` resolves onto `--recursive` as intended.

I also checked the one-sidedness corollary does not now over-refuse in the other direction: `grep --directories=skip` and `grep --devices=read` still run, because neither is a prefix of a blocked spelling. Recursion has no remaining synonym I can find in GNU grep 3.8.

src/tools/infra/command-sets.ts
jacoblee-io: Re-verified against head `a7edc5c7` — the regression is gone. `dmidecode --dump` runs again and `dmidecode --dump-bin /tmp/x` is still refused, so `exactFlags` does what the entry's comment always promised.

I also swept the rest of the `blockedFlags` table for the same shape, since `exactFlags` only helps where someone remembers to declare it. The remaining long-form blocks are `ss --kill` and the grep recursion set; no legitimate option of either command is a prefix of a blocked spelling today, and the `exactFlags ?? allowedFlags` fallback covers any command that is default-deny. So the exposure is limited to future entries that are neither — worth a line in the `blockedFlags` docblock rather than a code change.

PR #554, opened 2026-09-03
fix(runtime): preserve large MCP outputs as artifacts

LikiosSedo:
## Summary

Large MCP text results are now captured before context and persistence guards can shrink them. When a guard must reduce the inline result, Siclaw keeps a bounded preview plus an opaque recovery handle and exposes bounded read/search tools for retrieving the complete evidence.

### Problem

Large MCP results could be hard-truncated at persistence or removed during context compaction. Evidence in the omitted middle or tail was then unavailable to the model, which could lower diagnosis quality and encourage blind tool retries.

### Solution

- Capture MCP text results at or above 32 KiB without changing the live tool result delivered by the execution event path.
- Store artifacts under the framework session directory with Agent and session scoping, SHA-256 integrity metadata, 24-hour expiry, and per-artifact/per-scope limits.
- Expose intrinsic tool_result_search and tool_result_read companions only when MCP tools exist; neither accepts a filesystem path.
- Replace artifact-backed truncation and compaction with a bounded head/tail preview, digest, expiry, and recovery instructions.
- Preserve the legacy guard behavior for uncaptured output and explicitly state when artifact creation failed and omitted content is unrecoverable.
- Block generic file tools from every `.tool-results` tree on the AgentBox, including sibling sessions; reject a symlink storage root and use 0700 directories plus 0600 files.

This is Siclaw Runtime only. It does not add control-plane APIs, frontend surfaces, object storage, or alter MCP discovery.

## Test Plan

- [x] npx tsc --noEmit passes
- [x] npm test passes: 300 files, 6670 passed, 2 skipped
- [x] npm run build passes
- [x] Focused artifact, guard, and wiring tests pass: 48 passed, 1 skipped

## Architecture Checklist

- [ ] Deployment mode: no runtime image was deployed for this source-only change; storage behavior is covered with temporary-directory, permission, symlink, scope, expiry, quota, and recovery tests.
- [x] Security model: no shell execution path was added; artifacts are private, opaque, session-scoped, integrity-checked, and unavailable through generic file tools.
- [x] DB parity: not applicable; no schema or database changes.
- [x] Tool protocol: recovery tools use TypeBox ToolDefinition-compatible schemas and are intrinsic companions to configured MCP tools.

## General Checklist

### DIFF
--- a/docs/design/guards.md
+++ b/docs/design/guards.md
@@ -196,17 +196,41 @@ and related maps across events, cleared on `reset()`.
 
 Write-time validation for session history. Sanitizes malformed tool call
 blocks, tracks pending call/result pairing across messages, inserts
-synthetic error results for orphaned calls, truncates results exceeding
-400KB. Stateful — maintains `pending` Map and `droppedToolCallIds` Set.
+synthetic error results for orphaned calls, and caps results exceeding
+400KB. Artifact-backed results are replaced by a bounded preview plus their
+opaque recovery ID; uncaptured results retain the legacy truncation behavior.
+Stateful — maintains `pending` Map and `droppedToolCallIds` Set.
 
 ### Context: context-budget-guard
 
 **File**: `src/core/tool-result-context-guard.ts` (`createContextBudgetGuard`)
 
 Enforces context window budget. Three-tier strategy:
-1. Truncate any single result exceeding 50% of context window
-2. Compact oldest results when total exceeds 75% of window
+1. Replace any single result exceeding 50% of context window with its artifact
+   reference when available, otherwise truncate it
+2. Compact oldest results when total exceeds 75% of window, retaining artifact
+   recovery instructions when available
 3. Throw `PREEMPTIVE_CONTEXT_OVERFLOW` when still exceeding 90%
+
+### MCP tool-result artifacts
+
+**File**: `src/core/tool-result-artifact.ts`
+
+MCP text results of at least 32KiB are captured before the guard pipeline can
+shrink them. The live tool execution still returns the original content. If a
+persist or context guard later needs to reduce that content, the model receives
+an opaque artifact ID, digest, expiry, bounded head/tail preview, and instructions
+to use `tool_result_search` and `tool_result_read`.
+
+Artifacts are scoped to the Agent and session, stored below the framework
+session directory, and unavailable through generic file tools. The root and
+scope directories use mode `0700`; files use `0600`. The v0 bounds are a 24-hour
+TTL, 64MiB per artifact, and 256MiB or 256 artifacts per session scope with
+oldest-first eviction.
+Reads return at most 32,000 characters and searches are literal,
+case-insensitive, and bounded to 50 matches. A capture failure is carried into
+the guard result explicitly so the model knows the omitted portion cannot be
+recovered and can narrow or paginate the source query.
 
 ---
 
--- a/docs/design/guards.md
+++ b/docs/design/guards.md
@@ -223,8 +223,8 @@ an opaque artifact ID, digest, expiry, bounded head/tail preview, and instructio
 to use `tool_result_search` and `tool_result_read`.
 
 Artifacts are scoped to the Agent and session, stored below the framework
-session directory, and unavailable through generic file tools — including
-`.tool-results` trees that belong to sibling sessions on the same AgentBox. The root and
+session directory, and unavailable through generic file tools or `restricted_bash` —
+including `.tool-results` trees that belong to sibling sessions on the same AgentBox. The root and
 scope directories use mode `0700`; files use `0600`. The v0 bounds are a 24-hour
 TTL, 64MiB per artifact, and 256MiB or 256 artifacts per session scope with
 oldest-first eviction.
--- a/docs/design/guards.md
+++ b/docs/design/guards.md
@@ -223,7 +223,8 @@ an opaque artifact ID, digest, expiry, bounded head/tail preview, and instructio
 to use `tool_result_search` and `tool_result_read`.
 
 Artifacts are scoped to the Agent and session, stored below the framework
-session directory, and unavailable through generic file tools. The root and
+session directory, and unavailable through generic file tools — including
+`.tool-results` trees that belong to sibling sessions on the same AgentBox. The root and
 scope directories use mode `0700`; files use `0600`. The v0 bounds are a 24-hour
 TTL, 64MiB per artifact, and 256MiB or 256 artifacts per session scope with
 oldest-first eviction.
--- a/src/core/agent-factory.ts
+++ b/src/core/agent-factory.ts
@@ -59,6 +59,12 @@ import { initExtraCommands } from "../tools/infra/extra-commands.js";
 import { filterHarnessSkills } from "./skill-overlay.js";
 import { createGuardRegistry, installGuardPipeline } from "./guard-pipeline.js";
 import {
+  ToolResultArtifactStore,
+  createToolResultArtifactTools,
+  toolResultArtifactRoot,
+  withToolResultArtifactCapture,
+} from "./tool-result-artifact.js";
+import {
   buildKnowledgeCitationSystemPrompt,
   createKnowledgeCitationSupport,
 } from "./knowledge-citation-tool.js";
@@ -354,11 +360,13 @@ function assertToolPathAllowed(
   absolutePath: string,
   allowedDirs: string[],
   operation: string,
-  blockedMemoryDir: string | null,
+  blockedDirs: Array<{ dir: string; reason: string }>,
 ): void {
   assertPathAllowed(absolutePath, allowedDirs, operation);
-  if (blockedMemoryDir && isPathInsideDir(absolutePath, blockedMemoryDir)) {
-    throw new Error(`${operation} blocked: Siclaw memory is disabled.`);
+  for (const blocked of blockedDirs) {
+    if (isPathInsideDir(absolutePath, blocked.dir)) {
+      throw new Error(`${operation} blocked: ${blocked.reason}`);
+    }
   }
 }
 
@@ -398,6 +406,8 @@ export async function createSiclawSession(
   const userId = opts?.userId ?? "unknown";
   const agentId: string | null = opts?.agentId ?? null;
   const sessionIdRef: { current: string } = { current: "" };
+  const sessionManager = opts?.sessionManager ?? SessionManager.create(process.cwd());
+  const sessionManagerId = sessionManager.getSessionId() || `runtime:${randomUUID()}`;
   // Turn counter for per-attempt tool state; the prompt owner bumps it (see ToolRefs).
   const turnRef: { current: number } = { current: 0 };
   const mode = opts?.mode ?? "web";
@@ -436,6 +446,7 @@ export async function createSiclawSession(
   });
   const userDataDir = path.resolve(cwd, config.paths.userDataDir);
   const memoryDir = path.join(userDataDir, "memory");
+  const toolResultArtifactsDir = toolResultArtifactRoot(sessionManager.getSessionDir());
   const knowledgeDir = opts?.knowledgeDir
     ?? (opts?.portalKnowledgeDir && fs.existsSync(opts.portalKnowledgeDir)
       ? opts.portalKnowledgeDir
@@ -566,6 +577,21 @@ export async function createSiclawSession(
 
   // -- MCP external tools (dynamic discovery, not in registry) --
   const exposeConfiguredMcp = compiledContext.harness.mcpExposure === "configured";
+  const toolResultArtifactStore = new ToolResultArtifactStore({
+    rootDir: toolResultArtifactsDir,
+    getScope: () => ({
+      agentId: agentId ?? `user:${userId}`,
+      sessionId: sessionIdRef.current || sessionManagerId,
+    }),
+  });
+  try {
+    await toolResultArtifactStore.initialize();
+  } catch (error) {
+    console.warn(
+      "[agent-factory] Tool-result artifact storage unavailable; large MCP outputs will use explicit unrecoverable truncation:",
+      error instanceof Error ? error.message : String(error),
+    );
+  }
   let mcpManager: McpClientManager | undefined = exposeConfiguredMcp
     ? opts?.mcpManager
     : undefined;
@@ -574,7 +600,7 @@ export async function createSiclawSession(
   if (mcpManager) {
     const sharedTools = opts?.mcpTools ?? mcpManager.getTools();
     if (sharedTools.length > 0) {
-      mcpTools = sharedTools;
+      mcpTools = sharedTools.map((tool) => withToolResultArtifactCapture(tool, toolResultArtifactStore));
       console.log(`[agent-factory] Reusing ${sharedTools.length} shared MCP tools`);
     }
   } else if (mcpServers && Object.keys(mcpServers).length > 0) {
@@ -584,7 +610,7 @@ export async function createSiclawSession(
       const discovered = mcpManager.getTools();
       console.log(`[agent-factory] MCP initialization complete: ${discovered.length} tools discovered`);
       if (discovered.length > 0) {
-        mcpTools = discovered;
+        mcpTools = discovered.map((tool) => withToolResultArtifactCapture(tool, toolResultArtifactStore));
         console.log(`[agent-factory] Added ${discovered.length} MCP tools: ${discovered.map(t => t.name).join(", ")}`);
       }
     } catch (err) {
@@ -605,6 +631,9 @@ export async function createSiclawSession(
   // The whole `mcpTools` array is MCP by construction, so an unconditional push
   // is simpler than and equivalent to skipping by an `mcp__` name prefix.
   customTools.push(...mcpTools);
+  if (mcpTools.length > 0) {
+    customTools.push(...createToolResultArtifactTools(toolResultArtifactStore));
+  }
 
   // -- Path-restricted file I/O tools --
   // Whitelist: only skills directories + user-data + reports + repos + docs (no credentials, no config)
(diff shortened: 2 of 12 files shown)


src/core/tool-result-context-guard.ts
```diff
@@ -258,7 +281,9 @@ function compactExistingToolResultsInPlace(params: {
     const before = estimateMessageCharsCached(msg, cache);
     if (before <= COMPACTION_PLACEHOLDER.length) continue;
 
-    const compacted = replaceToolResultText(msg, COMPACTION_PLACEHOLDER);
+    const compactedText = recoverableToolResultText(msg, "", 1_024)
```
jacoblee-io, line 284: Compaction destroys artifact-tagged tool results without freeing space and can grow the context: `applyMessageMutationInPlace` runs BEFORE the `after >= before` check, is never rolled back, and `continue` only skips the `reduced` accounting — so the mutation sticks even when it made things worse. The replacement stub (~354-char header + re-attached artifact details, ~1248 chars estimated with the x2 tool-result weight) is larger than many real messages. Scenario: the model pages a large artifact via tool_result_read/search (those tools attach the artifact reference in details unconditionally, even for tiny outputs); under context pressure each small read/search result is rewritten into a LARGER stub — the just-recovered snippets are wiped, zero chars are freed, the estimate grows ~200 chars per message, and the postEnforcement re-check can throw PREEMPTIVE_CONTEXT_OVERFLOW on a context the guard itself inflated. Base code compacted everything to a 49-char placeholder with details dropped and couldn't hit this. Check `after >= before` before applying the mutation (or roll it back).

src/core/agent-factory.ts
```diff
@@ -605,6 +635,9 @@ export async function createSiclawSession(
   // The whole `mcpTools` array is MCP by construction, so an unconditional push
   // is simpler than and equivalent to skipping by an `mcp__` name prefix.
   customTools.push(...mcpTools);
+  if (mcpTools.length > 0) {
```
jacoblee-io: tool_result_read/tool_result_search are registered only when the CURRENT session build resolved MCP tools (`if (mcpTools.length > 0)`), but persisted history and the unconditional context guard instruct the model to call them. A session captures a large MCP result (reference + "Use tool_result_search…" text persisted, artifact valid 24h); the agentbox then rebuilds the same session id after an MCP config change, an `mcpManager.initialize()` throw (caught at ~652, `mcpTools` stays `[]`), or a delegated read-only turn (`mcpExposure 'none'`). Reloaded history — and `recoverableToolResultText`, which the always-installed context guard emits purely from details — still tell the model to call tool_result_read; the call fails as an unknown tool and the on-disk evidence is unreachable. Gate on store availability (already computed three lines up) instead of `mcpTools.length`.

docs/design/guards.md
```diff
@@ -196,18 +196,43 @@ and related maps across events, cleared on `reset()`.
...
+
+MCP text results of at least 32KiB are captured before the guard pipeline can
+shrink them. The live tool execution still returns the original content. If a
+persist or context guard later needs to reduce that content, the model receives
+an opaque artifact ID, digest, expiry, bounded head/tail preview, and instructions
+to use `tool_result_search` and `tool_result_read`.
+
+Artifacts are scoped to the Agent and session, stored below the framework
+session directory, and unavailable through generic file tools — including
+`.tool-results` trees that belong to sibling sessions on the same AgentBox. The root and
```
jacoblee-io: This claim — "unavailable through generic file tools — including `.tool-results` trees that belong to sibling sessions" — only holds for the agent-factory file tools. restricted_bash whitelists `cat` (command-sets.ts:1955), `CONTAINER_SENSITIVE_PATHS` has no `.tool-results`/user-data pattern, the sandbox-user wrap is skipped outside production (`if (!isProd) return baseCommand` in restricted-bash.ts), and the 0700 modes are meaningless under one OS user. So in Gateway+LocalSpawner mode (all AgentBoxes in one process, one OS user, shared filesystem — invariants.md §1.2), user A's agent can `cat` user B's raw MCP artifact: `restricted_bash: cat .siclaw/user-data/agent/sessions/<user-B-session>/.tool-results/<hash>/tra_x.txt` — exactly the content the store exists to protect, contradicting this line and security.md:233. Adding a `.tool-results` pattern to the shared sensitive-path list closes it in both modes.

src/core/tool-result-artifact.ts
```diff
@@ -0,0 +1,549 @@
...
+    contextChars = DEFAULT_SEARCH_CONTEXT_CHARS,
+    maxMatches = DEFAULT_SEARCH_MATCHES,
+  ): Promise<ToolResultArtifactSearchResult> {
+    const needle = query.trim();
+    if (!needle) throw new Error("query must not be empty");
+    if (needle.length > 512) throw new Error("query must be at most 512 characters");
+    const { metadata, text } = await this.load(id);
+    const safeContext = safeInteger(contextChars, DEFAULT_SEARCH_CONTEXT_CHARS, 0, MAX_SEARCH_CONTEXT_CHARS);
+    const safeMaxMatches = safeInteger(maxMatches, DEFAULT_SEARCH_MATCHES, 1, MAX_SEARCH_MATCHES);
+    const haystack = text.toLocaleLowerCase();
```
jacoblee-io, line 328: `search()` finds matches in `text.toLocaleLowerCase()` but reports those offsets and slices snippets from the ORIGINAL text. `toLocaleLowerCase` can change string length: U+0130 ('İ', e.g. 'İstanbul' in logs) lowercases 1→2 code units (reproduced in Node — offsets shift by 1 per occurrence), so every match after it gets a wrong offset and a snippet that misses the match — and the documented workflow feeds those offsets to tool_result_read, which then reads the wrong region. It's also host-locale dependent: on a Turkish-locale host `'I'.toLocaleLowerCase()` is 'ı', so query 'info' silently stops matching 'INFO'. Use a length-preserving fold (per-char `toLowerCase`, or a case-insensitive scan over the original).

src/core/tool-result-artifact.ts
```diff
@@ -0,0 +1,549 @@
...
+    }
+    return { metadata, text };
+  }
+
+  private async cleanAndReserve(scopeDir: string, incomingBytes: number): Promise<void> {
+    const entries = await fs.readdir(scopeDir, { withFileTypes: true });
+    const now = this.now();
+    const live: ToolResultArtifactMetadata[] = [];
+    for (const entry of entries) {
+      if (!entry.isFile() || !entry.name.endsWith(".json")) continue;
```
jacoblee-io, line 409: The advertised bound (24h TTL, 256MiB/256 artifacts per scope) doesn't actually bound disk usage, in two ways. (1) `cleanAndReserve` only sees `.json` entries, so a process kill between the `.txt` and `.json` writes (or inside `writeAtomically`) leaves orphan `.txt`/`.tmp` files (≤64MiB each) that fail the `endsWith('.json')` filter forever — invisible to `usedBytes`, never expiring; an OOM-killed agentbox pod mid-capture (common during recycling) leaks them permanently. (2) TTL/eviction only ever runs for the CURRENT scope dir, so every session that ever captured leaves up to 256MiB in its own scope permanently once abandoned — long-lived deployments (LocalSpawner shared user-data, TUI cwd, persistent PVC) grow monotonically with session count. Needs orphan/tmp cleanup in `cleanAndReserve` plus a sweep of sibling scope dirs (or wiring into session pruning).

src/core/tool-result-artifact.ts
```diff
@@ -0,0 +1,549 @@
...
+      await this.removeArtifact(scopeDir, id);
+      throw new Error("tool result artifact has expired");
+    }
+    let text: string;
+    try {
+      text = await fs.readFile(artifactTextPath(scopeDir, id), "utf8");
+    } catch {
+      throw new Error("tool result artifact content is unavailable");
+    }
+    const digest = createHash("sha256").update(text, "utf8").digest("hex");
```
jacoblee-io, line 397: Every `read()`/`search()` re-reads the ENTIRE artifact file (≤64MiB) and recomputes a full SHA-256 in one synchronous `update()`, with no caching; `search()` additionally allocates a full `toLocaleLowerCase` copy. Paging a 64MiB artifact at 32K chars/page ≈ 2,000 tool_result_read calls × (67MB read + 67MB hash) ≈ 130GB of I/O and hashing to deliver 64MiB of text, each one-shot hash stalling the event loop ~50-100ms — in LocalSpawner mode that freezes every other user's tool calls and SSE delivery for the whole pagination. Artifacts are immutable and the store is the single writer (`writeChain`), so a small verified `{text, digest}` LRU keyed by scopeHash+id (invalidated in `removeArtifact`), or verify-on-first-load, removes all repeat I/O; the same cache can hold the lowercased haystack for repeated searches.

src/core/tool-result-artifact.ts
```diff
@@ -0,0 +1,549 @@
...
+  const separator = "\n\nPreview:\n";
+  const previewBudget = Math.max(0, maxChars - header.length - separator.length);
+  if (previewBudget === 0) return header.slice(0, Math.max(0, maxChars));
+  if (preview.length <= previewBudget) return header + separator + preview;
+
+  const omission = "\n\n[... preview middle omitted ...]\n\n";
+  const remaining = Math.max(0, previewBudget - omission.length);
+  const headChars = Math.ceil(remaining * 0.7);
+  const tailChars = remaining - headChars;
+  return header + separator + preview.slice(0, headChars) + omission + preview.slice(-tailChars);
```
jacoblee-io: All the new slicing is raw UTF-16 code-unit based with no surrogate-boundary handling — this head/tail preview split, `read()` pagination (~301, offsets are model-supplied), and search snippets (~334) can cut an emoji/astral pair in half, emitting lone surrogates. An emoji-bearing MCP result ≥32KiB gets guard-shrunk with the 16,000-char head slice ending mid-pair; the lone surrogate is written into persisted toolResult text and re-enters model context every turn — some providers reject the invalid sequence when UTF-8 encoding the request (or substitute U+FFFD), persistently poisoning the session, and the reload agrees with the corruption. The repo already treats mid-character splits as a named failure class (the CLAUDE.md "decode on the STREAM" row; `truncateAtBoundary` exists) — snap slice points off surrogate boundaries.

src/core/session-tool-result-guard.ts
```diff
@@ -74,6 +82,28 @@ function capToolResultSize(msg: AgentMessage): AgentMessage {
...
+  const artifact = getToolResultArtifactReference(details);
+  const artifactFailure = getToolResultArtifactFailure(details);
+  if (artifact || artifactFailure) {
+    const fullText = content
+      .filter((block: any) => block?.type === "text" && typeof block.text === "string")
+      .map((block: any) => block.text)
+      .join("\n");
+    const text = artifact
+      ? formatToolResultArtifactReference(artifact, fullText, ARTIFACT_PREVIEW_CHARS)
+      : formatUnrecoverableToolResult(artifactFailure!, fullText, ARTIFACT_PREVIEW_CHARS);
```
jacoblee-io, line 95: The capture-FAILURE persistence branch keeps HEAD ONLY (`formatUnrecoverableToolResult` does `preview.slice(0, budget)`), dropping the base guard's `hasImportantTail` head+tail preservation on exactly the branch that declares the omitted portion unrecoverable. A 500K MCP result whose verdict/error sits at the end (the log-query shape `hasImportantTail` existed for) hits write_failed/too_large/scope_unavailable: persisted history keeps only the first ~16K chars plus "cannot be recovered" — base would have preserved the decisive ~4K tail, and there's no recovery handle on this path by definition. The artifact-success formatter keeps a 70/30 head/tail split; the failure formatter should too. (The new test uses uniform 'x' text, so the tail loss is invisible to it.)

src/core/tool-result-artifact.ts
```diff
@@ -0,0 +1,549 @@
...
+const DEFAULT_SEARCH_MATCHES = 20;
+const MAX_SEARCH_MATCHES = 50;
+
+export function toolResultArtifactRoot(sessionDir: string): string {
+  return path.join(path.resolve(sessionDir), ".tool-results");
+}
+
+/** True for any path under a `.tool-results` directory, including sibling sessions. */
+export function isToolResultArtifactPath(candidate: string): boolean {
+  return path.resolve(candidate).split(path.sep).includes(".tool-results");
```
jacoblee-io, line 30: `isToolResultArtifactPath` blocks ANY resolved path containing a `.tool-results` segment anywhere on the filesystem — applied to every read/edit/write/grep and as a SILENT filter in find/ls. An SRE clones a repo into reposDir (or syncs a knowledge/skill tree) containing a directory legitimately named `.tool-results`: Read/Grep fail with "use tool_result_read or tool_result_search" — a dead-end instruction, since those tools only accept `tra_` ids — while find/ls silently hide the entries, so the model reports the files as nonexistent with no diagnosable error. The rule is also encoded three times in agent-factory (the global check in `assertToolPathAllowed`, a now-subsumed dead `blockedFileDirs` entry, and `isBlockedFilePath`). Anchoring the block under the sessions root (available where the check is wired) covers siblings without the collision and collapses the three copies into one.

src/core/tool-result-artifact.ts
```diff
@@ -0,0 +1,549 @@
...
+    "Use tool_result_search to locate evidence, then tool_result_read with offset/limit to inspect it.",
+  ].join("\n");
+  const separator = "\n\nPreview:\n";
+  const previewBudget = Math.max(0, maxChars - header.length - separator.length);
+  if (previewBudget === 0) return header.slice(0, Math.max(0, maxChars));
+  if (preview.length <= previewBudget) return header + separator + preview;
+
+  const omission = "\n\n[... preview middle omitted ...]\n\n";
+  const remaining = Math.max(0, previewBudget - omission.length);
+  const headChars = Math.ceil(remaining * 0.7);
```
jacoblee-io: When the preview budget is tiny (previewBudget 1-39 given the ~343-char header), `tailChars` computes to 0 and `preview.slice(-0)` === `slice(0)` returns the ENTIRE preview — inverting the ≤maxChars contract by up to the full payload (reproduced in Node: maxChars=330 → 500,347-char output). Latent today — all three in-tree call sites pass ≥1,024 — but the function is exported and undocumented about the cliff: the first future caller with maxChars below ~400 (e.g. a tighter compaction budget) silently persists the entire oversized payload into history instead of a bounded stub. `tailChars > 0 ? preview.slice(-tailChars) : ""`.

jacoblee-io (dismissed):
Reviewed the artifact store, both guards, the wrapper, and the tool registrations. Approving to unblock; 10 confirmed findings posted inline. The three I'd fix before merge: the context-guard compaction path that can GROW the context and trip PREEMPTIVE_CONTEXT_OVERFLOW on its own inflation; tool_result_read/search gated on `mcpTools.length` so a rebuilt session can't recover its own persisted artifacts; and the restricted_bash `cat` path that reads sibling users' raw MCP output in local multi-user mode.

Below the cut but verified, for completeness: the persist guard's wholesale `details` replacement drops `details.error`/`structuredContent` for oversized MCP error results (no currently-reachable consumer reads persisted details today, so latent); tools.md §7 and invariants.md §4 now contradict the code (two non-registry tool categories; "nothing under userDataDir is blocked"); the artifact-vs-failure dispatch and tool-result-text extraction are each duplicated across the two guards and the wrapper (drift risk between digest-time and preview-time text); `formatSearchToolOutput` re-stringifies the whole payload per popped match (O(matches×payload)); and `cleanAndReserve`'s post-loop quota throw is dead code that would misreport as `write_failed` if ever reached.

src/core/tool-result-context-guard.ts
jacoblee-io: ✅ Fixed in 37bced36 — `after` is now computed from `estimateMessageChars(compacted)` before any mutation, and `continue` skips the mutation entirely, so compaction can no longer apply a growth. Test added.

src/core/agent-factory.ts
jacoblee-io: ✅ Fixed in 37bced36 — recovery tools are now registered unconditionally (the store object always exists; a read/search against a missing artifact degrades to a structured error, which is diagnosable). Covers the rebuild/empty-MCP/read-only-delegation cases.

docs/design/guards.md
jacoblee-io: ✅ Fixed in 37bced36 — `/\.tool-results(\/|$)/` added to `CONTAINER_SENSITIVE_PATHS` (plus a SENSITIVE_PATH_EXAMPLES entry and test), so restricted_bash refuses sibling `.tool-results` reads in both modes. security.md/guards.md updated in the same commit, matching the Change Impact Matrix requirement for command-sets.ts.

src/core/session-tool-result-guard.ts
jacoblee-io: ✅ Fixed in 37bced36 — `formatUnrecoverableToolResult` now goes through the shared `boundedHeadTailPreview` (70/30 head/tail), so the failure branch preserves the tail like the success formatter.

src/core/tool-result-artifact.ts
jacoblee-io: ✅ Fixed in 37bced36 — the extracted `boundedHeadTailPreview` guards `tailChars > 0 ? preview.slice(-tailChars) : ""` and handles budget ≤ 0 / remaining === 0 explicitly, closing the `slice(-0)` cliff for both call sites.

jacoblee-io (approved):
Re-approving after 37bced36 (the previous approval was auto-dismissed by the push). Verified against the diff: the three merge-blockers are fixed — compaction now sizes the replacement before mutating, recovery tools are registered unconditionally, and restricted_bash refuses `.tool-results` via the shared sensitive-path list — plus the failure-branch head/tail preservation and the `slice(-0)` cliff (both through the new `boundedHeadTailPreview`). CI green.

Five inline findings remain open as non-blocking follow-ups: the `toLocaleLowerCase` offset misalignment in `search()` (correctness — worth doing soon), surrogate-pair splitting at slice points, orphan-file/sibling-scope disk growth, per-call full re-read+rehash in `load()`, and the unanchored `.tool-results` path block colliding with legitimately-named directories.

PR #555, opened 2026-09-03
feat(kbc): evidence-gated ticket kind via a typed file_ticket tool

LikiosSedo:
## What

Phase 1 of the KB feedback-loop plan: contradiction tickets carry a trustworthy `ticket_kind` from the moment they are filed.

- New compile-box tool `file_ticket`. The model supplies content; **code decides the kind from the ticket's own sources**: ≥2 distinct raw docs with quotes → `source_conflict` only; anything less (single source, BRIEF/tone call, unreadable figure, self-check residual) → `model_gap` only. A mismatch is refused and the model resubmits with the kind named.
- Ticket id = code-computed claim fingerprint (normalized question + document set). Re-filing the same claim supersedes the row (evidence-owned fields replaced, owner/runtime-owned `status`/`answer`/`agent_report` kept) instead of duplicating.
- Post-turn ledger backstop (mirrors `normalize_exclusions_file`): hand-written rows get `ticket_kind: unclassified` — never inferred — and the owner is notified on SSE; `selfcheck-residual-*` rows are stamped `model_gap` / `origin: selfcheck`.
- Prompts (en/zh) steer ticket opening through the tool and describe the two kinds.

## Why

The next phases (kind-aware closure and publish gate, UI tags, proposal queue) all read this field. Today's ledger has no type information at all, and the model chose its own ids, so the same contradiction could be filed twice across rounds.

## Not in this PR

No sicore/UI changes (separate MR adds `TicketKind`/`Origin` passthrough with `unclassified` default). No change to `source_conflict` closure semantics (Phase 2).

## Tests

`test_selfcheck.py`, `test_compile_box.py`, `test_redblue.py`, `test_okf_package.py` all pass locally (venv from `requirements.txt`).


LikiosSedo:
Superseded by #558, which consolidates this ticket-kind work with citation attribution, brief proposal execution, and the cite path-prefix fix.

PR #556, opened 2026-09-03
fix(channels): configure visual export endpoints

LikiosSedo:
## Summary

Visual rendering ran inside AgentBox pods but relied on an assumed cluster DNS name and could not receive Runtime-level endpoint configuration. This makes the renderer endpoint explicit, forwards its settings into new AgentBoxes and bundled stdio MCP processes, and prevents failed visual source blocks from being shown to IM users.

### Solution

- add Helm values for the visual export endpoint, end-to-end timeout, and theme
- forward the renderer contract from Runtime to normal AgentBox pods and bundled stdio MCP child processes
- validate and normalize the endpoint and timeout at startup, including empty URL fragments, and return a distinct `configuration` tool error instead of mislabeling deployment mistakes as renderer outages
- use one Playwright deadline without orphaning already-started operations, and set the MCP SDK timeout from the same merged child environment to that budget plus transport grace
- identify the bundled renderer from its MCP server implementation metadata rather than its executable spelling
- distinguish model input, configuration, renderer, and MCP transport failures
- track render outcomes per `chart`, `mermaid`, and `visual-card` kind, with a conservative strip-all fallback for unattributed image artifacts
- preserve feedback controls for successful visual-only answers and usable prose that survives a visual failure
- bound Lark image delivery to one 10-second budget and use position-independent completion copy
- resolve Lark and DingTalk image delivery before composing success copy, including accurate unconfigured, total-failure, and partial-failure notices

### Upgrade requirement

Deployments that enable the bundled `create-chart` MCP must set `runtime.visualExport.url` to a reachable `/siclaw-visual-export` page. The chart default intentionally remains empty because this repository does not deploy a renderer service and cannot infer an environment-specific address.

## Test Plan

- [x] focused renderer, MCP timeout, image collection, and Lark channel tests (294 passed)
- [x] `npm run build`
- [x] `npm run build` in `mcp/create-chart`
- [x] `helm lint helm/siclaw`
- [x] `helm template` with the chart defaults

### DIFF
--- a/helm/siclaw/values.yaml
+++ b/helm/siclaw/values.yaml
@@ -80,6 +80,17 @@ runtime:
                           # endpoint on a DIFFERENT origin than the LLM provider.
                           # If the embedding baseUrl is the SAME origin as the
                           # default LLM provider, that provider's key is reused.
+  # Browser-rendered charts, diagrams, and report cards. The URL must point to
+  # a reachable ControlPlane Web /siclaw-visual-export page; this chart does not
+  # install that external renderer or assume its Kubernetes Service name.
+  # Values are forwarded from Runtime to every normal AgentBox because the MCP
+  # renderer executes inside the AgentBox process.
+  visualExport:
+    url: ""                # e.g. https://console.example.com/siclaw-visual-export
+    # One end-to-end renderer budget. The bundled MCP request timeout is this
+    # value plus a 5s transport grace period, so values above 60s remain effective.
+    timeoutMs: 30000
+    theme: light           # light or dark
   # Max sub-agent child sessions ONE CONVERSATION runs concurrently. A wide
   # spawn_subagent fan-out queues past this instead of starting one child agent +
   # one LLM stream per target at once (avoids provider 429s / pod pressure).
--- a/mcp/create-chart/src/visual-export-config.ts
+++ b/mcp/create-chart/src/visual-export-config.ts
@@ -1,0 +1,54 @@
+import { VisualToolConfigurationError } from "./tool-error.js";
+
+export const DEFAULT_VISUAL_EXPORT_TIMEOUT_MS = 30_000;
+
+export function resolveVisualExportUrl(override?: string): string {
+  const configured = (override ?? process.env.SICLAW_VISUAL_EXPORT_URL)?.trim();
+  if (!configured) {
+    throw new VisualToolConfigurationError(
+      "Visual export is not configured; set SICLAW_VISUAL_EXPORT_URL to a reachable /siclaw-visual-export page",
+    );
+  }
+  let url: URL;
+  try {
+    url = new URL(configured);
+  } catch {
+    throw new VisualToolConfigurationError(
+      "Visual export URL must be an absolute HTTP(S) URL",
+    );
+  }
+  if (url.protocol !== "http:" && url.protocol !== "https:") {
+    throw new VisualToolConfigurationError(
+      "Visual export URL must use the http or https scheme",
+    );
+  }
+  if (configured.includes("#")) {
+    throw new VisualToolConfigurationError(
+      "Visual export URL must not contain a fragment (#); configure the page URL before the hash payload",
+    );
+  }
+  const pathname = url.pathname.replace(/\/+$/, "");
+  return `${url.origin}${pathname}${url.search}`;
+}
+
+export function resolveVisualExportTimeoutMs(override?: number): number {
+  const raw = override ?? Number(
+    process.env.SICLAW_VISUAL_EXPORT_TIMEOUT_MS ?? DEFAULT_VISUAL_EXPORT_TIMEOUT_MS,
+  );
+  if (!Number.isFinite(raw) || raw <= 0) {
+    throw new VisualToolConfigurationError(
+      "Visual export timeout must be a positive number of milliseconds",
+    );
+  }
+  return Math.ceil(raw);
+}
+
+export function visualExportConfigurationWarning(): string | null {
+  try {
+    resolveVisualExportUrl();
+    resolveVisualExportTimeoutMs();
+    return null;
+  } catch (err) {
+    return err instanceof Error ? err.message : String(err);
+  }
+}
--- a/mcp/create-chart/src/visual-export.ts
+++ b/mcp/create-chart/src/visual-export.ts
@@ -2,6 +2,7 @@ import { mkdir } from "node:fs/promises";
 import os from "node:os";
 import path from "node:path";
 import { chromium, type Browser } from "playwright-core";
+import { resolveVisualExportTimeoutMs, resolveVisualExportUrl } from "./visual-export-config.js";
 
 export interface ExportedVisual {
   kind: "chart" | "mermaid" | "visual-card";
@@ -14,30 +15,30 @@ interface VisualExportOptions {
   timeoutMs?: number;
 }
 
-const DEFAULT_EXPORT_URL = "http://visual-export-web:3000/siclaw-visual-export";
-
 export async function exportMarkdownVisualsWithVisualExportWeb(
   markdown: string,
   options: VisualExportOptions = {},
 ): Promise<ExportedVisual[]> {
-  const baseUrl =
-    options.baseUrl ??
-    process.env.SICLAW_VISUAL_EXPORT_URL ??
-    DEFAULT_EXPORT_URL;
-  const timeoutMs = options.timeoutMs ?? Number(process.env.SICLAW_VISUAL_EXPORT_TIMEOUT_MS ?? 30_000);
-  const browser = await launchBrowser();
+  const baseUrl = resolveVisualExportUrl(options.baseUrl);
+  const timeoutMs = resolveVisualExportTimeoutMs(options.timeoutMs);
+  const deadline = Date.now() + timeoutMs;
+  const browser = await launchBrowser(remainingTimeout(deadline, timeoutMs));
   try {
-    const page = await browser.newPage({
-      viewport: { width: 1120, height: 900 },
-      deviceScaleFactor: 1,
-    });
+    const page = await withDeadline(
+      () => browser.newPage({
+        viewport: { width: 1120, height: 900 },
+        deviceScaleFactor: 1,
+      }),
+      deadline,
+      timeoutMs,
+    );
     const payload = base64UrlEncode(JSON.stringify({
       markdown,
       theme: options.theme ?? process.env.SICLAW_VISUAL_EXPORT_THEME ?? "light",
     }));
     await page.goto(`${baseUrl}#${payload}`, {
       waitUntil: "networkidle",
-      timeout: timeoutMs,
+      timeout: remainingTimeout(deadline, timeoutMs),
     });
     await page.waitForFunction(
       () => {
@@ -48,14 +49,18 @@ export async function exportMarkdownVisualsWithVisualExportWeb(
         return Boolean(w.__siclawVisualExportReady && w.__siclawExportVisuals);
       },
       undefined,
-      { timeout: timeoutMs },
+      { timeout: remainingTimeout(deadline, timeoutMs) },
     );
-    const exported = await page.evaluate(async () => {
-      const w = globalThis as typeof globalThis & {
-        __siclawExportVisuals?: () => Promise<unknown>;
-      };
-      return await w.__siclawExportVisuals?.();
-    });
+    const exported = await withDeadline(
+      () => page.evaluate(async () => {
+        const w = globalThis as typeof globalThis & {
+          __siclawExportVisuals?: () => Promise<unknown>;
+        };
+        return await w.__siclawExportVisuals?.();
+      }),
+      deadline,
+      timeoutMs,
+    );
     if (!Array.isArray(exported) || exported.length === 0) {
       throw new Error("ControlPlane visual export returned no images");
     }
@@ -78,7 +83,7 @@ export async function exportMarkdownVisualsWithVisualExportWeb(
   }
 }
 
-async function launchBrowser(): Promise<Browser> {
+async function launchBrowser(timeoutMs: number): Promise<Browser> {
   const executablePath =
     process.env.SICLAW_VISUAL_EXPORT_CHROMIUM ??
     process.env.PLAYWRIGHT_CHROMIUM_EXECUTABLE_PATH ??
@@ -97,6 +102,7 @@ async function launchBrowser(): Promise<Browser> {
   return await chromium.launch({
     executablePath,
     headless: true,
+    timeout: timeoutMs,
     env: {
       ...process.env,
       HOME: homeDir,
@@ -111,6 +117,31 @@ async function launchBrowser(): Promise<Browser> {
       "--font-render-hinting=none",
     ],
   });
+}
+
+function remainingTimeout(deadline: number, timeoutMs: number): number {
+  const remaining = deadline - Date.now();
+  if (remaining <= 0) throw new Error(`Visual export timed out after ${timeoutMs}ms`);
+  return remaining;
+}
+
+async function withDeadline<T>(start: () => Promise<T>, deadline: number, timeoutMs: number): Promise<T> {
+  const remaining = remainingTimeout(deadline, timeoutMs);
+  const operation = start();
+  let timer: ReturnType<typeof setTimeout> | undefined;
+  try {
+    return await Promise.race([
+      operation,
+      new Promise<never>((_resolve, reject) => {
+        timer = setTimeout(
+          () => reject(new Error(`Visual export timed out after ${timeoutMs}ms`)),
+          remaining,
+        );
+      }),
+    ]);
+  } finally {
+    if (timer) clearTimeout(timer);
+  }
 }
 
 function pngFromDataUrl(dataUrl: string): Buffer {
--- a/src/core/mcp-client.ts
+++ b/src/core/mcp-client.ts
@@ -49,6 +49,52 @@ export interface McpServersConfig {
   mcpServers: Record<string, McpServerConfig>;
 }
 
+const MCP_STDIO_FORWARDED_ENV = [
+  "SICLAW_VISUAL_EXPORT_URL",
+  "SICLAW_VISUAL_EXPORT_TIMEOUT_MS",
+  "SICLAW_VISUAL_EXPORT_THEME",
+  "SICLAW_VISUAL_EXPORT_CHROMIUM",
+] as const;
+const DEFAULT_VISUAL_EXPORT_TIMEOUT_MS = 30_000;
+const VISUAL_EXPORT_MCP_GRACE_MS = 5_000;
+
+/**
+ * Add the small platform-owned environment contract required by bundled stdio
+ * MCPs. The MCP SDK intentionally inherits only a safe OS baseline, so values
+ * present in the AgentBox process do not otherwise reach the child process.
+ * Explicit per-server configuration wins over the inherited platform value.
+ */
+export function mergeMcpStdioEnv(
+  configuredEnv: Record<string, string> | undefined,
+  parentEnv: Record<string, string | undefined> = process.env,
+): Record<string, string> | undefined {
+  const inherited: Record<string, string> = {};
+  for (const name of MCP_STDIO_FORWARDED_ENV) {
+    const value = parentEnv[name];
+    if (value !== undefined && value !== "") inherited[name] = value;
+  }
+  const merged = { ...inherited, ...configuredEnv };
+  return Object.keys(merged).length > 0 ? merged : undefined;
+}
+
+export function visualMcpRequestTimeoutMs(
+  serverImplementationName: unknown,
+  toolName: string,
+  env: Record<string, string | undefined> = {},
+): number | undefined {
+  if (
+    serverImplementationName !== "mcp-create-chart" ||
+    !/^render_(?:chart|mermaid|visual_card)$/.test(toolName)
+  ) {
+    return undefined;
+  }
+  const configured = Number(env.SICLAW_VISUAL_EXPORT_TIMEOUT_MS ?? DEFAULT_VISUAL_EXPORT_TIMEOUT_MS);
+  const rendererBudget = Number.isFinite(configured) && configured > 0
+    ? Math.ceil(configured)
+    : DEFAULT_VISUAL_EXPORT_TIMEOUT_MS;
+  return rendererBudget + VISUAL_EXPORT_MCP_GRACE_MS;
+}
+
 interface ManagedMcpClient {
   serverName: string;
   client: any; // Client from @modelcontextprotocol/sdk
@@ -224,12 +270,13 @@ export class McpClientManager {
         const cfg = serverConfig as any;
         const detectedTransport: string = cfg.transport
           ?? (cfg.url ? "streamable-http" : cfg.command ? "stdio" : "");
+        const stdioEnv = detectedTransport === "stdio" ? mergeMcpStdioEnv(cfg.env) : undefined;
         switch (detectedTransport) {
           case "stdio":
             transport = new StdioClientTransport({
               command: cfg.command,
               args: cfg.args,
-              env: cfg.env,
+              env: stdioEnv,
             });
             break;
           case "sse":
@@ -255,13 +302,22 @@ export class McpClientManager {
 
         await client.connect(transport);
         console.log(`[mcp-client] Connected to "${serverName}" (${detectedTransport})`);
+        const serverImplementationName = client.getServerVersion()?.name;
 
         // Discover tools
         const { tools: mcpTools } = await client.listTools();
         console.log(`[mcp-client] "${serverName}" provides ${mcpTools.length} tools: ${mcpTools.map((t: any) => t.name).join(", ")}`);
 
         for (const mcpTool of mcpTools) {
-          const toolDef = this.createToolDefinition(serverName, cfg.description, mcpTool, client);
+          const toolDef = this.createToolDefinition(
+            serverName,
+            cfg.description,
+            mcpTool,
+            client,
+            detectedTransport === "stdio"
+              ? visualMcpRequestTimeoutMs(serverImplementationName, mcpTool.name, stdioEnv)
+              : undefined,
+          );
           this.tools.push(toolDef);
         }
 
@@ -312,6 +368,7 @@ export class McpClientManager {
     serverDescription: string | undefined,
     mcpTool: { name: string; description?: string; inputSchema?: any },
     client: any,
+    requestTimeoutMs?: number,
   ): ResolvedToolDefinition {
     const fullName = buildMcpToolName(serverName, mcpTool.name);
     const inputSchema = mcpTool.inputSchema ?? { type: "object", properties: {} };
@@ -343,10 +400,14 @@ export class McpClientManager {
       parameters,
       execute: async (_toolCallId, args) => {
         try {
-          const result = await client.callTool({
-            name: mcpTool.name,
-            arguments: args ?? {},
-          });
+          const result = await client.callTool(
+            {
+              name: mcpTool.name,
+              arguments: args ?? {},
+            },
+            undefined,
+            requestTimeoutMs === undefined ? undefined : { timeout: requestTimeoutMs },
+          );
 
           const isError = !!result.isError;
           const { content, text } = mcpContentToAgentContent(result.content);
@@ -364,7 +425,7 @@ export class McpClientManager {
           const errorMsg = err?.message ?? String(err);
           return {
             content: [{ type: "text" as const, text: `MCP tool error: ${errorMsg}` }],
-            details: { error: errorMsg },
+            details: { error: errorMsg, errorKind: "transport" },
           };
         }
       },
--- a/src/gateway/channels/dingtalk.ts
+++ b/src/gateway/channels/dingtalk.ts
@@ -42,7 +42,12 @@ import { sessionRegistry } from "../session-registry.js";
 import { sessionTurnLocks } from "../session-turn-lock.js";
 import { appendMessage, bindMessageTraceId, ensureChatSession, warnTraceBindFailure } from "../chat-repo.js";
 import { collectChannelResponse } from "./lark.js";
-import type { RenderedReplyImage } from "./visual-image.js";
+import {
+  stripVisualBlocks,
+  type RenderedReplyImage,
+  type VisualFailureReason,
+  type VisualSourceKind,
+} from "./visual-image.js";
 import { deliverImages } from "./dingtalk-image.js";
 import {
   buildMarkdownMessage,
@@ -62,6 +67,12 @@ const TOPIC_ROBOT = "/v1.0/im/bot/messages/get";
  * data-exfiltration channel. Pin replies to the DingTalk OpenAPI domains.
  */
 const ALLOWED_WEBHOOK_HOSTS = new Set(["oapi.dingtalk.com", "api.dingtalk.com"]);
+const VISUAL_ONLY_NOTICE = "报告图片已发送。";
+const VISUAL_RENDER_FAILED_NOTICE = "报告图片生成失败，请稍后重试。";
+const VISUAL_CONFIGURATION_FAILED_NOTICE = "报告图片功能暂不可用，请联系管理员检查配置。";
+const IMAGE_DELIVERY_FAILED_NOTICE = "报告图片发送失败，请稍后重试。";
+const IMAGE_DELIVERY_PARTIAL_NOTICE = "部分报告图片发送失败，请稍后重试。";
+const IMAGE_DELIVERY_UNCONFIGURED_NOTICE = "报告图片无法发送，请联系管理员检查渠道配置。";
 
 function isAllowedWebhookHost(host: string): boolean {
   return ALLOWED_WEBHOOK_HOSTS.has(host) || host.endsWith(".dingtalk.com");
@@ -308,6 +319,8 @@ export async function handleDingTalkMessage(
   // two boxes and both would run — two writers on one transcript.
   let resultText = "";
   let replyImages: RenderedReplyImage[] = [];
+  let visualRenderFailures = new Map<VisualSourceKind, VisualFailureReason>();
+  let renderedVisualKinds = new Set<VisualSourceKind>();
   let agentError: Error | null = null;
   // Acquired INSIDE the try so a busy session surfaces through the SAME path the
   // AgentBox's 409 already used — the friendly "still working" notice. Outside it the
@@ -387,6 +400,8 @@ export async function handleDingTalkMessage(
     });
     resultText = collected.text;
     replyImages = collected.images;
+    visualRenderFailures = collected.visualRenderFailures;
+    renderedVisualKinds = collected.renderedVisualKinds;
   } catch (err) {
     agentError = err instanceof Error ? err : new Error(String(err));
     console.error(`[dingtalk] Agent execution failed for session=${sessionId}:`, agentError);
@@ -404,40 +419,70 @@ export async function handleDingTalkMessage(
     return;
   }
 
+  // Resolve image delivery before composing the text reply. The source fence is
+  // never safe as an IM fallback, but the text must state when no image reached
+  // the user instead of claiming that an image follows.
+  let deliveredImageCount = 0;
+  if (!agentError && replyImages.length > 0) {
+    if (config) {
+      const target = {
+        routeType,
+        openConversationId: routeType === "group" ? conversationId : undefined,
+        senderStaffId: message.senderStaffId,
+      };
+      try {
+        deliveredImageCount = await deliverImages(config, target, replyImages);
+      } catch (err) {
+        console.error(`[dingtalk] Image delivery failed for session=${sessionId}: ${safeErrorMessage(err)}`);
+      }
+      if (deliveredImageCount < replyImages.length) {
+        console.warn(`[dingtalk] Delivered ${deliveredImageCount}/${replyImages.length} reply images for session=${sessionId}`);
+      }
+    } else {
+      console.warn(`[dingtalk] Image delivery unavailable without channel config for session=${sessionId}`);
+    }
+  }
+  const imageDeliveryUnconfigured = !agentError && replyImages.length > 0 && !config;
+
   // On failure, reply with a generic notice only — the raw error message can
   // leak internal endpoints / infra details to everyone in the chat. The full
   // error was already logged above for operators.
   const finalBody = agentError
     ? AGENT_ERROR_NOTICE
     : (resultText || EMPTY_RESULT_NOTICE);
+  const strippedBody = agentError
+    ? finalBody
+    : stripVisualBlocks(finalBody, {
+        stripSourceKinds: new Set<VisualSourceKind>([
+          ...renderedVisualKinds,
+          ...visualRenderFailures.keys(),
+        ]),
+      });
+  let displayBody = strippedBody;
+  if ([...visualRenderFailures.values()].includes("configuration")) {
+    displayBody = appendChannelNotice(displayBody, VISUAL_CONFIGURATION_FAILED_NOTICE);
+  } else if (visualRenderFailures.size > 0) {
+    displayBody = appendChannelNotice(displayBody, VISUAL_RENDER_FAILED_NOTICE);
+  }
+  if (imageDeliveryUnconfigured) {
+    displayBody = appendChannelNotice(displayBody, IMAGE_DELIVERY_UNCONFIGURED_NOTICE);
+  } else if (replyImages.length > 0 && deliveredImageCount === 0) {
+    displayBody = appendChannelNotice(displayBody, IMAGE_DELIVERY_FAILED_NOTICE);
+  } else if (deliveredImageCount < replyImages.length) {
+    displayBody = appendChannelNotice(displayBody, IMAGE_DELIVERY_PARTIAL_NOTICE);
+  }
+  if (!displayBody) displayBody = VISUAL_ONLY_NOTICE;
 
   // Errors and the empty-result notice go out as plain text; real answers as
   // markdown so formatting (code blocks, lists, bold) renders.
   const body = agentError
-    ? buildTextMessage(finalBody)
-    : (resultText ? buildMarkdownMessage(finalBody) : buildTextMessage(finalBody));
+    ? buildTextMessage(displayBody)
+    : (resultText ? buildMarkdownMessage(displayBody) : buildTextMessage(displayBody));
   await replyToDingTalk(sessionWebhook, body);
+}
 
-  // Deliver any agent-produced images as separate robot messages (mirrors
-  // Lark's replyVisualImages). Best-effort and post-text: image delivery needs
-  // an access token + robot-send permission (config), and a failed image must
-  // never break the primary reply. Skipped when the agent errored, produced no
-  // images, or the channel config was not threaded through (older callers).
-  if (!agentError && config && replyImages.length > 0) {
-    const target = {
-      routeType,
-      openConversationId: routeType === "group" ? conversationId : undefined,
-      senderStaffId: message.senderStaffId,
-    };
-    try {
-      const delivered = await deliverImages(config, target, replyImages);
-      if (delivered < replyImages.length) {
-        console.warn(`[dingtalk] Delivered ${delivered}/${replyImages.length} reply images for session=${sessionId}`);
-      }
-    } catch (err) {
-      console.error(`[dingtalk] Image delivery failed for session=${sessionId}: ${safeErrorMessage(err)}`);
-    }
-  }
+function appendChannelNotice(body: string, notice: string): string {
+  return body.trim() ? `${body.trim()}\n\n${notice}` : notice;
 }
 
 /**
--- a/src/gateway/channels/lark.test.ts
+++ b/src/gateway/channels/lark.test.ts
@@ -8,6 +8,7 @@ import {
   collectResponse,
   collectChannelResponse,
   extractInbound,
+  replyVisualImages,
   resetLarkBindingQueuesForTest,
 } from "./lark.js";
 import {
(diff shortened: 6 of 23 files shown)


mcp/create-chart/src/visual-export-config.ts
```diff
@@ -0,0 +1,9 @@
+export function resolveVisualExportUrl(override?: string): string {
+  const configured = (override ?? process.env.SICLAW_VISUAL_EXPORT_URL)?.trim();
```
jacoblee-io, line 6: **HIGH — the forwarded env never reaches the process that reads it, and the fallback it replaced is gone, so every visual render now hard-fails.**

`create-chart` is a **stdio** MCP server. It is launched in `src/core/mcp-client.ts:229` as:

```ts
transport = new StdioClientTransport({ command: cfg.command, args: cfg.args, env: cfg.env });
```

`cfg.env` comes only from the Portal `mcp_servers.env` column (`src/portal/adapter.ts:1391`), and the documented config for this server (`mcp/create-chart/src/index.ts:11-16`) has just `transport` + `command`, i.e. `env` is `undefined`. The MCP SDK then spawns the child with `{ ...getDefaultEnvironment(), ...undefined }`, and `getDefaultEnvironment()` inherits **only** `HOME, LOGNAME, PATH, SHELL, TERM, USER` — every other variable of the parent is deliberately dropped.

So the chain `values.yaml -> runtime Deployment env -> K8sSpawner -> AgentBox PodSpec` stops one process boundary short of the only code that reads the value: `process.env.SICLAW_VISUAL_EXPORT_URL` is `undefined` inside `mcp-create-chart`.

Concrete scenario: an operator sets `runtime.visualExport.url`, upgrades the chart and recycles the boxes. Because the old `DEFAULT_EXPORT_URL` fallback was deleted, every `render_chart` / `render_mermaid` / `render_visual_card` call now returns `isError` with *"Visual export is not configured; set SICLAW_VISUAL_EXPORT_URL…"* — including in deployments that render fine today via `http://visual-export-web:3000/siclaw-visual-export`. The same applies to `SICLAW_VISUAL_EXPORT_{TIMEOUT_MS,THEME,CHROMIUM}`, which are read only in `mcp/create-chart/src/visual-export.ts:23,32,79`.

Suggested fix: merge an allowlist of `SICLAW_VISUAL_EXPORT_*` from `process.env` into the stdio transport `env` in `mcp-client.ts` (`env: { ...pickVisualExportEnv(process.env), ...cfg.env }`), so the value actually crosses the last hop. Otherwise the requirement is "the operator must enter it in the Portal MCP server's `env` map", in which case the Helm values and the spawner allowlist added here are dead configuration and should be documented as such.

src/gateway/channels/lark.ts
```diff
@@ -2071,8 +2077,11 @@ async function processQueuedLarkMessage(ctx: QueuedLarkMessageContext): Promise<
       ? AGENT_ERROR_NOTICE_BY_LOCALE[locale]
       : (resultText || EMPTY_RESULT_NOTICE_BY_LOCALE[locale]);
   if (agentError || sessionBusy) replyImages = [];
-  const displayBody = stripVisualBlocks(finalBody, { stripSourceBlocks: replyImages.length > 0 })
-    || VISUAL_ONLY_NOTICE_BY_LOCALE[locale];
+  const displayBody = stripVisualBlocks(finalBody, {
+    stripSourceBlocks: replyImages.length > 0 || visualRenderFailed,
+  }) || (visualRenderFailed
```
jacoblee-io: **MEDIUM — the "generation failed" notice fires even when images were delivered.**

`visualRenderFailed` latches on the *first* failed render in the stream and is never reconciled against `replyImages`. Both flags can be true at once: the model calls `render_chart`, it fails on a transient renderer timeout, the model retries and succeeds; or it renders three charts and only one fails.

Scenario: renders 1 fails, render 2 succeeds, and the final answer body is visual-only (just the fence, which the model is free to produce). Then `replyImages.length > 0` **and** `visualRenderFailed`, `stripVisualBlocks(...)` returns `""`, and the card renders `报告图片生成失败，请稍后重试` while `replyVisualImages` still uploads the successfully rendered PNG. The user sees "image generation failed" directly above the image.

Suggested fix: keep the strip condition as is, but gate only the notice on the absence of a delivered image:

```ts
}) || (visualRenderFailed && replyImages.length === 0
  ? VISUAL_RENDER_FAILED_NOTICE_BY_LOCALE[locale]
  : VISUAL_ONLY_NOTICE_BY_LOCALE[locale]);
```

src/gateway/channels/lark.ts
```diff
@@ -3034,6 +3043,8 @@ async function replyVisualImages(
 export interface CollectedChannelResponse {
   text: string;
   images: RenderedReplyImage[];
+  /** True when a visual renderer failed, so source blocks must not leak into IM replies. */
+  visualRenderFailed: boolean;
```
jacoblee-io: **LOW — the new field is computed for DingTalk and thrown away, so the leak this PR closes stays open on that channel.**

`src/gateway/channels/dingtalk.ts:384` calls `collectChannelResponse(..., { includeImages: true, ... })`, so `visualRenderFailed` is populated on that path too, but line 387-388 assigns only `collected.text` / `collected.images`, and the DingTalk handler never calls `stripVisualBlocks` at all.

Scenario: the renderer is down, the agent's `render_visual_card` fails, and the model's answer still carries the ```` ```visual-card {"type":"report",…} ```` fence. A DingTalk user gets the raw JSON block in the markdown reply — exactly the "failed visual source blocks shown to IM users" this PR is meant to prevent. (The success path has the same gap, so it is partly pre-existing, but the signal is now available and unused.)

Suggested fix: apply the same `stripVisualBlocks(finalBody, { stripSourceBlocks: replyImages.length > 0 || collected.visualRenderFailed })` + localized notice in the DingTalk reply path, or state in this PR that DingTalk is intentionally out of scope.

src/gateway/channels/lark.test.ts
```diff
@@ -3309,6 +3309,55 @@ describe("handleLarkMessage — streaming card flow", () => {
     expect(lark.im.message.reply).toHaveBeenCalledTimes(1);
   });
 
+  it("hides visual-card source when its renderer failed", async () => {
+    resolveBindingMock.mockResolvedValue(makeBinding());
+    promptMock.mockResolvedValue({ sessionId: "s-visual-card-failed" });
+    streamEventsMock.mockImplementation(async function* () {
+      yield {
+        type: "tool_execution_end",
+        toolName: "_mcp-create-chart/render_visual_card",
```
jacoblee-io: **LOW — this test pins two shapes the runtime never emits, so the production path is not covered.**

Two fixture details do not match what this codebase produces:

1. `toolName` — MCP tool names are built by `buildMcpToolName` (`src/core/mcp-client.ts:150-154`) as `mcp__create-chart__render_visual_card`. Nothing emits `_mcp-create-chart/render_visual_card`.
2. `result.isError` — the MCP wrapper at `src/core/mcp-client.ts:352-358` **drops** `isError` and surfaces the failure only as `result.details.error` (a string):

```ts
const isError = !!result.isError;
return { content, details: { ...(isError ? { error: text } : {}), ... } };
```

The four-way `||` in `collectChannelResponse` happens to cover both, so the test passes — but the only shape production can produce (`details.error`, `mcp__…__` naming) is untested, and a later simplification of that guard guided by this test would silently disable the feature with the suite still green.

Suggested fix: add (or switch to) a case with `toolName: "mcp__create-chart__render_visual_card"` and `result: { details: { error: "renderer unavailable" }, content: [...] }`. Worth noting alongside it that `details.error` is set to the joined result *text*, so an `isError` result with empty text evaluates falsy and is not detected.

mcp/create-chart/src/visual-export-config.ts
LikiosSedo: Fixed in 20c352eb. The Runtime-to-AgentBox values now cross the final stdio process boundary through a narrow allowlist; explicit per-server MCP env still wins. Added regression coverage that also proves unrelated parent secrets are not inherited.

src/gateway/channels/lark.ts
LikiosSedo: Fixed in 20c352eb. The failure notice is now shown only when no image was delivered; a later successful render keeps the visual-only success notice. Added a fail-then-success regression case.

src/gateway/channels/lark.ts
LikiosSedo: Fixed in 20c352eb. DingTalk now consumes the shared failure signal, strips visual source blocks on renderer failure or successful image delivery, and shows the same sanitized retry notice. Added channel-level regression coverage.

src/gateway/channels/lark.test.ts
LikiosSedo: Fixed in 20c352eb. The fixture now uses the production MCP tool name and result.details.error shape. The focused and full suites pass with the real event contract.

src/gateway/channels/lark.ts
```diff
@@ -2071,8 +2077,11 @@ async function processQueuedLarkMessage(ctx: QueuedLarkMessageContext): Promise<
       ? AGENT_ERROR_NOTICE_BY_LOCALE[locale]
       : (resultText || EMPTY_RESULT_NOTICE_BY_LOCALE[locale]);
   if (agentError || sessionBusy) replyImages = [];
-  const displayBody = stripVisualBlocks(finalBody, { stripSourceBlocks: replyImages.length > 0 })
-    || VISUAL_ONLY_NOTICE_BY_LOCALE[locale];
+  const displayBody = stripVisualBlocks(finalBody, {
+    stripSourceBlocks: replyImages.length > 0 || visualRenderFailed,
```
jacoblee-io: **MEDIUM — a stripped fence with no notice: the answer loses content and says nothing about it.**

`stripSourceBlocks` is now true whenever `visualRenderFailed`, but the retry notice only appears when the *whole* stripped body comes back empty (`stripVisualBlocks(...) || (...)`). Any answer that mixes prose with a fence therefore loses the fence silently.

Scenario: the renderer is down, `render_mermaid` returns `isError`, and the model falls back to writing the diagram itself — `"Here is the topology:"` followed by a hand-written ```` ```mermaid ```` block. `stripVisualBlocks` deletes the fence and returns `"Here is the topology:"`, which is non-empty, so the `||` fallback never runs. The user gets a dangling sentence, no diagram, and no explanation. Before this PR the fence survived, so the reader at least had the source.

The same code and the same scenario apply to `src/gateway/channels/dingtalk.ts:420`.

Suggested fix: decide the notice from `visualRenderFailed` itself rather than from "the body happened to be empty" — e.g. append the retry notice whenever `visualRenderFailed && replyImages.length === 0`, whether or not prose survived the strip.

src/gateway/channels/lark.ts
```diff
@@ -2071,8 +2077,11 @@ async function processQueuedLarkMessage(ctx: QueuedLarkMessageContext): Promise<
...
+  }) || (visualRenderFailed && replyImages.length === 0
+    ? VISUAL_RENDER_FAILED_NOTICE_BY_LOCALE[locale]
+    : VISUAL_ONLY_NOTICE_BY_LOCALE[locale]);
```
jacoblee-io: **MEDIUM — 👍/👎 get attached to the render-failure notice, which is exactly what the comment below it forbids.**

This concerns `src/gateway/channels/lark.ts:2099` (`const isAnswer = !agentError && resultText.trim().length > 0;`). That line is unchanged by this PR and so cannot be anchored directly, but the failure branch added here is what breaks it.

`isAnswer` treats a non-empty `resultText` as a real answer. When the reply was only a visual fence and the renderer failed, `resultText` is that raw fence (non-empty) and `agentError` is null — so `isAnswer` is true while `displayBody` is `VISUAL_RENDER_FAILED_NOTICE_BY_LOCALE[locale]`. `finalizeCard` then renders feedback buttons under it, and a click writes a `message_feedback` row rating a failure notice. That is the precise case the comment at lines 2095-2097 rules out: "never under an error or empty-result notice, where a click would write a rating against a non-answer and skew the feedback signal Metrics aggregates."

Suggested fix: fold the new state into that gate, e.g. `const isAnswer = !agentError && !(visualRenderFailed && replyImages.length === 0) && resultText.trim().length > 0;`

mcp/create-chart/src/visual-export-config.ts
```diff
@@ -0,0 +1,9 @@
...
+  if (!configured) {
+    throw new Error(
```
jacoblee-io: **MEDIUM — the Helm default is `""`, so an upgrade turns rendering off silently and the only symptom is a per-call error the model retries.**

Removing `DEFAULT_EXPORT_URL` is the right call — nothing in this repo ships a `visual-export-web` Service, so the old default really was invented DNS. But `helm/siclaw/values.yaml` ships `visualExport.url: ""` and the template emits nothing when it is empty, so any existing deployment upgraded without explicitly setting `runtime.visualExport.url` loses all visual rendering.

Scenario: `helm upgrade` with existing values → the AgentBox has no `SICLAW_VISUAL_EXPORT_URL` → every `render_chart` / `render_mermaid` / `render_visual_card` call throws here, comes back to the model as `isError`, and the model typically retries the tool before giving up. Nothing is logged at startup and there is no operator-facing signal; the outage shows up only as slow, chart-less answers.

Suggested fix: pair this with an operator-visible signal — a one-time warning from the AgentBox when the `create-chart` MCP is configured but `SICLAW_VISUAL_EXPORT_URL` is unset — and call the required value out in the release notes, since this is a breaking configuration change rather than a defaulted one.

src/gateway/channels/lark.ts
```diff
@@ -3162,8 +3174,15 @@ export async function collectChannelResponse(
 
       if (ev.type === "tool_execution_end" || ev.type === "tool_end") {
         if (options.includeImages) collectImageAttachments(ev.result?.content, images, seenImageKeys);
+        const name = (ev.toolName as string) || (ev.name as string) || "tool";
+        if (
+          options.includeImages &&
+          /(?:^|[._/-])render_(?:chart|mermaid|visual_card)$/.test(name) &&
```
jacoblee-io: **LOW — the flag also fires for client-side argument validation, so the user is told to retry when a retry cannot help.**

The predicate treats every `isError` from a `render_*` tool as a renderer failure, but `mcp/create-chart/src/index.ts:77` wraps *all* throws in `isError: true` — including the validation errors raised before the renderer is ever contacted: `render_chart: type must be pie, bar, or line`, `render_visual_card: title is required`, and the rest of `handler.ts`'s `validate*` family.

Scenario: the model sends a malformed spec once, gives up on the image, and answers in prose plus a fence. `visualRenderFailed` is true, the fence is stripped, and the user is shown "报告图片生成失败，请稍后重试" — advice that cannot help, because nothing was ever wrong with the renderer.

Related: the flag is one-way and is never cleared by a later successful call. That is masked today only because a success always yields an image, so `replyImages.length === 0` happens to gate the notice — a coincidence of the current `includeImages: true` call sites rather than a rule.

Suggested fix: distinguish the two cases — the create-chart validation errors are all prefixed `render_*: `, or the MCP could mark transport/renderer failures distinctly — so that only a real renderer failure suppresses the source and offers a retry.

src/gateway/channels/dingtalk.ts
```diff
@@ -410,12 +414,19 @@ export async function handleDingTalkMessage(
   const finalBody = agentError
     ? AGENT_ERROR_NOTICE
     : (resultText || EMPTY_RESULT_NOTICE);
+  const displayBody = agentError
+    ? finalBody
+    : stripVisualBlocks(finalBody, {
+        stripSourceBlocks: replyImages.length > 0 || visualRenderFailed,
```
jacoblee-io: **LOW — stripping is keyed on images *collected*, but DingTalk delivery is best-effort and post-text, so a delivery failure leaves the user with nothing.**

`replyImages.length > 0` means "the tool returned an image", not "the user will see one". DingTalk delivery happens after this reply and can be skipped or fail outright: it is gated on `config` being present (the parameter is optional, and the comment at lines 445-448 acknowledges "older callers"), and `deliverImages` throwing or delivering partially only produces a `console.error` / `console.warn`.

Scenario: a deployment whose robot lacks send permission, or any transient `deliverImages` throw. The text reply has already gone out as `已生成图片如下。` — or as prose with the chart source removed — and no image follows. Before this PR DingTalk did not strip at all, so the ```` ```chart ```` source was still there: ugly, but the content survived.

Suggested fix: since the text goes out first, either deliver the images before the text reply, or keep the source block when delivery is known to be impossible up front (`!config`) and post a short follow-up notice when `deliverImages` fails.

mcp/create-chart/src/visual-export-config.ts
```diff
@@ -0,0 +1,9 @@
...
+      "Visual export is not configured; set SICLAW_VISUAL_EXPORT_URL to a reachable /siclaw-visual-export page",
+    );
+  }
+  return configured.replace(/\/+$/, "");
```
jacoblee-io: **LOW — the normalizer handles trailing slashes but not an existing fragment, and the caller appends one.**

`visual-export.ts` does `page.goto(baseUrl + "#" + payload)`. An operator who pastes a hash-router console URL — `https://console.example.com/#/siclaw-visual-export`, which is exactly what the browser address bar shows for such a route — ends up with `https://console.example.com/#/siclaw-visual-export#<payload>`. The page's own hash parsing sees `/siclaw-visual-export#<payload>`, never decodes the payload, so `__siclawExportVisuals` is never populated and every render burns the full `timeoutMs` in `waitForFunction` before failing.

This function exists precisely to reject bad configuration and already normalizes one URL shape, so the fragment case belongs here: reject a URL containing `#` with a message naming the problem, so the operator gets a clear error instead of a 30s timeout per render.

src/gateway/channels/lark.ts
LikiosSedo: Fixed in 5dd02d9c. A renderer failure with no delivered image now appends the localized retry notice even when prose survives source-block stripping. Added mixed prose + Mermaid failure coverage for both Lark and DingTalk.

src/gateway/channels/lark.ts
LikiosSedo: Fixed in 5dd02d9c. Renderer-failure notices are no longer treated as rateable answers, including mixed prose cases. Regression coverage verifies that no feedback controls are appended.

mcp/create-chart/src/visual-export-config.ts
LikiosSedo: Fixed in 5dd02d9c. The create-chart MCP emits a one-time startup warning when the endpoint is unset, and the required runtime.visualExport.url upgrade action is now explicit in both CHANGELOG and the PR description.

src/gateway/channels/lark.ts
LikiosSedo: Fixed in 5dd02d9c with a structured error contract rather than text-prefix inference. create-chart now classifies errors as input, renderer, or internal; MCP transport failures are marked separately. Channels only suppress source and offer a retry for renderer/transport failures. Added input-error and renderer-error regression cases.

src/gateway/channels/dingtalk.ts
LikiosSedo: Fixed in 5dd02d9c. DingTalk now resolves image delivery before composing its text reply. Full success gets an accurate sent confirmation; missing config, partial delivery, or failure gets a clean delivery-failure notice while source JSON remains suppressed. Tests verify delivery order and both outcomes.

mcp/create-chart/src/visual-export-config.ts
LikiosSedo: Fixed in 5dd02d9c. Endpoint values containing a fragment are rejected at configuration resolution with an actionable error instead of reaching page navigation and timing out. Added regression coverage.

src/gateway/channels/lark.ts
```diff
@@ -2071,8 +2077,13 @@ async function processQueuedLarkMessage(ctx: QueuedLarkMessageContext): Promise<
       ? AGENT_ERROR_NOTICE_BY_LOCALE[locale]
       : (resultText || EMPTY_RESULT_NOTICE_BY_LOCALE[locale]);
   if (agentError || sessionBusy) replyImages = [];
-  const displayBody = stripVisualBlocks(finalBody, { stripSourceBlocks: replyImages.length > 0 })
-    || VISUAL_ONLY_NOTICE_BY_LOCALE[locale];
+  const strippedBody = stripVisualBlocks(finalBody, {
```
jacoblee-io, line 2124: **MEDIUM — `stripSourceBlocks` is keyed on "some render failed this turn", so it also deletes fences the model wrote as its own text fallback.**

Builds on #3931172866 (now fixed): that comment asked for a notice when a fence is stripped, and 5dd02d9c adds one. The residual problem is the *deletion itself* — `visualRenderFailed` is turn-wide and carries no information about which fence came from a failed tool.

`RENDER_CHART_DESCRIPTION` (`mcp/create-chart/src/handler.ts:35`) explicitly instructs the model: *"For qualitative diagrams, workflows, topology, or decision trees, use a `` ```mermaid `` fenced block instead"*. A hand-written fence in the answer body is therefore a documented, expected output — not renderer debris.

Scenario: the model calls `render_chart`, the renderer times out, and the model recovers exactly as its tool description tells it to — answering in prose plus a hand-written `` ```mermaid `` topology it never asked the renderer to draw. `visualRenderFailed` is true and `replyImages` is empty, so this line strips that fence and line 2085 appends `报告图片生成失败，请稍后重试`. The user loses a diagram the failed tool never touched, and is told to retry a render that already succeeded in text form.

The sharpest version is a `visual-card` that *was* the whole answer: before this PR the raw JSON fence reached the chat — ugly, but readable during an incident — and now the reply is only the retry notice, so the diagnosis exists nowhere the user can see it.

Suggested fix: carry the failed kinds out of `collectChannelResponse` (e.g. `visualRenderFailedKinds: Set<"chart" | "mermaid" | "visual-card">`, derivable from the tool name already matched at line 3182) and strip only those fence languages, instead of passing one boolean into `stripVisualBlocks`.

src/gateway/channels/lark.ts
```diff
@@ -2087,7 +2098,7 @@ async function processQueuedLarkMessage(ctx: QueuedLarkMessageContext): Promise<
     // Only solicit 👍/👎 on a real answer — never under an error or
     // empty-result notice, where a click would write a rating against a
     // non-answer and skew the feedback signal Metrics aggregates.
-    const isAnswer = !agentError && resultText.trim().length > 0;
+    const isAnswer = !agentError && !sessionBusy && !visualFailureWithoutImage && resultText.trim().length > 0;
```
jacoblee-io: **MEDIUM — the `isAnswer` fix over-corrects: a recovered prose answer now loses its 👍/👎.**

Builds on #3931173323 (now fixed). That comment was about feedback buttons appearing *under a bare retry notice*; the fix adds `!visualFailureWithoutImage`, which is broader than the case it closes — it keys on "a render failed this turn", not on "the body is only a notice".

Scenario: the renderer is down, `render_mermaid` fails, and the model still answers with a full prose diagnosis (several hundred characters of real analysis, no fence). `visualFailureWithoutImage` is true, so `isAnswer` is false, `finalizeCard` is called with `undefined` feedback, and a genuine answer ships with no rating controls. The comment directly above this line says feedback must be withheld from "an error or empty-result notice" — this reply is neither, and the feedback signal Metrics aggregates is silently lost for every turn in which any visual failed.

Suggested fix: gate on the composed body rather than on the flag — the notice-only case is exactly `strippedBody === ""`, so `strippedBody.trim().length > 0` closes #3931173323 without silencing feedback on answers that survived the failure.

src/gateway/channels/lark.ts
```diff
@@ -2071,8 +2077,13 @@ async function processQueuedLarkMessage(ctx: QueuedLarkMessageContext): Promise<
       ? AGENT_ERROR_NOTICE_BY_LOCALE[locale]
       : (resultText || EMPTY_RESULT_NOTICE_BY_LOCALE[locale]);
   if (agentError || sessionBusy) replyImages = [];
-  const displayBody = stripVisualBlocks(finalBody, { stripSourceBlocks: replyImages.length > 0 })
-    || VISUAL_ONLY_NOTICE_BY_LOCALE[locale];
+  const strippedBody = stripVisualBlocks(finalBody, {
+    stripSourceBlocks: replyImages.length > 0 || visualRenderFailed,
+  });
+  const visualFailureWithoutImage = visualRenderFailed && replyImages.length === 0;
```
jacoblee-io: **MEDIUM — requiring `replyImages.length === 0` means a mixed turn drops a visual with no notice at all.**

This is the flip side of #3931000970 (now fixed). That comment reported the notice firing when an image *had* been delivered; the fix conditions it on `replyImages.length === 0`. But `replyImages` is a per-turn aggregate across *all* render calls, so "some image was delivered" is now read as "no render failed".

Scenario: one turn renders two visuals — `render_chart` succeeds and `render_mermaid` times out. `replyImages.length === 1`, so `visualFailureWithoutImage` is false and no notice is shown; meanwhile line 2081 (`replyImages.length > 0`) strips *every* fence, the mermaid one included. The user receives one chart image, no diagram, and nothing telling them a second visual was ever attempted. `dingtalk.ts:448` has the identical shape.

Worth noting that the fail-then-retry-succeed case #3931000970 was about is distinguishable from this one: a retry reuses the *same* kind, whereas here the failing and succeeding renders are different kinds. Tracking failures per kind (see the fix suggested on line 2080) separates them — a kind that later succeeded clears its own flag, while a kind that never produced an image keeps it and gets the notice.

helm/siclaw/values.yaml
```diff
@@ -80,6 +80,15 @@ runtime:
...
                           # If the embedding baseUrl is the SAME origin as the
                           # default LLM provider, that provider's key is reused.
+  # Browser-rendered charts, diagrams, and report cards. The URL must point to
+  # a reachable ControlPlane Web /siclaw-visual-export page; this chart does not
+  # install that external renderer or assume its Kubernetes Service name.
+  # Values are forwarded from Runtime to every normal AgentBox because the MCP
+  # renderer executes inside the AgentBox process.
+  visualExport:
+    url: ""                # e.g. https://console.example.com/siclaw-visual-export
+    timeoutMs: 30000
```
jacoblee-io, line 92: **MEDIUM — this knob is capped at 60 s by the MCP SDK's default request timeout, and even the shipped default can exceed it.**

The defect is in `src/core/mcp-client.ts:372`, which is unchanged by this PR and so cannot be anchored directly; commenting here because this is the operator-facing value the PR introduces.

`mcp-client.ts:372` calls `client.callTool({ name, arguments })` with no options, so the SDK applies `DEFAULT_REQUEST_TIMEOUT_MSEC = 60000` (`@modelcontextprotocol/sdk/dist/esm/shared/protocol.js:8`). Nothing in this PR raises it. `mcp/create-chart/src/visual-export.ts` then spends `timeoutMs` **twice, sequentially** — once in `page.goto({ timeout: timeoutMs })` and again in `page.waitForFunction(..., { timeout: timeoutMs })` — on top of an unbounded `launchBrowser()`.

Two concrete consequences:

1. Any `visualExport.timeoutMs` above 60000 is silently inert. An operator who raises it to 120000 for a slow renderer still gets a failure at 60 s, and no log says why.
2. Even at the shipped default of 30000, a cold Chromium plus a slow `goto` plus a slow `waitForFunction` can pass 60 s. `callTool` then throws `Request timed out`, which `mcp-client.ts:393` labels `errorKind: "transport"`. The renderer's own clean `VisualToolRendererError` — the point of the new `tool-error.ts` contract — is discarded, and the configured window never applies.

Suggested fix: pass the renderer budget to the SDK (`client.callTool(params, undefined, { timeout: … })`), or bound the two Playwright waits by one shared deadline so the export can never outlive the MCP request. Either way this comment block should state the effective ceiling.

src/gateway/channels/lark.ts
```diff
@@ -2071,8 +2077,13 @@ async function processQueuedLarkMessage(ctx: QueuedLarkMessageContext): Promise<
...
+  const displayBody = visualFailureWithoutImage
+    ? appendChannelNotice(strippedBody, VISUAL_RENDER_FAILED_NOTICE_BY_LOCALE[locale])
+    : (strippedBody || VISUAL_ONLY_NOTICE_BY_LOCALE[locale]);
```
jacoblee-io: **MEDIUM — the DingTalk delivery-confirmation fix was not mirrored in Lark, where the same claim-without-image is still reachable.**

The defect is at `src/gateway/channels/lark.ts:2135` (`await replyVisualImages(...)`) and in `replyVisualImages` itself (`lark.ts:3028-3040`), neither of which is in this diff; anchored here because this line composes the text that makes the claim.

`dingtalk.ts:419-435` now resolves delivery *before* composing the reply, precisely so the text cannot promise an image that never arrived (#3931174654). Lark still does the opposite: `finalizeCard` renders `VISUAL_ONLY_NOTICE_BY_LOCALE[locale]` ("已生成图片如下。" / "Image generated below.") at line 2103, and only afterwards does line 2135 call `replyVisualImages`, whose per-image failure is nothing but a `console.warn` at line 3037.

Scenario: `render_visual_card` succeeds, so the fence is stripped and `strippedBody` is empty; `im.image.create` or the image reply then fails (expired token, missing permission, oversized PNG). The card says "Image generated below.", nothing follows it, and the card content has already replaced the live step indicator — so the answer now exists only in Portal. That is the exact failure this PR fixed, one channel at a time.

Suggested fix: move `replyVisualImages` ahead of `finalizeCard`, have it return a delivered count the way `deliverImages` does, and reuse the DingTalk `IMAGE_DELIVERY_FAILED_NOTICE` shape (localized through `*_BY_LOCALE`) when the count falls short.

src/gateway/channels/dingtalk.ts
```diff
@@ -404,40 +409,60 @@ export async function handleDingTalkMessage(
...
+        console.error(`[dingtalk] Image delivery failed for session=${sessionId}: ${safeErrorMessage(err)}`);
+      }
+      if (deliveredImageCount < replyImages.length) {
+        console.warn(`[dingtalk] Delivered ${deliveredImageCount}/${replyImages.length} reply images for session=${sessionId}`);
+      }
+    } else {
+      console.warn(`[dingtalk] Image delivery unavailable without channel config for session=${sessionId}`);
+    }
+  }
+  const imageDeliveryFailed = !agentError && replyImages.length > 0 && deliveredImageCount < replyImages.length;
```
jacoblee-io: **LOW — three distinct outcomes collapse into one "please retry later", including one where retrying can never help.**

Builds on the answer to #3931174654, which states that missing config, partial delivery and failure all get the delivery-failure notice. Lumping the last two is defensible; the residual issues are the first case and the counting.

1. **Missing `config`** (line 429 logs "Image delivery unavailable without channel config"): `deliveredImageCount` stays 0, so this predicate is true and the user is told `报告图片发送失败，请稍后重试。` A deployment whose channel config was never threaded through will never deliver an image, so "retry later" is advice that cannot succeed — and the source fence has already been stripped at line 443, so the content is gone as well. The user is left with a notice and nothing else, permanently.
2. **Partial success**: 2 images, 1 delivered → the user receives one image *and* a total-failure notice sitting next to it.
3. **A throw part-way through `deliverImages`**: line 424 assigns only on normal return, so `deliveredImageCount` stays 0 even when earlier images landed. The notice then contradicts what is visibly in the chat.

Suggested fix: track the no-config case separately and give it copy an operator can act on rather than a retry prompt; distinguish `deliveredImageCount === 0` from `0 < deliveredImageCount < replyImages.length` in the wording. For (3), have `deliverImages` surface its running count on the failure path — it is already documented as never aborting the rest.

mcp/create-chart/src/visual-export-config.ts
```diff
@@ -0,0 +1,20 @@
+export function resolveVisualExportUrl(override?: string): string {
+  const configured = (override ?? process.env.SICLAW_VISUAL_EXPORT_URL)?.trim();
```
jacoblee-io, line 6: **LOW — a configuration error is reported to the model and the user as a renderer outage.**

Builds on #3931173720, which was answered with the startup warning. This is about how the same condition is classified *at call time* under the new `tool-error.ts` contract.

This throws a plain `Error`, and `handler.ts:393-399` (`exportVisual`) funnels everything out of the export through `rendererError(err)` — so an *unconfigured endpoint* becomes `VisualToolRendererError` → `error_kind: "renderer"` → `visualRenderFailed` at `lark.ts:3184`. On a deployment upgraded without `runtime.visualExport.url`, every user of every visual tool is told `报告图片生成失败，请稍后重试` indefinitely, the model is told the renderer is unavailable and keeps re-attempting it, and the one honest signal (the startup warning) sits in pod logs where no user or model can see it.

The same funnel swallows a second config-shaped mistake: there is no scheme check here, so `console.example.com/siclaw-visual-export` (no `https://`) passes both the empty and the fragment check and fails only inside `page.goto` — again surfacing as a renderer failure.

Suggested fix: add a fourth kind (`configuration`) to `VisualToolErrorKind`, throw it from this function, and let `rendererError` pass it through unchanged. Channels keep suppressing the source block — still correct — without telling users to retry, and the operator gets a distinguishable signal in the tool result rather than only at startup. A scheme check here would move the second case into the same bucket.

mcp/create-chart/src/visual-export-config.ts
```diff
@@ -0,0 +1,20 @@
...
+  }
+  if (configured.includes("#")) {
+    throw new Error(
+      "Visual export URL must not contain a fragment (#); configure the page URL before the hash payload",
+    );
+  }
+  return configured.replace(/\/+$/, "");
+}
+
+export function visualExportConfigurationWarning(): string | null {
```
jacoblee-io, line 46: **LOW — the startup warning covers only the empty case, so the misconfiguration this PR added a check for produces no warning at all.**

This function re-implements the empty-string test instead of exercising the validator right above it, so the fragment rejection added at line 9 (closing #3931175019) has no startup counterpart.

Scenario: an operator sets `runtime.visualExport.url: https://console.example.com/#/siclaw-visual-export` — the hash-router form the browser address bar actually shows, and precisely the input the new fragment check exists to reject. `visualExportConfigurationWarning()` sees a non-empty trimmed value and returns `null`, so `index.ts:82-83` prints nothing but `[mcp-create-chart] ready`. The misconfiguration then surfaces only as a per-call tool error — the failure mode #3931173720 was raised about, and the one this warning was added to prevent. The hole widens with every future check added to `resolveVisualExportUrl`.

Suggested fix:

```ts
export function visualExportConfigurationWarning(): string | null {
  try {
    resolveVisualExportUrl();
    return null;
  } catch (err) {
    return err instanceof Error ? err.message : String(err);
  }
}
```

That keeps the two definitions of "configured correctly" from drifting, and every rejection reason reaches the operator at startup instead of only the empty one.

src/gateway/channels/lark.ts
LikiosSedo: Fixed in 7fb78f84. Failures are retained per visual kind even when another kind returns an image, while a later successful image for the same kind clears that kind's failure. Mixed turns now deliver successful images and still show the missing-visual notice.

mcp/create-chart/src/visual-export-config.ts
LikiosSedo: Fixed in 7fb78f84. The startup warning now calls the same URL and timeout validators used at request time, so missing, fragment-bearing, relative/non-HTTP URLs, and invalid timeout values are all reported consistently.

src/gateway/channels/dingtalk.ts
LikiosSedo: Fixed in 7fb78f84. DingTalk now distinguishes missing channel configuration, total delivery failure, and partial delivery. deliverImages catches each image independently and preserves the running success count while continuing with later images.

src/gateway/channels/lark.ts
LikiosSedo: Fixed in 7fb78f84. Channel collection now tracks render outcomes per chart, mermaid, and visual-card kind, and stripVisualBlocks removes only the kinds that actually rendered or failed. A chart failure therefore no longer removes a separately authored Mermaid fallback.

helm/siclaw/values.yaml
LikiosSedo: Fixed in 7fb78f84. Playwright launch, navigation, readiness, and export now share one deadline. The bundled create-chart MCP call also gets the configured renderer budget plus 5 seconds of transport grace, so values above 60 seconds are effective instead of being capped by the SDK default.

src/gateway/channels/lark.ts
LikiosSedo: Fixed in 7fb78f84. Feedback now depends on strippedBody containing a substantive surviving answer. A bare stripped visual failure has no feedback controls, while a recovered prose answer keeps them.

mcp/create-chart/src/visual-export-config.ts
LikiosSedo: Fixed in 7fb78f84. Invalid or missing endpoints and invalid timeout values now raise VisualToolConfigurationError with error_kind=configuration. URL validation requires an absolute HTTP(S) URL and rejects fragments; rendererError preserves the configuration classification, and channels use operator-action copy rather than retry copy.

src/gateway/channels/lark.ts
LikiosSedo: Fixed in 7fb78f84. Lark now attempts all image replies before finalizing the card, counts successful deliveries, and composes success, total-failure, or partial-failure copy from the observed count. The source fence remains suppressed without making an unverified delivery claim.

mcp/create-chart/src/visual-export.ts
```diff
@@ -48,14 +49,18 @@ export async function exportMarkdownVisualsWithVisualExportWeb(
         return Boolean(w.__siclawVisualExportReady && w.__siclawExportVisuals);
       },
       undefined,
-      { timeout: timeoutMs },
+      { timeout: remainingTimeout(deadline, timeoutMs) },
+    );
+    const exported = await withDeadline(
```
jacoblee-io, line 54: **MEDIUM — `withDeadline` orphans an already-started promise, so an exhausted budget crashes the whole `mcp-create-chart` process instead of failing one call.**

Builds on #3931785865, which asked for the two Playwright waits to be bound by one shared deadline. The shared deadline is right; the helper that enforces it has a hole.

`withDeadline` takes an **already-created** promise, and its very first statement is `remainingTimeout(deadline, timeoutMs)` (line 129), which *throws* when the budget is gone. Because it throws before `Promise.race` runs, nothing is ever attached to the `page.evaluate(...)` promise built at the call site. The outer `finally` then runs `browser.close()`, that orphan rejects with `Target page, context or browser has been closed`, and there is no handler — so it is an unhandled rejection.

`mcp/create-chart/src/index.ts:88-89` registers only SIGINT/SIGTERM plus `main().catch`, and Node ≥15 defaults to `--unhandled-rejections=throw`, so the process exits. `McpClientManager` has no restart path for a stdio child, so **every** later `render_chart` / `render_mermaid` / `render_visual_card` in that AgentBox fails at the transport layer for the rest of the pod's life — one late render takes the tool down for the whole session.

Two reachable triggers:

1. `page.waitForFunction` (line 43) is itself given `remainingTimeout(...)`, so it can succeed a millisecond inside the deadline; by the time control reaches line 54, `Date.now() >= deadline`.
2. Line 27 has the identical shape, and `chromium.launch` is handed the **whole** budget as its timeout (line 25) — a launch that consumes it and still succeeds leaves `browser.newPage(...)` orphaned the same way.

Suggested fix: take a factory rather than a promise, so nothing is started before the guard runs —

```ts
async function withDeadline<T>(start: () => Promise<T>, deadline: number, timeoutMs: number): Promise<T> {
  const remaining = remainingTimeout(deadline, timeoutMs);
  const operation = start();
  ...
```

or, keeping the current signature, attach a no-op handler before rethrowing (`operation.catch(() => {})`). Either way the deadline check must not be able to run before the promise has a handler.

src/core/mcp-client.ts
```diff
@@ -49,6 +49,53 @@ export interface McpServersConfig {
...
+  env: Record<string, string | undefined> = process.env,
+): number | undefined {
+  if (
+    typeof command !== "string" ||
+    !/(?:^|[/\\])mcp-create-chart(?:\.cmd)?$/.test(command) ||
+    !/^render_(?:chart|mermaid|visual_card)$/.test(toolName)
+  ) {
+    return undefined;
+  }
+  const configured = Number(env.SICLAW_VISUAL_EXPORT_TIMEOUT_MS ?? DEFAULT_VISUAL_EXPORT_TIMEOUT_MS);
```
jacoblee-io, line 91: **MEDIUM — the host reads the renderer budget from the wrong environment, so a per-server override desyncs the MCP timeout from the deadline the child actually enforces.**

Builds on #3931785865 (the 60 s SDK cap). Raising the request timeout is the right fix; the value it is computed from is not.

`visualMcpRequestTimeoutMs` defaults `env` to `process.env`, but the child does **not** get `process.env` — it gets `mergeMcpStdioEnv(cfg.env)` (line 279), where an explicit per-server `env` deliberately **wins** over the inherited platform value. That precedence is introduced by this PR and pinned by its own test ("lets an explicit MCP server config override the inherited value"). The two sides therefore read different variables.

Scenario: a Portal `mcp_servers` row for the bundled renderer configured as

```json
{"transport":"stdio","command":"mcp-create-chart","env":{"SICLAW_VISUAL_EXPORT_TIMEOUT_MS":"120000"}}
```

on a Runtime that sets nothing. The child resolves a 120 s Playwright deadline; the host computes `30000 + 5000` and aborts the JSON-RPC call at 35 s. `mcp-client.ts:427` labels it `errorKind: "transport"`, the channel appends `报告图片生成失败，请稍后重试。`, and the source fence has already been stripped — for a render that was still progressing inside its configured window. The reverse pairing (server override of 10 s, Runtime at 120 s) leaves a 125 s request timeout over a 10 s renderer, which is harmless but equally unintentional.

Same root cause makes the guard at line 88 fragile: the match requires a basename of exactly `mcp-create-chart`. That is the normal case (the image symlinks the bin into PATH), but a hand-written entry of `command: "node", args: ["…/create-chart/dist/index.js"]` silently falls back to the SDK's 60 s default and re-caps any budget above 60 s — the exact defect #3931785865 reported, back again for a correctly configured deployment.

Suggested fix: compute the merged env once and use it for both sides —

```ts
const stdioEnv = mergeMcpStdioEnv(cfg.env);
transport = new StdioClientTransport({ command: cfg.command, args: cfg.args, env: stdioEnv });
// …
visualMcpRequestTimeoutMs(cfg.command, mcpTool.name, stdioEnv)
```

For the command match, consider keying on the resolved child instead (or having the server advertise its own budget) rather than on a filename the operator is free to spell differently.

src/gateway/channels/lark.ts
```diff
@@ -2071,8 +2098,38 @@ async function processQueuedLarkMessage(ctx: QueuedLarkMessageContext): Promise<
       ? AGENT_ERROR_NOTICE_BY_LOCALE[locale]
       : (resultText || EMPTY_RESULT_NOTICE_BY_LOCALE[locale]);
   if (agentError || sessionBusy) replyImages = [];
-  const displayBody = stripVisualBlocks(finalBody, { stripSourceBlocks: replyImages.length > 0 })
-    || VISUAL_ONLY_NOTICE_BY_LOCALE[locale];
+
+  // Resolve delivery before finalizing the text card. A source fence is stripped
+  // once its renderer returned an image, so the replacement copy must reflect
+  // whether that image actually reached Lark.
+  const deliveredImageCount = await replyVisualImages(
```
jacoblee-io, line 2114: **LOW–MEDIUM — moving delivery ahead of the card makes the visual-only copy point the wrong way, and puts image upload on the critical path of the answer.**

Builds on #3931786498, which asked for exactly this move so the text could not claim an undelivered image. The claim problem is fixed; two consequences of the new order are not.

1. **The copy now points at content that is already above it.** `VISUAL_ONLY_NOTICE_BY_LOCALE` is `已生成图片如下。` / `Image generated below.` (lines 76-79, unchanged by this PR), and line 2132 still uses it as the whole body when everything stripped away. But the images went out at this line, *before* `finalizeCard`, so in the chat the notice sits **under** the images it announces. Concretely: a turn whose entire answer was a ```` ```visual-card ```` fence — image posted, then a card reading "Image generated below." with nothing after it.

2. **The answer is now gated on `im.image.create` + `im.message.reply`, two calls per image.** The old order made this explicitly impossible; the comment deleted from `dingtalk.ts` stated the contract — *"Best-effort and post-text: … a failed image must never break the primary reply."* While these awaits run, `cardFinalizing` is still `false` (it is set at line 2139), so coalesced milestone flushes keep updating the card, and the user sits on the ⏳ placeholder for the duration. `replyImageToLark` has no timeout of its own beyond whatever the Lark SDK applies, so a slow media endpoint delays — and a crash mid-loop loses — an answer that previously had already been delivered.

Suggested fix: split the two halves of delivery. Upload each image and obtain its `image_key` *before* composing the copy (that is what gives you the honest delivered count #3931786498 wanted), then send the image messages *after* `finalizeCard`. If keeping the current order, reword the notice to past tense in both locales so it no longer describes a position it cannot guarantee.

src/gateway/channels/dingtalk.ts
```diff
@@ -404,40 +419,70 @@ export async function handleDingTalkMessage(
...
+  let deliveredImageCount = 0;
+  if (!agentError && replyImages.length > 0) {
+    if (config) {
+      const target = {
+        routeType,
+        openConversationId: routeType === "group" ? conversationId : undefined,
+        senderStaffId: message.senderStaffId,
+      };
+      try {
+        deliveredImageCount = await deliverImages(config, target, replyImages);
```
jacoblee-io, line 434: **LOW — the primary text reply is now sequenced behind N media uploads, over a webhook that expires.**

Residual cost of the reordering requested in #3931174654 and delivered here. The accuracy goal is met; this is about what the new order costs the reply itself.

`deliverImages` (`dingtalk-image.ts:105-121`) is strictly sequential and does two authenticated round trips per image — `uploadImageMedia` then `sendImageMessage` — each bounded only by `fetchWithTimeout`. All of that now runs before `replyToDingTalk(sessionWebhook, body)` at line 481. The deleted comment stated the invariant this replaces: *"Best-effort and post-text: image delivery needs an access token + robot-send permission (config), and a failed image must never break the primary reply."*

Scenario: a turn renders three visuals while `api.dingtalk.com` media upload is degraded. Six `fetchWithTimeout` expiries elapse before a single character of the answer is sent, and the answer travels over `sessionWebhook`, which DingTalk scopes to a short window from message receipt (`sessionWebhookExpiredTime`). Text that used to be delivered first and immediately can now arrive late, or — at the tail of that window — not at all, because image delivery consumed the budget.

Suggested fix: same shape as the Lark case — resolve `media_id` per image before composing the copy (that alone gives the delivered count this branch needs), and send the image messages after `replyToDingTalk`. Failing that, bound the whole `deliverImages` loop with one deadline so the reply can never be delayed by more than a fixed amount, rather than by `2 × images × fetchWithTimeout`.

src/gateway/channels/lark.ts
```diff
@@ -2087,7 +2144,7 @@ async function processQueuedLarkMessage(ctx: QueuedLarkMessageContext): Promise<
     // Only solicit 👍/👎 on a real answer — never under an error or
     // empty-result notice, where a click would write a rating against a
     // non-answer and skew the feedback signal Metrics aggregates.
-    const isAnswer = !agentError && resultText.trim().length > 0;
+    const isAnswer = !agentError && !sessionBusy && strippedBody.trim().length > 0 && resultText.trim().length > 0;
```
jacoblee-io: **LOW — the `strippedBody` gate also silences feedback on a fully successful visual-only answer.**

Builds directly on #3931784503, whose suggested fix was `strippedBody.trim().length > 0` — implemented verbatim here. It closes the over-correction it was aimed at, but it keys on "did prose survive the strip", and a *successful* visual answer strips to nothing by design.

Scenario: the user asks for a topology diagram, the model answers with a single ```` ```mermaid ```` fence and nothing else, `render_mermaid` **succeeds**, and the PNG is delivered. `renderedVisualKinds` then contains `mermaid`, so line 2115 strips the fence and `strippedBody === ""`; no failure occurred, so no notice is appended and line 2132 substitutes `已生成图片如下。`. `isAnswer` is false, `finalizeCard` is called with `undefined`, and a correct, delivered answer ships with no 👍/👎 — where before this PR it had them. The comment two lines above scopes the exclusion to "an error or empty-result notice"; this is neither.

Note the interaction with #3932039405's per-kind stripping: the more reliably the fences are stripped on the success path, the more often a visual-only answer lands in this branch, so the blind spot widens rather than staying an edge case. Metrics loses the feedback signal for exactly the answers this PR is about.

Suggested fix: gate on "nothing usable survived **and** something failed", not on the stripped body alone —

```ts
const usableAnswer = strippedBody.trim().length > 0
  || (visualRenderFailures.size === 0 && deliveredImageCount > 0);
const isAnswer = !agentError && !sessionBusy && usableAnswer && resultText.trim().length > 0;
```

That keeps #3931784503's notice-only case unrated while letting a delivered image count as the answer it is.

src/gateway/channels/lark.ts
```diff
@@ -3161,9 +3228,26 @@ export async function collectChannelResponse(
       }
 
       if (ev.type === "tool_execution_end" || ev.type === "tool_end") {
-        if (options.includeImages) collectImageAttachments(ev.result?.content, images, seenImageKeys);
+        const name = (ev.toolName as string) || (ev.name as string) || "tool";
+        const visualKind = visualSourceKindFromToolName(name);
```
jacoblee-io, line 3269: **LOW–MEDIUM — kind attribution is wired into only one of the three sites that fill `images`, so a collected image can leave every fence unstripped.**

Builds on #3932039405's per-kind tracking, which is the right model. The gap is coverage: `renderedVisualKinds` is populated **only** in this `tool_execution_end` / `tool_end` branch, while `collectImageAttachments` is called from two further places that are unchanged by this PR and feed the same `images` array —

- `lark.ts:3275-3277` — `message_end` with `role === "toolResult" | "tool"`
- `lark.ts:3282` — `message_end` with `role === "assistant"`

Those two existed precisely because different brains/replay paths surface tool output differently. Any image that reaches `replyImages` through them without a matching observed `render_*` tool end leaves `renderedVisualKinds` **and** `visualRenderFailures` both empty, so `stripSourceKinds` at line 2111 is an empty set and `stripVisualBlocks` strips **no** fence at all.

Scenario: a turn whose image arrives only via a `message_end` frame — a reconnected/replayed stream, or a renderer call wrapped by another tool so `visualSourceKindFromToolName` returns `null` for the observed name. The chart PNG is delivered *and* the raw ```` ```chart {"type":"bar","data":{…}} ```` spec is posted into the Feishu card beside it. The replaced `stripSourceBlocks: replyImages.length > 0` covered this case unconditionally, so this is a regression on the exact leak the PR closes, reachable through the collection paths the new tracking does not observe.

Suggested fix: keep the per-kind stripping as the precise path and add a conservative floor for unattributed images — e.g. when `replyImages.length > 0` and `renderedVisualKinds.size === 0 && visualRenderFailures.size === 0`, fall back to stripping all three kinds. Alternatively return the count of images collected *without* an attributed kind from `collectChannelResponse` so the caller can decide, rather than treating "no kind observed" as "no visual involved".

mcp/create-chart/src/visual-export-config.ts
```diff
@@ -0,0 +1,53 @@
...
+    throw new VisualToolConfigurationError(
+      "Visual export URL must be an absolute HTTP(S) URL",
+    );
+  }
+  if (url.protocol !== "http:" && url.protocol !== "https:") {
+    throw new VisualToolConfigurationError(
+      "Visual export URL must use the http or https scheme",
+    );
+  }
+  if (url.hash) {
```
jacoblee-io: **LOW — the fragment check validates the parsed URL but the function returns the raw string, so a trailing bare `#` slips through.**

Builds on #3931175019 / #3932039429. The check catches the hash-router form it was written for; it does not catch the degenerate one, because `URL` normalizes an empty fragment away:

```js
new URL("https://console.example.com/siclaw-visual-export#").hash   // "" → passes line 25
new URL("https://console.example.com/#/siclaw-visual-export").hash  // "#/siclaw-visual-export" → caught
```

Verified in node. The value then returns from line 30 as `configured.replace(/\/+$/, "")`, which does not touch a trailing `#`, so `visual-export.ts:39` navigates to `https://console.example.com/siclaw-visual-export##<payload>`. The page's hash parsing sees `#<payload>` prefixed by an empty segment, never decodes it, `__siclawExportVisuals` is never populated, and the call burns the full budget in `waitForFunction` before failing.

Consequence for the new error contract: this is a **deployment mistake** that reports itself as a renderer outage. `visualExportConfigurationWarning()` calls this same validator and returns `null`, so the startup warning added for #3931788531 says nothing; and because the failure happens inside the export, `handler.ts`'s `exportVisual` funnels it through `rendererError` to `error_kind: "renderer"` — the user is told `报告图片生成失败，请稍后重试。` forever and the model keeps re-attempting a renderer that is fine. That is precisely the misclassification #3931787962 was raised about, reachable again through this one input.

Suggested fix: derive the returned base from the parsed URL rather than the raw string, which normalizes trailing slashes and drops any empty fragment in one step:

```ts
if (url.hash || configured.endsWith("#")) throw new VisualToolConfigurationError(/* … */);
return `${url.origin}${url.pathname.replace(/\/+$/, "")}${url.search}`;
```

That keeps the raw-string normalization from being a second, weaker definition of the same rule.

src/gateway/channels/lark.ts
LikiosSedo: Fixed in bf08e5e3. Visual-only copy is now position-independent past tense, milestone updates stop before media delivery begins, and all Lark image operations share one 10-second delivery budget so the typing card cannot block indefinitely.

src/core/mcp-client.ts
LikiosSedo: Fixed in bf08e5e3. The stdio environment is merged once and reused by both the child transport and host timeout calculation. The bundled renderer is identified from the connected MCP server implementation name mcp-create-chart, so alternate command spellings such as node plus an index.js path no longer bypass the override.

src/gateway/channels/lark.ts
LikiosSedo: Fixed in bf08e5e3. Image collection now reports deduplicated additions; artifacts observed outside an attributable renderer event increment an unattributed count and trigger the conservative strip-all floor. Added a toolResult-only regression test.

src/gateway/channels/dingtalk.ts
LikiosSedo: Acknowledged. Per current scope, the DingTalk delivery-order follow-up is intentionally deferred and this thread remains unresolved. This update focuses on the renderer process and Lark channel path.

src/gateway/channels/lark.ts
LikiosSedo: Fixed in bf08e5e3. A successfully delivered visual now counts as a usable answer even when all prose was stripped, while failure-only notices remain unrated. The visual-only success test now verifies feedback controls are appended.

mcp/create-chart/src/visual-export.ts
LikiosSedo: Fixed in bf08e5e3. The withDeadline helper now accepts a promise factory, so the remaining-budget check runs before Playwright work starts and no rejected operation can be orphaned.

mcp/create-chart/src/visual-export-config.ts
LikiosSedo: Fixed in bf08e5e3. The validator rejects any configured fragment marker, including a trailing empty fragment, and returns a normalized URL reconstructed from the parsed origin, path, and query. Added trailing-fragment and normalization tests.

LikiosSedo:
Superseded by #563, which keeps only chart/Mermaid image rendering, removes the visual-card authoring path, preserves legacy reader compatibility, and carries the explicit renderer configuration and IM source-suppression fixes on a clean branch from current main.

PR #557, opened 2026-09-03
fix(delegation): persist what a delegated leg's own AgentBox writes

glow-Kev1n:
## Summary

A delegated leg's own AgentBox had every persistence write refused by the Runtime's ownership gate, so the sessions and transcripts of sub-agents spawned inside a delegated turn were never created. AgentBox treats persistence as best-effort and swallows the failure, so this was silent data loss. The gate now admits a session's delegation **target** alongside its **owner**, plus four decisions around that arm and one logging fix.

### Problem

A delegated leg is the one session whose owner is deliberately not its executor: the coordinator that created the leg keeps ownership, so it keeps the conversation, while the turn runs in the delegated peer's AgentBox, whose certificate names the peer.

`sessionBelongsToIdentity` compared only the owner. So everything that box persisted *for* that leg — a spawned sub-agent's session and its transcript, task-ledger events — was refused with `403 delegation parent session mismatch`, while the pointers to those rows, written by the coordinator-side stream, went through.

The visible result was a leg advertising sub-agent sessions that had never been created, and anyone opening that conversation finding nothing behind them. Nothing surfaced at the time, because the box logs persistence failures as best-effort and moves on.

`server.ts` was already receiving `target_agent_id` from Portal and dropping it on the floor, so no upstream change was needed to fix this.

This is still unfixed on `main` as of the rebase below: `sessionBelongsToIdentity` there is owner-only, `sessionOwnedByIdentity` does not exist, and `session-registry.ts` carries no `targetAgentId` at all.

### Solution

The gate accepts the session's delegation target as a second admission arm. This does not widen attribution: `target_agent_id` is the box the leg was handed to, so a box still cannot claim a session delegated to anyone else, and a top-level session has no target to fall back on.

**⚠️ This PR changes an authorization gate.** The arm was decided per write face rather than applied blanket:

| Write face | Arm | Why |
|---|---|---|
| `ensure_session` — subject | ❌ owner-only | It upserts a row. A Portal whose upsert touched more than `last_active_at` would otherwise let a delegated peer re-point the coordinator-owned leg row's agent/parent/origin at itself. |
| `ensure_session` — parent | ✅ | The fix itself: the peer's box creates a sub-agent session beneath the leg. |
| `append_message` / `append_event` | ✅ | The fix itself: sub-agent transcripts, task-ledger events, delegation events. |
| `emit_chat_event` | ✅ | Carries the live frames that fold a sub-agent card from Running to done — same session id, same problem. Revoking it would pin those cards on "Running…". |
| `update_message` / `update_tool_message` | ✅ | No caller today (row updates run on the coordinator side). Narrowing them would only plant a mine for whoever revives that path; the conclusion is recorded in a comment instead. |
| `channel.deliver_message` | ❌ owner-only | Carries no delegated state: the `channel_update` tool behind it is suppressed on a delegated turn by design (a delegated worker's output flows back through the coordinator, which owns the single visible identity), so a leg has no legitimate route there. The arm that lets a peer's box persist a sub-agent transcript should not double as a key to the conversation's channel. |

This is the only face where a permission is *removed*, so its blast radius is worth stating: `channel.deliver_message` has exactly **one** producer in the repo — `createChannelMessageExecutor` in `src/agentbox/session.ts`, fed only by the `channel_update` tool, which `available()` disables whenever `refs.delegation` is set. So the only traffic that can still reach this check is an agent pushing a progress card on a non-delegated session, where the caller *is* the owner and passes as before. (`channels/background-delivery.ts` is the in-process delivery registry downstream of the gate, not a sender through it.)

**Refuse only after re-reading the row.** The 3-arg `rememberSession` callers — `chat.send`, channels, scheduled tasks — know nothing about delegation fields, so whichever runs first after a restart or an eviction caches the leg owner-only, and a cache hit never re-reads. Without this, the gate would refuse that box for the entry's whole lifetime, reinstating the same silent loss with no signal; and with more than one Runtime in play the first caller is not reliably the one that knows the target.

A record now carries whether it came from the row itself, which separates *"this session has no target"* from *"the target was never known here"*. Only the second is re-read, only on the about-to-refuse path, and a failed re-read leaves the refusal standing rather than surfacing as a 500. One extra read per session per cache lifetime; nothing on the success path. Concurrent re-reads coalesce through the existing single-flight, since a restart delivers a burst of buffered callbacks at once.

**A refusal warning now names the sessions it was about.** The box swallows the failure, so that line is the only trace a dropped write leaves, and without the ids it cannot be tied to the conversation that lost content. A malformed payload no longer shares that wording, so a stream of real refusals stays readable.

Two small ones: `remember` treats an empty target as "not supplied" rather than as an instruction to clear (`??` only falls through on null/undefined), and `sessionOwnedByIdentity` documents that its extra strictness is routing-dependent — the same caveat `resolveUserForIdentity` already carries, since a leg relayed to another Runtime is cached there under the peer.

## Test Plan

Run on the current base — this branch is rebased onto `main`, and the one conflict was an import line (upstream added `ChatMessageMetadata` where this branch had widened the `session-registry` import; both are kept). **No production code changed in the rebase**: the gate, the registry and the resolver are byte-identical to what was verified below.

- [x] `npm test` — 299 files / 6664 passing. The 3 failures in `mcp/a2a-mcp-adapter/src/stdio.e2e.test.ts` are environmental and unrelated: it spawns an MCP subprocess and this machine's policy blocks executing binaries under `node_modules/`. They fail identically on a clean `main`.
- [x] Gateway suite in isolation — 75 files / 1492 passing
- [x] Manual verification on a live deployment (below)

New tests: the 7 write faces × 3 identities matrix on the gate; the cold-cache refuse-then-re-read sequence; that an authoritative "no target" costs no extra read; that a failed re-read yields 403 rather than 500; concurrent-refresh coalescing; the empty-target preservation; and that a delegated leg cannot deliver to the conversation's channel.

**Live verification.** A delegated turn that spawns a batch of sub-agents now persists every one of their sessions and transcripts, each message carrying the leg's trace id, with **zero gate refusals over 90 minutes**. That window included the cold-cache period right after a Runtime restart — the peer's box wrote its sub-agent rows 31 minutes after the restart, which is exactly the case the re-read exists for.

## Architecture Checklist

- **Deployment mode**: not applicable — no resource sync, skills, or filesystem writes.
- **Security model**: no new shell execution paths. This PR *does* change an authorization gate, which is why the per-face table above spells out where the new arm applies and where it deliberately does not.
- **DB parity**: no schema changes.
- **Tool protocol**: no new tools.

## General Checklist

## Known limitations, not addressed here

- Three near-duplicate session resolvers could be merged; that is polish and belongs in its own PR.
- `resolveUserForIdentity` stays narrow (it does not accept the delegation arm): those routes mutate cron schedules, so allowing them from inside a delegated leg is a product question about who may own a schedule, not the attribution gap this PR closes.
- Task-ledger events persist correctly now but still carry no trace id. Stamping one is a separate change with its own consumer-side decisions.

### DIFF
--- a/src/gateway/internal-api.ts
+++ b/src/gateway/internal-api.ts
@@ -19,7 +19,7 @@ import { randomUUID } from "node:crypto";
 import type { FrontendWsClient } from "./frontend-ws-client.js";
 import type { CertificateIdentity } from "./security/cert-manager.js";
 import type { ChatMessageMetadata } from "../shared/message-kinds.js";
-import { sessionRegistry } from "./session-registry.js";
+import { sessionRegistry, type SessionRecord } from "./session-registry.js";
 import {
   deliverBackgroundChannelMessage,
   deliverChannelVisibleMessage,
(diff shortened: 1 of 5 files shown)


glow-Kev1n:
Rebased onto `main` (70 commits) to clear the conflict. The only overlap was one import line — upstream added `ChatMessageMetadata` where this branch had widened the `session-registry` import — so both are kept. **No production code changed in the rebase**: `sessionBelongsToIdentity`, the registry and the gate are byte-identical to what was reviewed and verified.

Re-verified on the new base: `npx tsc --noEmit` clean, gateway suite 75 files / 1492 passing, full suite 6664 passing. The 3 `a2a-mcp-adapter` stdio e2e failures are unchanged and environmental (this machine's policy blocks executing binaries under `node_modules/`; they fail identically on a clean `main`).

Also confirmed while rebasing that `main` has not fixed this independently: `sessionBelongsToIdentity` there is still owner-only, `sessionOwnedByIdentity` does not exist, and `session-registry.ts` carries no `targetAgentId` at all.

src/gateway/internal-api.ts
```diff
@@ -58,7 +58,75 @@ function sendJson(res: http.ServerResponse, status: number, data: unknown): void
...
+ * Accepting the target does not widen attribution: `target_agent_id` is the box
+ * the leg was handed to, so a box still cannot claim a session delegated to
+ * anyone else, and a top-level session has no target to fall back on.
+ */
 async function sessionBelongsToIdentity(sessionId: string | null | undefined, identity: CertificateIdentity): Promise<boolean> {
+  if (!sessionId) return true;
+  const owner = await sessionRegistry.get(sessionId);
+  if (!owner) return true;
+  if (owner.agentId === identity.agentId) return true;
+  if (owner.targetAgentId) return owner.targetAgentId === identity.agentId;
```
jacoblee-io, line 85: **[P1] Keep delegated message updates owner-only**

This target-accepting helper is also used by `delegation.update_message` / `delegation.update_tool_message`, which lets the target rewrite any existing message ID in the coordinator-owned leg. The target needs append access to persist its own descendants, but update is a materially broader session-wide mutation primitive and the payload has no `fromAgentId` ownership check. Please authorize those two events against the authoritative session owner instead (not the cache-only owner helper, since cross-runtime `chat.send` may cache the peer as owner), and add a regression test showing a target can append but cannot rewrite the delegated leg.

jacoblee-io (commented):
One blocking authorization issue: delegated targets need append access for their own execution records, but they must not gain mutation access to the coordinator-owned leg history.

src/gateway/session-registry.ts
```diff
@@ -46,12 +74,54 @@ export class SessionRegistry {
...
+    // `||`, not `??`: an empty string is a caller that did not supply a target,
+    // not one asking to clear it.
+    const target = targetAgentId || cached?.targetAgentId;
+    this.write(sessionId, {
+      userId,
+      agentId,
+      ...(target ? { targetAgentId: target } : {}),
+      // Provenance survives too: a 3-arg caller cannot unlearn that the row was
+      // already read.
+      ...(cached?.authoritative ? { authoritative: true } : {}),
```
jacoblee-io: **[P1] Do not preserve authoritative provenance across an owner rewrite**

`remember()` carries `authoritative: true` forward even when a 3-arg caller overwrites `agentId`. One concrete sequence is: the resolver caches `{ agentId: "coordinator", targetAgentId: "peer", authoritative: true }`; the target Runtime later handles `chat.send` and calls `remember(sessionId, userId, "peer")`; the entry now says the peer is the owner while remaining authoritative. `sessionOwnedByIdentity()` then trusts that cache without re-reading the row, reopening `update_message`, `update_tool_message`, and channel delivery to the delegated peer. Please preserve authoritative provenance only while the row-sourced owner/user fields remain unchanged, or clear it when a non-source caller changes them, and add a regression test for this sequence.

src/gateway/internal-api.ts
```diff
@@ -58,10 +58,101 @@ function sendJson(res: http.ServerResponse, status: number, data: unknown): void
...
+  try {
+    fromRow = await sessionRegistry.refresh(sessionId);
+  } catch (err) {
+    // Nothing is written either way, so this must not reach the handler's outer
+    // catch as a 500: that would tell a box reading its own logs that the platform
+    // is broken when the answer is simply unavailable.
+    console.warn(`[internal-api] could not re-read session ${sessionId} for an owner-only check: ${String(err)}`);
+  }
+  if (fromRow?.authoritative) return fromRow.agentId === identity.agentId;
+  return cached.agentId === identity.agentId;
```
jacoblee-io: **[P1] Keep owner-only mutations closed when the authoritative lookup fails**

This fallback trusts `cached.agentId` after the row refresh throws. For a relayed leg, the ordinary sequence is: the resolver first caches `{ agentId: "coordinator", targetAgentId: "peer", authoritative: true }`; the target Runtime then handles `chat.send` and re-remembers the leg as `agentId: "peer"`, correctly clearing `authoritative` but preserving the target. If Portal is temporarily unavailable during the next owner-only check, this line returns true for the peer and allows `update_message`, `update_tool_message`, or channel delivery. I reproduced that sequence: the peer's `update_message` returned 200 instead of 403. Since the delegated peer has no legitimate route to these owner-only operations, an unavailable source of truth must not grant the permission; at minimum reject when the non-authoritative cache already identifies the caller as `targetAgentId`, and add this outage sequence as a regression test.

jacoblee-io (approved):

PR #559, opened 2026-09-04
refactor(timing): move model-call timing to the provider boundary

caihongwei2006:
## What this fixes

A turn's elapsed time could not be taken apart, which made the audit trail
unusable for the one question it exists to answer — where did the time go. Four
reasons, all on the collection side:

1. **A model call that returned only tool calls had no row at all** — its time
   was parked on the tool rows as `pre_thinking_ms`, so the bulk of model time
   lived on tools.
2. **The wait for the provider's first byte was never measured** — the clock
   started when the gateway saw an SSE event, not when the request left.
3. **Time-to-first-token and output duration are two different quantities**, and
   `thinking_ms` blended them with reasoning — so "the provider is slow", "the
   model reasoned a while" and "the answer was long" all looked the same.
4. `turn_total_ms` was cumulative despite its name, and ttft/thinking overlapped
   so badly they needed a `NOISE_FLOOR` to deduplicate.

Every number in the audit trail was therefore inferred. This records each request
at the `chat.completions` boundary instead, so the trail is measured.

## How

- **`src/core/llm-call-recorder.ts`** wraps `agent.streamFn` innermost, before the guard pipeline, so guard work lands in setup rather than in `net_ttft`. Per request, on **one clock**: request / headers / first-token / block-edge / response-end timestamps, the provider's `usage` (reasoning tokens included) and `stop_reason`. `ms.net_ttft + ms.thinking + ms.output == ms.total`, by construction.
- **One agent-loop call ⇒ one assistant row** carrying `metadata.llm_call`, **even when the content is empty**. Reasoning text becomes its own `kind: "thinking"` row written just before it. Tool rows carry `metadata.llm_round` + `metadata.tool_call_id`, so a tool's group is known without consulting a clock.
- **Failed and rolled-back calls are kept.** A call that errored rides the `error_response` row; a routing attempt that was rolled back rides `model_route_notice` as `discarded_llm_calls`. Both are time the user waited for, and neither had a carrier before.
- **Removed:** `turnStartTime` / `firstTokenTime` / `lastBoundaryTime` / `pendingThinkingMs` / `NOISE_FLOOR_MS`, `pre_thinking_ms`, `timing.*`, and the portal's cross-pod `turnStartMs` anchor. Not dual-written — the inference logic is the thing being removed.

## Tool durations use a single clock

Both ends of an interval come from the same clock: the agentbox's when the runtime stamped both `startedAt` and `endedAt`, the gateway's when it stamped neither.

**Defaulting a missing stamp to `Date.now()` is wrong**: that subtracts one pod's clock from another's, and the surrounding `Math.max(0, …)` dresses the skew up as a plausible measurement rather than surfacing it — it never even looks wrong. Regression test: `never mixes the agentbox and gateway clocks when one of the stamps is missing`.

## ⚠️ Data-format change for readers of `chat_messages`

Two new row shapes exist and both must be recognised before this ships, or a
turn reads wrong:

- `kind: "thinking"` — a model call's own reasoning, its own row. It is not a
  reply to the user: counting it as one inflates reply counts and makes an
  unanswered turn look answered.
- An **empty** model-call row — a call that produced only tool calls still gets
  a row, because it is the carrier for that call's timing, tokens and reasoning
  link. It is identified by `metadata.llm_call` with empty `content`, not by a
  kind.

Both are documented field by field on the envelope type. Historical rows carry
neither and are simply reported as having no envelope telemetry — nothing is
back-filled and nothing is inferred from the retired fields.

## Relationship to #536

Both PRs write **the same fact** onto tool rows — which model call dispatched this tool — through two different mechanisms. This PR writes `metadata.llm_round` + `metadata.tool_call_id` directly; #536 sends a `toolset` envelope for the consumer to unpack. Land both and the same round number is persisted twice.

#536 also conflicts with this PR across 6 files (`PiAgentBrain`'s constructor: one removes `toolsetsByName`, the other keeps it) and with `main` across 5. Worth settling which mechanism we keep before either merges.

## Tests

Full `vitest`: 6667 passed / 2 skipped. `tsc` clean (root + portal-web). CI green.


src/gateway/sse-consumer.ts
```diff
@@ -846,6 +833,21 @@ export async function consumeAgentSse(opts: ConsumeAgentSseOptions): Promise<Sse
       if (eventType === "message_end" || eventType === "turn_end") {
         const message = evt.message as Record<string, unknown> | undefined;
         if (message?.role === "assistant") {
+          // ── LLM-call envelope: the timing source of truth ──
+          // Stamped by LlmCallRecorder at the streamFn boundary; message_end and
+          // turn_end deliver the same object, so seenEnvelopes keeps the row
+          // single. Only agent-loop calls carry a round; aux (compaction) calls
+          // never surface here — they ride the next agent call's `aux_calls`.
+          const rawEnvelope = llmCallFromMessage(message);
+          const envelope = rawEnvelope && !seenEnvelopes.has(rawEnvelope) ? rawEnvelope : undefined;
```
jacoblee-io: **Confirmed by repro — this dedup never fires in production.** `seenEnvelopes` is an identity-based WeakSet, but `message_end` and `turn_end` arrive as separately JSON-parsed SSE frames: pi-agent-core emits `turn_end` with the SAME assistant message object (agent-loop.js:108,130), but the agentbox relays each brain event as its own `data: <json>` frame and `AgentBoxClient` parses each independently — at the gateway the two envelopes are always DISTINCT objects. Empirically reproduced through the real `consumeAgentSse`: feeding `message_end` + a JSON-round-tripped `turn_end` writes 2 model-call rows and 2 thinking rows instead of 1+1.

Consequences per production turn: an extra hidden empty assistant row with a duplicate `llm_call` (doubling `/metrics/timing` counts), a duplicate thinking row (which also duplicates reasoning text in delegate-api's finalText join), the envelope pushed twice into `attemptLlmCalls` (duplicated `discarded_llm_calls` on route notices), and `lastAssistantDbMessageId` overwritten so the `agent_end` `context_usage` snapshot is patched onto the empty duplicate instead of the real answer — the context meter is lost on reload.

The unit test covering this passes only because it reuses one object reference, i.e. it pins exactly the assumption production violates. The same identity-dedup pattern is copied into lark.ts:3223 (harmless there today only because its row writes are gated to `message_end`). Dedup must key on envelope VALUES (e.g. attempt+round+request_at), or the agentbox must strip `llmCall` from the `turn_end` copy.

src/gateway/sse-consumer.ts
```diff
@@ -877,6 +882,7 @@ export async function consumeAgentSse(opts: ConsumeAgentSseOptions): Promise<Sse
                     kind: "error_response",
                     error_code: ErrorCodes.MODEL_ERROR,
                     retriable: true,
+                    ...(errorEnvelope ? { llm_call: errorEnvelope } : {}),
                   },
```
jacoblee-io, line 887: Redaction bypass: `envelope.error_message` is persisted unredacted inside `metadata.llm_call` while the identical text in the row's `content` goes through `redactText`. `appendMessage` does not redact metadata (only tool-result metadata takes extractPersistableDetails' JSON round-trip redaction). A provider error that echoes request content or credentials — e.g. a proxy 4xx quoting the URL with an API key, exactly what the redaction patterns exist for — lands verbatim in `chat_messages.metadata` (sliced to 500 chars in the recorder, never redacted) and is returned by every history/metrics endpoint. Same gap in lark.ts:3243 and in the `discarded_llm_calls` attached to the route-switch notice (~601).

src/gateway/sse-consumer.ts
```diff
@@ -937,111 +943,69 @@ export async function consumeAgentSse(opts: ConsumeAgentSseOptions): Promise<Sse
...
-              firstAssistantPersisted = true;
-              // On a routed turn the primary streams live before we know it won;
-              // defer the durable write to the commit point so a failed primary's
-              // reply isn't left in the DB after a fallback takes over.
-              if (isRoutingTurn && !routingCommitted) pendingAssistantOps.push(persistAssistant);
-              else await persistAssistant();
-            }
-            assistantContent = "";
+                rowEnvelope.thinking_row_id = thinkingRowId;
+              }
```
jacoblee-io, line 980: ADR-018 claims thinking rows are "excluded from reply counts", but nothing implements that: `persistAssistant` now calls `incrementMessageCount` twice per model call (thinking row + model-call row), and tool-only calls that previously persisted nothing now count too. `chat_sessions.message_count` feeds the session list, and the /metrics daily `message_count` is a bare `COUNT(*)` over `chat_messages` with no kind filter (siclaw-api.ts:3611) — a 10-round reasoning investigation reports ~30 "messages" where it used to report ~10 (and the turn_end double-write doubles the new rows again). Either drop the increments for hidden kinds or filter `kind IN ('thinking', …)` out of the counting queries.

src/gateway/sse-consumer.ts
```diff
@@ -1113,7 +1077,7 @@ export async function consumeAgentSse(opts: ConsumeAgentSseOptions): Promise<Sse
             sessionId,
             role: "assistant",
             content: redactText(cleaned, redactionConfig),
-            metadata: { incomplete: true },
+            metadata: { incomplete: true, ...(currentRound > 0 ? { llm_round: currentRound } : {}) },
```
jacoblee-io: The stream-cut "incomplete" assistant row is stamped `llm_round = currentRound`, but `currentRound` only advances on a `message_end` envelope — and a mid-call stream break is exactly the case where the in-flight call's envelope never arrived. The partial text was produced by round `currentRound+1`, so it's misattributed to the PREVIOUS completed call: a timeline reader joins the incomplete narration onto the wrong model call's group. (For a first-call cut the spread correctly omits the key, making the field inconsistent between first-call and later-call cuts.)

src/gateway/channels/lark.ts
```diff
@@ -3193,9 +3225,29 @@ export async function collectChannelResponse(
...
+        const rawEnvelope = llmCallFromMessage(ev.message);
+        const envelope = rawEnvelope && !seenEnvelopes.has(rawEnvelope) ? rawEnvelope : undefined;
+        if (envelope) {
+          seenEnvelopes.add(envelope);
+          if (envelope.kind === "agent" && envelope.round > 0) currentRound = envelope.round;
+        }
+        // turn_end only contributes round tracking (its text was handled on
+        // message_end). A failed call's envelope rides an error row, as in sse-consumer.
+        if (ev.type === "message_end" && persist && envelope && ev.message?.stopReason === "error") {
+          await persistRow({
```
jacoblee-io: This persists an `error_response` row immediately on every errored `message_end`, violating the error-surfacing revocation contract (docs/design/2026-08-02-error-surfacing-contract.md: deduped, revocable, last-error-wins) that sse-consumer implements via `pendingErrorOps`. pi retries transient model errors in-turn: a failed attempt emits `message_end` with stopReason 'error', then a retry can succeed — which is why sse-consumer buffers the error row and drops it when a later message carries real output, and why its rollback path clears it. Writing on sight means a 429 pi recovers from, or a routed primary that falls back, leaves one permanent `error_response` row per failed attempt in the channel session — reload shows errors above a successful answer, and retries stack multiple rows where the contract is ONE row per run carrying the LAST error. The sse-consumer test "carries a rolled-back primary's model calls…" pins the opposite behavior for the web path.

jacoblee-io (commented):
Reviewed against pi-agent-core/pi-ai sources and with an empirical repro through the real `consumeAgentSse`. Withholding approval for now: the first finding is **confirmed by a failing repro** — every production turn double-processes its model call and persists duplicate rows — and two more findings leak content (chain-of-thought into delegation results; unredacted provider errors into metadata). Inline comments below; one finding can't be anchored because the file isn't in this diff:

**`src/gateway/delegate-api.ts:249` — `recoverRemoteResult`'s finalText join leaks chain-of-thought into every remote delegation result.** The durable readback is the AUTHORITATIVE final text for remote delegations (relay text is never trusted), and its filter is only `role==='assistant' && kind !== 'error_response' && content non-empty`. The new `kind: "thinking"` rows are role='assistant' with the full reasoning text as content, so a peer running a reasoning model returns `finalText = '<reasoning>\n\n<answer>'` to the coordinator on every delegated turn (doubled again by the dedup bug), and the coordinator quotes raw CoT to the user / into its own context. The filter needs `kind !== 'thinking'` (and to skip the empty model-call rows, which the content trim currently happens to catch).

src/gateway/channels/lark.ts
```diff
@@ -3093,12 +3095,31 @@ export async function collectChannelResponse(
...
   const toolInputs = new Map<string, string[]>();
-  const toolStarts = new Map<string, number[]>();
+  const toolStarts = new Map<string, Array<{ startedAt?: number; localStartMs: number }>>();
   const toolsets = new Map<string, Array<string | undefined>>();
+  const toolRounds = new Map<string, Array<Record<string, unknown>>>();
   const pushQ = <T,>(m: Map<string, T[]>, k: string, v: T): void => { const a = m.get(k) ?? []; a.push(v); m.set(k, a); };
   const shiftQ = <T,>(m: Map<string, T[]>, k: string): T | undefined => m.get(k)?.shift();
   const toolKey = (ev: Record<string, any>, name: string): string =>
     ev.toolCallId != null ? `id:${String(ev.toolCallId)}` : `name:${name}`;
+  // LLM-call timeline (same contract as sse-consumer.ts): the latest agent-loop
```
jacoblee-io: Reuse: this re-implements sse-consumer's llm-call bookkeeping (eventClock, `roundMetaFor` ≅ `toolRoundMetadata`, seenEnvelopes/currentRound tracking, the one-clock duration rule) as inline copies, even though the PR itself extracted `buildThinkingRow`/`isModelCallRowEnvelope` into `src/gateway/llm-call-rows.ts` precisely because two persistence paths must agree. The round/clock/dedup half of the contract is held together only by a comment ("same contract as sse-consumer.ts"), and the copies have already diverged structurally (`Array<{startedAt?, localStartMs}>` queues vs `PendingToolCall`) — so the next contract change, e.g. the value-based dedup fix for the WeakSet bug, must be discovered and applied twice or the channel path silently drifts. Worth lifting the shared half into llm-call-rows.ts.

src/core/llm-call-recorder.ts
```diff
@@ -0,0 +1,511 @@
...
+      stop_reason: typeof message?.stopReason === "string" ? message.stopReason : undefined,
+      error_message: typeof message?.errorMessage === "string" ? message.errorMessage.slice(0, 500) : undefined,
+      thinking_visible: thinkingVisible,
+      tool_call_ids: call.toolCallIds,
+    };
+
+    if (call.kind === "aux") {
+      this.pendingAux.push(envelope);
+    } else {
+      this.round += 1;
```
jacoblee-io, line 500: A pi in-turn auto-retry consumes a round for the failed attempt, but the failed envelope's only carrier (the error row) is revoked when the retry succeeds — so round numbering gets holes and the failed call vanishes from the timeline. Each attempt is a separate `streamFn` call, so `sealCall` assigns it round N; its envelope rides only the deferred error row, which sse-consumer deliberately drops when a later attempt produces output (the revocation contract). Persisted rows then show rounds …, N−1, N+1 with no N — contradicting ADR-018's "rounds are numbered from 1 per prompt" contiguity that sicore's timeline decoder consumes, and its "✅ failed calls … appear on the timeline" claim (the retry's 30s is silently folded into the next call's `since_prev_ms`, which readers label as tool-group span). Routing failures get `discarded_llm_calls` as a carrier; internal-retry failures get nothing.

src/gateway/sse-consumer.ts
caihongwei2006: Fixed in ab568de6. Dedup is now centralized in LlmCallTimeline and keyed by envelope values rather than object identity, so independently JSON-parsed message_end and turn_end frames collapse to one call. The regression test explicitly JSON-round-trips the second frame. Terminal error buffering also preserves the first deduplicated envelope instead of replacing it on turn_end.

src/gateway/sse-consumer.ts
caihongwei2006: Fixed in ab568de6. Every persisted llm_call copy now goes through redactLlmCallEnvelope, including nested aux_calls and discarded_llm_calls. Provider error_message values are redacted before they enter metadata in both SSE and Lark persistence paths.

src/gateway/sse-consumer.ts
caihongwei2006: Clarified in ab568de6 after checking the production Sicore persistence contract. siclaw_chat_sessions.message_count is transactionally incremented for every appended physical row, and the timeline design treats it as a physical-row count. The local-only exclusion was removed because it would make standalone/local behavior diverge from production; the ADR now states that thinking and empty carrier rows remain included.

src/gateway/sse-consumer.ts
caihongwei2006: Fixed in ab568de6. A stream-cut partial row is now attributed to timeline.nextRound(), i.e. the in-flight currentRound + 1 call, instead of the preceding completed round. Regression coverage includes a cut after an earlier completed call.

src/gateway/channels/lark.ts
caihongwei2006: Fixed in ab568de6. Lark now buffers model errors with last-error-wins semantics, drops them when a later attempt produces output or routing rolls back, and flushes only the terminal failure. Failed retries that recover remain as empty model-call rows rather than stale error_response rows.

src/gateway/channels/lark.ts
caihongwei2006: Fixed in ab568de6. Value-based envelope dedup, current-round tracking, event-clock extraction, tool metadata, duration calculation, redaction, and model-call row eligibility now live in the shared llm-call-rows module and are used by both SSE and Lark paths.

src/core/llm-call-recorder.ts
caihongwei2006: Fixed in ab568de6. Failed in-turn retries are retained in pendingFailedLlmCalls. On recovery they are persisted as empty model-call rows; on terminal failure all earlier calls are written before the final error_response carrier. The persisted rounds are contiguous and the retry time no longer becomes an unexplained tool-group span.

caihongwei2006:
Also addressed the unanchored delegation finding in ab568de6: recoverRemoteResult now excludes kind=thinking rows, with regression coverage. Follow-up review gaps are covered as well: terminal message_end plus turn_end keeps the failed call envelope; tool-only live calls no longer overwrite an older visible bubble timing; Lark route rollback persists redacted discarded_llm_calls on the switch notice; and synchronous, stream-creation, iterator, or result rejection paths seal the provider call and attach it to the synthetic pi-agent failure message. Trailing post-answer compaction is explicitly retired at endPrompt and logged as prompt-tail residual under the existing cross-service contract; it is not mislabeled as an agent round or carried into the next prompt.

PR #560, opened 2026-09-04
fix(knowledge): knowledge_cite accepts page paths carrying the mount prefix

LikiosSedo:
## Why

In live traces (siflow-test, hardware KB) every answer spent two `knowledge_cite` calls: the first passed a page path copied from the model's own `read` call, `.siclaw/knowledge/repos/<repo>/topics/招摇B30X.md`, and was rejected as unread because the path was joined literally onto `knowledgeDir`; the retry with `repos/...` passed.

## What

`resolveCitedPagePath` strips leading path segments until the path lands on a page **read this turn**; otherwise it falls back to the literal resolution, so an unread page still fails as unread however it is spelled. Applied to both `pages[].path` and `evidence_refs`.

## Test

- new case in `src/core/knowledge-citation-tool.test.ts`: mount-prefixed relative path cites; the same prefix on an unread page still fails.
- `npm test`: 299 files, 6652 passed.

### DIFF
--- a/src/core/knowledge-citation-tool.test.ts
+++ b/src/core/knowledge-citation-tool.test.ts
@@ -63,6 +63,22 @@ describe("knowledge_cite", () => {
     expect(events).toHaveLength(1);
   });
 
+  it("tolerates a mount-prefixed relative path copied from the model's own read call", async () => {
+    const { dir, page } = fixture();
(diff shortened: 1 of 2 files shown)


LikiosSedo:
Superseded by #558, which now includes this cite path-prefix fix on the consolidated KB branch.

PR #562, opened 2026-09-04
refactor(timing): move model-call timing to the provider boundary

caihongwei2006:
## Summary

Record model-call timing at the provider boundary so tool-only calls, provider latency, reasoning, output, and failed routing attempts can be accounted for independently. Persist one assistant carrier per model call, bounded/redacted thinking rows, and tool-to-call attribution; remove the former gateway timing inference.

Transcript queries preserve legacy NULL-metadata messages and filter hidden telemetry before pagination in both chat and task-run traces. Thinking storage failures cannot prevent answer persistence, and stopped tools retain their call linkage.

## Timing and persistence contract

- `metadata.llm_call.ms` records `net_ttft`, `thinking`, `output`, and `total` on one clock. Provider usage and stop reason travel with the envelope.
- The chat bubble badge now shows that message's single model-call duration, not cumulative turn time. For a multi-round tool turn, the final answer badge covers the final call only; the UI does not currently display the whole turn's wall-clock duration.
- SSE and Lark persist failed calls on error/carrier rows and rolled-back calls on route notices. Buffered failed attempts send only their envelopes with the switch event; discarded text is not replayed.
- Thinking text is redacted and clipped on a UTF-8 boundary to the existing MySQL TEXT limit (65,535 bytes including the marker). Truncation is explicit in metadata.
- Thinking rows and empty assistant carriers with `llm_call` and no `kind` are hidden in transcripts. Explicit error/route kinds remain visible. Physical database message counts still count physical rows.
- Synthetic notification turns now preserve recorder attempt state across fallback, but retain their existing conditional report persistence; they do not have the full SSE/Lark database timeline contract.
- Historical rows without envelopes have no timing telemetry. There is no backfill or fallback to retired timing fields.

## Deploy requirement

Rebuild the agentbox image, bump `SICLAW_AGENTBOX_IMAGE`, then recycle running agentbox pods. `imagePullPolicy: Always` pulls only when a pod is created. A gateway-only rollout leaves old pods emitting no envelopes and their sessions reporting timing count 0 until recycled.

## Relationship to #536

Both changes carry tool-to-model-call attribution through different mechanisms. Reconcile that overlap before merging both.

## Test Plan

- [x] Backend full suite: 6,694 passed, 2 skipped; adapter/transcript/recovery targeted suite: 280 passed.
- [x] Frontend suite: 249 passed.
- [x] Root TypeScript build and frontend production build.
- [x] Real node:sqlite chat/count/trace queries; multi-byte reasoning cap and failed insert; parallel tool abort linkage; buffered routing envelope delivery and redaction; synthetic recorder lifecycle.
- [ ] Live MySQL and production rollout validation.

### DIFF
--- a/portal-web/src/hooks/usePilotChat.ts
+++ b/portal-web/src/hooks/usePilotChat.ts
@@ -382,26 +382,49 @@ export function dropFailedAttemptOutput(messages: PilotMessage[]): PilotMessage[
   return [...head, ...tailHidden];
 }
 
-function extractTiming(metadata: Record<string, unknown> | undefined, durationMs: number | null | undefined): import("../components/chat/types").MessageTiming | undefined {
-  // Tool rows: duration_ms (⚙️ exec) is its own column; pre_thinking_ms
-  // (💭 model-thinking-before-this-tool) lives at metadata.pre_thinking_ms.
-  // Assistant rows: timing sub-object inside metadata holds ⏳ ttft, 💭
-  // thinking-before-text, and turn-total.
+/** Map a runtime `llm_call.ms` partition onto the chat's MessageTiming shape. */
+function timingFromLlmCall(llmCall: unknown): import("../components/chat/types").MessageTiming {
   const t: import("../components/chat/types").MessageTiming = {}
-  const rawTiming = metadata?.timing as Record<string, unknown> | undefined
-  if (rawTiming) {
-    if (typeof rawTiming.ttft_ms === "number") t.ttftMs = rawTiming.ttft_ms
-    if (typeof rawTiming.thinking_ms === "number") t.thinkingMs = rawTiming.thinking_ms
-    if (typeof rawTiming.output_ms === "number") t.outputMs = rawTiming.output_ms
-    if (typeof rawTiming.turn_total_ms === "number") t.turnTotalMs = rawTiming.turn_total_ms
+  const ms = (llmCall as { ms?: Record<string, unknown> } | undefined)?.ms
+  if (!ms || typeof ms !== "object") return t
+  if (typeof ms.net_ttft === "number") t.netTtftMs = ms.net_ttft
+  if (typeof ms.thinking === "number") t.thinkingMs = ms.thinking
+  if (typeof ms.output === "number") t.outputMs = ms.output
+  if (typeof ms.total === "number") t.totalMs = ms.total
+  return t
+}
+
+/** Attach a live call only to the bubble produced by that call. Tool-only calls
+ * have no streaming assistant bubble and therefore stay as hidden rows after
+ * reload instead of overwriting the previous visible answer's timing. */
+export function attachLiveAssistantCall(
+  messages: PilotMessage[],
+  timing?: import("../components/chat/types").MessageTiming,
+  route?: ModelRouteMetadata | null,
+): PilotMessage[] {
+  if (!timing && !route) return messages
+  const idx = findLastMessageIndex(messages, (m) =>
+    m.role === "assistant" &&
+    m.metadata?.kind !== "model_route_notice" &&
+    m.metadata?.kind !== "delegation_status_notice",
+  )
+  if (idx < 0) return messages
+  const current = messages[idx]
+  const applyTiming = timing && current.isStreaming
+  if (!applyTiming && !route) return messages
+  const mergedTiming = applyTiming ? { ...(current.timing ?? {}), ...timing } : current.timing
+  const next = {
+    ...current,
+    ...(mergedTiming && Object.keys(mergedTiming).length > 0 ? { timing: mergedTiming } : {}),
+    ...(route ? { metadata: { ...(current.metadata ?? {}), model_route: route } } : {}),
   }
-  // Pre-tool thinking gets surfaced as the same `thinkingMs` field — it's
-  // semantically the same emoji (💭 model reasoning) just attached to a
-  // tool row instead of a text row. UI uses one badge code path either way.
-  // No threshold: a 0/small 💭 on the 2nd-Nth tool of a batch is the visible
-  // proof that they came from one thinking burst (not N independent ones).
-  const preThinking = metadata?.pre_thinking_ms
-  if (typeof preThinking === "number") t.thinkingMs = preThinking
+  return [...messages.slice(0, idx), next, ...messages.slice(idx + 1)]
+}
+
+function extractTiming(metadata: Record<string, unknown> | undefined, durationMs: number | null | undefined): import("../components/chat/types").MessageTiming | undefined {
+  // Assistant model-call rows: metadata.llm_call.ms (measured at the provider
+  // boundary by the runtime). Tool rows: duration_ms is its own column.
+  const t = timingFromLlmCall(metadata?.llm_call)
   if (typeof durationMs === "number" && durationMs >= 0) t.durationMs = durationMs
   return Object.keys(t).length > 0 ? t : undefined
 }
@@ -556,6 +579,11 @@ export function toPilotMessage(m: ChatMessage): PilotMessage {
   // A background exec job's completion marker — folded into the launching tool's box
   // (annotateExecJobCompletions), never shown as its own row.
   const isExecJobEvent = metadata?.kind === "exec_job_event"
+  // A model call's reasoning text — its own row, folded into the timeline, never a bubble.
+  const isThinkingRow = metadata?.kind === "thinking"
+  // A model call that produced only tool calls still gets a row (it carries the
+  // call's timing/tokens in metadata.llm_call); there is nothing to render for it.
+  const isEmptyModelCallRow = m.role === "assistant" && !m.content?.trim() && metadata?.llm_call != null
   const staleRunning = isStaleRunningTool(m)
   const toolIsDelegation = isDelegationTool(m.tool_name)
   const toolStatus = toolStatusFromMessage(m)
@@ -598,7 +626,7 @@ export function toPilotMessage(m: ChatMessage): PilotMessage {
     toolDetails: m.role === "tool" ? (recoveredMetadata ?? undefined) : undefined,
     metadata: recoveredMetadata,
     timing,
-    hidden: m.hidden || isDelegationEvent || isTaskEvent || isTaskNotification || isExecJobEvent || isTaskTool(m.tool_name),
+    hidden: m.hidden || isDelegationEvent || isTaskEvent || isTaskNotification || isExecJobEvent || isThinkingRow || isEmptyModelCallRow || isTaskTool(m.tool_name),
     fromAgentId: m.from_agent_id ?? null,
     parentSessionId: m.parent_session_id ?? null,
     delegationId: m.delegation_id ?? null,
@@ -1710,11 +1738,6 @@ export function usePilotChat({ agentId, sessionId, forceLive }: UsePilotChatOpti
           const toolInput = formatToolInput(toolName ?? "", args)
           const hidden = toolName === "update_plan" || isTaskTool(toolName)
           const dbMessageId = evt.dbMessageId as string | undefined
-          // 💭 Pre-tool thinking time (gap from previous tool_end / turn-start
-          // to now). Server-stamped; matches the value persisted in the row's
-          // metadata.pre_thinking_ms, so live and reload render identically.
-          const preThinkingMs = typeof evt.preThinkingMs === "number" ? evt.preThinkingMs : undefined
-          const showThinking = preThinkingMs != null
 
           setMessages((prev) => [
             ...prev,
@@ -1730,7 +1753,6 @@ export function usePilotChat({ agentId, sessionId, forceLive }: UsePilotChatOpti
               timestamp: timeNow(),
               isStreaming: true,
               hidden,
-              ...(showThinking ? { timing: { thinkingMs: preThinkingMs } } : {}),
             },
           ])
           break
@@ -1875,38 +1897,16 @@ export function usePilotChat({ agentId, sessionId, forceLive }: UsePilotChatOpti
           const endMsg = evt.message as
             | { role?: string; toolName?: string; details?: Record<string, unknown> }
             | undefined
-          // ⏳/💭/✍️ Server-stamped per-message timing block, attached to
-          // the event by sse-consumer right before persistence.
-          const evtTiming = evt.timing as
-            | { ttft_ms?: number; thinking_ms?: number; output_ms?: number; turn_total_ms?: number }
-            | undefined
+          // The model call's timing partition, stamped by the runtime at the
+          // provider boundary and attached to the event by sse-consumer; the same
+          // values land in the row's metadata.llm_call, so live and reload agree.
+          const evtTiming = evt.llmCall ? timingFromLlmCall(evt.llmCall) : undefined
           const evtRoute = endMsg?.role === "assistant" ? (modelRouteFromEvent(evt) ?? currentModelRouteRef.current) : null
           if (endMsg?.role === "assistant" && evtRoute) {
             currentModelRouteRef.current = evtRoute
           }
           if (endMsg?.role === "assistant" && (evtTiming || evtRoute)) {
-            setMessages((prev) => {
-              const idx = findLastMessageIndex(prev, (m) =>
-                m.role === "assistant" &&
-                m.metadata?.kind !== "model_route_notice" &&
-                m.metadata?.kind !== "delegation_status_notice",
-              )
-              if (idx < 0) return prev
-              const current = prev[idx]
-              const timing = {
-                ...(current.timing ?? {}),
-                ...(typeof evtTiming?.ttft_ms === "number" ? { ttftMs: evtTiming.ttft_ms } : {}),
-                ...(typeof evtTiming?.thinking_ms === "number" ? { thinkingMs: evtTiming.thinking_ms } : {}),
-                ...(typeof evtTiming?.output_ms === "number" ? { outputMs: evtTiming.output_ms } : {}),
-                ...(typeof evtTiming?.turn_total_ms === "number" ? { turnTotalMs: evtTiming.turn_total_ms } : {}),
-              }
-              const next = {
-                ...current,
-                ...(Object.keys(timing).length > 0 ? { timing } : {}),
-                ...(evtRoute ? { metadata: { ...(current.metadata ?? {}), model_route: evtRoute } } : {}),
-              }
-              return [...prev.slice(0, idx), next, ...prev.slice(idx + 1)]
-            })
+            setMessages((prev) => attachLiveAssistantCall(prev, evtTiming, evtRoute))
           }
           if (endMsg?.role === "toolResult" && endMsg.details && Object.keys(endMsg.details).length > 0) {
             // Pi-agent brain: tool result details arrive via message_end (not tool_execution_end).
--- a/portal-web/src/hooks/usePilotChat.ts
+++ b/portal-web/src/hooks/usePilotChat.ts
@@ -583,7 +583,7 @@ export function toPilotMessage(m: ChatMessage): PilotMessage {
   const isThinkingRow = metadata?.kind === "thinking"
   // A model call that produced only tool calls still gets a row (it carries the
   // call's timing/tokens in metadata.llm_call); there is nothing to render for it.
-  const isEmptyModelCallRow = m.role === "assistant" && !m.content?.trim() && metadata?.llm_call != null
+  const isEmptyModelCallRow = m.role === "assistant" && !m.content?.trim() && metadata?.llm_call != null && metadata?.kind == null
   const staleRunning = isStaleRunningTool(m)
   const toolIsDelegation = isDelegationTool(m.tool_name)
   const toolStatus = toolStatusFromMessage(m)
--- a/src/gateway/channels/lark.ts
+++ b/src/gateway/channels/lark.ts
@@ -3205,12 +3205,13 @@ export async function collectChannelResponse(
         const discardedId = lastAssistantMessageId;
         lastAssistantMessageId = null;
         lastAssistantText = "";
+        const { knowledge_citations: _citations, ...retainedMetadata } = lastRowMetadata;
         try {
           await updateMessage({
             messageId: discardedId,
             sessionId,
             content: redact(lastRowContent),
-            metadata: { ...lastRowMetadata, knowledge_citations: undefined, discarded_route_attempt: true },
+            metadata: { ...retainedMetadata, discarded_route_attempt: true },
           });
         } catch (err) {
           console.warn(`[${logPrefix}] discard rolled-back assistant row failed session=${sessionId}:`, err);
--- a/src/gateway/channels/lark.ts
+++ b/src/gateway/channels/lark.ts
@@ -68,6 +68,16 @@ import {
   normalizeKnowledgeSourceCitations,
   type KnowledgeSourceCitation,
 } from "../../shared/knowledge-citations.js";
+import type { LlmCallEnvelope } from "../../core/llm-call-recorder.js";
+import {
+  buildThinkingRow,
+  eventClock,
+  isModelCallRowEnvelope,
+  LlmCallTimeline,
+  redactLlmCallEnvelope,
+  toolDurationMs,
+} from "../llm-call-rows.js";
+import { messageProducedOutput, modelRouteNoticeContent, modelRouteSwitchMetadata } from "../sse-consumer.js";
 import { registerBackgroundChannelDelivery } from "./background-delivery.js";
 
 const VISUAL_ONLY_NOTICE_BY_LOCALE = {
@@ -3088,6 +3098,7 @@ export async function collectChannelResponse(
   let lastAssistantMessageId: string | null = null;
   // Exact text persisted on lastAssistantMessageId (updateMessage requires content).
   let lastRowContent = "";
+  let lastRowMetadata: Record<string, unknown> = {};
   let pendingKnowledgeSources: unknown = null;
   // Same contract as sse-consumer: the citations appended to THIS assistant row
   // ride on its metadata for feedback attribution.
@@ -3109,18 +3120,65 @@ export async function collectChannelResponse(
   // Invocation-id queues pair parallel start↔end events. Older runtimes without
   // toolCallId fall back to per-name FIFO for backward compatibility.
   const toolInputs = new Map<string, string[]>();
-  const toolStarts = new Map<string, number[]>();
+  const toolStarts = new Map<string, Array<{ startedAt?: number; localStartMs: number }>>();
   const toolsets = new Map<string, Array<string | undefined>>();
+  const toolRounds = new Map<string, Array<Record<string, unknown>>>();
   const pushQ = <T,>(m: Map<string, T[]>, k: string, v: T): void => { const a = m.get(k) ?? []; a.push(v); m.set(k, a); };
   const shiftQ = <T,>(m: Map<string, T[]>, k: string): T | undefined => m.get(k)?.shift();
   const toolKey = (ev: Record<string, any>, name: string): string =>
     ev.toolCallId != null ? `id:${String(ev.toolCallId)}` : `name:${name}`;
+  // Shared with sse-consumer: value-based envelope dedup, current round, and
+  // one-clock tool duration rules must stay identical across persistence paths.
+  const timeline = new LlmCallTimeline();
+  // The ONLY bucket of envelopes awaiting a carrier. An agent call that produced
+  // output gets its own row inline, so it is never in here; a call that FAILED
+  // gets no row (isModelCallRowEnvelope excludes stop_reason "error"), so it waits
+  // here until exactly one of three things claims it: a switch notice
+  // (discarded_llm_calls), a recovery (call-only row), or a terminal failure (the
+  // error_response row). One carrier each, never two.
+  let pendingFailedLlmCalls: LlmCallEnvelope[] = [];
+  let pendingError: { content: string; envelope?: LlmCallEnvelope } | null = null;
   const persistRow = async (msg: Parameters<typeof appendMessage>[0]): Promise<string | null> => {
     try { return await appendMessage({ ...msg, traceId: persist?.traceId ?? msg.traceId }); }
     catch (err) {
       console.warn(`[${logPrefix}] audit persist failed session=${sessionId}:`, err);
       return null;
     }
+  };
+  const persistCallOnlyRows = async (envelopes: LlmCallEnvelope[]) => {
+    for (const envelope of envelopes) {
+      await persistRow({
+        sessionId,
+        role: "assistant",
+        content: "",
+        metadata: { llm_call: redactLlmCallEnvelope(envelope, redact) },
+      });
+    }
+  };
+  const flushRecoveredLlmCalls = async () => {
+    const failedCalls = pendingFailedLlmCalls;
+    pendingFailedLlmCalls = [];
+    pendingError = null;
+    await persistCallOnlyRows(failedCalls);
+  };
+  const flushTerminalError = async () => {
+    if (!pendingError) return;
+    const failedCalls = pendingFailedLlmCalls;
+    pendingFailedLlmCalls = [];
+    const terminal = pendingError;
+    pendingError = null;
+    await persistCallOnlyRows(terminal.envelope ? failedCalls.slice(0, -1) : failedCalls);
+    await persistRow({
+      sessionId,
+      role: "assistant",
+      content: terminal.content,
+      metadata: {
+        kind: "error_response",
+        error_code: "MODEL_ERROR",
+        retriable: true,
+        ...(terminal.envelope ? { llm_call: redactLlmCallEnvelope(terminal.envelope, redact) } : {}),
+      },
+    });
   };
 
   try {
@@ -3131,6 +3189,12 @@ export async function collectChannelResponse(
         pendingKnowledgeSources = null;
         renderedKnowledgeSourceUrls.clear();
         pendingRowCitations = [];
+        if (ev.type === "model_route_rollback") {
+          // The failed call is NOT dropped here: the switch notice that follows a
+          // rollback is its carrier. Only the user-visible error goes, because a
+          // rollback means the turn has not failed — a candidate has.
+          pendingError = null;
+        }
       }
       if (ev.type === "model_route_rollback" && lastAssistantMessageId && persist) {
         // Lark persists every assistant turn as it lands (sse-consumer buffers
@@ -3146,10 +3210,32 @@ export async function collectChannelResponse(
             messageId: discardedId,
             sessionId,
             content: redact(lastRowContent),
-            metadata: { discarded_route_attempt: true },
+            metadata: { ...lastRowMetadata, knowledge_citations: undefined, discarded_route_attempt: true },
           });
         } catch (err) {
           console.warn(`[${logPrefix}] discard rolled-back assistant row failed session=${sessionId}:`, err);
+        }
+      }
+      if (ev.type === "model_route_switch" && persist && redaction) {
+        const metadata = modelRouteSwitchMetadata(ev, redaction);
+        if (metadata) {
+          // ONLY the calls that have no row of their own. This path writes each
+          // successful call's row inline as it arrives and a rollback cannot
+          // unwrite them, so attaching the whole attempt here — as it used to —
+          // recorded those calls TWICE and made channel turns over-count model
+          // calls against identical web turns. The failed call is the one with no
+          // row, and this notice is its carrier.
+          const discarded = pendingFailedLlmCalls;
+          pendingFailedLlmCalls = [];
+          if (discarded.length > 0) {
+            metadata.discarded_llm_calls = discarded.map((call) => redactLlmCallEnvelope(call, redact));
+          }
+          await persistRow({
+            sessionId,
+            role: "assistant",
+            content: modelRouteNoticeContent(metadata),
+            metadata,
+          });
         }
       }
       if (ev.type === "knowledge_sources") {
@@ -3193,8 +3279,9 @@ export async function collectChannelResponse(
         const name = (ev.toolName as string) || (ev.name as string) || "tool";
         const key = toolKey(ev, name);
         pushQ(toolInputs, key, ev.args ? JSON.stringify(ev.args) : "");
-        pushQ(toolStarts, key, Date.now());
+        pushQ(toolStarts, key, { startedAt: eventClock(ev, "startedAt"), localStartMs: Date.now() });
         pushQ(toolsets, key, typeof ev.toolset === "string" && ev.toolset.length > 0 ? ev.toolset : undefined);
+        pushQ(toolRounds, key, timeline.toolMetadata(ev));
       }
 
       if (ev.type === "tool_execution_end" || ev.type === "tool_end") {
@@ -3209,9 +3296,14 @@ export async function collectChannelResponse(
           else if (ev.result?.details?.error) outcome = "error";
           const key = toolKey(ev, name);
           const input = shiftQ(toolInputs, key) || "";
-          const startedAt = shiftQ(toolStarts, key);
+          const start = shiftQ(toolStarts, key);
+          const endedAt = eventClock(ev, "endedAt");
+          const startedAt = start?.startedAt ?? start?.localStartMs;
           const toolset = shiftQ(toolsets, key) ??
             (typeof ev.toolset === "string" && ev.toolset.length > 0 ? ev.toolset : undefined);
+          const roundMeta = shiftQ(toolRounds, key) ?? timeline.toolMetadata(ev);
+          const metadata: Record<string, unknown> = { ...roundMeta };
+          if (startedAt != null) metadata.started_at = new Date(startedAt).toISOString();
           await persistRow({
             sessionId,
             role: "tool",
@@ -3220,7 +3312,8 @@ export async function collectChannelResponse(
             toolset: toolset ?? null,
             toolInput: input ? redact(input) : null,
             outcome,
-            durationMs: startedAt != null ? Date.now() - startedAt : null,
+            durationMs: start ? toolDurationMs(start, endedAt) : null,
+            metadata: Object.keys(metadata).length > 0 ? metadata : null,
           });
         }
       }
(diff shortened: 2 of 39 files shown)


src/gateway/channels/lark.ts
```diff
@@ -3193,10 +3275,29 @@ export async function collectChannelResponse(
...
         const blocks = Array.isArray(ev.message.content) ? ev.message.content : [];
         if (options.includeImages) collectImageAttachments(blocks, images, seenImageKeys);
+        const rowEnvelope = isModelCallRowEnvelope(envelope, ev.message?.stopReason)
+          ? redactLlmCallEnvelope(envelope, redact)
+          : undefined;
         let turnText = contentBlocksToMarkdown(blocks);
+        const producedOutput = ev.message?.stopReason !== "error" && ev.message?.stopReason !== "aborted" && (
+          turnText.trim().length > 0 || blocks.some((block: any) => block?.type === "toolCall")
+        );
+        if (pendingError && producedOutput) await flushRecoveredLlmCalls();
```
jacoblee-io: `flushRecoveredLlmCalls` is the only new write path NOT gated on `persist` (unlike ~3231/3285/3327/3353) — a non-audited binding (auditable=false, `options.persist` undefined) that hits an errored `message_end` followed by an in-turn recovery writes chat_messages rows for a session that has no chat_sessions row: FK violation on MySQL swallowed as a console.warn (so the write also silently never lands for deployments that later enable audit), or orphan rows where the FK isn't enforced. And since `redaction = persist ? … : null` (~3101), `redact` is the identity function on this path — the provider's error_message is written verbatim, bypassing the redaction contract every other write honors.

src/gateway/channels/lark.ts
```diff
@@ -3193,10 +3275,29 @@ export async function collectChannelResponse(
...
+        }
+      }
       if (ev.type === "message_end" && ev.message?.role === "assistant") {
         const blocks = Array.isArray(ev.message.content) ? ev.message.content : [];
         if (options.includeImages) collectImageAttachments(blocks, images, seenImageKeys);
+        const rowEnvelope = isModelCallRowEnvelope(envelope, ev.message?.stopReason)
+          ? redactLlmCallEnvelope(envelope, redact)
+          : undefined;
         let turnText = contentBlocksToMarkdown(blocks);
+        const producedOutput = ev.message?.stopReason !== "error" && ev.message?.stopReason !== "aborted" && (
```
jacoblee-io: This inline `producedOutput` re-derives sse-consumer's exported `messageProducedOutput` but omits the string-content case: the original (sse-consumer.ts:144) treats `typeof content === "string" && content.trim().length > 0` as output, while this copy reads only `Array.isArray(ev.message.content) ? blocks : []` — false for a string-content assistant message. A channel turn that errors then recovers with a string-content final message: the web path suppresses the error and commits the recovered call rows; this path keeps `pendingError` and `flushTerminalError` persists a terminal error_response row over a successful answer — the Feishu transcript reports failure for a turn that recovered. Use `messageProducedOutput(ev.message)`.

src/gateway/channels/lark.ts
```diff
@@ -3193,10 +3275,29 @@ export async function collectChannelResponse(
       }
       // pi-agent-brain emits the final assistant reply as message_end with
       // a content array of blocks; collect the text blocks only.
+      let envelope: LlmCallEnvelope | undefined;
+      if ((ev.type === "message_end" || ev.type === "turn_end") && ev.message?.role === "assistant") {
+        envelope = timeline.take(ev.message);
+        if (envelope?.kind === "agent") attemptLlmCalls.push(envelope);
```
jacoblee-io: Two divergences from sse-consumer's routing bookkeeping here: (a) this push lacks the `isRoutingTurn && !routingCommitted` gate (sse-consumer.ts:836), and (b) the channel path persists rows INLINE with no commit gating, yet `model_route_rollback` (~3166) clears only pending state and never the persisted rows — while the same envelopes then ride `metadata.discarded_llm_calls` on the switch notice, whose comment ("their rows were (or will be) discarded") is copied from sse-consumer and is false here. A discarded routing attempt's intermediate narration/tool calls are recorded TWICE for channel sessions — as persisted llm_call rows AND as discarded_llm_calls — so any timeline decoder following ADR-018's model double-counts the attempt, and channel turns over-count model calls relative to identical web turns.

src/gateway/channels/lark.ts
```diff
@@ -3193,10 +3275,29 @@ export async function collectChannelResponse(
...
+        if (ev.type === "message_end" && ev.message?.stopReason === "error") {
+          if (envelope) pendingFailedLlmCalls.push(envelope);
+          pendingError = {
```
jacoblee-io, line 3346: `pendingError` is set for ANY `stopReason === 'error'` message_end, coercing a missing `errorMessage` to `""` — and `flushTerminalError` (~3142) has no non-empty guard, so it persists a content-empty `kind:'error_response'` row. pi's synthetic error message_ends can genuinely lack errorMessage — sse-consumer.ts:846 guards on `message.errorMessage` truthy for exactly this reason (its own comment at ~855 says so). With no envelope attached, `isEmptyModelCallRow` doesn't hide the row → a blank assistant bubble in the channel transcript on reload, and readers scanning error_response content for failure text misreport the turn as empty rather than failed.

src/gateway/sse-consumer.ts
```diff
@@ -846,6 +825,19 @@ export async function consumeAgentSse(opts: ConsumeAgentSseOptions): Promise<Sse
...
         if (message?.role === "assistant") {
+          // ── LLM-call envelope: the timing source of truth ──
+          // Stamped by LlmCallRecorder at the streamFn boundary. message_end and
+          // turn_end arrive as separately parsed JSON values, so the shared
+          // timeline deduplicates by envelope values rather than object identity.
+          // Aux (compaction) calls ride the next agent call's `aux_calls`.
+          const rawEnvelope = llmCallFromMessage(message);
+          const envelope = timeline.take(message);
+          if (envelope) {
+            if (isRoutingTurn && !routingCommitted) attemptLlmCalls.push(envelope);
```
jacoblee-io, line 860: `attemptLlmCalls` is reset only at `model_route_success` — never at `model_route_start` or `model_route_exhausted` (`discardRoutedAttempt` at ~429 clears `pendingFailedLlmCalls` but not this array). One consume run carries several turns (steer — this file's own documented contract at ~407): turn 1's final routing attempt fails, its envelope is pushed here and already rides turn 1's error_response row; when steered turn 2's primary rolls back, `takeDiscardedLlmCalls()` drains turn 1's stale envelope into turn 2's model_route_switch notice — double-persisted, attributed to the wrong turn. lark.ts:3170 explicitly does `attemptLlmCalls = []` on model_route_start, confirming the missing reset on this side.

src/gateway/sse-consumer.ts
```diff
@@ -937,111 +945,71 @@ export async function consumeAgentSse(opts: ConsumeAgentSseOptions): Promise<Sse
...
-                // switch to.
-                dbMessageId = id;
+          // ── Persist: one model-call row per envelope, plus its thinking row ──
+          // A call that produced only tool calls still gets a row (content ""):
+          // it is the carrier of that round's timing, tokens and thinking. A
+          // failed call's envelope already rides the error row above, so here
+          // only its partial text (if any) is written, without the envelope.
+          // Rows without an envelope (runtime predating the recorder, or a
+          // duplicate turn_end delivery) keep the old rule: text or nothing.
+          const cleaned = assistantContent ? stripEmptyResponseMarkers(assistantContent) : "";
```
jacoblee-io, line 984: Deploy-skew blackout: the recorder lives in the AGENTBOX image, and `imagePullPolicy: Always` only pulls on pod CREATE — so after the runtime/portal rolls out, live pods keep old code. New gateway + old agentbox: message_end carries no envelope, this rewritten path writes no timing at all ("text or nothing"), and metrics-timing reads only `metadata.llm_call.ms` — `/metrics/timing` shows count 0 for every session on a not-yet-recycled pod, indistinguishable from no traffic, with no operator-visible indicator. ADR-018 concedes only HISTORICAL rows; per the repo's own convention for agentbox-spanning changes, the PR should state the deploy requirement explicitly: agentbox rebuild + `SICLAW_AGENTBOX_IMAGE` bump + pod recycle.

portal-web/src/hooks/usePilotChat.ts
```diff
@@ -556,6 +579,11 @@ export function toPilotMessage(m: ChatMessage): PilotMessage {
   // A background exec job's completion marker — folded into the launching tool's box
   // (annotateExecJobCompletions), never shown as its own row.
   const isExecJobEvent = metadata?.kind === "exec_job_event"
+  // A model call's reasoning text — its own row, folded into the timeline, never a bubble.
+  const isThinkingRow = metadata?.kind === "thinking"
+  // A model call that produced only tool calls still gets a row (it carries the
+  // call's timing/tokens in metadata.llm_call); there is nothing to render for it.
+  const isEmptyModelCallRow = m.role === "assistant" && !m.content?.trim() && metadata?.llm_call != null
```
jacoblee-io: The new hidden rows (thinking + empty model-call) are fetched in every history page and poller refetch, then hidden client-side — the server query is `SELECT * … LIMIT ?` with no kind exclusion (siclaw-api.ts:2281) and PAGE_SIZE=20 counts physical rows. A tool-heavy reasoning-model turn adds ~2 hidden rows per model call (the thinking row carrying the full joined reasoning text), so page 1 of a reopened session can render only a handful of visible bubbles — possibly zero assistant text — forcing repeated load-more, while every background-poller refetch re-downloads tens of KB of never-rendered reasoning. Pre-existing hidden kinds were sparse per-job markers; these scale with every model round. Cheaper: exclude these kinds from the default pagination query server-side.

jacoblee-io (commented):
Re-reviewed the reworked version (the PR 559 findings — identity dedup, delegate-api CoT leak, lark revocation — all look addressed, and the recorder's failed-call FIFO was verified safe against the vendored pi retry semantics). Still withholding approval: verification confirmed 9 correctness bugs, and the top cluster is the Lark channel path's hand-copied persistence state machine diverging from sse-consumer in four separately confirmed ways (persist gating, recovery predicate, rollback semantics, empty-error guard). Inline comments below; three findings can't be anchored because their lines aren't in this diff:

1. **`src/portal/siclaw-api.ts:2674` — the task-run trace endpoint renders chain-of-thought.** `GET .../runs/:runId/messages` selects chat_messages WITHOUT the metadata column and with no kind filter, and TraceView.tsx:83-91 renders every assistant row as a Markdown bubble. A scheduled task run on a reasoning model persists `kind:"thinking"` rows (full reasoning text as content) and content-empty model-call rows via the new consumer — the operator sees the model's complete reasoning duplicated ahead of each reply plus blank bubbles per tool-only call, with no field the frontend could even filter on. Violates ADR-018's own rule that every chat_messages reader must treat `kind:'thinking'` as hidden; delegate-api and usePilotChat were patched, this reader was not.

2. **`src/agentbox/session.ts:2184` — routed background/synthetic turns never wire the recorder's attempt protocol.** `handleRouteEvent` never touches `brain.llmCalls` (no `beginAttempt`/`rollbackAttempt`/explicit `beginPrompt`), so with `promptExplicit=false` each per-candidate `brain.prompt()` fires an implicit `beginPrompt` that fully RESETS the recorder (attempt=1, round=0, new prompt_received_at). The surviving fallback's envelopes claim attempt=1, `since_prev_ms` no longer spans the discarded attempt (violating `rollbackAttempt`'s own documented invariant), and trailing aux calls are dropped by the implicit endPrompt — the audit trail is wrong on exactly the failure turns it was built to explain.

3. **`src/gateway/sse-consumer.ts:1054` — stopped tool rows lose `llm_round`/`tool_call_id`.** The abort finalizer builds `stoppedMeta` as only `{status:'stopped', started_at}` without spreading `pendingCall.roundMeta`, and `updateMessage` (adapter.ts:2907) runs `SET metadata = ?` wholesale — exactly the replacement pitfall the struct's own field doc ("must be re-stamped at end") and the normal tool_end path guard against. After reload, an aborted turn's stopped tool row can no longer be attributed to its model call, breaking the PR's "a tool's group is known without consulting a clock" contract.

Also confirmed but below the cap: the metrics timing endpoint JSON.parses each row's metadata up to 4× (≈200k parses per request at the 50k row limit); call-only model-call rows carry no `metadata.kind`, so each reader re-invents shape detection; `docs/design/2026-08-02-error-surfacing-contract.md` and ADR-014's `ttft_ms` claim are now factually false and weren't updated.

src/gateway/channels/lark.ts
caihongwei2006: Fixed in `aba2646`. The write is now gated on `persist` together with every other write in that block:

```ts
if (persist && ev.type === "message_end" && ev.message?.stopReason === "error"
    && typeof ev.message.errorMessage === "string" && ev.message.errorMessage) {
```

Gating at the point the error is *queued* rather than at each flush closes both halves you identified at once: a non-audited binding now queues nothing, so `flushRecoveredLlmCalls` and `flushTerminalError` have nothing to write, and the identity-`redact` path is unreachable because `redaction == null` implies `persist` is falsy on the same line.

src/gateway/channels/lark.ts
caihongwei2006: Fixed in `aba2646` — the local re-derivation is gone and the shared rule is imported:

```ts
if (pendingError && messageProducedOutput(ev.message)) await flushRecoveredLlmCalls();
```

You were right that string content was the miss, and it was the one that mattered: a recovered turn whose final message carries a string got a terminal `error_response` written over a successful answer.

One further divergence went with it, in the other direction. The copy tested `contentBlocksToMarkdown(blocks).trim().length > 0`, i.e. whether the blocks *rendered* to anything, while the shared rule requires a `toolCall` block or a `text` block with non-empty text. Any block type that markdown-renders to a non-empty string but delivers no answer would have counted as recovery. Both paths now answer the question the same way, which is the actual point — this is a rule about the provider's verdict, and two copies of it drift.

Covered by `treats a string-content recovery as output and revokes the pending error`.

src/gateway/channels/lark.ts
caihongwei2006: Fixed in `aba2646`, but **not** by adding the `isRoutingTurn && !routingCommitted` gate — I think that gate would be the wrong fix on this path, and it is worth being explicit about why.

The double-count you describe is real, and its cause is the one you named: this path persists each call's row *inline* and a rollback cannot unwrite them. Given that, `attemptLlmCalls` had no reason to exist here at all — gating *when* it accumulates would still have it accumulate calls that already own a row, and `discarded_llm_calls` would still carry them. So the bucket is removed and the notice now carries only the calls that have **no** row of their own:

```ts
const discarded = pendingFailedLlmCalls;
pendingFailedLlmCalls = [];
```

`pendingFailedLlmCalls` holds exactly the errored calls — `isModelCallRowEnvelope` excludes `stop_reason: "error"`, so those never get an inline row. One carrier each, never two, and no commit-gating machinery on a path that does not commit-gate. The invariant is now stated at the declaration.

One thing your comment implies that I want to correct for the record: the `model_route_rollback` branch no longer clears the failed call. Dropping it there was the bug behind the first version of this fix — the switch notice that *follows* a rollback is that call's only carrier, so clearing on rollback deleted the row-less call outright. Only `pendingError` is cleared there, because a rollback means a candidate failed, not the turn. Covered by `moves a rolled-back attempt's failed call onto the fallback notice` and `puts only the row-less call on the notice when a rolled-back attempt also narrated`.

src/gateway/channels/lark.ts
caihongwei2006: Fixed in `aba2646` — same truthiness guard as `sse-consumer.ts:846`, and for the reason its comment gives:

```ts
&& typeof ev.message.errorMessage === "string" && ev.message.errorMessage
```

The coercion to `""` was the actual defect: it turned "no error text" into "an error with empty text", which is precisely the state a reader cannot tell from an empty turn. With the guard, a synthetic error `message_end` sets no `pendingError` at all, so no `error_response` row is written — the turn's failure is still visible through the envelope's `stop_reason`, which is where a decoder should be reading it from anyway. Covered by `does not persist an error row for a failure that carries no message`.

src/gateway/sse-consumer.ts
caihongwei2006: Fixed in `aba2646`. Reset added at `model_route_start`, and — beyond what you flagged — at `model_route_exhausted` as well:

```ts
if (eventType === "model_route_exhausted") {
  routingCommitted = true;
  attemptLlmCalls = [];
```

`exhausted` is terminal for the attempt sequence and the final attempt's calls ride the error row it commits, so leaving them in the bucket is the same cross-turn leak by a second route.

Your parenthetical about `discardRoutedAttempt` is the right call and I have kept it that way deliberately: a *rollback's* calls are exactly what the following switch notice is supposed to carry, so clearing there would delete the payload. The reset belongs at the sequence boundaries, not in the discard.

Verified by removing both resets and watching `does not leak a previous turn's envelope into a later turn's switch notice` fail — 2 envelopes on the second turn's notice instead of 1.

src/gateway/sse-consumer.ts
caihongwei2006: Agreed, and the PR body now states it as a deploy-order requirement under **Deploy requirement**: rebuild the agentbox image, bump `SICLAW_AGENTBOX_IMAGE`, recycle the running pods — in that order, with the reason spelled out (`imagePullPolicy: Always` pulls only on pod create, so a gateway-only rollout leaves live pods on the old runtime and their sessions reporting a count of 0, which reads the same as no traffic).

I did not add a runtime indicator for the blackout, and I would rather not in this PR: an "envelope missing" counter is a metric about a deploy state, and it would keep firing forever on historical rows, which are legitimately envelope-less. The window is bounded by a step in the rollout, so documenting the step is the proportionate fix. Happy to be argued out of that if you would rather see the gauge.

portal-web/src/hooks/usePilotChat.ts
caihongwei2006: Fixed in `aba2646`, server-side as you suggested — the same predicate on both the list and the `COUNT(*)`, so pagination stays consistent with what comes back:

```ts
const HIDDEN_ROW_PREDICATE =
  `(metadata LIKE '%"kind":"thinking"%'` +
  ` OR (role = 'assistant' AND (content IS NULL OR content = '')` +
  ` AND metadata LIKE '%"llm_call"%' AND metadata NOT LIKE '%"kind":%'))`;
```

Your point about scaling is the one that decided it: the pre-existing hidden kinds were sparse per-job markers, while these are ~2 rows per model round with the reasoning text on one of them, so a tool-heavy reopened session could spend a whole page budget on rows nobody renders.

Two properties of this fix worth stating rather than leaving to be discovered:

1. It is `LIKE` on a TEXT column, not JSON extraction — exact for rows this runtime writes with a compact `JSON.stringify`, and identical on MySQL and SQLite. It is deliberately a cheap predicate rather than a correct-for-all-inputs one; a hand-edited or differently-serialised `metadata` would slip through. For a transcript pagination filter that is the right trade, and the client-side hiding is still in place as the backstop, so a slip renders as before rather than as a leak.
2. The `NOT LIKE '%"kind":%'` clause on the carrier arm is load-bearing: an `error_response` or a route notice can also be content-empty, and those must stay visible. That is why the arm keys off the *absence* of a kind rather than off emptiness alone.

This does mean the reasoning rows are no longer delivered to the transcript at all, so a future collapsed-thinking view will need to opt back in (a query flag) rather than just stop hiding them. That is a deliberate default: not shipping tens of KB per poll to render nothing.

src/portal/siclaw-api.ts
```diff
@@ -2273,12 +2273,39 @@ export function registerSiclawRoutes(router: RestRouter, config: SiclawConfig, c
...
-      db.query("SELECT COUNT(*) AS count FROM chat_messages WHERE session_id = ?", [params.sid]),
+      db.query(
+        `SELECT COUNT(*) AS count FROM chat_messages WHERE session_id = ? AND NOT ${HIDDEN_ROW_PREDICATE}`,
+        [params.sid],
+      ),
       db.query(
         // Fetch newest N messages (DESC + LIMIT), then reverse in app to get chronological order.
         // This ensures page=1 returns the most recent messages (for initial load at bottom of chat).
-        "SELECT * FROM chat_messages WHERE session_id = ? ORDER BY (seq IS NULL) DESC, seq DESC, created_at DESC, id DESC LIMIT ? OFFSET ?",
+        `SELECT * FROM chat_messages WHERE session_id = ? AND NOT ${HIDDEN_ROW_PREDICATE}` +
```
jacoblee-io: **HIGH — `NOT (… LIKE …)` drops every row whose `metadata` is NULL, so all user prompts disappear from the transcript**

`HIDDEN_ROW_PREDICATE` is applied as `WHERE session_id = ? AND NOT (…)` on both queries below. That is not a boolean filter for a row with `metadata IS NULL` — it is `UNKNOWN`, and `UNKNOWN` does not satisfy a `WHERE`:

- `NULL LIKE '%"kind":"thinking"%'` → `UNKNOWN`
- second disjunct: `role = 'assistant'` is `FALSE` for a user row, so the whole `AND` chain short-circuits to `FALSE`
- `UNKNOWN OR FALSE` → `UNKNOWN`
- `NOT UNKNOWN` → `UNKNOWN` → row **not returned**

Every `chat_messages` row written without metadata is silently removed. The primary user prompt row is exactly such a row: `src/gateway/server.ts:843` calls `appendMessage({ sessionId, role: "user", content: text, deferSequence: true })` with no metadata, and `adapter.ts`'s `jsonParam(undefined)` returns `null`, so `metadata` is stored as SQL NULL.

Verified against real `node:sqlite` (MySQL's `LIKE`/`NOT` 3VL is identical):

```
rows: 1 user(NULL), 2 assistant(NULL), 3 assistant ''+llm_call, 4 assistant thinking, 5 assistant answer+llm_call
kept:  [ { id: '5' } ]
count: [ { c: 1 } ]
```

Rows 1 and 2 — the user's question and a plain assistant reply — are gone, and only the one row this runtime happened to write with an envelope survives. This hits the `COUNT(*)` on line 2301 as well as the list on line 2307, so `total` agrees with the truncated page and nothing looks anomalous to the caller.

Concretely: after this ships, reopening **any** session loses every user prompt, and loses **all** pre-rollout assistant rows too (they carry no metadata at all). No test catches it — `siclaw-api.misc.test.ts` mocks `db.query`, so the predicate's SQL semantics are never executed.

Suggested fix — make the predicate two-valued before negating it:

```sql
AND (metadata IS NULL OR NOT (…))
```

or `AND COALESCE(<predicate>, 0) = 0` (works on MySQL and SQLite). Worth a test that runs the real statement against `node:sqlite` with a NULL-metadata user row, since the mocked-query tests structurally cannot cover this.

src/gateway/llm-call-rows.ts
```diff
@@ -0,0 +1,151 @@
...
+ */
+export function buildThinkingRow(
+  message: unknown,
+  envelope: LlmCallEnvelope,
+  redact: (text: string) => string,
+): ThinkingRowSpec | null {
+  const blocks = thinkingBlocksFromMessage(message).filter((b) => b.text.length > 0 || b.redacted);
+  if (blocks.length === 0) return null;
+  return {
+    content: redact(blocks.map((b) => b.text).join("\n\n")),
```
jacoblee-io: **MEDIUM — full reasoning text goes into `content TEXT` (64 KB on MySQL) with no cap, and the write is not best-effort, so an over-long thinking chain loses the answer row and kills the turn**

This line joins **all** thinking blocks of one model call and stores the result in `chat_messages.content`. That column is declared `content TEXT` in `src/portal/migrate.ts:341` — note `tool_input` right below it is `MEDIUMTEXT`, `content` is not. `TEXT` is 65,535 bytes on MySQL, and an extended-thinking turn (32k+ reasoning tokens) exceeds that comfortably. With MySQL's default `sql_mode` including `STRICT_TRANS_TABLES` the INSERT **errors** rather than truncating.

The throw is not contained. Trace from here:

1. `buildThinkingRow` returns the oversized row;
2. `sse-consumer.ts` `persistAssistant` does `const thinkingRowId = await appendRow({ sessionId, role: "assistant", ...thinkingRow });` — unwrapped, and it is the **first** write in that closure, before the model-call row;
3. `appendRow` → `appendMessage` → `chat.appendMessage` RPC → the adapter's `INSERT` rejects, so the promise rejects;
4. `persistAssistant` rejects, and the call site is `await persistAssistant()` **inside** the `for await (const event of client.streamEvents(...))` loop;
5. that loop's `try` has only a `finally { await flushTerminalError(); }` — no `catch` — so the rejection propagates out of `consumeAgentSse`.

Result: the assistant row that carries the actual answer is never written (step 2 aborts before it), the remaining events of the turn are never relayed or persisted, and the caller sees a rejected `consumeAgentSse` — a turn reported as failed even though the provider answered successfully. Lark's `collectChannelResponse` has the same ordering (thinking row first, then the model-call row), though there `persistRow` swallows the error, so the channel path loses only the thinking row.

Suggested fix: cap the joined thinking text before storing it (the repo already has boundary-clipping `truncateAtBoundary` in `agentbox/delegation-summary.ts`) and/or widen `content` to `MEDIUMTEXT` in `migrate.ts` — remembering `safeAlterTable` only ADDs, so an existing MySQL table needs the idempotent `widenColumn` path. Independently, the thinking-row write should be best-effort relative to the model-call row: a failure to store reasoning must not be able to delete the answer.

src/portal/siclaw-api.ts
```diff
@@ -2273,12 +2273,39 @@ export function registerSiclawRoutes(router: RestRouter, config: SiclawConfig, c
...
+    // Fetching them anyway meant page 1 of a reopened tool-heavy session could
+    // spend its whole page budget on rows nobody sees — sometimes rendering no
+    // assistant text at all — and every background poll re-downloaded tens of KB
+    // of reasoning that is then thrown away.
+    //
+    // metadata is TEXT written with a compact JSON.stringify, so these LIKEs are
+    // exact for rows this runtime writes and behave the same on MySQL and SQLite.
+    // The carrier clause requires NO `kind`: an error_response or route notice can
+    // also be content-empty, and those must stay visible.
+    const HIDDEN_ROW_PREDICATE =
```
jacoblee-io: **MEDIUM — the scheduled-task run trace renders `kind: "thinking"` rows as ordinary assistant replies** *(the defect is at `src/portal/siclaw-api.ts:2700`, the `GET …/tasks/:taskId/runs/:runId/messages` query — that endpoint is untouched by this PR, so this comment is anchored to the sibling endpoint that DID get the new filter.)*

The user-facing chat endpoint here learned to exclude the two new row shapes. The scheduled-task run trace did not, and it is worse off than "unfiltered": its query is

```sql
SELECT id, role, content, tool_name, tool_input, outcome, duration_ms, created_at
FROM chat_messages WHERE session_id = ?
ORDER BY created_at DESC, id DESC LIMIT ?
```

— **no `metadata` column**. `portal-web/src/pages/TaskRunDetail.tsx:120` fetches that payload and renders `{role, content}` rows verbatim; it never goes through `toPilotMessage`, so the `isThinkingRow` / `isEmptyModelCallRow` hiding added in `usePilotChat.ts` does not apply, and it *cannot* be made to apply because the discriminator (`metadata.kind === "thinking"`) is not in the response.

Concretely, after this ships, opening a cron run's trace shows:
- the model's raw chain-of-thought as a normal assistant message, indistinguishable from the task's answer (this is the same reasoning text the PR body says is "hidden in transcripts");
- one blank assistant entry per tool-only model call;
- an earlier `truncated: true` cut, since `MAX_TRACE_MESSAGES` now fills 2–3× faster with rows nobody is meant to see.

Suggested fix: apply the same predicate to that query (once corrected per the NULL-metadata issue above), or select `metadata` and let the view filter — the former is cheaper and keeps `truncated` honest. The PR body's "both must be recognised before this ships, or a turn reads wrong" is exactly this endpoint's situation.

src/gateway/sse-consumer.ts
```diff
@@ -656,27 +680,26 @@ export async function consumeAgentSse(opts: ConsumeAgentSseOptions): Promise<Sse
...
-          preThinkingMs != null
-            ? { ...(detailsMeta ?? {}), pre_thinking_ms: preThinkingMs }
-            : detailsMeta;
+        // Re-stamp the round markers — extractPersistableDetails only looks at
+        // the tool *result*, and updateMessage REPLACES metadata wholesale.
+        const roundMeta = pendingCall?.roundMeta ?? timeline.toolMetadata(evt);
+        const merged: Record<string, unknown> = { ...(detailsMeta ?? {}), ...roundMeta };
+        // started_at survives onto the final row (it did not before: updateMessage
+        // replaces metadata) so the parallel view can place the bar without a join.
+        if (pendingCall) merged.started_at = new Date(pendingCall.startMs).toISOString();
```
jacoblee-io, line 711: **LOW — the abort finalizer rewrites a stopped tool row without `roundMeta`, erasing exactly the attribution this line re-stamps** *(the defect is at `src/gateway/sse-consumer.ts:1067`, the `stoppedMeta` object in the `persist && signal?.aborted` block; that hunk is not in the diff, so this comment is anchored to the re-stamp site it should mirror.)*

This line exists because `updateMessage` REPLACES metadata wholesale, so the tool-end handler has to re-send the round markers captured at `tool_execution_start` (`const roundMeta = pendingCall?.roundMeta ?? timeline.toolMetadata(evt)` a few lines above, plus `started_at` here).

The abort finalizer performs the same `updateMessage` on the very same rows, but builds its metadata from scratch:

```ts
const stoppedMeta: Record<string, unknown> = {
  status: "stopped",
  started_at: new Date(pendingCall.startMs).toISOString(),
};
```

`pendingCall.roundMeta` is in scope and is dropped. Concrete scenario: a user hits Stop while three parallel tools are running. Their `tool_execution_start` rows were written with `metadata.llm_round` + `metadata.tool_call_id`; the finalizer overwrites that metadata with `{status, started_at}`, so on reload those three tools have no `llm_round` and cannot be attributed to the model call that issued them — the interrupted turn is precisely the one an operator wants to read, and it is the only turn whose tool rows lose their call linkage.

Suggested fix: spread the captured markers in, same as the tool-end path:

```ts
const stoppedMeta: Record<string, unknown> = {
  ...pendingCall.roundMeta,
  status: "stopped",
  started_at: new Date(pendingCall.startMs).toISOString(),
};
```

src/gateway/sse-consumer.ts
```diff
@@ -846,6 +838,19 @@ export async function consumeAgentSse(opts: ConsumeAgentSseOptions): Promise<Sse
...
         if (message?.role === "assistant") {
+          // ── LLM-call envelope: the timing source of truth ──
+          // Stamped by LlmCallRecorder at the streamFn boundary. message_end and
+          // turn_end arrive as separately parsed JSON values, so the shared
+          // timeline deduplicates by envelope values rather than object identity.
+          // Aux (compaction) calls ride the next agent call's `aux_calls`.
+          const rawEnvelope = llmCallFromMessage(message);
+          const envelope = timeline.take(message);
+          if (envelope) {
+            if (isRoutingTurn && !routingCommitted) attemptLlmCalls.push(envelope);
```
jacoblee-io, line 860: **LOW — a *buffered* failed attempt's envelope never reaches the gateway, so "failed and rolled-back calls are kept" does not hold for background/synthetic turns**

This push can only ever see envelopes that arrived as SSE events, and a buffering candidate's events are discarded rather than relayed when it fails and is replaced. In `src/core/model-routing.ts`:

- `runAttempt`'s subscription (≈L1095) relays live only `if (streaming)`; otherwise it pushes into a local `events[]` and goes live only once an `isToolExecutionEvent` arrives.
- `flushBrainEvents(attemptResult.events, …)` is called on exactly two paths: success (≈L844) and `!willSwitch` / exhausted (≈L887).
- The `willSwitch` path — a failure that *does* have a next candidate — never flushes. It emits `model_route_rollback` only `if (attemptResult.emittedLive)` (≈L953), then `model_route_switch`.

So for a candidate that failed before its first tool call while buffering, the `message_end` carrying `message.llmCall` is dropped inside the runner. `timeline.take` never sees it, `attemptLlmCalls` stays empty, `takeDiscardedLlmCalls()` returns `undefined`, and the `model_route_notice` row is written with no `discarded_llm_calls`.

That is fine for an interactive turn (the primary streams live, so `emittedLive` is true and the envelope did arrive). But `optimisticPrimaryStream: false` for background/synthetic turns — every attempt buffers there — so on those turns **every** non-final failed attempt loses its timing entirely, which is the case the feature is advertised for ("Both are time the user waited for, and neither had a carrier before").

Suggested fix: either flush the failed attempt's `message_end` envelopes (or just their `llmCall` payloads) on the `willSwitch` path so the following switch notice can carry them, or scope the claim in the PR body and ADR-018 to attempts that streamed live, and say plainly that a buffered failed attempt's span shows up only as the survivor's `since_prev_ms`.

src/portal/siclaw-api.ts
caihongwei2006: Fixed in 4dc3d9fd. Chat list and COUNT now share transcriptVisiblePredicate, which uses the existing dialect-aware JSON extraction helper and COALESCE instead of negated LIKE. NULL/legacy metadata stays visible; formatted JSON and nested kind fields are handled correctly too.

Regression coverage executes the predicate and the actual chat endpoint SQL against node:sqlite with NULL-metadata user/assistant rows, hidden carriers, thinking, and explicit error/route notices. The endpoint test also verifies total and pagination after filtering.

src/gateway/llm-call-rows.ts
caihongwei2006: Fixed in 4dc3d9fd. buildThinkingRow now clips the redacted text at a UTF-8 boundary to 65,535 bytes including the truncation marker, and records metadata.truncated. This respects the existing MySQL TEXT column without depending on a schema rollout.

The SSE thinking insert is now best-effort: failures are logged, the answer insert still runs, and no thinking_row_id is attached unless that insert succeeded. Tests cover multi-byte oversized reasoning and an injected thinking-insert failure followed by a persisted answer and continued event consumption. No live MySQL instance was used; the byte limit is tested directly.

src/portal/siclaw-api.ts
caihongwei2006: Fixed in 4dc3d9fd. The task-run messages query now uses the same transcriptVisiblePredicate as chat list/count, before ORDER BY/LIMIT. It excludes thinking and empty kind-less model-call carriers while preserving NULL-metadata messages and explicit notices.

An endpoint regression runs the real SQL against node:sqlite with 205 thinking rows plus a legacy user prompt and answer. Both visible messages survive, thinking is absent, and truncated remains false.

src/gateway/sse-consumer.ts
caihongwei2006: Fixed in 4dc3d9fd. The abort finalizer now spreads pendingCall.roundMeta into stoppedMeta before replacing the row metadata. A regression stops three parallel tools and verifies that all three retain their llm_round and distinct tool_call_id after finalization.

src/gateway/sse-consumer.ts
caihongwei2006: Addressed in 4dc3d9fd. The routing runner now puts buffered message_end envelopes on model_route_switch.discardedLlmCalls. It does not replay discarded message text or error events. SSE and Lark consumers redact these envelopes and persist them on the existing notice as discarded_llm_calls; live attempts retain their existing carrier path. Tests exercise buffered fallback, SSE persistence/redaction, and both buffered/live Lark notices.

The synthetic notification path now explicitly brackets the whole routing run and applies attempt/rollback lifecycle events, fixing the separate recorder-reset finding from the review summary. Scope clarification: synthetic notifications still use their existing conditional report persistence, not the SSE/Lark full timeline persistence. Both ADR-018 and the PR description state this limitation explicitly; this change does not claim full synthetic-notification database telemetry.

caihongwei2006:
Addressed the remaining review findings in 4dc3d9fd, with replies on all five open inline threads.

The additional summary finding is fixed too: synthetic notification turns explicitly open/close the recorder around the entire routing run and apply attempt/rollback events. A regression verifies that implicit per-candidate prompt boundaries do not reset fallback attempt 2 to attempt 1.

For the lower-priority summary notes:
- Timing metrics decode each row metadata once and reuse it for all four values.
- ADR-014 and the error-surfacing contract now describe the current envelope/carrier behavior.
- Kept the existing kind-less empty-call shape rather than introducing another format change. Portal readers share one SQL predicate, the frontend now preserves explicit error/route kinds even when empty, and ADR-018 states the discriminator and precedence. Synthetic notification reports remain outside the full SSE/Lark timeline persistence contract.

Validation: backend suite 6,691 passed / 2 skipped, updated Lark suite 248 passed, frontend suite 249 passed, root TypeScript build and frontend production build passed. SQLite statements were executed; no live MySQL or production deployment was exercised.

jacoblee-io (commented):
Read the whole change. The core refactor is clean: measurement moved to the provider boundary, the inferred fields are removed outright rather than dual-written, `net_ttft + thinking + output == total` genuinely holds by construction, and the thinking-row clip is correct (I ran the byte math separately — stays within 65,535 including the marker and never splits a character). Test coverage is solid.

Findings below, worst first. Only the first one I'd consider merge-blocking.

## 1. The upstream-mode transcript readers were not updated

`transcriptVisiblePredicate` reached two call sites (`siclaw-api.ts:2291` pagination/count, `:2696` task-run traces). The two mirrors in `src/portal/adapter.ts` are still bare queries:

- `chat.getMessages` (RPC handler, `adapter.ts:2957`)
- `POST /api/internal/siclaw/chat/messages` (`adapter.ts:1885`)

Both are `WHERE session_id = ? ORDER BY created_at DESC, seq DESC, id DESC LIMIT ?` with no kind filter, and in upstream mode that is the path the chat transcript actually comes from. After this ships those readers get, per model round, one `kind: "thinking"` row rendered as an assistant reply (up to 64KB of reasoning text) plus one empty assistant carrier row for every tool-only call. A `limit: 50` page can now be almost entirely telemetry, so paging degrades as well.

ADR-018 states the requirement itself — *"Any reader of `chat_messages` must treat `kind: "thinking"` as a hidden assistant kind before this runtime ships"* — and these two readers are in this repo. `CLAUDE.md` also calls out that adapter's HTTP + WS mirrors have to move together.

If some consumers legitimately want the telemetry rows, an opt-in flag on those handlers is fine; silently returning them is not.

## 2. Row inflation vs. `message_count`

`persistCallOnlyRows` (`sse-consumer.ts:149`, `lark.ts:3131`) is the only write path that does not call `incrementMessageCount`; thinking rows, carrier rows and error rows all do. Two consequences:

1. A session's `message_count` (shown as "N messages" in the session list and audit views) now diverges from the number of rows the transcript renders by roughly 2–3×. The ADR acknowledges the daily `message_count` metric (`siclaw-api.ts:3632`) stepping on rollout, but not this one, and this is the number a user looks at directly.
2. Call-only rows being the single exception to the increment rule makes `message_count` mean neither "physical rows" nor "visible rows". Please make it consistent one way or the other.

## 3. `metadata` has no size bound while `content` now does

`content` got a 65,535-byte UTF-8-boundary clip. `metadata` is the same `TEXT` column and is unbounded: `llm_call.blocks` (two ISO timestamps per content block), `tool_call_ids`, nested `aux_calls`, and `discarded_llm_calls` (an array of envelopes) all grow without a cap.

Under MySQL strict mode an overflow is a thrown `Data too long for column 'metadata'`. The thinking-row insert is wrapped in try/catch (second commit), but the model-call row's `appendRow` is not, so the throw escapes the `for await` and takes the rest of the turn's SSE consumption with it. Realistically the envelope is a few KB, so this is unlikely rather than impossible — but having capped `content` it seems odd to leave the sibling column open. A cap on `blocks` length would cover the unbounded axis.

## 4. The chat bubble's "model time" changed meaning, and the PR description doesn't say so

`combinedModelMs` is now a single call's `total`. On a multi-round tool-using turn the badge shows only the last call — the one that produced the final text — so the number drops sharply, and **nothing in the UI shows the turn's wall clock any more**. `since_prev_ms` and the per-round envelopes are persisted but nothing renders them. The description covers the dashboard split; this user-visible change deserves a line too.

## 5. The live envelope is not redacted

`sse-consumer.ts:850` attaches the original envelope to the event, so `error_message` (up to 500 chars of provider error text, which can echo a key) reaches the browser verbatim; only the persisted copy goes through `redactLlmCallEnvelope`. `stream_error` has the same pre-existing gap, so this isn't a regression — but it is a second channel, and the `redactionConfig` is right there. Attaching a redacted clone costs nothing.

## 6. Smaller things

- **Commit message contradicts the code.** It says *"prevResponseEndAt is not advanced, so the next call's `since_prev_ms` spans it and it is counted exactly once"*. `sealFailedCall` calls `sealCall`, whose agent branch unconditionally does `this.round += 1` and `this.prevResponseEndAt = responseEndAt`. So a failed call that never finds a carrier has its span dropped rather than reattributed, and leaves a hole in the round numbering. The ADR's "round numbering has no holes" only holds for the retries that get replayed as empty rows.
- **Wall clock for durations.** The partition identity is stated as holding by construction, but every interval comes from `Date.now()`. An NTP step backwards makes a `Math.max(0, …)` clamp break the identity. These are all same-process durations — `performance.now()` for the deltas, with the ISO timestamps kept as-is, would make the guarantee real.
- **`transcriptVisiblePredicate` de-sargables pagination and count.** `idx_chat_messages_session` used to let `COUNT(*)` stay index-only; now every candidate row reads the `metadata` TEXT and runs `JSON_VALID` + `JSON_EXTRACT`. Worth measuring on a large session before rollout.
- `LlmCallRecorder.attemptStarted` is declared in the middle of the class, after the methods that use it. Works, but reads badly.
- `TimingStatsCard`'s `entry` prop is kept but unused ("call-site compatibility"). Dropping it and updating the call site is cleaner than keeping a dead prop.

caihongwei2006:
Thanks for separating the merge blocker from the other findings. This revision fixes #1; the remaining points are clarified below without expanding the runtime refactor.

1. **Adapter readers:** fixed in 689d7e9b. Both `POST /api/internal/siclaw/chat/messages` and `chat.getMessages` now use the same NULL-safe `transcriptVisiblePredicate` before LIMIT. Tests execute both handlers' SQL against SQLite with 60 hidden rows, legacy user/assistant rows, an explicit empty error notice, another session, and a before cursor. All three visible rows survive a limit of 3. No telemetry opt-in was added: the in-repo caller uses this as transcript/recovery data, while timing metrics query their own rows. These are Portal's upstream-compatible mirrors; this does not change an external upstream service's implementation.

2. **Counts:** the claimed inconsistency is not present. `src/gateway/chat-repo.ts:400` implements `incrementMessageCount()` as a no-op. Both adapter append handlers increment `chat_sessions.message_count` after INSERT (HTTP append-message and RPC `chat.appendMessage`), including calls from `persistCallOnlyRows`. Adding another increment at those callers would change nothing. The count is physical rows, not visible bubbles; ADR-018 explicitly names both `chat_sessions.message_count` and the generic Portal metric. The divergence from visible bubbles is real and intentional. Changing that established count contract is outside this timing PR.

3. **Metadata size:** accepted as a possible limit, but not a demonstrated P1 workload. Reasoning text alone can exceed TEXT in an ordinary extended-thinking response; a call envelope has no reasoning body and grows with block/tool/attempt structure instead. Arbitrarily cutting `blocks` would discard measurement detail and would not bound `aux_calls` or `discarded_llm_calls`, so it is not a complete fix. Keeping the current envelope contract here; a supported metadata budget/schema decision should address the entire payload if actual workloads require it. This is not a claim that overflow is impossible.

4. **Badge meaning:** agreed; added an explicit line to the PR description. The badge is per model call, so the final answer's badge covers only its final call. It no longer claims cumulative turn time, and there is currently no full-turn wall-clock UI. This is the intended move from inferred turn figures to measured call figures; a turn timeline would be additional UI scope.

5. **Live redaction:** agreed that live error text is not uniformly redacted. Redacting only `evt.llmCall` would not close that boundary: the event also retains `message.llmCall` and `message.errorMessage`, and `stream_error` carries the error too. Persisted envelopes already use the configured redaction path. A consistent live-event redaction contract should cover all those fields together; leaving that pre-existing live-stream behavior outside this P1 fix rather than treating one cloned alias as sufficient.

6. **Smaller points:**
   - The earlier commit prose is inaccurate: `sealFailedCall` does advance the recorder through `sealCall`. Failed time is represented by its retained carrier, not reattributed into `since_prev_ms`. If a carrier is absent, that is missing telemetry; there is no reconstruction guarantee. No history rewrite or new inference logic proposed.
   - Same-process clocks avoid cross-pod subtraction, but do not make `Date.now()` monotonic. The partition statement assumes no wall-clock step during a request. A backward step is a valid limitation; changing the recorder's clock model is separate from the transcript blocker.
   - The filter loses index-only counting, but does not remove the indexed `session_id` restriction. Reading metadata for candidate rows is the cost of filtering before pagination; filtering after LIMIT reintroduces the demonstrated empty-page problem. No large-session benchmark is claimed, and adding stored flags/indexes without workload evidence would be another schema change.
   - Field placement and the unused compatibility prop are cleanup, not behavioral blockers. Leaving them out of this revision.

Validation: 280 targeted tests passed and the TypeScript build passed. Full-suite result: 6,694 passed / 2 skipped.

jacoblee-io (approved):
Approving. The blocking finding is closed and the rebase holds up under review.

**What I re-verified on `fff04a60`:**

- `689d7e9b` applies `transcriptVisiblePredicate` before `LIMIT` in both adapter mirrors, which was the one merge-blocker. The test covers HTTP and RPC separately, keeps NULL-metadata legacy rows visible, keeps a content-empty `error_response` visible, and pushes 60 hidden rows through to prove they can no longer crowd out a page.
- The conflict resolution in `sse-consumer.ts` is not mechanical: main's per-row knowledge-citation attribution is correctly folded into the rewritten `assistantRowMetadata`. I checked the one hazard I could see — `pendingRowCitations = []` now also runs on the empty-carrier branch — and it is safe, because `pendingRowCitations` is only ever set inside the `base.trim()` path, which sets `assistantContent` non-empty in the same iteration. Set and consumed in one event; nothing can leak onto a hidden row.
- `fff04a60` is a real fix, not conflict noise: main's rollback wrote `{...lastRowMetadata, knowledge_citations: undefined, discarded_route_attempt: true}`, and `lastRowMetadata` now carries `llm_call`. Destructuring only the citations out keeps the envelope on the discarded row, which stays consistent with the Lark switch notice only carrying calls that have no row of their own — no double count.
- CI green across all four jobs.

**Left open deliberately** (my earlier comment, items 2–6 — none blocking, but they shouldn't be lost):

1. `persistCallOnlyRows` is still the only write path that skips `incrementMessageCount`. Please settle this one way or the other before or shortly after merge: as it stands `message_count` means neither "physical rows" nor "visible rows", and if the asymmetry is intentional it deserves a line in ADR-018 rather than being inferred from the code.
2. `metadata` has no size bound while `content` now does, and the model-call `appendRow` has no try/catch (the thinking row does).
3. The chat bubble's "model time" now shows one call's total, and nothing in the UI shows a turn's wall clock any more — worth a line in the description.
4. The live event carries the unredacted envelope; only the persisted copy goes through `redactLlmCallEnvelope`.
5. Smaller: the commit message's "prevResponseEndAt is not advanced" claim contradicts `sealCall`; wall-clock deltas make the partition identity breakable under an NTP step; `transcriptVisiblePredicate` de-sargables the chat count/pagination and is worth measuring on a large session; `TimingStatsCard`'s `entry` prop is dead.

One observation from the rebase, pre-existing rather than introduced here: a rolled-back *successful* call is represented differently by the two paths — SSE writes no row and nests the envelope in `model_route_notice.discarded_llm_calls`, Lark writes the row and marks it `discarded_route_attempt: true` with `llm_call` retained. The metrics timing query reads rows and not `discarded_llm_calls`, so a channel session contributes one extra sample where an identical web session contributes none. Worth a follow-up issue.

**Deploy**: per the description this needs an agentbox rebuild + `SICLAW_AGENTBOX_IMAGE` bump + pod recycle. `imagePullPolicy: Always` only pulls on pod create, so verifying against a live session proves nothing.

PR #563, opened 2026-09-04
fix(channels): keep visual output supplemental

LikiosSedo:
## Summary

- remove the bundled `render_visual_card` authoring path so final conclusions remain natural-language answers
- keep chart and Mermaid rendering as optional images for IM while retaining source blocks required by the existing web renderer; renderer metadata is no longer returned
- require an explicit visual export URL and forward that narrow runtime contract into AgentBox MCP processes
- suppress legacy visual-card source blocks in IM delivery while leaving existing web readers untouched

## Deployment note

Deployments that use the bundled chart or Mermaid renderer must set `runtime.visualExport.url` to a reachable ControlPlane Web `/siclaw-visual-export` endpoint. The chart no longer guesses an internal Service DNS name.

## Verification

- `npm test -- --run` (302 files, 6719 passed, 2 skipped)
- focused Lark, visual-image, and MCP-client tests (281 passed)
- `npm run build`
- `npm --prefix mcp/create-chart run build`
- `helm lint helm/siclaw`
- default and configured `helm template` renders
- `git diff --check`

### DIFF
--- a/mcp/create-chart/src/handler.ts
+++ b/mcp/create-chart/src/handler.ts
@@ -1,8 +1,5 @@
 import { exportMarkdownVisualsWithVisualExportWeb } from "./visual-export.js";
-import type { RenderChartArgs, RenderChartResult, RenderChartToolResponse } from "./types.js";
-
-const CHART_SPEC_VERSION = 1;
-const VISUAL_SPEC_VERSION = 1;
+import type { RenderChartArgs, RenderChartToolResponse } from "./types.js";
 
 export const RENDER_CHART_INPUT_SCHEMA = {
   type: "object",
@@ -31,9 +28,9 @@ export const RENDER_CHART_INPUT_SCHEMA = {
 export const RENDER_CHART_DESCRIPTION =
   [
     "Render a pie/bar/line chart only when finalized structured numeric data is already in context and can be passed as valid tool arguments. This includes requests such as 画图, 画饼图, 柱状图, 趋势图 when the required numeric data is available.",
-    "For qualitative diagrams, workflows, topology, or decision trees, use a ```mermaid fenced block instead; xychart-beta is suitable for simple bar charts.",
+    "For qualitative diagrams, workflows, topology, or decision trees, use render_mermaid instead; xychart-beta is suitable for simple bar charts.",
     "Arguments must be one JSON object. data must be an object, never a JSON string. Use only literal finite numbers; never use placeholders, expressions, previous-message references, or bare tokens.",
-    "The tool renders through ControlPlane Web's own chart renderer/export path and returns a READY_TO_PASTE chart block as plain markdown, metadata, and a PNG image artifact. In your final reply, paste the READY_TO_PASTE block exactly as returned and preserve the image artifact. Do not rewrite, escape, quote, or wrap the chart JSON; the frontend renders ```chart fenced JSON blocks as SVG, while IM channels forward the PNG artifact.",
+    "After a successful render, the tool returns a web-renderable ```chart block plus a PNG image artifact. In web replies, include the returned chart block so the UI can render it. In IM channel sessions, preserve the image artifact and follow the channel instructions not to paste source. On every surface, make the natural-language answer complete on its own and never expose renderer metadata.",
   ].join(" ");
 
 export const RENDER_MERMAID_INPUT_SCHEMA = {
@@ -54,64 +51,13 @@ export const RENDER_MERMAID_INPUT_SCHEMA = {
 } as const;
 
 export const RENDER_MERMAID_DESCRIPTION = [
-  "Render a Mermaid diagram through ControlPlane Web's own Mermaid renderer/export path and return a PNG image artifact.",
+  "Render a Mermaid diagram through ControlPlane Web's own Mermaid renderer/export path and return an image/png artifact.",
   "Use this in Feishu/Lark channel replies whenever the user asks for a flowchart, sequence diagram, timeline, topology, remediation flow, or Mermaid diagram image.",
-  "Arguments must contain Mermaid source only, not fenced markdown. The tool returns READY_TO_PASTE ```mermaid markdown plus an image/png content block. Paste READY_TO_PASTE exactly and preserve the image artifact.",
+  "Arguments must contain Mermaid source only, not fenced markdown. After a successful render, the tool returns a web-renderable ```mermaid block plus an image/png artifact. In web replies, include the returned Mermaid block so the UI can render it. In IM channel sessions, preserve the image artifact and follow the channel instructions not to paste source. On every surface, make the natural-language answer complete on its own and never expose renderer metadata.",
 ].join(" ");
-
-export const RENDER_VISUAL_CARD_INPUT_SCHEMA = {
-  type: "object",
-  required: ["type", "title"],
-  properties: {
-    type: {
-      type: "string",
-      enum: [
-        "report",
-        "final_report",
-        "health_check",
-        "incident_timeline",
-        "root_cause_chain",
-        "metric_snapshot",
-        "status_distribution",
-        "action_plan",
-      ],
-      description: "ControlPlane Web visual-card type.",
-    },
-    title: { type: "string" },
-    label: { type: "string" },
-    subtitle: { type: "string" },
-    conclusion: { type: "string" },
-    summary: { type: "string", description: "Alias accepted by ControlPlane Web as conclusion." },
-    tone: { type: "string" },
-    status: { type: "string" },
-    footer: { type: "string" },
-    items: { type: "array" },
-    events: { type: "array" },
-    nodes: { type: "array" },
-    root_cause: { type: "string" },
-    rootCause: { type: "string" },
-    metrics: { type: "array" },
-    segments: { type: "array" },
-    total: { type: "number" },
-    actions: { type: "array" },
-    sections: { type: "array" },
-  },
-  additionalProperties: false,
-} as const;
-
-export const RENDER_VISUAL_CARD_DESCRIPTION = [
-  "Render a ControlPlane Web visual-card conclusion card through ControlPlane Web's own visual-card renderer/export path and return a PNG image artifact.",
-  "Use this for Feishu/Lark final diagnosis cards, health checks, root-cause summaries, incident timelines, metric snapshots, status distributions, and action plans when the group needs a conclusion-card image.",
-  "Arguments must be one visual-card JSON object, not Markdown and not a JSON string. The tool returns READY_TO_PASTE ```visual-card markdown plus an image/png content block. Paste READY_TO_PASTE exactly and preserve the image artifact.",
-].join(" ");
-
-function newChartId(type: string): string {
-  return `${type}-${Date.now()}-${Math.random().toString(36).slice(2, 8)}`;
-}
 
 export async function handleRenderChart(rawArgs: unknown): Promise<RenderChartToolResponse> {
   const args = validate(rawArgs);
-  const id = newChartId(args.type);
 
   const spec = JSON.stringify(args);
   const markdownEmbed = "```chart\n" + spec + "\n```";
@@ -120,33 +66,11 @@ export async function handleRenderChart(rawArgs: unknown): Promise<RenderChartTo
   if (!visual?.image) throw new Error("render_chart: ControlPlane Web export returned no chart image");
   const png = visual.image;
 
-  const result: RenderChartResult = {
-    schema_version: CHART_SPEC_VERSION,
-    chart_id: id,
-    type: args.type,
-    artifact_kind: "chart_spec",
-    spec_path: "",
-    svg_path: "",
-    png_path: "",
-    bytes: Buffer.byteLength(spec, "utf8"),
-    image_bytes: png.byteLength,
-    image_mime: "image/png",
-    renderer: "visual-export-web",
-    embed_instructions:
-      "Paste the READY_TO_PASTE block above verbatim into your reply where the chart should appear, and preserve the returned image artifact. Do not modify the JSON, add backslashes, escape non-ASCII characters, convert to ```svg, or inline an <img>.",
-  };
-
   return {
     content: [
       {
         type: "text",
-        text: [
-          "READY_TO_PASTE:",
-          markdownEmbed,
-          "",
-          "METADATA_JSON:",
-          JSON.stringify(result, null, 2),
-        ].join("\n"),
+        text: markdownEmbed,
       },
       {
         type: "image",
@@ -159,25 +83,16 @@ export async function handleRenderChart(rawArgs: unknown): Promise<RenderChartTo
 
 export async function handleRenderMermaid(rawArgs: unknown): Promise<RenderChartToolResponse> {
   const args = validateMermaid(rawArgs);
-  const id = newChartId("mermaid");
   const markdownEmbed = "```mermaid\n" + args.source + "\n```";
   const exported = await exportMarkdownVisualsWithVisualExportWeb(markdownEmbed);
   const visual = exported.find((item) => item.kind === "mermaid") ?? exported[0];
   if (!visual?.image) throw new Error("render_mermaid: ControlPlane Web export returned no Mermaid image");
 
-  const meta = visualMetadata(id, "mermaid", markdownEmbed, visual.image);
-
   return {
     content: [
       {
         type: "text",
-        text: [
-          "READY_TO_PASTE:",
-          markdownEmbed,
-          "",
-          "METADATA_JSON:",
-          JSON.stringify(meta, null, 2),
-        ].join("\n"),
+        text: markdownEmbed,
       },
       {
         type: "image",
@@ -185,60 +100,6 @@ export async function handleRenderMermaid(rawArgs: unknown): Promise<RenderChart
         data: visual.image.toString("base64"),
       },
     ],
-  };
-}
-
-export async function handleRenderVisualCard(rawArgs: unknown): Promise<RenderChartToolResponse> {
-  const spec = validateVisualCard(rawArgs);
-  const id = newChartId("visual-card");
-  const specJson = JSON.stringify(spec);
-  const markdownEmbed = "```visual-card\n" + specJson + "\n```";
-  const exported = await exportMarkdownVisualsWithVisualExportWeb(markdownEmbed);
-  const visual = exported.find((item) => item.kind === "visual-card") ?? exported[0];
-  if (!visual?.image) throw new Error("render_visual_card: ControlPlane Web export returned no visual-card image");
-
-  const meta = visualMetadata(id, "visual-card", markdownEmbed, visual.image);
-
-  return {
-    content: [
-      {
-        type: "text",
-        text: [
-          "READY_TO_PASTE:",
-          markdownEmbed,
-          "",
-          "METADATA_JSON:",
-          JSON.stringify(meta, null, 2),
-        ].join("\n"),
-      },
-      {
-        type: "image",
-        mimeType: "image/png",
-        data: visual.image.toString("base64"),
-      },
-    ],
-  };
-}
-
-function visualMetadata(
-  id: string,
-  kind: "mermaid" | "visual-card",
-  markdownEmbed: string,
-  image: Buffer,
-): Record<string, unknown> {
-  return {
-    schema_version: VISUAL_SPEC_VERSION,
-    visual_id: id,
-    type: kind,
-    artifact_kind: `${kind}_spec`,
-    spec_path: "",
-    png_path: "",
-    bytes: Buffer.byteLength(markdownEmbed, "utf8"),
-    image_bytes: image.byteLength,
-    image_mime: "image/png",
-    renderer: "visual-export-web",
-    embed_instructions:
-      "Paste the READY_TO_PASTE block above verbatim into your reply and preserve the returned image artifact. Do not expose escaped JSON or describe Feishu upload mechanics.",
   };
 }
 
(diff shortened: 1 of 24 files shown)


mcp/create-chart/src/handler.ts
```diff
@@ -120,33 +66,11 @@ export async function handleRenderChart(rawArgs: unknown): Promise<RenderChartTo
...
       {
         type: "text",
-        text: [
-          "READY_TO_PASTE:",
-          markdownEmbed,
-          "",
-          "METADATA_JSON:",
-          JSON.stringify(result, null, 2),
-        ].join("\n"),
+        text: "Chart image rendered. Keep it as optional supporting material and answer the user in natural language; do not paste chart source or renderer metadata.",
```
jacoblee-io: **medium** — This removes the only path by which a chart reaches the Portal web UI.

`portal-web/src/components/chat/Markdown.tsx` (`hasLanguageClass(className, "chart")` / `"mermaid"`) renders ```chart / ```mermaid fences through `ChartRenderer`. That fence used to arrive because the tool returned a `READY_TO_PASTE` block and the description told the model to paste it verbatim. Now the description says *"do not paste chart source, renderer metadata, or tool output into the final text"*, and the PNG lives only in the MCP tool result — which `portal-web/src/components/chat/PilotArea.tsx` renders as a plain `<pre>` of `message.content` (text parts only; image content blocks are never displayed there).

Concrete scenario: a Portal **web** user asks for a pie chart. The model calls `render_chart`, a PNG is produced, and the user sees prose with no chart at all. The new IM-only guidance in `CHANNEL_SECTION` does not apply on the web path, so nothing offsets it — the tool description is global.

The PR summary states existing web readers are left untouched; as written they lose the feature. Either keep a web-visible artifact path (render tool-result images in `PilotArea`) or scope the "never paste source" instruction to channel sessions.

src/gateway/channels/lark.ts
```diff
@@ -2072,7 +2072,9 @@ async function processQueuedLarkMessage(ctx: QueuedLarkMessageContext): Promise<
       : (resultText || EMPTY_RESULT_NOTICE_BY_LOCALE[locale]);
   if (agentError || sessionBusy) replyImages = [];
   const displayBody = stripVisualBlocks(finalBody, { stripSourceBlocks: replyImages.length > 0 })
```
jacoblee-io: **low** — This new fallback makes 👍/👎 appear under an empty-result notice.

`stripVisualBlocks` now strips `visual-card` / `siclaw-card` / `conclusion-card` unconditionally, so an answer whose text was only such a legacy block collapses to `""` and `displayBody` becomes `EMPTY_RESULT_NOTICE_BY_LOCALE[locale]`. But `isAnswer` a few lines below (`const isAnswer = !agentError && resultText.trim().length > 0`) is still **true** in exactly that case, because `resultText` is non-empty before stripping.

Result: `finalizeCard` attaches the feedback buttons under the empty-result notice, contradicting the invariant stated in the comment right above `isAnswer` ("never under an error or empty-result notice") and writing ratings against a non-answer into the Metrics feedback signal. Judge `isAnswer` from `displayBody` (or from the stripped text), not from raw `resultText`.

src/gateway/channels/visual-image.ts
```diff
@@ -35,12 +35,15 @@ export function collectImageAttachments(
 
 export function stripVisualBlocks(markdown: string, options: StripVisualBlocksOptions = {}): string {
   let output = stripDataImages(markdown);
+  // These legacy report containers are no longer authored by Siclaw. Existing
+  // web readers may remain tolerant, but never expose their JSON source in an
+  // IM channel when an older session or skill still produces one.
+  output = stripFences(output, "visual-card");
```
jacoblee-io, line 41: **low** — Making these unconditional silently swallows one delivery path.

`src/gateway/channels/lark.ts:1926` calls `stripVisualBlocks(backgroundMessage.text)` with no options, so it now also strips these three fences. If a background message of `kind: "final"` had a body consisting only of a legacy `visual-card` block, `display` becomes empty and the very next line does `return true` — reporting the message as **delivered** while nothing is sent to the group.

The sibling branch at `lark.ts:1948` handles the same collapse with `|| EMPTY_RESULT_NOTICE_BY_LOCALE[locale]`; the text branch has no such fallback. Suppressing the JSON is right, but the turn should still surface *something* rather than being reported as delivered.

mcp/create-chart/src/visual-export.ts
```diff
@@ -14,30 +15,30 @@ interface VisualExportOptions {
...
 ): Promise<ExportedVisual[]> {
-  const baseUrl =
-    options.baseUrl ??
-    process.env.SICLAW_VISUAL_EXPORT_URL ??
-    DEFAULT_EXPORT_URL;
-  const timeoutMs = options.timeoutMs ?? Number(process.env.SICLAW_VISUAL_EXPORT_TIMEOUT_MS ?? 30_000);
-  const browser = await launchBrowser();
+  const baseUrl = resolveVisualExportUrl(options.baseUrl);
+  const timeoutMs = resolveVisualExportTimeoutMs(options.timeoutMs);
+  const deadline = Date.now() + timeoutMs;
```
jacoblee-io, line 24: **low** — The shared deadline now has to cover cold Chromium startup, shrinking the effective render budget.

Before this change `chromium.launch` used Playwright's own default and `goto` / `waitForFunction` each got a full `timeoutMs`. Now one `deadline` spans launch + `newPage` + `goto(networkidle)` + `waitForFunction` + `evaluate`, and the client side caps the whole call at `timeoutMs + 5s` (`src/core/mcp-client.ts`, `VISUAL_EXPORT_MCP_GRACE_MS`).

Concrete scenario: on a CPU-throttled AgentBox pod a cold `chromium.launch` takes 8-10s and the ControlPlane SPA needs another 10-15s to reach `networkidle`, leaving `waitForFunction` a couple of seconds — renders that previously succeeded now fail with `Visual export timed out after 30000ms`. The `values.yaml` comment calls 30000 the "renderer budget", which understates what it now covers. Consider either raising the default or excluding browser launch from the budget.

mcp/create-chart/src/visual-export-config.ts
```diff
@@ -0,0 +1,48 @@
...
+      "Visual export URL must not contain a fragment (#); configure the page URL before the hash payload",
+    );
+  }
+
+  const pathname = url.pathname.replace(/\/+$/, "");
+  return `${url.origin}${pathname}${url.search}`;
+}
+
+export function resolveVisualExportTimeoutMs(override?: number): number {
+  const raw = override ?? Number(
```
jacoblee-io: **low** — An empty-string timeout is treated as `0` and hard-fails, unlike the URL which treats empty as unset.

`??` does not catch `""`, so `Number("")` is `0` and this throws `"Visual export timeout must be a positive number of milliseconds"`. `resolveVisualExportUrl` above deliberately does `?.trim()` + `if (!configured)` for exactly this case.

Concrete scenario: a Portal MCP server config carrying `env: { SICLAW_VISUAL_EXPORT_TIMEOUT_MS: "" }` survives `mergeMcpStdioEnv` (configured env is spread last and is not filtered for empty values, unlike the inherited names), reaches the child process, and every render throws immediately — while `visualMcpRequestTimeoutMs` reads the same `""`, falls back to the default, and makes the client wait the full 35s. Normalise empty to unset here the same way the URL does.

src/core/mcp-client.ts
```diff
@@ -49,6 +49,50 @@ export interface McpServersConfig {
...
+  "SICLAW_VISUAL_EXPORT_URL",
+  "SICLAW_VISUAL_EXPORT_TIMEOUT_MS",
+  "SICLAW_VISUAL_EXPORT_THEME",
+  "SICLAW_VISUAL_EXPORT_CHROMIUM",
+] as const;
+const DEFAULT_VISUAL_EXPORT_TIMEOUT_MS = 30_000;
+const VISUAL_EXPORT_MCP_GRACE_MS = 5_000;
+
+/**
+ * Forward only the platform-owned renderer contract into bundled stdio MCPs.
```
jacoblee-io: **low** — Comment and behaviour disagree: this is forwarded to *every* stdio MCP, not just bundled ones.

`mergeMcpStdioEnv(cfg.env)` is applied to every server whose detected transport is `stdio`, including arbitrary operator- or third-party-configured MCP commands. Those subprocesses receive `SICLAW_VISUAL_EXPORT_URL` (an internal ControlPlane console endpoint) and `SICLAW_VISUAL_EXPORT_CHROMIUM` with no opt-out and no way for an operator to see it happened.

Either gate the forwarding on the server identity (name/command), or reword the comment to say the allowlist is process-wide so the next reader does not assume a scope the code does not enforce.

mcp/create-chart/src/handler.ts
LikiosSedo: Fixed in `817726cc`. The successful tool result again contains the fenced `chart`/`mermaid` block required by the existing web renderer, alongside the PNG. IM sessions still receive the PNG and are instructed not to paste source; channel adapters strip paired source blocks. Renderer metadata remains removed.

src/gateway/channels/visual-image.ts
LikiosSedo: Fixed in `817726cc`. A background message with `kind: final` that collapses after legacy-card sanitization now delivers the localized empty-result notice instead of returning success with no visible message. Added a regression test.

src/core/mcp-client.ts
LikiosSedo: Fixed in `817726cc`. Process-level visual-export settings are now inherited only when the stdio command basename is exactly `mcp-create-chart`; unrelated stdio MCPs receive only their explicitly configured environment. Added command/path and non-inheritance tests.

src/gateway/channels/lark.ts
LikiosSedo: Fixed in `817726cc`. Feedback eligibility is now computed from the sanitized final body, so a legacy-card-only result becomes the empty-result notice without attaching rating controls. Added a regression test.

mcp/create-chart/src/visual-export.ts
LikiosSedo: Addressed in `817726cc`. The single end-to-end deadline remains intentional so the operation is bounded, but the default is now 60s (65s MCP transport deadline) and the Helm comment explicitly says it covers browser launch plus rendering.

mcp/create-chart/src/visual-export-config.ts
LikiosSedo: Fixed in `817726cc`. A blank `SICLAW_VISUAL_EXPORT_TIMEOUT_MS` is normalized as unset and uses the 60s default, matching URL handling and the client-side timeout calculation. Covered by tests.

mcp/create-chart/src/handler.ts
jacoblee-io: Verified on `817726cc` — resolved. I traced the whole Portal web round trip rather than just checking that a fence string is present: the tool text is the bare ```chart / ```mermaid fence carrying the same `JSON.stringify(validatedArgs)` shape as before the PR, `mcpContentToAgentContent` keeps the text block, `createToolDefinition.execute` returns it in `content`, and `Markdown.tsx` `pre()` dispatches `hasLanguageClass(className, "chart")` -> `ChartFence` and `"mermaid"` -> `MermaidFence`. The surface split in the description is coherent with `CHANNEL_SECTION`, which is only injected for `mode: "channel"`.

src/gateway/channels/visual-image.ts
jacoblee-io: Verified on `817726cc` — resolved. `stripVisualBlocks(...) || (backgroundMessage.kind === "final" ? EMPTY_RESULT_NOTICE_BY_LOCALE[locale] : "")` restores delivery for a collapsed final while still letting a collapsed milestone/artifact fall through to the silent `return true`, which is the right split. `cleanupMarkdown` trims, so the collapse is exactly `""` and the fallback reliably fires.

mcp/create-chart/src/visual-export.ts
jacoblee-io: Verified on `817726cc` — accepted. Keeping one end-to-end deadline is reasonable now that the default is 60s, and the constants agree everywhere that can name the number (`visual-export-config.ts`, `mcp-client.ts`, `values.yaml`, and the template's `default 60000`). 65s also clears the MCP SDK's own 60s `DEFAULT_REQUEST_TIMEOUT_MSEC`, so the grace genuinely extends rather than being clamped. I also checked the one thing that could have re-capped it: `AgentBoxClient(..., 120_000, ...)` in lark/dingtalk bounds only the `/api/prompt` POST, which returns a sessionId immediately — the render runs under the unbounded SSE stream.

mcp/create-chart/src/visual-export-config.ts
jacoblee-io: Verified on `817726cc` — resolved. `?.trim()` plus `configured ? Number(configured) : DEFAULT` makes both `""` and `"   "` fall to the 60s default, while `"0"` and `"abc"` still throw. The client mirror in `visualMcpRequestTimeoutMs` reaches the same number by a different route (`Number("") === 0` -> not `> 0` -> default), so the two sides now agree for `""`, `"   "` and `" 120000 "`.

src/gateway/channels/lark.ts
jacoblee-io: **Follow-up on `817726cc`** — the reported case is fixed, but the new predicate is wrong for a different notice and only survives by accident.

`isAnswer = !agentError && !sessionBusy && sanitizedFinalBody.trim().length > 0`, and `sanitizedFinalBody` derives from `finalBody`, which is already `resultText || EMPTY_RESULT_NOTICE_BY_LOCALE[locale]`. So for a plain empty agent result `sanitizedFinalBody` **is** the notice, non-empty, and `isAnswer` evaluates **true**. There are three ways `finalBody` can become a notice; the fix guards two of them (`agentError`, `sessionBusy`) and misses the third.

It is not observable today only because `collectChannelResponse` leaves `lastAssistantMessageId` null when no assistant text was persisted (the `if (turnText)` guard, and the delta-only fallback also requires `text.trim()`), and the call site is `isAnswer && assistantMessageId`. So the guard that is supposed to express "this is a real answer" is broken, and an unrelated condition in another function is what actually suppresses the buttons — any future change that persists an assistant row for an empty turn brings them back. Deriving from `stripVisualBlocks(resultText, { stripSourceBlocks: replyImages.length > 0 })` instead of from `finalBody` would make the predicate say what it means.

Separately, a side effect worth a deliberate decision rather than an accident: a legitimately image-only answer (chart rendered, text stripped, `displayBody = VISUAL_ONLY_NOTICE_BY_LOCALE[locale]`) now also gets `isAnswer === false`, so a real answer delivered as a PNG can no longer be rated at all. That is a behaviour change from the previous `resultText.trim().length > 0`.

src/core/mcp-client.ts
jacoblee-io: **Follow-up on `817726cc`** — the scope fix is right (unrelated stdio MCPs no longer inherit the renderer env, and the comment now matches the behaviour). One thing to weigh before this ships.

On whether a basename check is too weak an identity test: for *this* payload it does not really matter. A false positive forwards four non-secret ops values (renderer URL, timeout, theme, chromium path) to a process the operator already chose to run inside the AgentBox with full tool privileges.

The direction that does matter is the **false negative**. `Dockerfile.agentbox` installs the binary as `/usr/local/bin/mcp-create-chart` and the example config in `mcp/create-chart/src/index.ts` uses `"command": "mcp-create-chart"`, so the happy path matches — but there is no seeded config anywhere in the tree; this is entirely operator-entered in Portal. Any equivalent spelling silently skips the forwarding: `command: "node", args: [".../dist/index.js"]`, `command: "/usr/bin/env", args: ["mcp-create-chart"]`, or a wrapper script under another name.

The failure mode is the bad part: every call then fails with `Visual export is not configured; set SICLAW_VISUAL_EXPORT_URL...` **while the Runtime env is set correctly**, and the startup warning on the MCP's own stderr says the same thing. Nothing anywhere names the command-name gate, so there is no path from symptom to cause. Consider also matching an `args` entry whose basename is `mcp-create-chart`, or logging once when a stdio MCP misses the gate while `SICLAW_VISUAL_EXPORT_URL` is set in the parent env.

src/gateway/channels/lark.ts
```diff
@@ -1923,7 +1923,8 @@ async function processQueuedLarkMessage(ctx: QueuedLarkMessageContext): Promise<
   };
   registerBackgroundChannelDelivery(sessionId, async (backgroundMessage) => {
     if ("text" in backgroundMessage) {
-      const display = stripVisualBlocks(backgroundMessage.text);
+      const display = stripVisualBlocks(backgroundMessage.text)
```
jacoblee-io: **low, pre-existing but newly reachable** — anchored here because the real site, `condenseMilestone` (`src/gateway/channels/lark.ts:3260`), is outside this diff.

`collectChannelResponse` surfaces an intermediate assistant turn as a live-card milestone via `options.onMilestone(condenseMilestone(lastAssistantText))`, and `condenseMilestone` takes the first non-empty line **without** running it through `stripVisualBlocks` — unlike the background path right here, which does strip. Everything else in this PR now routes through `stripVisualBlocks`; this is the one channel-facing path that does not.

What makes it newly reachable: `render_chart` / `render_mermaid` now hand the model a bare ```chart / ```mermaid fence as their entire text output. An intermediate turn that echoes it right after the tool call puts ```` ```chart ```` — or the spec JSON's first line — on the group card. The same applies to a legacy ```visual-card block from an older session or skill, which is exactly the source this PR sets out never to expose in IM.

It is transient (the final card rebuilds with `milestones: []`), hence low, but it is the remaining hole in the "never expose renderer source in an IM channel" contract. Running the milestone text through `stripVisualBlocks` before `condenseMilestone` would close it.

LikiosSedo: Fixed in ae0958bd. `condenseMilestone` now applies the IM visual-source filter, including `chart`, `mermaid`, and legacy visual-card fences, before selecting the first visible line. This closes the transient live-card path without changing Web rendering. The regression test covers both a chart block followed by visible narration and a legacy-card-only intermediate turn.

jacoblee-io: **Follow-up on `ae0958bd`** — the fix is right at `condenseMilestone`, but it covers only one of that function's two callers.

`condenseMilestone` now strips with `{ stripSourceBlocks: true }` before picking the first visible line, and the new test covers the `message_end` milestone path. The sibling caller does not benefit:

```ts
// channelActivityMilestone, src/gateway/channels/lark.ts:3294
return condenseMilestone(clean) || clean;
```

When the activity text is entirely renderer source, `condenseMilestone` returns `""` and the `|| clean` fallback puts the **unstripped** fence on the card — exactly what this PR is closing off. It is reachable: `activity` is the joined `text` blocks of `ev.partialResult.content` (`lark.ts:3144-3150`), so a sub-agent that renders a chart and echoes the fence in its own narration lands here.

The root cause is that `""` now means two different things. Before this change it meant only "the first line condensed to nothing" (e.g. a bare `###`), for which falling back to the original text is correct. It now also means "nothing survived source-stripping", for which falling back publishes the source.

Either return `""` when stripping emptied the text (emit no milestone), or have the fallback use the stripped text rather than `clean`.

Verified on `ae0958bd`; `lark`, `visual-image` and `mcp-client` tests pass (278).

src/gateway/channels/lark.ts
LikiosSedo: Fixed in 8e33bcd8. Feedback eligibility is now derived from the sanitized original result, not from a fallback notice. Empty and legacy-card-only results cannot be rated, while a successfully delivered PNG-only result remains rateable. Both boundaries now have regression coverage.

src/core/mcp-client.ts
LikiosSedo: Addressed in 8e33bcd8. The inheritance gate now recognizes the canonical executable, an exact `mcp-create-chart` wrapper argument (for example `/usr/bin/env mcp-create-chart`), and the package entrypoint path used by a direct Node launch. Unrelated stdio MCP commands still do not inherit the visual-export environment, and custom launchers can continue to set explicit MCP env.

src/gateway/channels/lark.ts
LikiosSedo: Fixed in 8e33bcd8. The activity fallback now uses the stripped text rather than the original `clean` value, so a renderer-source-only activity produces no milestone. Added an end-to-end Lark regression for the `tool_execution_update` path.

jacoblee-io: **low, same contract as the two paths this PR just closed** — background delivery still strips only the legacy card fences.

Both call sites here (`lark.ts:1926` and `lark.ts:1949`) use bare `stripVisualBlocks(...)`, so `chart` / `mermaid` source survives. The non-final branch feeds it straight to `addMilestone(display)`, which means one card can now show two different behaviours: a milestone from the SSE path is source-stripped (`condenseMilestone`), a milestone from the background path is not.

It is also the worst place for the source to land: `ChannelDeliveryPayload` (`DelegationAppendMessagePayload` / `ChannelDeliverMessagePayload`) carries no image attachments, so unlike `lark.ts:2070` there is no rendered picture the source could be the fallback for — the reader gets JSON and nothing else.

`{ stripSourceBlocks: true }` on both, matching `condenseMilestone`.

jacoblee-io (dismissed):
Verified on `8e33bcd8`. All nine findings from the earlier rounds are resolved in the code, not just in the replies:

- `isAnswer` is now derived from `sanitizedResultBody` (built from `resultText`), so no notice can satisfy it, and the image-only regression is gone — a PNG answer is ratable again.
- The inheritance gate recognises the `args` spelling and the `mcp/create-chart/(dist/)?index.js` entrypoint, covering `/usr/bin/env mcp-create-chart` and the `node …/dist/index.js` form.
- The activity fallback uses the stripped text, so a source-only activity yields no milestone while the degenerate-heading case still falls back as before.

Full suite on this head: 302 files, 6718 passed / 2 skipped; `tsc --noEmit` clean.

Approving. One same-contract gap is left, inline below — not a regression from this PR, and fine as a follow-up.

Two smaller residues, both pre-existing and neither blocking:
- A wrapper under any other name still misses the gate silently, and the resulting "Visual export is not configured" names nothing an operator can act on. Widening covered the common spellings; the diagnostic half is still open.
- `lark.ts:2070` keeps renderer source in the final reply when no image was produced (`stripSourceBlocks: replyImages.length > 0`). That predates this PR, but it is now the only ordinary path that can put source in an IM message, which is worth stating explicitly somewhere rather than leaving as an implicit exception to "never expose source in IM".

src/gateway/channels/lark.ts
LikiosSedo: Fixed in ca1f0196. Both background-delivery branches now strip `chart` and `mermaid` source in addition to legacy visual-card fences. A source-only milestone is ignored, while a source-only final/background answer becomes the localized empty-result notice. The regression test covers both payload shapes.

jacoblee-io (approved):
Verified on `ca1f0196` — the background-delivery gap from the previous round is closed.

Both call sites now pass `{ stripSourceBlocks: true }`, so a source-only background payload renders the ⚠ empty-result notice on a final and produces no milestone on a non-final, instead of putting `chart` / `mermaid` JSON on the card. With this, every path that reaches an IM card strips renderer source, and the one deliberate exception is `lark.ts:2070`, where the source is kept only when no image was produced.

Full suite on this head: 302 files, 6719 passed / 2 skipped; `tsc --noEmit` clean.

Approving. The only thing still open is the diagnostic half of the inheritance-gate note — a wrapper under an unrecognised name skips the forwarding silently and the resulting "Visual export is not configured" names nothing an operator can act on. Not blocking.

PR #564, opened 2026-09-06
feat(mcp): add product-support-result stdio MCP

LikiosSedo:
## Why

`src/core/agent-types.ts` already ships the `product_support` agent type, and its built-in contract tells the agent to call a result-submission tool exactly once per turn with a machine-readable outcome. No such tool shipped in this repo, so a deployment had to layer one in from a private overlay and build the agentbox image from two sources.

This adds the tool the type already contracts for, so an agentbox image built from this repo alone can serve a Product Support agent.

## What

`mcp/product-support-result/` — a stateless stdio MCP exposing one tool:

`submit_product_support_result(label, info)` validates and canonicalizes one turn's handoff decision, returning it as both text content and MCP `structuredContent`.

- `ticket_type` is one of `consultation` / `incident` / `llm_incident` / `requirement` / `unknown`. `llm_incident` covers faults while calling an LLM / model API or inference endpoint (gateway errors, auth, rate limits, quotas, a named model misbehaving); `incident` is every other fault
- `info.llm = { region, aspect, model }` is optional on input (absent means all three empty) and always present in the result; it carries best-effort intake details for an `llm_incident`: `region` is `""` / `domestic` / `overseas`, `aspect` is `""` / `platform_api` / `network` / `model`, `model` is free text. These are hints for first-line support, not gates: a ticket-ready `llm_incident` does not require any of them, so the agent is never pushed to invent a value. For every other ticket type all three must be empty
- `label=true` requires a resolved type, non-empty summary and description, and no missing fields; a requirement additionally requires a concrete product
- `missing_fields` accepts lowercase `snake_case` machine identifiers only, never user-facing questions
- unknown keys are rejected at both schema levels

It has no network access, no ticket/group/notification side effect, and no session or turn identity, so "exactly one successful result per turn" is deliberately left to the caller that consumes the result.

Registered in `mcp/MCP_LIST.txt`, so `Dockerfile.agentbox` builds it and links `mcp-product-support-result` onto `PATH` alongside `create-chart`.

## Notes for review

- `tsconfig.json` excludes `src/**/*.test.ts` (mirroring `mcp/create-chart`) so the in-image `npm ci && npx tsc` needs no dev dependencies; `dist/` contains no test output. The tests run under the repo's vitest instead.
- The package pins `@modelcontextprotocol/sdk` `^1.27.1` with the lockfile resolving 1.27.1, the same version the repo root tests resolve, so the tested and shipped SDK are identical.

## Second commit

`c539adca` adds `llm_incident` and the `llm` block described above and bumps the package to 0.2.0. With it this package is now the single home of the tool; the private overlay copy is being removed and will not be maintained further.

`72df8e27` addresses the review: SDK pinned to the repo-root version (1.27.1), per-field size limits enforced in schema and parser, `missing_fields` errors at the caller's index, `unknown` exempt from the llm-emptiness rule, uniform enum case handling. Package 0.2.1.

`b6a6fbb5` addresses the follow-up on the approving review: the parser mirrors the advertised schema exactly (exact enum values, length checked on the raw value), `llm` is optional with absent meaning all-empty, and the post-trim de-duplication of `evidence` / `missing_fields` is documented. Package 0.2.2.

## Verification

- `npm ci && npx tsc` inside `mcp/product-support-result` — reproduces the image build step; `dist/` has `index.js`, `result.js`, `server.js` and no test output
- `npx vitest run mcp/product-support-result` — 24 passed (2 files), with the package's own `node_modules` removed so resolution goes to the repo root
- `npx vitest run` (full suite) — 304 files, 6748 passed, 2 skipped, 0 failed
- `npx tsc --noEmit` at the repo root — clean
- `npm audit` inside the package — 0 vulnerabilities

Not verified: no image was built and no deployment was exercised in this PR.

### DIFF
--- a/mcp/product-support-result/package.json
+++ b/mcp/product-support-result/package.json
@@ -1,0 +1,30 @@
+{
+  "name": "mcp-product-support-result",
+  "version": "0.1.0",
+  "description": "Stateless MCP validator for product-support handoff results",
+  "type": "module",
+  "bin": {
+    "mcp-product-support-result": "./dist/index.js"
+  },
+  "main": "./dist/index.js",
+  "files": [
+    "dist",
+    "README.md"
+  ],
+  "scripts": {
+    "build": "tsc",
+    "start": "node dist/index.js",
+    "clean": "rm -rf dist"
+  },
+  "engines": {
+    "node": ">=22.12.0"
+  },
+  "dependencies": {
+    "@modelcontextprotocol/sdk": "^1.29.0"
+  },
+  "devDependencies": {
+    "@types/node": "^22.0.0",
+    "typescript": "^5.9.0"
+  },
+  "license": "Apache-2.0"
+}
--- a/mcp/product-support-result/package.json
+++ b/mcp/product-support-result/package.json
@@ -1,6 +1,6 @@
 {
   "name": "mcp-product-support-result",
-  "version": "0.2.0",
+  "version": "0.2.1",
   "description": "Stateless MCP validator for product-support handoff results",
   "type": "module",
   "bin": {
@@ -20,7 +20,7 @@
     "node": ">=22.12.0"
   },
   "dependencies": {
-    "@modelcontextprotocol/sdk": "^1.29.0"
+    "@modelcontextprotocol/sdk": "^1.27.1"
   },
   "devDependencies": {
     "@types/node": "^22.0.0",
--- a/mcp/product-support-result/package.json
+++ b/mcp/product-support-result/package.json
@@ -1,6 +1,6 @@
 {
   "name": "mcp-product-support-result",
-  "version": "0.2.1",
+  "version": "0.2.2",
   "description": "Stateless MCP validator for product-support handoff results",
   "type": "module",
   "bin": {
--- a/mcp/product-support-result/package.json
+++ b/mcp/product-support-result/package.json
@@ -1,6 +1,6 @@
 {
   "name": "mcp-product-support-result",
-  "version": "0.1.0",
+  "version": "0.2.0",
   "description": "Stateless MCP validator for product-support handoff results",
   "type": "module",
   "bin": {
--- a/mcp/product-support-result/src/result.ts
+++ b/mcp/product-support-result/src/result.ts
@@ -1,11 +1,38 @@
 export const TICKET_TYPES = [
   "consultation",
   "incident",
+  "llm_incident",
   "requirement",
   "unknown",
 ] as const;
 
 export type TicketType = (typeof TICKET_TYPES)[number];
+
+/**
+ * Deployment region the user is calling the LLM API from, as classified by the
+ * operator's own rules. Empty when the conversation does not establish it.
+ */
+export const LLM_REGIONS = ["domestic", "overseas"] as const;
+export type LlmRegion = (typeof LLM_REGIONS)[number];
+
+/**
+ * Which part of the LLM API path the user reports as failing. Empty when the
+ * conversation does not establish it.
+ */
+export const LLM_ASPECTS = ["platform_api", "network", "model"] as const;
+export type LlmAspect = (typeof LLM_ASPECTS)[number];
+
+/**
+ * Best-effort intake details for an `llm_incident`. Every field may stay empty:
+ * the block exists so first-line support sees what the conversation did
+ * establish, not so the agent is forced to guess. Only the enum shape is
+ * validated; none of the fields is required for a ticket-ready result.
+ */
+export interface LlmIncidentInfo {
+  region: "" | LlmRegion;
+  aspect: "" | LlmAspect;
+  model: string;
+}
 
 export interface ProductSupportInfo {
   ticket_type: TicketType;
@@ -14,6 +41,7 @@ export interface ProductSupportInfo {
   description: string;
   evidence: string[];
   missing_fields: string[];
+  llm: LlmIncidentInfo;
 }
 
 export interface ProductSupportResult {
@@ -29,8 +57,12 @@ const INFO_KEYS = new Set([
   "description",
   "evidence",
   "missing_fields",
+  "llm",
 ]);
+const LLM_KEYS = new Set(["region", "aspect", "model"]);
 const TICKET_TYPE_SET = new Set<string>(TICKET_TYPES);
+const LLM_REGION_SET = new Set<string>(LLM_REGIONS);
+const LLM_ASPECT_SET = new Set<string>(LLM_ASPECTS);
 const MISSING_FIELD_PATTERN = /^[a-z][a-z0-9_]*$/;
 
 function isRecord(value: unknown): value is Record<string, unknown> {
@@ -61,6 +93,22 @@ function readString(value: unknown, path: string): string {
   return value.trim();
 }
 
+function readOptionalEnum<T extends string>(
+  value: unknown,
+  allowed: ReadonlySet<string>,
+  options: readonly T[],
+  path: string,
+): "" | T {
+  const text = readString(value, path).toLowerCase();
+  if (text.length === 0) {
+    return "";
+  }
+  if (!allowed.has(text)) {
+    throw new Error(`${path} must be empty or one of: ${options.join(", ")}`);
+  }
+  return text as T;
+}
+
 function readStringArray(value: unknown, path: string): string[] {
   if (!Array.isArray(value)) {
     throw new Error(`${path} must be an array of strings`);
@@ -79,6 +127,18 @@ function readStringArray(value: unknown, path: string): string[] {
     }
   });
   return canonical;
+}
+
+function readLlmInfo(value: unknown, path: string): LlmIncidentInfo {
+  if (!isRecord(value)) {
+    throw new Error(`${path} must be an object`);
+  }
+  assertExactKeys(value, LLM_KEYS, path);
+  return {
+    region: readOptionalEnum(value.region, LLM_REGION_SET, LLM_REGIONS, `${path}.region`),
+    aspect: readOptionalEnum(value.aspect, LLM_ASPECT_SET, LLM_ASPECTS, `${path}.aspect`),
+    model: readString(value.model, `${path}.model`),
+  };
 }
 
 export function parseProductSupportResult(input: unknown): ProductSupportResult {
@@ -114,6 +174,7 @@ export function parseProductSupportResult(input: unknown): ProductSupportResult 
         input.info.missing_fields,
         "input.info.missing_fields",
       ),
+      llm: readLlmInfo(input.info.llm, "input.info.llm"),
     },
   };
 
@@ -124,6 +185,17 @@ export function parseProductSupportResult(input: unknown): ProductSupportResult 
       );
     }
   });
+
+  // The llm block belongs to llm_incident only. A stray region or model on a
+  // consultation would be read by first-line support as an established fact.
+  if (result.info.ticket_type !== "llm_incident") {
+    const { region, aspect, model } = result.info.llm;
+    if (region !== "" || aspect !== "" || model !== "") {
+      throw new Error(
+        "input.info.llm fields must be empty unless ticket_type is llm_incident",
+      );
+    }
+  }
 
   if (!result.label) {
     return result;
--- a/mcp/product-support-result/src/result.ts
+++ b/mcp/product-support-result/src/result.ts
@@ -77,8 +77,12 @@ const INFO_KEYS = new Set([
   "description",
   "evidence",
   "missing_fields",
-  "llm",
 ]);
+// Optional: absent means "all three empty". All-empty is the correct value for
+// four of the five ticket types, so requiring the block would only turn the
+// most common shape into a host-side schema rejection on the turn's one
+// mandatory tool call.
+const INFO_OPTIONAL_KEYS = new Set(["llm"]);
 const LLM_KEYS = new Set(["region", "aspect", "model"]);
 const TICKET_TYPE_SET = new Set<string>(TICKET_TYPES);
 const LLM_REGION_SET = new Set<string>(LLM_REGIONS);
@@ -94,9 +98,10 @@ function assertExactKeys(
   value: Record<string, unknown>,
   expected: ReadonlySet<string>,
   path: string,
+  optional: ReadonlySet<string> = new Set(),
 ): void {
   for (const key of Object.keys(value)) {
-    if (!expected.has(key)) {
+    if (!expected.has(key) && !optional.has(key)) {
       throw new Error(`${path}.${key} is not allowed`);
     }
   }
@@ -111,11 +116,13 @@ function readString(value: unknown, path: string, maxChars: number): string {
   if (typeof value !== "string") {
     throw new Error(`${path} must be a string`);
   }
-  const text = value.trim();
-  if ([...text].length > maxChars) {
+  // Length is checked on the raw value, exactly as the advertised `maxLength`
+  // is applied by a host that validates arguments before dispatch, so this
+  // parser is never more lenient than the schema the model was shown.
+  if ([...value].length > maxChars) {
     throw new Error(`${path} must be at most ${maxChars} characters`);
   }
-  return text;
+  return value.trim();
 }
 
 function readEnum<T extends string>(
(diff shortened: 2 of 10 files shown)


mcp/product-support-result/package.json
```diff
@@ -0,0 +1,30 @@
...
+  "scripts": {
+    "build": "tsc",
+    "start": "node dist/index.js",
+    "clean": "rm -rf dist"
+  },
+  "engines": {
+    "node": ">=22.12.0"
+  },
+  "dependencies": {
+    "@modelcontextprotocol/sdk": "^1.29.0"
```
jacoblee-io: **MEDIUM — the SDK version this package ships with is the one nothing verifies.**

`"@modelcontextprotocol/sdk": "^1.29.0"` (the lockfile resolves 1.30.0) is what `Dockerfile.agentbox`'s in-image `npm ci` installs, and therefore what the shipped binary runs. But no `node_modules` is committed here and no CI job installs or builds inside `mcp/` — CI runs root `npm ci` + `npm test` only. So `server.test.ts`, the only test that exercises the SDK at all (`listTools`, `callTool`, output-schema validation), resolves the repo root's **1.27.1**, exactly as the PR description states.

The consequence is that the artifact runs a dependency version no test covers: a change in SDK 1.29/1.30's `Server` result handling or client-side `outputSchema` validation ships green and fails only inside the image, where the failure surfaces as a tool that no longer answers rather than as a red test.

Nothing in this package needs ≥1.29 — I ran all 19 tests against 1.27.1 and `outputSchema` round-trips fine. Both `mcp/create-chart` and the repo root pin `^1.27.1`.

Fix: pin `^1.27.1` so the tested version and the shipped version are the same, or add a CI step that runs `npm ci && npx tsc && vitest` inside the package.

mcp/product-support-result/src/server.ts
```diff
@@ -0,0 +1,129 @@
...
+          enum: ["consultation", "incident", "llm_incident", "requirement", "unknown"],
+          description:
+            "llm_incident covers failures while calling an LLM / model API or inference endpoint (gateway errors, auth, rate limits, quotas, a named model misbehaving). incident is every other fault.",
+        },
+        product: {
+          type: "string",
+          description:
+            "Concrete, knowledge-grounded product for requirements. For incidents and consultations, use a concrete product only when established by the conversation or authoritative product knowledge; otherwise leave this empty and preserve the user-visible entry in description.",
+        },
+        summary: { type: "string" },
```
jacoblee-io: **MEDIUM — no length bound on any field, so an oversized result kills the INSERT of the one row that carries the outcome.**

`summary` / `description` / `product` / `llm.model` have no `maxLength`, and `evidence` / `missing_fields` have no `maxItems` — neither in this JSON Schema nor in the parser in `result.ts`.

Measured against this branch: a result with two 300,000-char strings and 5,000 `evidence` entries validates and comes back as a 639 KB `content[0].text` **and** a 639 KB `structuredContent`. Neither is bounded downstream:

- `src/core/mcp-client.ts:352` truncates neither, so the whole blob re-enters the model's context on the next turn.
- `src/gateway/sse-consumer.ts:653` persists `structuredContent` into `chat_messages.metadata`, declared `TEXT` in `src/portal/migrate.ts:346` — 65,535 bytes on MySQL. (`tool_input` next to it is already `MEDIUMTEXT`.)

Scenario: a support user pastes a long log, the model copies it into `description` or `evidence`, and under the default `STRICT_TRANS_TABLES` the tool row's INSERT fails with MySQL error 1406. The row that is lost is precisely the turn's one machine-readable outcome.

This tool is the single gate in front of a downstream ticket-creation robot, so an oversized field should fail closed here — a rejection the model can correct by shortening the field — rather than being forwarded and dropped later. Suggest `maxLength` on the string fields and `maxItems` on the arrays, in the schema and enforced in the parser.

mcp/product-support-result/src/result.ts
```diff
@@ -0,0 +1,221 @@
...
+      evidence: readStringArray(input.info.evidence, "input.info.evidence"),
+      missing_fields: readStringArray(
+        input.info.missing_fields,
+        "input.info.missing_fields",
+      ),
+      llm: readLlmInfo(input.info.llm, "input.info.llm"),
+    },
+  };
+
+  result.info.missing_fields.forEach((field, index) => {
```
jacoblee-io: **LOW — the index in this error message is not the caller's index, so the model is pointed at the wrong element.**

The pattern check runs over `result.info.missing_fields`, which is the array `readStringArray` has already **deduplicated**. The `index` from this `forEach` is therefore a position in the deduplicated array, while the error text names it as `input.info.missing_fields[<index>]` — a coordinate in the input the caller sent.

Measured: input `missing_fields: ["a", "a", "Bad Field"]` produces

```
input.info.missing_fields[1] must be a lowercase snake_case field identifier
```

while the offending entry is the caller's `[2]`. Every duplicate earlier in the array widens the skew, and at `[1]` the model finds `"a"`, which is valid — so the correction it makes is to the wrong element.

Fix: validate the pattern inside `readStringArray`'s own `forEach`, which already holds the source index, or carry that index through to here.

mcp/product-support-result/src/result.ts
```diff
@@ -0,0 +1,221 @@
...
+    if (!MISSING_FIELD_PATTERN.test(field)) {
+      throw new Error(
+        `input.info.missing_fields[${index}] must be a lowercase snake_case field identifier`,
+      );
+    }
+  });
+
+  // The llm block belongs to llm_incident only. A stray region or model on a
+  // consultation would be read by first-line support as an established fact.
+  if (result.info.ticket_type !== "llm_incident") {
```
jacoblee-io: **LOW — this rule and the `unknown` rule contradict each other during intake, so a fact the user already stated cannot be recorded.**

The emptiness check runs before the `label` branch and rejects the whole call, so it also governs the phase where the agent is still gathering. But during intake `ticket_type` is `unknown` by construction — line 204 forbids `unknown` only for `label: true` — which makes the two rules mutually exclusive in practice.

Measured: `{label: false, ticket_type: "unknown", llm: {region: "domestic"}}` is rejected outright.

Scenario: the user names the region in their first message, before the agent can tell `incident` from `llm_incident`. The agent either burns a round trip on the rejection, or drops an established fact because there is nowhere to record it until the type resolves — and the type may only resolve later in the conversation.

The comment above is right about the case it describes: a stray region on a *consultation* would be read by first-line support as an established fact. That reasoning applies to the resolved types, not to `unknown`. Suggest exempting `unknown`, or enforcing emptiness only once the type has resolved to a non-LLM one.

mcp/product-support-result/src/result.ts
```diff
@@ -0,0 +1,221 @@
...
+
+  if (typeof input.label !== "boolean") {
+    throw new Error("input.label must be a boolean");
+  }
+  if (!isRecord(input.info)) {
+    throw new Error("input.info must be an object");
+  }
+  assertExactKeys(input.info, INFO_KEYS, "input.info");
+
+  const ticketType = readString(input.info.ticket_type, "input.info.ticket_type");
```
jacoblee-io: **LOW — one validator, two tolerances for the same kind of field, and the strict one is on the primary classification field.**

`ticket_type` is trimmed but not case-normalized, while `llm.region` / `llm.aspect` are lowercased by `readOptionalEnum` (line 102).

Measured: `" Overseas "` and `"MODEL"` are accepted and canonicalized, but `"Incident"` is rejected with `input.info.ticket_type must be one of: …`.

So the leniency lands on the optional hint fields and the strictness on the field that decides which ticket gets created — the opposite of where each would be worth having. A model that has just been told `"Overseas"` was fine has no way to predict that `"Incident"` is not.

Suggest picking one rule. Lowercasing here costs nothing and does not affect the output enum, since the accepted set is already the lowercase spelling.

mcp/product-support-result/package.json
LikiosSedo: Fixed in 72df8e27. Pinned `^1.27.1` and regenerated the lockfile so it resolves 1.27.1, the same version the repo root tests against. Verified the in-package `npm ci && npx tsc` (the image step) now installs 1.27.1 and `npm audit` is clean; the 24 package tests run at the repo root against the same version.

mcp/product-support-result/src/server.ts
LikiosSedo: Fixed in 72df8e27. Bounds are now enforced in both the JSON Schema and the parser: `summary` 200, `description` 2000, `product` / `llm.model` 128 characters; `evidence` 20 items x 300, `missing_fields` 20 items x 64. Limits are counted in characters (CJK is not penalized) and sized so a result filled to every limit with 3-byte UTF-8 stays under ~30 KB, well inside the 65,535-byte TEXT column. An oversized field is rejected here with the field named, so the model can shorten it. Tests cover each limit and the character-vs-byte point.

mcp/product-support-result/src/result.ts
LikiosSedo: Fixed in 72df8e27. The pattern check moved into `readStringArray` and runs on the original index before dedup. Test added with your exact input `["a", "a", "Bad Field"]`, which now reports `missing_fields[2]`.

mcp/product-support-result/src/result.ts
LikiosSedo: Agreed and fixed in 72df8e27. `unknown` is now exempt: the emptiness rule applies only once the type has resolved to `consultation`, `incident` or `requirement`, so a region or model stated in the first message can be recorded while the type is still unresolved. Your case `{label: false, ticket_type: "unknown", llm: {region: "domestic"}}` is a test now. README and the schema description updated to match.

mcp/product-support-result/src/result.ts
LikiosSedo: Fixed in 72df8e27. All three enum fields (`ticket_type`, `llm.region`, `llm.aspect`) now go through one `readEnum` that trims and lower-cases on input and returns the canonical lowercase value. `" Incident "` is accepted and canonicalized to `incident`; test added.

mcp/product-support-result/src/result.ts
```diff
@@ -0,0 +1,271 @@
...
+function readEnum<T extends string>(
+  value: unknown,
+  allowed: ReadonlySet<string>,
+  options: readonly T[],
+  path: string,
+  optional: boolean,
+): "" | T {
+  // Enum fields are case-insensitive on input and canonical lowercase on
+  // output, uniformly: the model must not have to guess which field is strict.
+  const text = readString(value, path, 64).toLowerCase();
```
jacoblee-io: **LOW — the enum trim/lower-casing cannot take effect on the primary agentbox path, so the README's claim holds on only one of two host paths.**

`readEnum` accepts `" Incident "`, but `src/core/mcp-client.ts:71` passes the advertised schema through verbatim as the tool's `parameters`, and pi validates arguments against it *before* dispatch. I compiled the actually-advertised schema with the repo's own `typebox/compile`: `"Incident"` and `" incident "` both fail with `must be equal to one of the allowed values`, and so does `region: "Overseas"`.

So on the pi-brain path the model gets a generic validator rejection instead of this tool's guidance, and the normalization only ever runs on the SDK-brain native-MCP path. The README states it unconditionally ("Enum fields are trimmed and lower-cased on input"), and the new test `"normalizes ticket_type case like the other enum fields"` calls the parser directly, bypassing the host — so it can never surface the divergence.

Same root cause at `result.ts:110`: `readString` trims *before* the length check, while the advertised `maxLength` applies to the raw string. A 200-char summary with a trailing newline is rejected upstream for length even though this parser would accept it.

Either state the path-dependence where the tolerance is documented, or drop the tolerance and let the schema be the single answer.

mcp/product-support-result/src/server.ts
```diff
@@ -0,0 +1,134 @@
...
+            "Blocking machine field identifiers only; never user-facing questions or diagnostic instructions.",
+          maxItems: LIMITS.missingFieldsMaxItems,
+          items: {
+            type: "string",
+            minLength: 1,
+            maxLength: LIMITS.missingFieldMaxChars,
+            pattern: MISSING_FIELD_PATTERN,
+          },
+        },
+        llm: {
```
jacoblee-io, line 64: **LOW — the one tool a turn must land is also the most schema-brittle, and all-empty is the correct value for four of the five ticket types.**

`llm` is `required` with all three sub-keys required, on top of `additionalProperties: false` at three nesting levels. A model that emits a `consultation` result without the block is rejected host-side with `must have required properties llm` (verified against the advertised schema), which burns a retry on the call that produces the turn's only machine-readable outcome.

Making `llm` optional with "absent ⇒ all three empty" would remove that whole class of rejected calls without weakening anything: the cross-field emptiness rule at `result.ts:243` is what actually protects first-line support from a stray region or model name, and it keeps working unchanged.

mcp/product-support-result/src/result.ts
```diff
@@ -0,0 +1,271 @@
...
+  value.forEach((item, index) => {
+    const itemPath = `${path}[${index}]`;
+    const text = readString(item, itemPath, rules.itemMaxChars);
+    if (text.length === 0) {
+      throw new Error(`${itemPath} must not be blank`);
+    }
+    if (rules.itemPattern && !rules.itemPattern.re.test(text)) {
+      throw new Error(`${itemPath} ${rules.itemPattern.message}`);
+    }
+    if (!seen.has(text)) {
```
jacoblee-io, line 182: **LOW — `evidence` and `missing_fields` are silently deduplicated, and that is the one canonicalization the docs don't mention.**

`readStringArray` drops post-trim duplicates, so `evidence: ["x", " x "]` comes back as `["x"]` (confirmed end to end through the built binary). Neither the README's size-limit/canonicalization section nor the schema descriptions say so, while trimming and enum lower-casing are both documented.

A consumer that diffs the persisted `tool_input` against `structuredContent` — exactly the pairing the README's "Portal MCP config" section tells a control plane to rely on — then sees an item count it cannot account for. Either document it alongside the other canonicalizations, or return the caller's list unchanged and reject duplicates instead.

jacoblee-io (dismissed):
Approving. All five findings from the previous round are fixed, and I verified each rather than taking the commit message for it: the SDK is pinned back to `^1.27.1` with the lockfile in sync (so the in-image `npm ci` matches what the tests run against), the string fields carry `maxLength` and the arrays `maxItems`, the `missing_fields` pattern check moved inside `readStringArray`'s `forEach` so the reported index is the caller's, `unknown` is exempted from the `llm`-emptiness rule with the reasoning recorded in the comment, and `ticket_type` is trimmed and lower-cased like the other enums.

I also went past reading the diff: 24 tests pass, `tsc` is clean under the package's own tsconfig, `dist/` has no test output and keeps its shebang, and the built binary answers `listTools`, a success call and a rejection correctly over real stdio through `StdioClientTransport`. The `outputSchema: inputSchema` reuse is safe — the SDK client compiles it with Ajv in non-strict mode, Ajv counts `maxLength` in code points exactly as the parser's `[...text].length` does (a 200-emoji summary passes both), and `Client.callTool` exempts the `isError` path that carries no `structuredContent`.

No correctness bug found — no inverted condition, off-by-one, null deref, missing await or dropped error. The three comments below are contract-level and none of them blocks the merge. The first is the one I would actually act on: the enum tolerance you just added cannot take effect on the pi-brain path, because the advertised schema is validated before dispatch, so the README's unconditional claim about it is true on only one of the two host paths.

mcp/product-support-result/src/result.ts
LikiosSedo: Fixed in b6a6fbb5 by dropping the tolerance rather than documenting the path-dependence: the schema is the single answer and the parser mirrors it. Enum values are exact (`" Incident "` and `region: "Overseas"` are now rejected here too, with the field named), and `readString` checks the length on the raw value before trimming, as the host applies `maxLength`, so the parser is never more lenient than the advertised schema. README Canonicalization section rewritten accordingly; the old normalization test is replaced by one that asserts the strict behaviour.

mcp/product-support-result/src/server.ts
LikiosSedo: Agreed, fixed in b6a6fbb5. `llm` is no longer in `required`; an absent block is read as all three fields empty, and the returned result always carries the block. When present, all three keys are still required and the cross-field emptiness rule for resolved non-LLM types is unchanged. The schema description now tells the model to omit it for anything other than `llm_incident`.

mcp/product-support-result/src/result.ts
LikiosSedo: Documented in b6a6fbb5, keeping the behaviour: the README Canonicalization section and the schema descriptions of `evidence` and `missing_fields` now state that items are trimmed and post-trim duplicates are removed, so the returned arrays can be shorter than the caller's. I kept dedup rather than rejecting duplicates because a duplicate evidence line is not something worth a rejected call on the turn's one mandatory tool.

jacoblee-io (approved):

PR #568, opened 2026-09-07
feat(mcp): report per-server connection outcomes and add mcp.probe

LikiosSedo:
## Why

A configured MCP server that failed to connect left one `console.error` inside the AgentBox and nothing else. `/api/sync-status` reported the configured names (config keys), so Sicore showed every server as "installed" — including a platform MCP whose URL pointed at a web page. Its agents never had the tools and nobody could see why without reading a Pod log.

## What

- `McpClientManager` keeps a per-server connection outcome and exposes `getServerConnections()`: `connected` with tool inventory, or `failed` with a classified error (`classifyMcpConnectError`: `invalid_config | dns | connection_refused | timeout | tls | auth | not_found | not_mcp | http | protocol | unknown`, plus `httpStatus` / `contentType` / a short stripped message). An HTML body is reduced to `HTML page: <title>` instead of a 15 KB response.
- `BoxSyncStatus` schema **v4**: `mcp.servers[]` next to `mcp.names`. Absent = old box / MCP not initialized yet; `[]` = reported, nothing configured. The normalizer keeps those apart.
- The box records outcomes when a session is **created**, so a server that failed to connect is visible even if the turn that followed also failed (e.g. the model rejecting an oversized tool list). `/api/sync-status` prefers the newest live session's manager.
- `agent.syncStatus`: connection outcomes (name/state/toolCount/errorKind) are part of the replica identity; the flat `mcp.servers` is only published when every running box agrees.
- New Runtime RPC **`mcp.probe { server, timeoutMs? }`** dials one server config from the Runtime and returns the same outcome shape. Runs on the Runtime on purpose (network position of the boxes, per-runtime address overrides). `stdio` is refused.
- Design note: `docs/design/mcp-connection-observability.md`; cross-reference in `mcp-session-lifecycle.md`.

Sicore counterpart (stores the outcomes, shows them on the harness panel and the MCP page, adds a "Test connection" probe and Dev MCP tools) is in a separate MR.

## Tests

- `src/core/mcp-client-connections.test.ts` — real local HTTP endpoints: connected server, HTML 404 → `not_found`/`text/html`, connection refused, unknown transport, copies.
- `src/shared/agentbox-sync-status.test.ts` — v4 normalization.
- `src/gateway/server-sync-status.test.ts` — consensus with/without agreeing outcomes.
- `src/gateway/server-mcp-probe.test.ts` — probe happy path, HTML 404, stdio refusal, timeout, validation.
- Full `vitest run`: 6784 passed; 5 failures in `server-chat-abort.test.ts` / `node-logs-skill-scripts.test.ts` pass when run alone on both this branch and `origin/main` (load-dependent, unrelated).

### DIFF
--- a/src/agentbox/http-server.ts
+++ b/src/agentbox/http-server.ts
@@ -32,6 +32,7 @@ import {
   readBoxSyncStatus,
   type KnowledgeSyncHandler,
 } from "./sync-handlers.js";
+import type { ObservedMcpServer } from "../shared/agentbox-sync-status.js";
 import { GATEWAY_SYNC_DESCRIPTORS, type AgentBoxSyncHandler, type GatewaySyncType } from "../shared/gateway-sync.js";
 import { detectLanguage } from "../shared/detect-language.js";
 import { stripLanguageDirective } from "../shared/strip-language-directive.js";
@@ -653,6 +654,14 @@ export function createHttpServer(
     candidatesRevision: string | null;
     observedAt: string;
   } | null = null;
+  /**
+   * Per-server MCP connection outcome from the most recent session that
+   * initialized MCP in this box. Recorded when the session is CREATED, not when
+   * a turn succeeds: a server that fails to connect is exactly the observation
+   * a developer needs, and it must not disappear because the turn that followed
+   * also failed (e.g. the model rejecting an oversized tool list).
+   */
+  let observedMcpServers: ObservedMcpServer[] | null = null;
   if (sessionManager.credentialBroker) {
     perServerHandlers.cluster = createClusterHandler(sessionManager.credentialBroker);
     perServerHandlers.host = createHostHandler(sessionManager.credentialBroker);
@@ -808,11 +817,21 @@ export function createHttpServer(
    * intended to send.
    */
   addRoute("GET", "/api/sync-status", async (_req, res) => {
+    const inventory = readBoxSyncStatus({
+      knowledgeDir: sessionManager.knowledgeDir,
+      knowledgeHandler: perServerKnowledgeHandler,
+    });
+    // Prefer a live session's manager over the last recorded snapshot: an MCP
+    // reload invalidates sessions, so the newest session is the one built from
+    // the current config. Fall back to the snapshot when every session has been
+    // released, and to "no servers field" when this box never initialized MCP.
+    const liveManager = sessionManager.list()
+      .filter((session) => session.mcpManager)
+      .sort((a, b) => b.createdAt.getTime() - a.createdAt.getTime())[0]?.mcpManager;
+    const servers = liveManager ? liveManager.getServerConnections() : observedMcpServers;
     sendJson(res, 200, {
-      ...readBoxSyncStatus({
-        knowledgeDir: sessionManager.knowledgeDir,
-        knowledgeHandler: perServerKnowledgeHandler,
-      }),
+      ...inventory,
+      mcp: { ...inventory.mcp, ...(servers ? { servers } : {}) },
       model: observedModel,
       harness: observedHarness,
       tiers: observedTiers,
@@ -916,6 +935,9 @@ export function createHttpServer(
       delegation,
       body.userId,
     );
+    if (managed.mcpManager) {
+      observedMcpServers = managed.mcpManager.getServerConnections();
+    }
     if (!managed._promptDone || managed._promptInflight) {
       // _promptInflight covers the synthetic-parent-prompt path that may
       // be holding the brain even when _promptDone has already flipped
--- a/src/core/mcp-client-connections.test.ts
+++ b/src/core/mcp-client-connections.test.ts
@@ -1,0 +1,160 @@
+/**
+ * McpClientManager connection observations.
+ *
+ * Until these existed a server that failed to connect left one console.error
+ * and nothing else: the control plane saw the configured names and reported
+ * "installed" for a server no session had ever reached. These tests pin the
+ * per-server outcome the manager now keeps, and the error classification the
+ * control plane renders, against real HTTP endpoints rather than a mocked SDK
+ * so the shape the SDK actually throws is what gets classified.
+ */
+import { describe, it, expect, afterEach } from "vitest";
+import http from "node:http";
+import type { AddressInfo } from "node:net";
+import { McpClientManager, classifyMcpConnectError } from "./mcp-client.js";
+
+type Responder = (req: http.IncomingMessage, body: string, res: http.ServerResponse) => void;
+
+const servers: http.Server[] = [];
+afterEach(async () => {
+  await Promise.all(servers.splice(0).map((s) => new Promise<void>((r) => { s.closeAllConnections(); s.close(() => r()); })));
+});
+
+async function listen(responder: Responder): Promise<string> {
+  const server = http.createServer((req, res) => {
+    let body = "";
+    req.on("data", (chunk) => { body += chunk; });
+    req.on("end", () => responder(req, body, res));
+  });
+  servers.push(server);
+  await new Promise<void>((r) => server.listen(0, "127.0.0.1", () => r()));
+  return `http://127.0.0.1:${(server.address() as AddressInfo).port}/mcp`;
+}
+
+/** The smallest streamable-http MCP server the SDK client will complete a handshake with. */
+const fakeMcpServer = (tools: string[]): Responder => (req, body, res) => {
+  if (req.method !== "POST") {
+    res.writeHead(405).end();
+    return;
+  }
+  const msg = JSON.parse(body);
+  if (msg.id === undefined) { // notification
+    res.writeHead(202).end();
+    return;
+  }
+  const result = msg.method === "initialize"
+    ? { protocolVersion: msg.params?.protocolVersion ?? "2025-03-26", capabilities: { tools: {} }, serverInfo: { name: "fake", version: "0" } }
+    : msg.method === "tools/list"
+      ? { tools: tools.map((name) => ({ name, inputSchema: { type: "object", properties: {} } })) }
+      : undefined;
+  res.writeHead(200, { "Content-Type": "application/json" });
+  res.end(JSON.stringify(result !== undefined
+    ? { jsonrpc: "2.0", id: msg.id, result }
+    : { jsonrpc: "2.0", id: msg.id, error: { code: -32601, message: "Method not found" } }));
+};
+
+const HTML_404 = `<!DOCTYPE html><html lang="zh-CN"><head><title>Simate</title></head><body><h2>页面不存在</h2></body></html>`;
+
+describe("classifyMcpConnectError", () => {
+  it("recognises an HTML 404 page as not_found with a text/html hint and a stripped message", () => {
+    const err = Object.assign(new Error(`Streamable HTTP error: Error POSTing to endpoint: ${HTML_404}`), { code: 404 });
+    expect(classifyMcpConnectError(err)).toEqual({
+      kind: "not_found", httpStatus: 404, contentType: "text/html", message: "HTML page: Simate",
+    });
+  });
+
+  it("recognises an HTML body with no status as not_mcp", () => {
+    const err = new Error(`Error POSTing to endpoint: ${HTML_404}`);
+    expect(classifyMcpConnectError(err)).toMatchObject({ kind: "not_mcp", contentType: "text/html" });
+    expect(classifyMcpConnectError(err).httpStatus).toBeUndefined();
+  });
+
+  it("maps 401/403 to auth and other statuses to http", () => {
+    expect(classifyMcpConnectError(Object.assign(new Error("Unauthorized"), { code: 401 })).kind).toBe("auth");
+    expect(classifyMcpConnectError(Object.assign(new Error("Forbidden"), { code: 403 })).kind).toBe("auth");
+    expect(classifyMcpConnectError(Object.assign(new Error("Bad Gateway"), { code: 502 }))).toEqual({
+      kind: "http", httpStatus: 502, message: "Bad Gateway",
+    });
+  });
+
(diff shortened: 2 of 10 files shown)


src/core/mcp-client.ts
```diff
@@ -270,16 +421,28 @@ export class McpClientManager {
     const { StreamableHTTPClientTransport } = await import("@modelcontextprotocol/sdk/client/streamableHttp.js");
 
     for (const [serverName, serverConfig] of entries) {
+      const startedAt = Date.now();
+      // Auto-detect transport when not explicitly set: url → streamable-http, command → stdio
+      const cfg = serverConfig as any;
+      const detectedTransport: string = cfg.transport
```
jacoblee-io, line 427: Transport auto-detect (`cfg.transport` access) was hoisted out of the per-server try/catch, so one malformed (null) `mcpServers` entry now rejects `initialize()` entirely and the session loses ALL MCP tools — plus the very observability this PR adds. Config `{"mcpServers": {"bad": null, "good": {...}}}` reaches McpClientManager unvalidated (sync-handlers persists `payload?.mcpServers` verbatim; agent-factory passes it straight in). `cfg.transport` on null throws TypeError BEFORE `try {`; in the base the same access sat inside the try, was caught, and other servers still connected. Now `initialize()` rejects, agent-factory's catch sets `mcpManager = undefined`, so the healthy server's tools are lost too, no observedMcpServers snapshot is recorded, and /api/sync-status has no live manager — the failure mode is worse than pre-PR. Move the detect back inside the per-server try (recording the throw as that server's failed connection).

src/gateway/server.ts
```diff
@@ -2300,6 +2316,61 @@ export async function startRuntime(opts: StartRuntimeOptions): Promise<RuntimeSe
...
+      : typeof server.url === "string" ? "streamable-http" : typeof server.command === "string" ? "stdio" : "";
+    const startedAt = Date.now();
+    const failed = (kind: McpConnectErrorKind, message: string): McpServerConnection => ({
+      name, transport, state: "failed", toolCount: 0, toolNames: [],
+      durationMs: Date.now() - startedAt, observedAt: new Date().toISOString(),
+      error: { kind, message },
+    });
+    if (transport === "stdio") {
+      return { ok: true, probe: failed("invalid_config", "stdio servers run inside the AgentBox and are not probed from the Runtime") };
+    }
```
jacoblee-io, line 2348: mcp.probe dials an operator-supplied URL from the Runtime process with no address screening and echoes back HTTP status, content-type hint, and page `<title>`. A control-plane caller probing `{url: 'http://169.254.169.254/…'}` or an internal admin page gets status/content-type/'HTML page: <title>' leaked — and the Runtime sits in a different network/NetworkPolicy position than the AgentBox pods where MCP dials normally run in K8s mode. The repo's own analogous endpoint (provider-model-listing.ts) applies `isBlockedIpLiteral` with `allowLoopback: true` at the same trust level, keeping metadata ranges blocked "even though only an admin can set base_url" per its own comment — by that documented bar the same guard belongs here. Severity bounded by control-plane-authenticated callers, but the stdio refusal in this same handler shows Runtime-process risk was considered for command execution and not for network reach.

src/core/mcp-client.ts
```diff
@@ -334,14 +510,65 @@ export class McpClientManager {
         }
 
         this.clients.push({ serverName, client, transport });
+        this.connections.push({
+          name: serverName, transport: detectedTransport, state: "connected",
+          toolCount: mcpTools.length,
+          toolNames: mcpTools.map((t: any) => String(t.name)).sort(),
+          durationMs: Date.now() - startedAt,
+          observedAt: new Date().toISOString(),
```
jacoblee-io, line 518: `classifyMcpConnectError` runs the dns/refused/timeout/tls message-regexes over the raw message — which embeds the full HTTP response body (`Error POSTing to endpoint: <body>`, streamableHttp.js:364) — BEFORE consulting the structured `httpStatus`. A Cloudflare 502 body "The connection to the origin web server timed out" matches `/timed? ?out/i` → kind 'timeout'; "SSL handshake failed" in an error page → 'tls' — a permanent misconfiguration (this PR's flagship "page URL pasted as MCP endpoint" case) renders as a transient network error. Related: when `.code` is absent, `httpStatus` is regex-mined from message prose ("upstream returned status: 200" → httpStatus=200), fabricating a status the transport never surfaced. Classify by the structured status first; message regexes only as fallback.

src/core/mcp-client.ts
```diff
@@ -309,15 +472,28 @@ export class McpClientManager {
             break;
           default:
             console.warn(`[mcp-client] Unknown transport for "${serverName}": ${detectedTransport}`);
+            recordFailure({
+              kind: "invalid_config",
```
jacoblee-io, line 476: **Verified by execution.** `compactErrorMessage` collapses ANY message containing a tag-like token (`/<[a-z][^>]*>/i`) to the bare string 'HTML page': "unexpected content type <text/plain>", "expected <object>", and a SOAP fault body all become exactly 'HTML page' (no `<title>` → no suffix), destroying the actual error text an operator needs. And the predicate diverges from the classifier's stricter html test (`looksLikeHtml || /<html[\s>]/i`), so a fragment like `<div>Login required</div>` gets kind 'unknown' + message 'HTML page' — the message asserts HTML while the closed vocabulary says nothing. Gate the collapse on the same looksLikeHtml-strength predicate the classifier uses.

src/core/mcp-client.ts
```diff
@@ -309,15 +472,28 @@ export class McpClientManager {
...
+          continue;
+        }
         console.log(`[mcp-client] Connected to "${serverName}" (${detectedTransport})`);
         const serverImplementationName = client.getServerVersion()?.name;
 
         // Discover tools
         const { tools: mcpTools } = await client.listTools();
+        if (this.disposed) {
+          await this.closeLate(serverName, client, startedAt, detectedTransport);
+          continue;
```
jacoblee-io, line 495: A malformed URL (no scheme) throws `ERR_INVALID_URL` 'Invalid URL' from `new URL(cfg.url)` and classifies as 'unknown', although the vocabulary's own doc comment promises `invalid_config` for 'bad URL'. Operator saves url `example.com/mcp`: the throw lands inside the per-server try, matches no branch (ERR_INVALID_URL is none of the sys-code prefixes, no keyword in 'Invalid URL', no status) → kind 'unknown' on both the box report and mcp.probe — the second of only two examples the invalid_config doc names lands in the wrong kind. `URL.canParse` pre-validation or an ERR_INVALID_URL branch → invalid_config.

src/agentbox/http-server.ts
```diff
@@ -808,11 +817,21 @@ export function createHttpServer(
...
   addRoute("GET", "/api/sync-status", async (_req, res) => {
+    const inventory = readBoxSyncStatus({
+      knowledgeDir: sessionManager.knowledgeDir,
+      knowledgeHandler: perServerKnowledgeHandler,
+    });
+    // Prefer a live session's manager over the last recorded snapshot: an MCP
+    // reload invalidates sessions, so the newest session is the one built from
+    // the current config. Fall back to the snapshot when every session has been
+    // released, and to "no servers field" when this box never initialized MCP.
+    const liveManager = sessionManager.list()
```
jacoblee-io, line 828: GET /api/sync-status can pair NEW `mcp.names` (from disk) with OLD-config `mcp.servers`: `invalidateSessions()` only marks `_invalidated` — a session with an in-flight/detached turn remains in `sessionManager.list()` and is the 'newest' with an mcpManager; a subagent child spawned by that parent shares the parent's old manager but carries its own newer createdAt, so it wins the sort even later; and the observedMcpServers snapshot is never cleared on MCP reload. Until a fresh session is created the payload shows the new config's names beside the previous config's connection outcomes (a removed server 'connected', an added one missing); after all sessions release, the stale never-cleared snapshot serves the old observation indefinitely. Replicas reloading at different moments then differ in identity → `agent.syncStatus consistent:false` flapping that withholds the flat harness/model/tiers for a converged deployment. Skip invalidated sessions in the pick (and clear/re-stamp the snapshot on MCP reload).

src/core/mcp-client.ts
```diff
@@ -261,6 +410,8 @@ export class McpClientManager {
    */
   async initialize(): Promise<void> {
     const entries = Object.entries(this.config.mcpServers);
```
jacoblee-io, line 412: `initialize()` now resets `this.connections` and `this.disposed` — advertising re-initialization — but not `this.clients`/`this.tools`, so a second `initialize()` without an intervening `shutdown()` doubles every client and tool while `getServerConnections()` reports one clean run. No current caller re-initializes (all sites build a fresh manager), but the half-reset is new in this diff and creates the trap: a future retry-after-partial-failure path gets every previously-discovered tool registered twice in `getTools()` and two live clients per server, while the observation list claims a single run. One line: reset clients/tools alongside connections — or throw on re-init of a used manager.

src/gateway/server.ts
```diff
@@ -2300,6 +2316,61 @@ export async function startRuntime(opts: StartRuntimeOptions): Promise<RuntimeSe
...
+    const manager = new McpClientManager({ mcpServers: { [name]: { ...server, transport } as any } });
+    let timer: NodeJS.Timeout | undefined;
+    const timedOut = new Promise<"timeout">((resolve) => { timer = setTimeout(() => resolve("timeout"), timeoutMs); });
+    try {
+      const outcome = await Promise.race([manager.initialize().then(() => "done" as const), timedOut]);
+      if (outcome === "timeout") {
+        return { ok: true, probe: failed("timeout", `no MCP handshake within ${timeoutMs}ms`) };
+      }
+      const probe = manager.getServerConnections().find((item) => item.name === name)
+        ?? failed("unknown", "probe produced no observation");
```
jacoblee-io, line 2360: The probe's timeout path never aborts the in-flight dial — `shutdown()` only flips `disposed` and closes `this.clients`, which is empty mid-connect — so each timed-out probe leaves a socket and pending promise chain lingering in the Runtime until the SDK's internal DEFAULT_REQUEST_TIMEOUT_MSEC (60s) fires (an endpoint that accepts TCP but never answers the initialize POST is exactly what the probe's own test simulates). Repeated operator retries against a broken config hold overlapping sockets in the long-lived Runtime process. Bounded, not unbounded — but plumbing an AbortSignal into the manager (fired in the probe's finally) would tear the dial down at the probe's own deadline.

src/shared/agentbox-sync-status.ts
```diff
@@ -22,8 +22,41 @@ export interface ObservedKnowledgeRepo {
...
+ * about whether it was ever reached. `servers` is the per-server connection
+ * outcome from the process that dialled them. A v4 box reports it whenever a
+ * session has initialized MCP; a v3 box cannot, so a consumer must not read an
+ * absent `servers` as "all connected".
  */
-export const AGENT_SYNC_STATUS_SCHEMA_VERSION = 3;
+export const AGENT_SYNC_STATUS_SCHEMA_VERSION = 4;
+
+/** Closed vocabulary shared with core/mcp-client.ts; see McpConnectErrorKind. */
+export type ObservedMcpErrorKind =
```
jacoblee-io, line 36: `ObservedMcpErrorKind` hand-duplicates `McpConnectErrorKind`'s 11-member closed vocabulary instead of importing it (src/shared already type-imports from src/core in 4 files), with no test pinning the two lists in sync. Adding a new kind in core (e.g. 'oauth') compiles cleanly while `normalizeObservedMcp`'s OBSERVED_MCP_ERROR_KINDS set silently rewrites it to 'unknown' at the gateway boundary — the classification the whole PR exists to surface is quietly discarded and nothing fails. `export type ObservedMcpErrorKind = McpConnectErrorKind` and derive the runtime Set from one list.

src/core/mcp-client-connections.test.ts
```diff
@@ -0,0 +1,217 @@
...
+
+const servers: http.Server[] = [];
+afterEach(async () => {
+  await Promise.all(servers.splice(0).map((s) => new Promise<void>((r) => { s.closeAllConnections(); s.close(() => r()); })));
+});
+
+async function listen(responder: Responder): Promise<string> {
+  const server = http.createServer((req, res) => {
+    let body = "";
+    req.on("data", (chunk) => { body += chunk; });
```
jacoblee-io, line 26: Both new test files accumulate HTTP request bodies with `body += chunk` and no `setEncoding` — the exact anti-pattern the repo CLAUDE.md stream-decoding rule pins ("ANY new reader that accumulates a stream into a string… `body += chunk` decodes each fragment independently, so a multibyte character straddling a chunk boundary becomes two U+FFFD"; the rule even prescribes behavior for test doubles). Same at server-mcp-probe.test.ts:29. These fixtures `JSON.parse` bodies from the real MCP SDK; a multibyte value split across chunks corrupts, JSON.parse throws, the fake server never responds, and the test times out flakily. `req.setEncoding('utf8')` in both copies — or extract the duplicated ~40-line fakeMcpServer helper once and fix it there.

jacoblee-io (approved):
Reviewed the connection-outcome reporting, the classifier, mcp.probe, and the sync-status plumbing. Approving to unblock; 10 findings inline. The two I'd fix before merge: the transport auto-detect hoisted out of the per-server try/catch (one malformed entry now kills ALL MCP tools — a regression that defeats the observability this PR adds), and mcp.probe dialing operator-supplied URLs from the Runtime with no address screening while the repo's own analogous endpoint (provider-model-listing) guards the same trust level.

Cut by the 10-finding cap (low severity): connections sorted per poll instead of once, transport-detect logic duplicated between initialize and probe, the 'failed' literal tripled, and the ~40-line fake MCP server helper copy-pasted between the two new test files. One candidate self-refuted during verification: the claimed unhandled-rejection on the probe timeout race — `Promise.race` keeps handlers attached to the losing promise and `initialize()` swallows per-server errors anyway.

PR #569, opened 2026-09-08
fix(capability): never roll a live kbc box mid-run; stamp watchdog closes

LikiosSedo:
## Problem

A capability run's box is single-use and keyed by the run id, but `capabilityBoxClient` routed every (re)attach through `agentBoxManager` acquisition. That path compares the running box against the profile's current image and **rolls** it when stale (`getOrCreateK8s` rollReason → `markStaleBoxesDraining`): the old pod is marked draining, a replacement is spawned, and the relay attaches to the empty replacement. For a kbc box this abandons the compile it is executing — the old pod is deleted at `DRAIN_DEADLINE_MS` and its pending candidate commit is lost (`_flush_on_shutdown` only drains to a live relay).

**2026-09-07 incident (prod, kb `660ed03c…`)**: the v0.3.17 release restarted the runtime at 13:41Z; boot recovery (`onAdopt`) re-attached a live compile run through acquisition; the box was rolled for its image (the release changed nothing under `kbc/`); the run idled on the new pod; the watchdog closed it as `done` at 15:42Z (`idleTtlMs` = 2h). Sicore recorded the operation as failed, but the page showed "documents ready, generate" with no trace of it.

## Fix

- `capabilityBoxClient` reuses the run's **live** box as-is (`getAsync`) and only spawns when no box exists. The rule lives in `acquireCapabilityBox` (`src/gateway/capability/box-acquire.ts`) so it is unit-tested against a fake manager. Image changes still reach kbc boxes the way they always did: every run is a new pod.
- The watchdog stamps `close_reason: "idle_timeout"` into the run checkpoint when **it** writes the `done`, so a consumer whose operation is still open can tell an undriven run from one the box finished. A `done` emitted by the box carries no `close_reason`. Sicore side consumes it in the companion MR (error message on the failed operation + a visible `compile_failed` phase).

## Verification

- `vitest run src/gateway/capability`: 95 passed (3 new in `box-acquire.test.ts`, 1 new + 1 extended in `run-manager.test.ts`). `capability-metrics.test.ts` fails to load locally only because my linked `node_modules` lacks `prom-client`; unrelated to this change.
- `tsc --noEmit`: no errors in touched files (remaining errors are missing local deps).
- Not yet exercised against a real rollout; the reproduction requires a kbc run in flight during a runtime restart.

### DIFF
--- a/src/gateway/capability/box-acquire.ts
+++ b/src/gateway/capability/box-acquire.ts
@@ -1,0 +1,70 @@
+/**
+ * Box acquisition for capability runs (kb-compile and friends).
+ *
+ * A capability run's box is SINGLE-USE and keyed by the runId: the pod's identity
+ * is the run. That makes it fundamentally different from a chat agent's pool,
+ * and the difference matters on one path — acquisition. `getOrCreate*` compares
+ * a running box against the image the profile would spawn today and ROLLS a box
+ * it judges stale (markStaleBoxesDraining / getOrCreateK8s rollReason): the old
+ * pod is marked draining, a replacement comes up, and the caller is handed the
+ * replacement. For a chat pool that is a rolling upgrade. For a kbc box it is
+ * abandoning the run it is executing: the relay attaches to an empty replacement,
+ * the old pod is deleted at the drain deadline, and whatever the compile still
+ * had to emit — the candidate commit included — dies with it.
+ *
+ * 2026-09-07 incident: a siclaw release restarted the runtime; boot recovery
+ * re-attached a live compile run through acquisition, the box was rolled for its
+ * (unchanged) image, the run idled on the new pod and the watchdog closed it as
+ * "done" two hours later. The knowledge base showed "ready to generate" with no
+ * trace of the failure.
+ *
+ * Rule: a LIVE box for this run is reused as-is, regardless of image, cert
+ * freshness or anything else acquisition would roll on. Only a missing/dead box
+ * goes through spawn. Image changes reach kbc boxes the way they always did —
+ * every run is a new pod.
+ */
+
+export interface CapabilityBoxHandleLike {
+  endpoint: string;
+}
+
+export interface CapabilityBoxAcquirer {
+  /** The run's live box, if one is running right now. Never spawns, never rolls. */
+  getAsync(agentId: string, profile?: string): Promise<CapabilityBoxHandleLike | undefined>;
+  getOrCreateWithDisposition?(
+    agentId: string,
+    config: { profile: string; orgId?: string },
+  ): Promise<{ handle: CapabilityBoxHandleLike; created: boolean }>;
+  getOrCreate(
+    agentId: string,
+    config: { profile: string; orgId?: string },
+  ): Promise<CapabilityBoxHandleLike>;
+}
+
+export interface CapabilityBoxAcquisition {
+  endpoint: string;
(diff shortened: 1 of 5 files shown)


src/gateway/capability/box-acquire.ts
```diff
@@ -0,0 +1,70 @@
...
+ * the old pod is deleted at the drain deadline, and whatever the compile still
+ * had to emit — the candidate commit included — dies with it.
+ *
+ * 2026-09-07 incident: a siclaw release restarted the runtime; boot recovery
+ * re-attached a live compile run through acquisition, the box was rolled for its
+ * (unchanged) image, the run idled on the new pod and the watchdog closed it as
+ * "done" two hours later. The knowledge base showed "ready to generate" with no
+ * trace of the failure.
+ *
+ * Rule: a LIVE box for this run is reused as-is, regardless of image, cert
```
jacoblee-io, line 21: **low** — this rule is stated more broadly than the guard implements it: cert freshness stops mattering only while the pod is judged `running`.

`getAsync` requires `info.status === "running"`, and `mapPodStatus` (`k8s-spawner.ts:1376`) maps **phase=Running with Ready=False** to `starting`. For `kb-compile` the readiness probe is an in-container **exec** probe (`healthProbeFor`, `k8s-spawner.ts:186`) that starts a python interpreter every 2s with `timeoutSeconds: 3` and the default failure threshold of 3 — so it goes unready after roughly six seconds of the box being too busy to answer, which is the state a compile produces.

From there the run falls through the very path this file exists to avoid:

- `getAsync` → `undefined`, so `acquireCapabilityBox` spawns instead of reusing;
- `getOrCreateK8s`: `rollReason` requires `status === "running"`, so no roll (good), but warm reuse requires it too, so the call reaches `spawner.spawn()`;
- `spawn()` reads the existing pod — Running, same profile, not terminating, no `recreate` — and the only remaining test is the certificate. Fine cert ⇒ `waitForPodReady` and reuse, which is the right outcome. But `certificateNeedsRenewal` ⇒ `deleteNamespacedPod` on the pod that is running the compile.

Narrow (it needs an unready window and a certificate already a third through its life, so a long compile), and still far better than before, when every attach went through acquisition. But it is the same failure shape as the incident, reached without anything rolling.

Either gate on the pod existing and not being terminating rather than on readiness, or treat `starting` here as "wait and retry" instead of "no box" — the current mapping sends a live-but-busy box into the spawn path.

jacoblee-io (approved):
Verified on `f0457345`.

The fix holds where it matters: `markStaleBoxesDraining` has exactly one caller in the tree (`manager.ts:463`, inside `getOrCreatePooled`), and the only way in is `getOrCreateK8s`'s `rollReason` branch — which `acquireCapabilityBox` now bypasses for a live box. The `close_reason` half is clean too: it is stamped only in the non-failure arm of `reapStale`, a box-emitted `done` carries none, and both directions are pinned by tests.

On the verification gaps in the description — they are your local environment, not the change. With a complete `node_modules`: `tsc --noEmit` is silent, `vitest run src/gateway/capability` is 8 files / 96 passed (including the `capability-metrics.test.ts` that would not load for you), and the full suite is 310 files / 6819 passed / 2 skipped. Nothing to chase there.

Approving. One residual inline, low and narrow — the guard is readiness rather than liveness, and kb-compile's readiness probe is likeliest to fail exactly while the box is busy.

PR #574, opened 2026-09-10
fix: preserve source citations through final answer delivery

LikiosSedo:
## Summary

Knowledge answers could show duplicate references after a conversation relay, lose references after intermediate commentary, or persist citation attribution on the commentary row instead of the final answer.

The destination runtime owns citation rendering for relayed events. Direct consumers retain the registered source union for each independently delivered answer and skip known progress messages. Citation metadata follows the text item containing the references; the model-call billing envelope remains on one row. The shared renderer recognizes identical complete footers under either supported heading, including mixed-language answers and source titles.

## Test Plan

- [x] Reproduced the duplicate footer, mixed-language footer, missing final sources, and misplaced native-item attribution before their respective fixes.
- [x] 367 focused tests passed across citation rendering, SSE consumption, Feishu collection, and native-item persistence.
- [x] `npm test`: 328 files passed; 7,064 tests passed, 2 skipped.
- [x] `npx tsc --noEmit` and `npm run build` passed.
- [ ] Live deployment and Feishu acceptance.

### DIFF
--- a/src/gateway/sse-consumer.ts
+++ b/src/gateway/sse-consumer.ts
@@ -13,7 +13,7 @@
 
 import { randomUUID } from "node:crypto";
 import { AssistantItemStream } from "./assistant-item-stream.js";
-import type { AssistantItem } from "../shared/assistant-items.js";
+import { assistantTextBlocks, type AssistantItem } from "../shared/assistant-items.js";
 import type { ChatMessageMetadata } from "../shared/message-kinds.js";
 import { ErrorCodes } from "../lib/error-envelope.js";
 import { AgentBoxClient } from "./agentbox/client.js";
@@ -436,18 +436,8 @@ export async function consumeAgentSse(opts: ConsumeAgentSseOptions): Promise<Sse
   let isRoutingTurn = false;
   let routingCommitted = false;
   let pendingKnowledgeSources: unknown = null;
-  // Citations rendered into the assistant row being committed — persisted on
-  // that row's metadata as `knowledge_citations`, so feedback on the row can be
-  // attributed to the cited repos/pages. Exactly the delta appended to the text.
-  let pendingRowCitations: KnowledgeSourceCitation[] = [];
-  // URLs already rendered into an assistant message THIS TURN. Consumers ASSIGN
-  // (not merge) the knowledge_sources event, and a source must appear exactly
-  // once per turn. Nulling pending after the first ended message lost the
-  // references when the model narrated or re-cited before its final answer
-  // (zero-fresh re-cite emits nothing), and duplicated them across bubbles when
-  // it cited twice. Instead keep pending and append only the not-yet-rendered
-  // delta to each message; reset at the user-message turn boundary.
-  const renderedKnowledgeSourceUrls = new Set<string>();
+  // Keep the registered union until the user turn ends. Each independently
+  // delivered answer needs its sources even if an earlier message cited them.
   const pendingAssistantOps: Array<() => Promise<void>> = [];
   const pendingErrorOps: Array<() => Promise<void>> = [];
   const flushOps = async (ops: Array<() => Promise<void>>) => {
@@ -507,8 +497,6 @@ export async function consumeAgentSse(opts: ConsumeAgentSseOptions): Promise<Sse
     pendingErrorOps.length = 0;
     pendingFailedLlmCalls = [];
     pendingKnowledgeSources = null;
-    renderedKnowledgeSourceUrls.clear();
-    pendingRowCitations = [];
     pendingStreamError = null;
     errorMessage = "";
   };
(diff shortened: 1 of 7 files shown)


src/shared/knowledge-citations.ts
```diff
@@ -101,5 +101,9 @@ export function appendKnowledgeSourceCitations(text: string, value: unknown): st
     const destination = source.url.replaceAll("(", "%28").replaceAll(")", "%29");
     return `- [${source.title.replace(/[\[\]]/g, "")}](${destination})`;
   });
-  return `${text.trimEnd()}\n\n### ${heading}\n\n${lines.join("\n")}`;
+  const footer = `\n\n### ${heading}\n\n${lines.join("\n")}`;
+  // A relayed or pre-rendered answer can already carry this exact footer.
+  // Match the complete generated block, not arbitrary URLs in answer prose.
+  if (text.trimEnd().endsWith(footer)) return text;
```
jacoblee-io: [P2] Preserve footer idempotence when source titles change language detection

The heading is inferred from the entire input text, including any previously appended footer. With an English answer and a Chinese source title, the first call produces `### Original sources`; the second call detects the Chinese title and builds `### 参考原文`, so this suffix check fails and appends the same sources again.

Reproduced against this commit:
```ts
const sources = [{ title: "GPU 故障排查手册", url: "https://example.com/runbook" }];
const once = appendKnowledgeSourceCitations("All nodes are healthy.", sources);
const twice = appendKnowledgeSourceCitations(once, sources);
// twice contains both headings and the same link twice
```

The ConversationClient guard fixes the main relay path, but the new renderer-level deduplication still misses this pre-rendered/raw-stream case. Please recognize an identical complete footer under either supported heading, or determine the language from the answer excluding an existing footer. Add an English-answer/Chinese-title regression test; the current English and Chinese cases both use English source titles.

src/gateway/sse-consumer.ts
```diff
@@ -1134,9 +1137,11 @@ export async function consumeAgentSse(opts: ConsumeAgentSseOptions): Promise<Sse
                     if (lastText) lastText.text += cited.slice(base.length);
                     message.content = parts;
                   } else message.content = cited;
-                  for (const source of freshSources) renderedKnowledgeSourceUrls.add(source.url);
-                  pendingRowCitations = freshSources;
                 }
+                // An identical footer may already exist; attribution still
+                // belongs to this row even when rendering is a no-op.
+                for (const source of freshSources) renderedKnowledgeSourceUrls.add(source.url);
```
jacoblee-io: [P2] Preserve sources in the independently delivered final answer

This turn-wide rendered set still consumes a source when it is appended to commentary. Reproduced on this head with `knowledge_sources(A) -> commentary + toolUse -> tool result -> knowledge_sources(A) -> final_answer`: the final `resultText` is only `All nodes are healthy.`, without A's URL. The same sequence fails against `collectChannelResponse` because it also marks A rendered on the progress message and returns only the last assistant text. Scheduled conversation tasks use `resultText`, and Lark/Dingtalk use the channel collector, so preserving A somewhere in the transcript does not preserve it in the delivered answer. The ConversationClient guard prevents another append downstream; it cannot restore sources missing from the owning runtime's final answer.

Please separate deduplication of a relayed/rendered message from final-answer source attribution, so the final answer retains its applicable references even when commentary already displayed them. Add assertions against the final `resultText` and channel `text`, not only the concatenation of all messages. This behavior also exists on main; it is an additional gap in the citation flow rather than a regression introduced by the latest language fix.

src/gateway/sse-consumer.ts
```diff
@@ -1134,9 +1137,11 @@ export async function consumeAgentSse(opts: ConsumeAgentSseOptions): Promise<Sse
...
+                pendingRowCitations = freshSources;
```
jacoblee-io: [P2] Attach citation metadata to the item that contains the cited answer

Moving attribution outside the renderer-change condition handles an existing footer, but `pendingRowCitations` still goes into `headMetadata` (line 1189), which is applied only to `index === 0` (line 1198). For one native assistant message containing a commentary text item followed by a final-answer text item, the footer is appended to the last text item while its `knowledge_citations` metadata is persisted on the first, commentary row. Reproduced on this head: the final `resultText` contains the source URL, but the final persisted row has `metadata.knowledge_citations === undefined`. Feedback on that answer therefore lacks its knowledge-repository/page attribution.

Please associate citations with the actual answer item when splitting the message into rows, independently of the call-level billing envelope. Copying the whole envelope to every row would double-count usage. Add a commentary-plus-final native-item regression checking both the footer and the final row's `repo_ids`/pages. This is also reproducible on main and remains unresolved by this PR.

jacoblee-io (commented):
Rechecked 507b3d31: the mixed-language footer idempotence issue is fixed, and all 360 existing tests in the citation renderer, SSE consumer, and Lark suites pass. Two additional pre-existing boundary issues remain in the citation flow; three isolated regression tests reproduce them on this head (SSE final delivery, channel final delivery, and item-level persistence). Details inline below.

jacoblee-io (approved):

PR #575, opened 2026-09-10
fix(kbc): keep evidence references consistent with frozen sources

LikiosSedo:
Draft source IDs could be corrected from the frozen manifest while authored evidence markers kept old IDs. Flow-style source rows also skipped identity stamping. Recreated source identities on otherwise unchanged Raw paths could remain on pages outside an incremental changeset.

Rebind declarations and authored markers across the draft when the Raw resource has an unambiguous frozen identity. Run this after restoring unauthorized body edits, and advance only the identity-normalized baseline for untouched pages. Section attribution remains scoped; identity repair does not broaden model edit permissions or migrate inherited format debt. Later repair turns retain the corrected identities while still restoring out-of-scope body edits.

Check Raw evidence against the frozen manifest during self-check and include manifest changes in the check key, so internally consistent but untrusted IDs enter repair before publication.

Recognize external URI schemes before path normalization and use the same managed-resource boundary for coverage, identity stamping, marker remapping, and frozen-ID validation. Preserve external/package resource declarations and mixed-page attribution. Flow mapping insertion reads YAML tokens so trailing commas followed by comments stay valid, and rewritten frontmatter must parse before any page is saved.

Validation:
- CI script entrypoints (`test_compile_box.py`, `test_selfcheck.py`, `test_incremental.py`) pass, with the new regressions wired into their runners.
- Compile-box, self-check and incremental suites: 221 passed; two failures reproduced unchanged on clean main (212 passed): charset corruption line numbering and self-check wiring idempotency.
- Added regression coverage for external and package resources, Raw-name collisions, mixed attribution, commented flow mappings, and atomic rejection of unsafe anchor rewrites.
- The incremental regression covers recreated identity on an untouched page, simultaneous unauthorized body edits, inherited format debt, and a subsequent repair turn.
- Cross-repository smoke: feed the actual consumer frozen manifest into the compiler, process the scoped incremental turn, synchronize its generated pages, and verify readiness plus successful publication through the consumer service.

These checks establish the corrected code paths; they do not reconstruct a specific production draft history.

### DIFF
--- a/kbc/platform/pod/selfcheck.py
+++ b/kbc/platform/pod/selfcheck.py
@@ -23,6 +23,7 @@ import posixpath
 import re
 import unicodedata
 import uuid
+from collections.abc import Collection
 from datetime import date, datetime, timezone
 from functools import lru_cache
 from pathlib import Path
@@ -220,14 +221,27 @@ def _explicit_external_source(resource: str) -> bool:
     managed-source namespace. Bare values remain a legacy compatibility case
     when (and only when) they exactly match the current Raw inventory.
     """
-    value = posixpath.normpath(resource.strip().replace("\\", "/"))
+    value = resource.strip().replace("\\", "/")
+    # Filesystem normalization collapses URI separators (https:// -> https:/).
+    # Recognize schemes before interpreting the value as a path.
+    if re.match(r"^[A-Za-z][A-Za-z0-9+.-]*:", value):
+        return True
+    value = posixpath.normpath(value)
     lowered = value.lower()
     return (
-        "://" in value
-        or lowered.startswith(("mailto:", "urn:", "doi:"))
-        or value.startswith(("/", "../"))
+        value.startswith(("/", "../"))
         or lowered.startswith("references/")
     )
+
+
+def _managed_source_entry(resource: str, managed_sources: Collection[str] | None) -> str:
+    """Resolve explicit Raw paths or exact legacy inventory matches only."""
+    if _explicit_external_source(resource):
+        return ""
+    entry = _norm_source_entry(resource)
+    if _explicit_managed_source(resource) or (managed_sources is not None and entry in managed_sources):
+        return entry
+    return ""
 
 
 def parse_okf_sources(
@@ -256,20 +270,12 @@ def parse_okf_sources(
             resource = item.get("resource")
             if not isinstance(resource, str):
                 continue
-            if _explicit_managed_source(resource):
-                entry = _norm_source_entry(resource)
-                if entry:
-                    resources.append(entry)
-                continue
-            if _explicit_external_source(resource):
-                continue
             # Clean v0.2 packages use an explicit managed namespace. Preserve
             # the old bare-path dialect only when it exactly identifies a Raw
             # inventory entry; every other value is an OKF scope descriptor.
-            if managed_sources is not None:
-                entry = _norm_source_entry(resource)
-                if entry in managed_sources:
-                    resources.append(entry)
+            entry = _managed_source_entry(resource, managed_sources)
+            if entry:
+                resources.append(entry)
     return resources, derived, has_key
 
 
@@ -2051,6 +2057,7 @@ def _stamp_source_ids_text(
     frontmatter = "".join(lines[1:end])
     try:
         root = yaml.compose(frontmatter, Loader=yaml.SafeLoader)
+        tokens = list(yaml.scan(frontmatter, Loader=yaml.SafeLoader))
     except yaml.YAMLError:
         return None
     if not isinstance(root, yaml.MappingNode):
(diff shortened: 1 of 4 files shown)


kbc/platform/pod/selfcheck.py
```diff
@@ -2114,6 +2118,74 @@ def _stamp_source_ids_text(
...
+            if not isinstance(source, dict):
+                continue
+            source_id, resource = source.get("id"), source.get("resource")
+            if not isinstance(source_id, str) or not isinstance(resource, str) or source_id.strip() not in used:
+                continue
+            entry = _norm_source_entry(resource)
+            if _explicit_external_source(resource):
+                continue  # External citation declarations are inspected by the package boundary.
+            if manifest_ids.get(entry) != source_id.strip():
+                violations.append({"page": rel, "kind": "evidence_source_mapping",
```
jacoblee-io: [P2] Preserve external URL evidence when checking frozen Raw identities

A valid page declaring `resource: https://example.com/manual` with `id: external`, and referencing that ID from an evidence marker, now receives `evidence_source_mapping`. `_explicit_external_source()` calls `posixpath.normpath()` before checking for `://`, which turns `https://example.com/manual` into `https:/example.com/manual`; the external-source exemption therefore fails, and this new check requires the URL to exist in the Raw manifest. The same page passes Layer-1 on the base. This sends valid external citations into unnecessary repair and can leave them unconverged.

Please recognize URI schemes before filesystem normalization and restrict frozen-ID validation to resources that actually belong to the managed Raw namespace. The existing resource contract also allows package-relative paths and scope descriptors outside Raw.

kbc/platform/pod/selfcheck.py
```diff
@@ -2092,7 +2092,11 @@ def _stamp_source_ids_text(
             # row tells the model what to fix). Leave the row out of the map.
             continue
         ids[entry] = wanted
-        if id_pair is None:
+        if id_pair is None and item.flow_style:
+            insert_at = item.end_mark.index - 1
+            prefix = " " if frontmatter[:insert_at].rstrip().endswith(",") else ", "
+            edits.append((insert_at, insert_at, f"{prefix}id: {json.dumps(wanted)}"))
```
jacoblee-io, line 2109: [P2] Handle comments after a flow mapping's trailing comma before inserting an ID

This valid frontmatter row is corrupted by the new insertion branch when `a.md` has a frozen identity:

```yaml
sources: [{resource: raw/a.md, # Original document
}]
```

Because the text before `}` ends with the comment rather than the existing comma, `rstrip().endswith(",")` is false and another comma is inserted before `id`. The resulting YAML fails to parse with `expected the node content, but found ','`. `normalize_evidence_source_ids()` writes these invalid bytes without validating them, and its global scope means an untouched page outside the incremental changeset can also be damaged. The base leaves this valid row intact.

Please determine the separator from YAML tokens, or leave unsupported layouts unchanged, and validate the rewritten frontmatter before committing it.

jacoblee-io (commented):
Found two reproducible regressions in source identity normalization and validation. Both minimal reproductions pass on base 403395ad and fail on this head. The three existing script entrypoints pass; pytest reports 218 passed and 2 failed, with the same two failures reproduced on the base (212 passed).

kbc/platform/pod/selfcheck.py
jacoblee-io: Verified fixed in 3b8dcc5. The original HTTPS evidence reproduction now passes. I also checked the added external/package-resource, Raw-name collision, and mixed-attribution regressions; the managed-resource boundary preserves the external declarations.

kbc/platform/pod/selfcheck.py
jacoblee-io: Verified fixed in 3b8dcc5. The original commented trailing-comma reproduction now passes. An additional 40-case check covering flow layouts, nested values, comments, and LF/CRLF preserved the frontmatter semantics and body. The unsafe-anchor regression also confirms that invalid rewrites leave the original page intact.

jacoblee-io (approved):
Re-reviewed 3b8dcc5744476ac325b3427b2729c2aec8522d9a. Both previous findings are fixed, and I found no new blocking issues in the revised source-identity and incremental-guard paths.

Validation:
- Both original minimal reproductions now pass.
- The compile-box, self-check, incremental, OKF-package, and batching script entrypoints pass.
- The compile-box, self-check, and incremental pytest suites report 221 passed and 2 failed. The failures remain the previously reproduced baseline issues in test_charset_corruption_detection and test_wiring; the KBC base files are unchanged from that comparison.
- An additional 40-case YAML check preserved frontmatter semantics, prose, and idempotency across flow layouts, comments, nested values, and LF/CRLF.
- All current PR CI checks pass.

Approved for this revision.

PR #581, opened 2026-09-10
feat(trace): add bilingual interactive request timelines

jacoblee-io:
## Summary

Add reusable interactive request timelines to `render_chart` for HTTP services, gateway retries, agent/tool executions and other observed spans. Callers and diagnostic skills supply evidence through a shared span contract. The compact Portal card stays beside the answer and survives history reload, with English/Chinese/automatic language selection, responsive details, zoom/selection and full PNG export.

### Problem

Text-only timings obscure overlapping spans, retries and missing observations. Copying chart JSON into an answer also makes rendering depend on the model repeating the payload.

### Solution

- Return a versioned attachment and short summary; preserve existing pie/bar/line and Mermaid fence/PNG behavior.
- Validate bounded span data, subtract UTC timestamps before numeric conversion, preserve actual parents and unknown ends, and keep coverage visible.
- Persist redacted tool metadata across Web, Lark, delegated and synthetic turns. Render completed attachments automatically and suppress matching repeated fences.
- Forward direct Lark PNG results and request an optional normal authenticated host link. Standalone Portal returns no link. Add a stateless Portal export surface; no new production dependencies, database schema or access tokens.
- Synchronize shared frontend files through an explicit host component directory, independent of product names or repository layout.

## Test Plan

- [x] Generic service-request fixture and renderer contracts: 40 targeted tests passed after the example/documentation update.
- [x] Runtime TypeScript/build and create-chart build with Pi SDK 0.85.1, based on `main` at `e0b7626f`.
- [x] Full `npm test -- --maxWorkers=2`: 7,101 passed, 2 skipped.
- [x] OCR backend: 15 Python unit tests passed.
- [x] Portal: 268 tests and TypeScript/Vite build.
- [x] Hosted Lark link collection uses remote persisted tool IDs without duplicate writes or duplicate link requests.
- [x] Normalization and final visual-ID byte limits, including an exact-limit attachment round trip through the Portal reader.
- [x] Portal and companion-host DOM regressions cover initial visual focus, asynchronous history, later answers, new user turns and session changes; Portal also covers StrictMode/unmount cleanup. Portal adds jsdom only as a development test dependency.
- [x] Shared contract and interaction synchronization check against the companion host. The generic `--host-dir` interface also passed temporary-host copy/check, drift rejection for all seven shared files and seven invalid-argument checks.
- [x] Current-worktree Portal browser verification with synthetic history: initial visual focus, history arrival, later answers preserving position and new questions resuming auto-follow.
- [x] Prior release-base synthetic-data browser validation: both languages, light/dark, desktop/mobile layouts, compact card, selection/zoom, history, read-only behavior and all-span PNG.

This head has not been redeployed. Companion-host transcript behavior is covered by real chat-component DOM regressions; the new browser run covers Portal only. Direct live Lark delivery is not yet verified. Deploy readers/exporters before Runtime/MCP and skills. Background notifications remain text-only; older visuals require loading their history page.

## Architecture Checklist

- [x] No new shell execution, database migration or core prompt changes.
- [x] AgentBox uses the existing Runtime persistence path; it does not import host persistence.
- [x] Hosted links retain existing login/session authorization and never grant access.

## General Checklist

- [x] One logical request-timeline feature, developed in an isolated worktree.
- [x] Conventional Commit; no deployment-specific infrastructure or credentials included.

### DIFF
--- a/docs/design/chart-rendering.md
+++ b/docs/design/chart-rendering.md
@@ -14,8 +14,9 @@ description: "Contract between MCP chart tools and the Portal frontend renderer.
 
 ## How It Works
 
-The frontend recognises chart output through a single convention: a fenced
-Markdown code block with the language tag `chart`.
+The frontend recognises legacy pie/bar/line output through a fenced Markdown
+block tagged `chart`. Request timelines additionally use versioned structured
+tool attachments, rendered directly from persisted message metadata.
 
 ```
 ```chart
@@ -51,7 +52,8 @@ ChartRenderer  (portal-web/src/components/chat/ChartRenderer.tsx)
   └── wrapped in React.memo — skips re-render when spec is unchanged
         ├── spec.type === "pie"  → renderPie
         ├── spec.type === "bar"  → renderBar
-        └── spec.type === "line" → renderLine
+        ├── spec.type === "line" → renderLine
+        └── spec.type === "waterfall" → TraceTimelineRenderer
 ```
 
 ## Mermaid Diagram Rendering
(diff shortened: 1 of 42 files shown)


LikiosSedo:
Reviewed commit `821a1e080e2c8096a66aa731df77aa8562363e12`. Three P2 issues need follow-up:

1. **Hosted Lark conversations never collect visual links** — [`src/gateway/channels/lark.ts:3406`](https://github.com/scitix/siclaw/blob/821a1e080e2c8096a66aa731df77aa8562363e12/src/gateway/channels/lark.ts#L3406)

   When the host supports conversations, `processQueuedLarkMessage` deliberately calls `collectChannelResponse` without `persist`, because the destination Runtime already owns persistence. The new `getVisualLink` loop is inside `if (persist)`, so it is never reached on that path. A reproduction with `conversationEvents: true`, a completed visual attachment and a persisted tool message ID returns the final answer but `visualLinks: []`. Please decouple link collection from local persistence and use the persisted message identity supplied by the remote conversation event.

2. **Initial transcript auto-scroll overrides visual deep-link focus** — [`portal-web/src/components/chat/TraceTimelineRenderer.tsx:68`](https://github.com/scitix/siclaw/blob/821a1e080e2c8096a66aa731df77aa8562363e12/portal-web/src/components/chat/TraceTimelineRenderer.tsx#L68)

   The renderer schedules `scrollIntoView` for the requested visual, while the parent `PilotArea` schedules its initial scroll to the latest message afterward. If the attachment is already loaded and later answers follow it, the card expands but the transcript scrolls away from it. The equivalent host transcript component reproduces the callback order `visual -> latest`; Portal has the same child-focus/parent-auto-follow pattern. Please coordinate visual navigation with transcript scrolling so a requested visual takes precedence. This concerns an already loaded attachment, not the documented limitation for older history pages.

3. **Accepted input can produce an attachment rejected by the reader** — [`mcp/create-chart/src/waterfall-spec.ts:94`](https://github.com/scitix/siclaw/blob/821a1e080e2c8096a66aa731df77aa8562363e12/mcp/create-chart/src/waterfall-spec.ts#L94)

   The byte limit is checked before normalization adds `status`, `layer` and `schema_version`; the handler adds `visual_id` afterward. A valid 200-span input with eight bounded evidence references per span is 256,516 bytes and passes validation, but the resulting spec is 264,150 bytes. `traceAttachments` invokes the same normalizer and rejects it against the 262,144-byte limit, returning `spec: null`. Please validate the final emitted spec, including its visual ID, so a successful tool response is readable by supported clients.

Validation: 384 existing Runtime/persistence/Lark tests and 35 MCP tests passed. Additional reproductions confirmed the missing hosted link and the normalization/reader size mismatch; the host transcript DOM reproduction confirmed the scroll ordering. No live browser or Lark end-to-end delivery was exercised.

nightmeng (dismissed):

LikiosSedo (approved):
Reviewed structured timeline attachments, persistence, rendering, and channel compatibility. No blocking findings. Focused tests and current CI checks pass.

PR #582, opened 2026-09-10
feat(kbc): add selectable Pi execution and local authoring kit

LikiosSedo:
## Summary

Knowledge compilation can use Pi or Claude Agent SDK with the same host-owned file/domain tools, quality checks, artifact checkpoints, and execution observations. Pi reuses the ordinary agent's execution core and model compatibility translation, including OpenAI Chat Completions and Responses. Compiler engine/model configuration is independent and directly saved by the paired control plane; it does not use the interactive Agent Type release workflow. Existing frozen configurations remain recoverable. Autonomous PK testing and repair is off by default and can be explicitly enabled.

Large text inputs have bounded, complete UTF-8 pagination. The local authoring kit exports the same checker and standards for external authoring tools.

## Review fixes

- Failed owner turns reconcile their scoped changes and clear pending operation state before returning to idle, preserving subsequent unrelated edits.
- Compiler context protection keeps PDF images intact and fails explicitly on overflow instead of silently replacing visual evidence with text.
- Fresh read-only sessions receive their allowed source roots. Claude content blocks sharing one message ID are assembled before structured output is consumed, preserving full JSON and accurate call counts.

- Empty legacy boxes are replaced for either supported engine only after configuration and source revision checks; current Claude boxes are recognized as compatible.
- The non-root image smoke checks that both supported SDKs are installed and retains the real Pi worker startup and HTTP health checks.

## Validation

- Real Pi and Claude SDK process tests cover host tools, results, provider failures, cancellation, and metadata observations.
- Real-worker regressions reproduce and fix both review comments.
- API-key Responses fixtures preserve reasoning and all file-tool results.
- Go configuration, continuation, migration and vet checks pass in the paired control plane. UI interaction tests cover both engines and incompatible model replacement.
- Native production-container suite before the PK-default follow-up: 60 passed, including PDF tools, both SDK processes and both review regressions. The message-assembly follow-up suite passed 39 cases.
- Real-provider acceptance completed through both SDKs on a small synthetic corpus: compilation, preserved engine snapshots, persisted observations and 8/8 final PK questions. The fresh Claude run passed its first full PK round.
- PK-default follow-up: 93 Python checks passed, including default-off gating and explicit enablement; two existing self-check failures were reproduced on the unchanged prior commit. Seven form interactions, Go capability/knowledge/router/MCP checks, vet, TypeScript and lint passed.
- Latest follow-up: 107 Runtime capability tests and 61 native-container tests passed. The non-root image permission, real worker startup and HTTP health smoke passed. All five PR checks passed at the pushed revision.
- Matching API/Web and Runtime/KBC/AgentBox images are deployed by immutable digest to the test environment. Healthy rollout, actual image IDs and saved PK-off settings were verified. No production rollout is included.

Historical real-provider protocol and checkpoint-recovery acceptance remains recorded in prior verification. The full synthetic corpus semantic/PK evaluation is supplementary; this PR does not claim it passed.


kbc/platform/pod/compile_box.py
```diff
@@ -5666,18 +5268,25 @@ async def _consume_turn_stream(
...
+                if fail_on_error_result:
+                    raise ModelResultError(f"model result failed (subtype={subtype}, api_status={status})")
+                note = _loc(run, "The model request failed; this session remains available for retry.",
+                            "模型请求失败，此会话仍可重试。")
+                await run.emit({"type": "error", "code": "model_request_failed", "error": note,
+                                "message": "model_request_failed", "recoverable": True})
+                await run.emit({"type": "turn_done", "text": note})
+                if stop_on_result:
+                    return
+                continue
```
jacoblee-io, line 5303: [P1] Finalize the failed turn's scope before returning to idle

This branch bypasses `_emit_message()` and `_post_turn_selfcheck()`, leaving `_incr_pending` armed after a failed scoped operation. I reproduced this through the real Pi worker: `compile.apply_proposal` scoped to `a.md` receives HTTP 400; the next ordinary owner message edits `b.md`, and `Write` succeeds, but the post-turn check restores `b.md` from the previous proposal's `before_bytes` while the assistant reports success. Reconcile and clear the failed turn's scope before emitting `turn_done`, so subsequent owner edits cannot inherit an obsolete restoration snapshot.

src/kbc/pi-worker.ts
```diff
@@ -0,0 +1,303 @@
...
+          signal?.removeEventListener("abort", cancel);
+          this.pendingTools.delete(callId);
+        }
+      },
+    }));
+    this.execution = await createPiExecutionSession({
+      services, sessionManager: SessionManager.create(config.cwd, path.join(config.state_dir, "sessions")),
+      model: resolved, thinkingLevel: config.thinking_level, customTools: tools,
+      onModelEnvelope: manifest => this.emit("model_envelope", { manifest }),
+    });
```
jacoblee-io, line 165: [P2] Preserve PDF images when applying the compiler context budget

The default context-budget guard installed here replaces an oversized tool result with text-only content, removing its image blocks. With a 128,000-token model and a `Read` result containing text plus 20 PNG page blocks (the supported PDF read size), I verified through the real worker that the next provider request contains zero images; a 400,000-token model retains all 20. No truncation notice or tool error is added, so scanned PDFs lose all their visual evidence before the model can inspect it. Use an image-aware compiler guard or coordinate PDF pagination with this budget instead of silently dropping the pages.

jacoblee-io (commented):
Two reproducible findings from reviewing this revision.

src/gateway/capability/materialize.ts
```diff
@@ -206,6 +227,16 @@ export async function materializeCapabilityInputs(opts: {
   if (src?.llm && typeof src.llm === "object") result.llm = src.llm;
   if (src?.settings && typeof src.settings === "object") result.settings = src.settings;
 
+  if (emptyLegacyBox && src?.llm?.engine === "pi_agent") {
```
jacoblee-io: [P2] Replace incompatible empty boxes for Claude execution too

When an old-image pod restarts during rollout, its health reports `runs: 0, test_sessions: 0` without an engine. If the resolved configuration selects `claude_agent_sdk`, this condition skips replacement: I reproduced zero calls to `replaceEmptyLegacyBox` and `/sources` being sent to the old client. Runtime then forwards the new `llm.execution` payload to the old `/session` handler, whose `_apply_session_config()` only reads top-level `base_url`, `model`, `api_key` and `auth_token`. With the new payload it ignores the resolved roles and clears the endpoint and credentials, preventing the Claude attempt from resuming. Apply the compatibility/replacement check to both supported engines, retaining the existing empty-box and revision checks.

kbc/platform/pod/test_image_permissions.sh
```diff
@@ -45,6 +46,27 @@ for path in Path("/app").rglob("*"):
         unreadable.append(f"{path}: {exc}")
 assert not unreadable, "unreadable application files:\n" + "\n".join(unreadable)
 import compile_box  # noqa: F401
+from importlib.util import find_spec
+assert find_spec("claude_agent_sdk") is None
```
jacoblee-io: [P2] Update the image smoke check for the restored Claude dependency

`requirements.txt` now installs `claude-agent-sdk==0.2.110` in the runtime image, so this assertion fails on every correctly built image. The current [KBC Box Test](https://github.com/scitix/siclaw/actions/runs/34571665267/job/103174899801) fails at `<string>:18`, which is this exact assertion, after the Docker build succeeds. Update the smoke check to verify the supported SDKs are available; otherwise CI never reaches the worker startup and container health checks.

jacoblee-io (commented):
Follow-up review of c15021be. The two earlier findings are addressed and their regression tests pass. Two remaining issues are detailed inline. Validation passed: npm run build, 27 targeted SDK/worker tests using local HTTP fixtures, and 40 materialization/session-driver tests. The current KBC image CI failure is linked below.

jacoblee-io (commented):
Re-reviewed the changes from c15021be through 741f1f14. Both findings from the previous review are addressed: empty legacy boxes are replaced for either selected engine, with revision and active-test-session checks preserved; the image smoke check now verifies the installed Claude SDK version. The PK default change also preserves explicit opt-in and the operational stop.

No new actionable findings in this incremental diff. Local validation passed: 42 materialization/session-driver tests, the complete test_selfcheck.py script, and the session configuration/PK-stop test. All five checks on this head are green, including KBC Box Test.

jacoblee-io (dismissed):
Reviewed 741f1f14. The previously reported issues are fixed, targeted verification passed, and all CI checks are green.

jacoblee-io (approved):

PR #578, opened 2026-09-10
feat(knowledge): retrieve cross-library evidence in one call

LikiosSedo:
## Summary

Label routing misses facts that appear only in compiled Wiki bodies and requires separate reads before answering. Add `knowledge_lookup` to search mounted libraries and return bounded complete pages with original paths, library IDs and versions, registering only the returned evidence with existing citation validation.

The same lazy in-memory FTS5 engine powers `siclaw knowledge search`. Catalog navigation and library-aware label search remain available. The branch is integrated with the current CLI and capability contracts. No embedding service, production dependency, mount schema or deployment variable is added.

An offline corpus comparison command accepts original questions, supplied alternate queries and multiple expected pages. It compares label navigation, body lookup and equal-weight query fusion, reporting page recall and complete-question coverage. Supplied alternates are an evaluation intervention; runtime ranking remains unchanged. Documentation explains distractor-library, collection-order, applicability and no-answer evaluation boundaries.

## Validation

- Full backend suite: 379 files, 7,621 tests passed with the current pinned SDK installation patch.
- Backend build, TypeScript no-emit check and AgentBox TypeScript check.
- Focused retrieval, citation, capability, CLI snapshot and handoff contracts.
- Portal capability catalog tests with an isolated configuration.
- CLI help and deterministic synthetic retrieval comparison.
- Corpus comparison smoke, known-baseline reconciliation and invalid-input rejection.

## Remaining acceptance

This remains an experimental retrieval implementation. Representative multi-library Agent evaluation must establish normal tool selection, complete evidence coverage, applicable conditions, answer and citation correctness, and end-to-end latency. Lexical matches and relevance ranks are not answer confidence. Large pages still require separate reading.

Downstream analytics that count only native `read` inputs must recognize complete `knowledge_lookup` results before broad rollout. Private corpus questions, traces and measurements are retained separately.

PR #586, opened 2026-09-11
feat(knowledge): unify compilation recovery and library-aware QA

LikiosSedo:
Knowledge-base compilation now recovers from interrupted execution while question answering discovers available knowledge before asking for clarification and routes label navigation within the correct mounted library.

## Review scope

- Runtime/compiler recovery: acknowledged event replay, bounded persistence retries, planless resume, retained failure evidence, and cancellation retries addressed to the original compiler box. Existing Pi and Claude compiler execution remain supported.
- Knowledge QA: one bounded discovery step before clarification, with migration of the previous built-in default.
- Library navigation: per-library label scoring, explicit roots and names, deterministic exact-name selection, explicit ambiguity, compact catalogs and bounded output. Existing single-library behavior remains supported.

This PR consolidates #547 and #572 into the existing recovery change. Experimental body-evidence retrieval in #578 remains separate. Review recovery, discovery and library routing as the three sections of this knowledge-base delivery.

## Review corrections

- Completion recovery rejects produced-page references that do not match actual candidate pages before accepting the command id or dispatching a model turn. Corrected requests can reuse the rejected id, and pending batch plans/reset markers retain ownership.
- Only the materializer manifest declares library boundaries. A root library or legacy flat mount stays single-library even when its catalog links to section indexes.
- Pod listing, watching and status projection share the registered KB profile names.
- Search omissions describe actual trimming, and ambiguity hints use the same margin as library selection.
- The labeled-package fixture explicitly archives its files and disables AppleDouble metadata for that subprocess.

## Compatibility and rollout

Reliable replay/adoption requires the compatible Runtime and compiler-box handshake. Drain active compiles when upgrading or rolling back across that boundary, or recover the persisted workspace in a new compatible pair. Deploy the recovery-context producer and diagnostic receiver before Runtime. Rebuild Runtime, the compiler box and the conversational AgentBox; update both box image references and recycle existing conversational AgentBox Pods. The existing batch-disable setting prevents new automatic batching while preserving resume of an existing plan.

The protocol document describes the consumer's content-based reply deduplication and its cross-process crash limitation, the distinct artifact/event retry windows, and the reconnect budget per driver invocation. This change does not promise exactly-once assistant-message creation across process crashes.

## Validation

- Current head `69f745d7`, based on main `abfb08f5`: TypeScript no-emit check and backend build passed.
- Recovery/relay regressions: 35 passed. Eleven added regression cases fail before the provenance fix and pass afterward, covering both inline and file-backed lists, path mismatches, bounded diagnostics and corrected-request retries.
- Pi/Claude compiler contract suite: 64 passed locally; two PDF cases require Poppler, which is installed by KBC CI. SDK and model-usage contract tests: 23 passed.
- Pytest selfcheck collection: 83 passed, with two failures also reproduced on the preceding head (`test_charset_corruption_detection` and `test_wiring`). The KBC CI workflow separately runs its standalone selfcheck entry point.
- Prior head `e6c75b7f`: full backend suite without global `COPYFILE_DISABLE` passed 375 files (7,581 tests, 1 skipped); targeted knowledge/materialization/profile/overview suite passed 171 tests. This follow-up changes only Python recovery, its tests and the protocol documentation.
- No new live provider fault injection, deployment or experimental retrieval acceptance is claimed.


LikiosSedo:
Full box suite in the kbc-box image on dev02 (`siclaw-kbc-box:evidence-3b8dcc574-0910`, `KBC_PK_MODE=off`, `asyncio_mode=auto`):

- this branch (`6d6af89b`): `test_compile_box.py` + `test_selfcheck.py` → 204 passed, 2 failed
- baseline `main` (`d78fd7b`): `test_selfcheck.py` → 83 passed, the same 2 failed

The two failures (`test_selfcheck.py::test_wiring`, `test_selfcheck.py::test_charset_corruption_detection`) are pre-existing on main and untouched by this change (selfcheck.py is not modified here). GitHub CI (KBC Box Test, Test, Type Check, Portal Web Test) is green.

LikiosSedo:
Review follow-up pushed (`b6014d07`): the box now takes Sicore's `recovery_mode` / `produced_pages` on `compile.resume` and no longer selects completion from file existence; in complete mode the directive names inherited pages to regenerate. Companion sicore change in !1180 computes progress against the base draft and injects the parameters into the command receipt.

jacoblee-io (commented):
Both halves address real gaps and the tests are substantive. I ran `container-evidence.test.ts` and `session-driver.test.ts` on the head commit (63bedad) — 35 tests pass. `tsc --noEmit` is clean for the 14 changed files (the errors I saw came from a stale shared `node_modules` with a pi-coding-agent version skew, none in this diff). I did not run the Python suite.

One finding I would treat as blocking, then a few I would like answered before merge.

## Blocking: reconnect can persist a failed run as `done`

`?replay=1` replays artifacts only — `_pending_sync_events` plus the workspace diff (`compile_box.py:7346-7357`). Every other event travels the bare queue:

```python
ev = await asyncio.wait_for(run.events.get(), timeout=25)   # 7361
await resp.write(...)                                        # ← the socket resets here
run.events.task_done()                                       # 7368
```

The event already dequeued when the write fails is gone permanently; nothing retains it and nothing replays it. And `error` is always followed by `end` (`compile_box.py:5743-5745`), so losing the `error` frame means the reconnected stream delivers `end`, `relayBoxEvent` takes the `end` branch, and `manager.endRun(runId, "done")` records a failed compile as done — with no `checkpoint.failure` and an empty auto-resume detail. The same window loses `turn_done`: the reply is never persisted and the run never leaves `running` for `idle`.

Before this PR that drop surfaced as a loud `relay_failed`. After it, the loudest failure mode in the system becomes a silently wrong terminal state. The window is narrow in wall-clock terms but it is exactly "the connection broke", which is the only scenario this feature runs in.

Two ways out, either is fine:

- Box side: retain the terminal event until the relay acknowledges it, mirroring `_pending_sync_events`, and replay it under `replay=1`.
- Runtime side: remember that no terminal was seen before the drop, and refuse to terminalize on a bare `end` arriving on a reconnected stream.

## Questions before merge

**1. The container-evidence emitter has no switch.** `spawner?.observeContainers?.(...)` (`server.ts:2966`) is unconditional for `K8sSpawner`. The PR body says to deploy the receiver "before enabling the emitter", but there is nothing to enable or disable — deploying this Runtime turns it on. Against a control plane without `capability.observeContainer`, every observation rejects: `console.error`, then retained in `failed` (cap 256). Each WS reconnect calls `retry()`, which re-`record()`s all 256 — one full-JSON `console.info` each — and they all fail again. That is a permanent log loop with no backoff and no way to stop it short of rolling back the Runtime. An env gate, or a breaker that disables the queue after N consecutive method-not-found, would make the stated deployment order actually enforceable.

Separately, `CAPABILITY_OBSERVE_CONTAINER` is declared in `container-evidence.ts` while every other `capability.*` method name lives in `contract.ts`.

**2. The informer restart loop is unbounded.** `onError` schedules `setTimeout(5s)` → `informer.start().catch(onError)`, forever, with an error logged each time. The Helm chart does grant `watch` on pods, so a chart-deployed Runtime is fine — but a Runtime deployed outside it without that verb logs every 5 seconds for the pod's lifetime. Exponential backoff plus a give-up would bound it.

**3. `complete` state over an above-threshold corpus discards the new directive.** `compile.resume` is in `_FULL_COMPILE_ACTIONS`, so after `_render_command` builds the produced/inherited `resume_complete` text, `_should_route_to_batch` sees a large `raw/` with no resumable plan and spawns `_run_batch_compile(run, text)` — which ignores `trigger_text`, calls `_plan_batches` over the whole inventory, and emits "batch compile engaged". The produced/inherited classification this PR just built is never shown to the model, and the run restarts from scratch, which is the outcome the change exists to prevent.

Does the control plane ever send `recovery_mode: "complete"` for an above-threshold KB? `test_resume_without_plan_routes_by_workspace` only exercises this with a two-file `raw/`, so the routing interaction is uncovered either way.

## Smaller points

- `server.ts:1531` — `async () => { await capabilityRunManager.reconcile(); retryContainerEvidence?.(); }` couples the evidence replay to reconcile succeeding. A reconcile rejection is only caught by `notifyConnected`, so the replay is skipped entirely until the next reconnect. `try`/`finally` keeps the documented ordering without the coupling.
- `_load_recovery_provenance` is looser than the inline path it is preferred over: inline entries go through `_bounded_string(limit=512)` and the list is capped at 500, while the file path accepts unbounded strings from an unbounded `read_text`, and those strings are rendered verbatim into the model prompt. Worth applying the same bounds.
- `ContainerEvidenceQueue.record` — `if (!this.pending.has(key) && this.pending.size >= 256)`: the first conjunct is dead, the function already returned on `pending.has(key)`.
- `record()` re-logs the full observation JSON at info level on every `retry()`; a flapping WS replays up to 256 of them per reconnect.
- The backoff comment says "about three minutes"; 2+4+8+16+32+60 is 122s.
- The informer selector hardcodes `kb-compile,kb-compile-codex,kb-compile-pi,kb-test` while `containerObservation` filters on `startsWith("kb-")`. A new kb profile silently stops producing evidence. Deriving both from one list would keep them from drifting.

## Verified

- `relayBoxEvent` is line-for-line the previous inline switch; the only changes are the wrapper, matching the PR body's claim.
- `containerObservation` really does exclude env, args, annotations and messages, and the test pins that with a negative match.

jacoblee-io (commented):
Second pass on the same head (63bedad). Two more findings I would treat as blocking — one of them makes the terminal-loss problem from my first review considerably worse than I described it — plus a few smaller ones.

## Blocking: a reconnect creates a second relay that splits the event queue

`handle_events` counts relays but never fences them (`compile_box.py:7341`, `7372`, and nothing else reads `_relays` except the shutdown drain at `7771`). There is no relay id, the new attachment does not cancel the previous one, and a second attachment for the same run is not refused.

So when the stream drops silently — partition, LB idle timeout, half-open socket, i.e. the cases where the peer never sends RST — the old `handle_events` coroutine is still sitting in `await run.events.get()` while the Runtime reconnects and starts a second one. Both now consume from the same single `asyncio.Queue`, and whichever wins each `get()` takes that event. The dead one writes to a dead socket, `task_done()` is never reached, and the event is gone.

The old coroutine only discovers the death on its next write, which is the next event it steals or the 25s heartbeat — so the window is bounded at roughly 25 seconds, during which **roughly half of all events are lost**, not the single in-flight frame I described in my first review. Everything in that review about `error` being followed by `end` still applies, just at much higher probability: over a 25s window a compile can easily emit the `error`, and `end` arriving on the surviving relay then records the run as `done`.

Before this PR nothing ever re-attached to a live run's stream within a process, so this hazard was latent. The reconnect makes it routine — it fires on every transport blip.

This wants a fence, not a bigger buffer: the box should treat the newest relay for a run as the only valid one and terminate the previous coroutine on attach (or reject the second attach and let the Runtime retry). Combined with retaining the terminal event until ACK, that closes both halves.

## Blocking: `produced_pages` has no path-convention validation, and the failure mode is "rewrite the whole draft"

`selfcheck.candidate_pages` is keyed **relative to `candidate/`, posix** (`selfcheck.py:286-288`), and `_render_command` compares against that keyspace directly:

```python
produced_set = set(produced)
inherited = [page for page in all_pages if page not in produced_set]
```

Nothing checks that `produced` is drawn from that keyspace, and nothing checks that any of it exists. A control plane that emits `candidate/a.md` instead of `a.md` — the obvious mistake, since the rendered output is `f"- candidate/{p}"` and the doc row never states the convention — produces an empty intersection. Every page then lands in `inherited`, and the model is told they "were inherited from the previous published draft and are NOT this compile's output … Regenerate each of them from raw/ (rewrite, do not keep)."

That is a full-draft rewrite of a knowledge base, triggered by a prefix, and every existing check passes: `produced_count` matches `len(produced_pages)`, the provenance file is well-formed, arity agrees. The same thing happens at smaller scale for a single typo'd page name.

The PR already identified this exact hazard for the partial-list case — "a page missing from it is told to be rewritten as inherited" is why an incomplete list is refused with 409. But arity is not membership. The guard should be a subset check against `selfcheck.candidate_pages(workdir)`, refusing when a produced page is not a real candidate page (or at minimum when the intersection is empty while `produced_count > 0`). The path convention should also be stated in the `compile.resume` row of `kb-authoring-command-protocol.md`; right now it exists only in the test fixtures.

## `attempts` is a per-run lifetime budget, and the two comments disagree about that

`attempts` is never reset after a reconnect that succeeds. `StreamReconnectPolicy.maxAttempts` documents this correctly ("per run"), but the rationale block above `defaultStreamReconnectPolicy` reads as per-outage — "about three minutes of transport trouble tolerated" (also 2+4+8+16+32+60 is 122s, not three minutes).

The practical consequence is the part worth deciding on: a multi-hour batch compile that hits an unrelated blip roughly once an hour dies on its seventh one, hours apart, with a `relay_failed` that has nothing to do with the transport state at that moment. Resetting the counter once a reconnected stream yields at least one event makes the budget mean what the rationale says and keeps the guard against a genuinely flapping connection.

## Smaller

- **Inconsistent eviction, in the wrong direction.** `record()` drops the *newest* observation when `pending` is full, while `retainForReconnect` drops the *oldest* when `failed` is full. For lifecycle evidence the newest snapshot is the valuable one — it carries the terminal exit status — so under backlog the pending queue preserves 256 stale `observed` updates and discards the `deleted` frame the feature exists to capture. Evicting oldest in both places would be consistent and keeps the useful end of the stream.
- **`isBoxAlive` inherits the client's 30s timeout.** The default probe is `client.getJson("/health")`, which takes no deadline and falls back to the client's (`30000` for the capability client, `server.ts:1351`/`1360`). "A dead box is not worth waiting for" then means waiting 30 seconds on a hung box. The `observeContainer` RPC already uses an explicit `3_000`; the probe deserves the same.
- **The reconnect metric is incremented before the post-backoff settle check**, so a reconnect that is abandoned during the backoff still counts as one.
- **Queue-full logs the same payload twice** — `record()` emits the full observation JSON at info, then the queue-full branch emits it again at error.
- **The two spawn-path deletes** (`k8s-spawner.ts:547`, `572`) do not call `beforeStop`, unlike `stopPod`. These remove `Failed`/`Succeeded`/stale-CA pods, which is exactly the interesting evidence. The informer's relist and `delete` handler do cover them, so this is consistency rather than loss — but if `beforeStop` is worth doing on the clean path it is worth doing here.

jacoblee-io (commented):
Third pass, on `05912b7a`. The `EventRelay` rework is the right shape — single-consumer fencing plus one unacknowledged durable frame with an explicit persistence ACK actually closes both halves of the terminal-loss problem, rather than widening a buffer. `_should_route_to_batch` now short-circuits `complete` and forces `plan`, with a 201-page above-threshold fixture behind it. Eviction direction, the metric placement, the dead conjunct and the backoff comment are all fixed.

Ran `src/gateway/capability/**` plus `server-capability-setup-failure.test.ts` on this head: 139 tests pass. I could not run the Python suites locally (no aiohttp/pytest-asyncio, and local Python is 3.9 while `asyncio.timeout` needs 3.11) — `kbc-ci.yml` now covers both new files on 3.11, which matches the box image.

Three new findings on the new code, then what is still open from earlier.

## The 4-attempt retry re-emits live frames

The retry loop wraps `relayBoxEvent`, and `relayBoxEvent` emits before it persists. For `turn_done` that is now load-bearing, because a `persistTurn` failure throws when `event_id` is present (`if (evt.event_id) throw err;`) instead of being logged — so the retry re-enters the case and re-runs `emit("turn", …)`. `error` and `done` have the same shape via `emit("lifecycle", …)` ahead of `commitRelayEvent`.

Measured on this head, with two `persistTurn` failures on one `turn_done`:

```
>>> persistTurn failures=2  live 'turn' frames emitted=3
```

`session-driver-delivery.test.ts:101` asserts `turns).toHaveLength(1)`, which pins the persistence side and passes — the live duplication is not asserted anywhere, which is why it survived. Hoisting the `emit` above the retried body would fix it; if repeated live frames are considered acceptable because the consumer renders from a refetch, an explicit `emitEvent` count assertion in that test would make it a decision rather than an oversight.

## The rolling-upgrade behaviour is fail-closed in both directions — please put it in the deployment note

Two independent latches now make a mixed Runtime/box pair unrecoverable rather than degraded:

- Runtime side, adoption: with `replayWorkspace: true` against a box that never sends `relay_ready`, the first event throws `box cannot safely replay lifecycle events` and the run fails. Pinned at `session-driver-delivery.test.ts:184`. Before this PR the adopt path replayed artifacts and carried on.
- Box side: `_ack_mode` is latched for the life of the run and a mismatch is a permanent 409. A run whose first relay was legacy can never be taken over by a new Runtime, and vice versa.

`CLAUDE.md` already records that "a rolled-out runtime does NOT pick up a new box image for existing live sessions", which is exactly the state a Runtime rollout produces. So every in-flight compile on a pre-existing box now dies on adoption instead of resuming. I think that is the right trade — a wrongly-recorded `done` is worse than a visible failure — but the PR's deployment note still only mentions the recovery-context producer and the container-evidence receiver. It should say that in-flight KB compiles do not survive this upgrade, so the rollout can be scheduled accordingly.

## One extra run-state write and one box round trip per durable frame

`syncArtifacts` is durable, and `KBC_SYNC_INTERVAL_SECS` defaults to 20. Each sync batch now costs `persistArtifacts` → `/artifacts/ack` → `commitRelayEvent` (a `persistRunState` write) → `/events/ack`. Previously a sync batch caused no run-state write at all, so this is a new periodic write against the control plane for every live compile.

The part worth deciding explicitly is `failFast: true` on that write: four retries and then the run fails. For `done`/`error` that coupling is clearly worth it. For a 20-second artifact sync it means a transient control-plane hiccup now kills a compile that previously would have kept going and synced again 20 seconds later. Is `syncArtifacts` meant to be in `DURABLE_TYPES` with the same strictness as the terminal frames, or would "durable id, best-effort commit" be enough for it?

## Still open from the earlier passes

- **`produced_pages` membership is documented but not enforced.** The protocol table now says "POSIX paths relative to `candidate/`", which is the right fix for the contract, but `_prepare_command` still validates only arity. A control plane that sends `candidate/topic-1.md` still yields an empty intersection, every page lands in `inherited`, and the model is told to rewrite the whole draft — and every existing check passes. A subset assertion against `selfcheck.candidate_pages(workdir)` is a few lines and turns the documented convention into an enforced one. `test_planless_complete_delivers_preservation_directive` passes all 201 pages in the correct form, so the wrong form is still uncovered.
- **`isBoxAlive` inherits the capability client's 30s timeout** (`server.ts:1351`/`1360`); the probe deserves its own short deadline like the `3_000` on `observeContainer`.
- **`onConnected` still couples the evidence replay to `reconcile()` succeeding** — `try`/`finally` keeps the documented ordering without it.
- **`retry()` still re-logs every retained observation** at info with the full JSON, up to 256 per reconnect. (The duplicate queue-full log is gone — that one is fixed.)
- **`CAPABILITY_OBSERVE_CONTAINER` still lives in `container-evidence.ts`** rather than `contract.ts` with the other method names.

## Smaller, on the new code

- `_batch_mode_enabled()` is no longer a complete kill switch: the check moved below the `compile.resume` branch, so a `plan`-state resume runs the orchestrator even with `KBC_BATCH_MODE=off`. `test_existing_batch_ownership_precedes_size_and_mode` pins this deliberately and finishing a pinned plan is defensible, but the flag went from "never run the orchestrator" to "never start one" and it is an ops rollback lever, so it is worth a line in the docs.
- `EventRelay.stream` suppresses only `CancelledError` and `ConnectionError` while awaiting the cancelled previous consumer. Any other exception from that handler propagates out of `stream()` before `self._task` and `self._ack_mode` are assigned, so the new attachment 500s. The next reconnect recovers because `previous.done()` is then true, so the cost is one wasted attempt out of six — widening to `Exception` removes it.
- `acknowledge()` remembers only `_last_ack`, so an ACK for anything older raises `ValueError` → 409. The one-pending-frame invariant makes that hard to reach, but if it happens the driver cannot distinguish it from a transport failure and burns a reconnect attempt on it.

jacoblee-io (commented):
Fourth pass on `05912b7a` (unchanged since my last review), covering ground the earlier passes did not reach: the Python relay tests, the run-manager write policies, and the container-evidence queue's steady state.

First, what I checked and found sound, so you know it has been looked at:

- `test_event_relay.py` is stronger than I expected. Per-durable-type write-loss replay, fencing in both positions (cancelled mid-write and cancelled while parked on `events.get()`), ordering across takeover, the 409 protocol latch, the ack-route status codes, and a real `TestServer` reconnect that holds the first connection open while the second attaches. That last one is the actual half-open case.
- The `asyncio.shield` on the ACK wait is the subtlest line in the module — without it a takeover cancels `self._ack`, the successor's `while not self._ack.done()` sees a cancelled future as done, and the frame is marked acknowledged with no ACK ever sent, which is precisely what the design exists to prevent. It *is* covered, deterministically: in `test_artifact_turn_and_terminal_stay_ordered_across_takeover` the old task has reached the ACK wait by the time `writing` is observed, and the `relay._pending is event` assertion is what would fail. Worth a comment naming that, because it reads as incidental and a later refactor of the test would drop it without anyone noticing.
- Epoch/sequence identity holds: a box restart mints a fresh `uuid4().hex`, so a stale `persistedRelayEventId` from a checkpoint can never false-dedupe a new run's frames.
- `server-capability-setup-failure.test.ts` now pins `?ack=1&replay=1`, the ACK call, and the absence of a spurious `failed` write on the adopt path.

Three findings.

## `commitRelayEvent` and `endRun` apply opposite policies to the same write, and the strict one governs the 20-second sync

`endRun` states its policy explicitly and implements it:

> When the terminal persist fails, the record STAYS in memory with its terminal status — the reconcile loop retries the same terminal write (flushTerminal) until it lands, so the store converges to the true outcome instead of a later blanket "failed".

`commitRelayEvent` awaits `this.persist(rec, { failFast: true })` unguarded, so the throw reaches the driver, which retries four times (250 + 500 + 1000 ms) and then sets `relayError` and fails the relay.

For `done`/`error` this lands safely by accident: the delete is after the await, so the record survives in memory with a terminal status, `endRun`'s stickiness makes the caller's `relay_failed` a no-op, and `flushTerminal` converges to the true outcome.

For the non-terminal durable events it does not. `syncArtifacts` is in `DURABLE_TYPES` and `KBC_SYNC_INTERVAL_SECS` defaults to 20, so every twenty seconds a live compile now performs a `persistRunState` write whose failure, after 1.75 seconds of retries, kills the run — and there the status is not terminal, so `endRun(runId, "failed")` is not a no-op and the compile is recorded as failed. A transient control-plane hiccup during a routine artifact sync now ends a multi-hour compile that previously would have synced again twenty seconds later.

The two functions are answering the same question — "the store did not accept this write, now what" — with opposite answers, and the strict answer is applied to the most frequent and least consequential event. Giving `syncArtifacts` the `endRun` treatment (leave it to reconcile, do not fail the relay) while keeping strictness for the terminal frames would match the policy the file already documents. If `syncArtifacts` genuinely needs the ACK barrier, the barrier and the failure policy can still be separated: withhold the box ACK without failing the run.

## A permanently-declined container observation is retried for the life of the process

The dedupe branch reads:

```ts
if ((result as { observed?: boolean } | null)?.observed === true) this.acknowledged.set(key, true);
else this.retainForReconnect(key, observation);
```

with the comment "Another Runtime may own this run, or its initial write may still be in flight." Those are two causes with opposite lifetimes — one clears on its own, one never does — and both take the same branch. There is no attempt counter and no give-up, so an observation for a run this Runtime does not own occupies a slot in the 256-entry `failed` map for the process's lifetime and is re-sent on every WS reconnect, refused every time.

The informer selector has no Runtime scoping — `${prefix}/app=agentbox,${prefix}/boxType in (…)` — and the spawner stamps only `app`, `agent`, `boxType` and `ca-fp`, with `labelPrefix` defaulting to the constant `siclaw.io`. So in any namespace holding more than one Runtime's boxes, every foreign kb pod produces exactly this. Combined with the new oldest-eviction in `record()`, foreign observations that can never succeed will push legitimate ones out of the pending queue — the eviction fix makes the queue prefer the newest, and under a foreign-pod stream the newest are the ones that cannot land.

Two questions: can two Runtimes share a namespace in your deployments? And is `observed: false` ever meant to be permanent, given the comment says it can be? Either way a bounded attempt count on `retainForReconnect` costs nothing and removes the class, whatever the topology turns out to be.

## The event-id format is Runtime-enforced but protocol-unspecified, and the regex exists twice

`session-driver.ts` rejects any durable frame whose id fails `/^[a-f0-9]{32}:[1-9][0-9]{0,15}$/` with `box omitted a valid durable event id` — a `relayError`, so the run fails rather than degrades. `run-manager.ts`'s `relayEventIdFromCheckpoint` carries a second, independent copy of the same regex.

The protocol section you added says only "an `event_id` scoped to that live box run". So `f"{uuid4().hex}:{self._sequence}"` is a load-bearing contract across two languages and two repos-worth of deployment skew, with nothing written down: changing it to `str(uuid4())` for the dashes, or uppercasing the hex, breaks every run on a box/Runtime version mismatch, and breaks it hard rather than falling back.

Either state the shape in `kb-authoring-command-protocol.md` beside the handshake, or have the Runtime validate what it actually needs — a non-empty bounded opaque string that differs between frames — and keep the strict shape only at the point of generation. The second copy in `run-manager.ts` should then read from one exported constant either way.

jacoblee-io (dismissed):
Reviewed the three sections of this delivery against the highest-risk failure modes for each: replay idempotency for recovery, silent mis-selection for library routing, and fail-open degradation for discovery.

**Recovery / replay — the protocol is fail-closed on both ends, which is the part that matters.**

`session-driver.ts` refuses to request replay unless the box advertised `event_ack: 1` in `relay_ready`, and throws `box cannot safely replay lifecycle events` instead of degrading to a blind replay against an older box. On the box side `event_relay.py` keeps exactly one unacknowledged durable frame per run, makes repeat acks idempotent (`event_id == self._last_ack`), rejects an ack that is not the current pending frame, and refuses to switch `_ack_mode` mid-run — the comment gives the right reason, that switching would incorrectly claim a complete history. `event_id` carrying the epoch prefix keeps ids from colliding across a box restart. The two ends line up.

This line is the one I'd point at as the load-bearing contract:

> Explicit full-compile commit. Replayed file presence alone is not a commit.

Recovery logic usually fails by treating a resumed intermediate state as success; stating it as a contract next to the field is the right place for it.

**Library routing — ambiguity is surfaced, not resolved by guessing.**

An unknown or ambiguous explicit selector returns an error plus the available libraries rather than silently picking one, and `crossLibraryTie` (top-two score gap < 0.1) tells the caller to compare domains or scope the search instead of proceeding. Library roots come from the materializer's `.citation-manifest.json` rather than a hardcoded layout, and `libraryRootForFile` matches on a `/` boundary so `a` cannot swallow `a-b`. The comment that model-written `name`/`domain` are matching/display metadata and never executed is worth keeping as the file grows.

One small note, not blocking: `readManifestLibraryRoots` returns `null` on any read/parse failure and routing then falls back to the root catalog, which is model-written text. That is a reasonable degradation, but it means a corrupted manifest quietly downgrades the authority of routing rather than surfacing. A warn on that path would make it diagnosable.

**Coverage of this review.** I read the recovery/replay path end to end (driver + relay), the library discovery and selection path, and the label scoring/ambiguity surface. I did not review `container-evidence.ts`, `compile_box.py`, `sync-handlers.ts`, or the remaining `knowledge-search.ts` rendering in the same depth, and I did not run the suite — the stated 7,559 passed / 1 skipped is taken as reported.

Approving on the sections I reviewed. Please keep the rollout ordering in the description (recovery-context producer and diagnostic receiver before this Runtime, drain active compiles across the protocol boundary) in the deploy notes rather than only in the PR body — it is the kind of constraint that gets lost after merge.

src/knowledge/libraries.ts
```diff
@@ -0,0 +1,118 @@
...
+ * single-library mount (or a mount without manifest and without library links),
+ * so callers can treat "one library" and "no library dimension" identically.
+ */
+export function discoverKnowledgeLibraries(knowledgeDir: string): KnowledgeLibraryInfo[] {
+  let rootIndex = "";
+  try { rootIndex = fs.readFileSync(path.join(knowledgeDir, "index.md"), "utf-8"); } catch { /* no catalog */ }
+  const fromCatalog = parseRootCatalogLibraries(rootIndex);
+  const manifestRoots = readManifestLibraryRoots(knowledgeDir);
+
+  const roots = manifestRoots && manifestRoots.some((root) => root !== "")
```
jacoblee-io: **This discards the manifest's authoritative answer for every single-library mount, and the catalog fallback then invents libraries.**

`sync-handlers.ts` writes `root: ""` for the single-repo branch, so `manifestRoots` is `[""]`, `some(root => root !== "")` is false, and we fall through to parsing `index.md`. But on a single-library mount `index.md` *is* that library's own page catalog, and a hierarchical catalog linking to section indexes (`- [GPU](gpu/index.md)`) is a perfectly normal shape — catalog trails are multi-hop by design. Every second-level directory then becomes a "library".

Reproduced locally with a manifest of exactly `repos:[{root:""}]` plus two section-index links:

```
multiLibrary: true   selected: ["gpu","net"]
pages: [{file:"gpu/a100.md",library:"gpu",score:1,pc:[1]},
        {file:"net/switch.md",library:"net",score:1,pc:[1]}]
```

Consequences on a mount that is *not* multi-library: results gain a `library` field, the `libraries[]` block appears, routing can restrict `results` to one section, and per-library term frequency splits the shared label (`pageCount` 2 → 1+1), which changes the `0.6 + 0.4/√pageCount` dilution and therefore the scores. That is exactly what "Single-library output is unchanged, field for field." in the CHANGELOG promises will not happen.

It also puts two definitions of "multi-library" in the same PR: `overview-generator.ts` decides from the manifest (`filter(root !== "").length > 1`), so the **injected Wiki prompt says single-library while the tool says multi-library** for the same mount.

The TUI / `portal-knowledge-materializer` path writes no manifest at all and unpacks flat, so it hits the same fallback.

Suggested fix: when the manifest is present (non-null), trust it — a manifest declaring only `""` means one library, so return `[{ root: "", … }]`. Fall back to catalog parsing only when the manifest is **absent**. Worth a regression covering "single library whose catalog links to section indexes".

src/agentbox/sync-handlers.test.ts
```diff
@@ -1349,6 +1349,47 @@ describe("knowledgeHandler multi-repo identity", () => {
...
+  function packageBase64WithLabeledPages(title: string, pages: Array<[file: string, pageTitle: string, labels: string]>): string {
+    const sourceDir = fs.mkdtempSync(path.join(os.tmpdir(), "knowledge-package-labels-test-"));
+    const archive = path.join(os.tmpdir(), `knowledge-package-labels-${process.pid}-${Date.now()}-${Math.random()}.tar.gz`);
+    try {
+      fs.writeFileSync(path.join(sourceDir, "index.md"), `# ${title}\n\n${pages.map(([file, pageTitle]) => `- [${pageTitle}](${file})`).join("\n")}\n`);
+      for (const [file, pageTitle, labels] of pages) {
+        fs.mkdirSync(path.dirname(path.join(sourceDir, file)), { recursive: true });
+        fs.writeFileSync(path.join(sourceDir, file), `---\ntype: Topic\ntitle: ${pageTitle}\nlabels:\n${labels}---\n# ${pageTitle}\n`);
+      }
+      execFileSync("tar", ["-czf", archive, "-C", sourceDir, "."]);
```
jacoblee-io: **This test fails deterministically on macOS.**

Archiving `.` makes bsdtar emit AppleDouble entries, and `validateKnowledgePackage` then throws `Unsupported knowledge file type: ._.`. The other three helpers in this file (`packageBase64`, `packageBase64WithRoutes`) list entries explicitly, which is why only this one breaks.

```
$ npx vitest run src/agentbox/sync-handlers.test.ts -t "computed label sample"
Error: Unsupported knowledge file type: ._.
 ❯ validateKnowledgePackage src/shared/knowledge-package.ts:83:13

$ COPYFILE_DISABLE=1 npx vitest run … -t "computed label sample"
 ✓ 1 passed
```

CI is Linux so it stays green there; the whole suite is 372 files / 7559 passed for me too, but only with `COPYFILE_DISABLE=1`.

Either list the entries explicitly like the neighbouring helpers, or pass `{ env: { ...process.env, COPYFILE_DISABLE: "1" } }` to `execFileSync`.

src/gateway/capability/container-evidence.ts
```diff
@@ -0,0 +1,236 @@
...
+    if (this.pending.size) console.error("[capability-container] shutdown dropped pending persistence", this.pending.size);
+    this.pending.clear();
+    this.failed.clear();
+    await this.draining;
+  }
+}
+
+export function observeContainerLifecycle(kc: k8s.KubeConfig, api: k8s.CoreV1Api, namespace: string, prefix: string,
+  send: (observation: ContainerObservation) => Promise<unknown>) {
+  const selector = `${prefix}/app=agentbox,${prefix}/boxType in (kb-compile,kb-compile-codex,kb-compile-pi,kb-test)`;
```
jacoblee-io: `kb-compile-pi` appears **nowhere else in the repository**. `BUILTIN_PROFILES` holds `kb-compile`, `kb-compile-codex` and `kb-test`, and `getBoxProfile` fails closed on an unknown name — so a pod with that label could never be served anyway.

That leaves three different definitions of "a KB box" in this feature: this hand-written selector, `BUILTIN_PROFILES`, and `containerObservation`'s `profile?.startsWith("kb-")`. They already disagree, and adding a fourth `kb-` profile would make the informer blind to it while `containerObservation` happily accepts it.

Deriving the selector from `BUILTIN_PROFILES` keys (filtered to `kb-`) would collapse all three into one source of truth.

src/knowledge/labels.ts
```diff
@@ -333,20 +497,101 @@ export class KnowledgeLabelIndex {
...
+      const domainScore = multiLibrary && domainTerms.length > 0
+        ? queryCoverage(query, matchedDomainTerms)
+        : 0;
+      for (const term of matchedDomainTerms.slice(0, 2)) why.push(`domain: ${term}`);
+      const density = Math.min(1, pages.length / 3);
+      const score = pages.length === 0 && domainScore === 0
+        ? 0
+        : Math.min(1, 0.7 * bestPage + 0.1 * density + 0.2 * domainScore);
+      if (score === 0 && !multiLibrary) continue;
+      if (score === 0) continue;
```
jacoblee-io, line 575: The line above (`if (score === 0 && !multiLibrary) continue;`) is fully subsumed by this one — same guard, strictly narrower condition. Drop it.

src/tools/query/knowledge-search.ts
```diff
@@ -23,6 +25,98 @@ function truncateUtf16Safe(value: string, maxLength: number): string {
...
+  if (measure(rows) <= budget) return { results: rows, omitted };
+  if (rows.some((row, index) => index >= KEEP_FULL_RESULTS && row.labels !== undefined)) {
+    rows = rows.map((row, index) => index >= KEEP_FULL_RESULTS ? { ...row, labels: undefined } : row);
+    omitted.push(`labels beyond the top ${KEEP_FULL_RESULTS} results`);
+    if (measure(rows) <= budget) return { results: rows, omitted };
+  }
+  rows = rows.map((row) => typeof row.description === "string" && row.description.length > 200
+    ? { ...row, description: truncateUtf16Safe(row.description, 200) }
+    : row);
+  omitted.push("descriptions shortened to 200 characters");
```
jacoblee-io: This (and the `catalog trails reduced…` pushes at the two later steps) is unconditional, so `omitted` can report trimming that never happened — no description longer than 200 characters, or every trail already ≤ 2 steps and `rootAndLeaf` returning the row unchanged.

`omitted` is model-facing and is the thing that tells the agent whether to re-ask with a larger `topK`/`includeLabels`, so a false entry there is a small lie in the tool contract. Push only when the map actually changed a row.

src/tools/query/knowledge-search.ts
```diff
@@ -97,20 +236,65 @@ export function createKnowledgeSearchTool(resolver: KnowledgeResolver): ToolDefi
...
+          Buffer.byteLength(JSON.stringify({ results: rows, ...librariesBlock }, null, 2), "utf8");
+        const fitted = fitResultsToBudget(fullResults, measure, KNOWLEDGE_SEARCH_PAYLOAD_BUDGET_BYTES);
+        const results = fitted.results.map((row) =>
+          Object.fromEntries(Object.entries(row).filter(([, v]) => v !== undefined)) as RenderedResult & { score: number });
         const hasMore = result.matchedPages > results.length;
+        // Two libraries scoring within 0.1 of each other is a routing tie whether
+        // both were selected or none was: the Agent should compare domains (or
+        // search each library) before reading a leaf from the wrong one.
+        const crossLibraryTie = multiLibrary && result.libraries.length > 1 && !library &&
+          result.libraries[0].score - result.libraries[1].score < 0.1;
```
jacoblee-io: The tie threshold `0.1` is inlined here while the routing margin is the named `LIBRARY_ROUTE_MARGIN = 0.15` in `labels.ts`. A gap in [0.1, 0.15) therefore routes to **both** libraries but reports no tie, so the agent gets results drawn from two libraries with no hint that they were near-equal.

Either name this constant next to the other one, or state why the two thresholds are deliberately different.

jacoblee-io (commented):
Reviewed the full diff (37 files, +3476/-241) and ran the suite locally in a worktree.

**Summary:** the recovery half — acknowledged Runtime↔box event delivery — is in good shape; the protocol doc, the Python relay tests and the TS delivery tests all line up, and I could not find a real defect there. The library-routing half has **one real bug that contradicts the CHANGELOG's own compatibility claim**, plus a new test that fails on macOS.

### Verification

- `npx tsc --noEmit`: clean.
- `COPYFILE_DISABLE=1 npm test`: 372 files, 7559 passed, 1 skipped — matches the PR description exactly.
- Without `COPYFILE_DISABLE`, one new test fails on macOS (inline comment below); that is why the PR's own run was green.
- Python tests not run locally (no aiohttp here); CI now covers `test_event_relay.py` and `test_resume_routing.py`.

### Must fix

1. **A single-library mount is misdetected as multi-library** (`src/knowledge/libraries.ts`) — reproduced locally, details inline. This directly contradicts "Single-library output is unchanged, field for field." in the CHANGELOG, and makes the injected Wiki prompt and `knowledge_search` disagree about the same mount.
2. **New test fails on macOS** (`src/agentbox/sync-handlers.test.ts`) — AppleDouble entries from `tar -C dir .`.
3. **Three different definitions of "a KB box"**, one of which names a profile that does not exist (`kb-compile-pi`), in `container-evidence.ts`.

### Questions

4. **The `turn_done` dedup window is narrower than the commit message claims.** The order is `persistTurn` → `commitRelayEvent` → `/events/ack`. If the Runtime *process* dies between the first two, a reconnect still sees the old `persistedRelayEventId` and re-issues `persistTurn`. The design doc acknowledges this ("the existing authoring sink provides this deduplication"), so this is a hard dependency on the consumer — what is that sink's dedup key? If it is "adjacent identical content", two genuinely identical consecutive replies would be swallowed. Worth aligning the commit message ("cannot duplicate replies") with the doc.

5. **Two durability barriers on the same event stream use two different retry policies.** `persistTurn`/`commitRelayEvent` go through `persistEventStep` (120s cap, then the run fails); `persistArtifacts` and its ACK keep the original *unbounded* loop that only exits when the run turns terminal. A >120s consumer outage therefore fails the run at a `turn_done` but waits indefinitely at a `syncArtifacts`. Probably deliberate, but the doc only states the 120s window and never explains why the other half differs.

### Minor

- `.citation-manifest.json` now has **four** readers/constants (`knowledge-citation-tool.ts`, `overview-generator.ts`, `knowledge-package.ts`, and the new `libraries.ts`, which adds its own parser).
- `unlabeledByFile` is unbounded: every unlabeled page's path and title is retained forever to serve five samples per library.
- Reconnect budget: `attempts` is local to `driveCapabilitySession`, so it is per drive-call, not "per run" as the doc says — a re-attach resets it. Separately, is 6 attempts enough for a compile that runs for hours?
- `knowledgeLabelSample` walks every `.md` of every library and YAML-parses its frontmatter on **every** materialization, which runs at session start. Worth measuring on a several-thousand-page library.
- `ContainerEvidenceQueue.record` logs the full observation JSON for every distinct snapshot. In a namespace with frequent pod status churn that is a lot of log volume; the full dump could be limited to non-Running phases or snapshots carrying a termination.

### Deployment

The rollout section only covers the Runtime + compiler-box handshake, but this change spans **three images**:

- runtime/gateway — session-driver, run-manager, container-evidence, server.ts
- kbc box — `compile_box.py` / `event_relay.py` (`SICLAW_COMPILE_BOX_IMAGE`)
- **agentbox** — `src/knowledge/*`, `src/tools/query/knowledge-search.ts`, `src/agentbox/sync-handlers.ts`, `src/core/agent-types.ts`

The agentbox half needs its own rebuild, an `SICLAW_AGENTBOX_IMAGE` bump and a pod recycle — `imagePullPolicy: Always` only pulls on pod CREATE, so a live pod keeps the old code. Worth adding to the rollout notes.

### What is good

- `kb-authoring-command-protocol.md` states the ACK protocol, retry windows, the legacy boundary and the "no transparent in-place conversion across the protocol boundary" rule explicitly. Above-average doc work.
- `test_event_relay.py` targets the right things: replaying the same frame after a write loss, fencing the previous consumer mid-write, duplicate ACK idempotence, a legacy stream refusing to switch protocol (409), and a real aiohttp server exercising an HTTP-level reconnect.
- Making the cancel path fail closed on profile resolution (error instead of falling back to the default pod-name prefix) is the right call.
- `_resume_workspace_state`'s three-way split, and the rule that the control plane's decision beats local inference, is well argued: a regeneration clones the stable draft into staging, so file existence genuinely cannot distinguish an inherited page from this lineage's output.

kbc/platform/pod/compile_box.py
```diff
@@ -1723,11 +1772,36 @@ def _prepare_command(run: "CompileRun", command: dict) -> None:
...
+            # control plane wrote (never truncated); accept the inline list only
+            # when it is whole; otherwise refuse rather than misclassify.
+            produced = _load_recovery_provenance(run.workdir, count)
+            if produced is None and inline is not None and len(inline) == count:
+                produced = [_bounded_string(p, f"parameters.produced_pages[{i}]", required=True, limit=512) for i, p in enumerate(inline)]
+            if produced is None:
+                raise CommandRejected(
+                    f"recovery provenance is incomplete: {RECOVERY_PROVENANCE_PATH} is missing or does not list all "
+                    f"{count} produced page(s) and the inline list is partial", 409)
+        run._recovery = {"recovery_mode": mode, "produced_count": count, "produced_pages": produced or []}
```
jacoblee-io, line 1811: **`produced_pages` is never checked against the pages that actually exist, and a path-shape mismatch silently produces the most destructive outcome.**

The comment right above states the rule exactly — "a page missing from it is told to be rewritten as inherited ... otherwise refuse rather than misclassify" — and the partial-list refusal implements one half of it. The other half is missing.

`selfcheck.candidate_pages()` keys pages by **posix path relative to `candidate/`** (`a.md`), and `_render_command` prepends `candidate/` itself for display. If the control plane ever emits `candidate/a.md`, a backslash separator, or a `./` prefix, then in `_render_command`:

```python
produced_set = set(produced)
inherited = [page for page in all_pages if page not in produced_set]
```

`produced_set` matches nothing, every page becomes `inherited`, and the directive tells the model *"Regenerate each of them from raw/ (rewrite, do not keep)"* for the entire draft.

The only guard today is count agreement in `_load_recovery_provenance`, and a producer's own list always agrees with its own count — so the shape error is invisible by construction. The contract ("POSIX paths relative to `candidate/`") lives only in the protocol doc's prose; nothing enforces it.

Suggestion: in this branch, after `produced` is resolved, require `set(produced) <= set(selfcheck.candidate_pages(run.workdir))` and raise `CommandRejected(..., 409)` naming a few non-matching entries otherwise. That is the same "refuse rather than misclassify" call you already made for the partial list.

The new tests cover a partial inline list, a count mismatch and a pending plan overriding `complete`, but there is no case where a produced name matches no candidate page.

jacoblee-io (commented):
Second pass on `e6c75b7f` (rebased onto `abfb08f5`). **All six inline findings from the previous review are resolved, and resolved correctly** — please go ahead and mark those threads resolved. One new suggestion inline, and one note on the test run.

### Re-verified

**Single-library misdetection** — fixed the right way: roots now come from the manifest only, a section index stays navigation inside its library, and a legacy flat mount without a manifest stays one library. Re-ran my repro against this head:

```
manifest=true   multiLibrary: false  pageCounts: [[2],[2]]  libs: [""]
manifest=false  multiLibrary: false  pageCounts: [[2],[2]]  libs: [""]
genuine repos/a + repos/b   multiLibrary: true  selected: ["repos/a","repos/b"]
```

`pageCount` is back to the undivided per-mount count, so the scoring dilution is restored and the CHANGELOG's "field for field" claim now holds. Both directions have regressions.

**macOS archive fixture** — fixed. I re-ran the whole suite **without** `COPYFILE_DISABLE` this time and `sync-handlers.test.ts` is green.

**KB profile names** — fixed. `KB_BOX_PROFILE_NAMES` derives from `BUILTIN_PROFILES`, the selector and the projection share it, the phantom `kb-compile-pi` is gone, and the new test asserts all three paths agree per registered profile. `box-profile.ts` has no imports, so no cycle.

**Redundant branch / false `omitted` entries / tie threshold** — all fixed. Two things you caught that I had not flagged and that are worth keeping:

- rounding the margin difference before comparing (`0.7 - 0.55` is `0.15000000000000002`, which would have failed the `<= 0.15` test);
- making `compactRow` return the original row when nothing changes, which is what makes the new "only report an actual trim" rule in `trim()` correct rather than approximately correct.

### The two questions

Both answered in the doc, and the first answer is more candid than I expected:

> The existing authoring sink compares trimmed reply content with the most recent assistant reply in the same organization, repository and attempt; it does not use a durable event-id key. Consequently, separate identical replies can also be suppressed by that sink.

That is the honest framing — at-least-once persistence delivery, not exactly-once message creation — and it correctly identifies that the missing piece is a durable turn identity on the consumer contract. Worth a follow-up issue rather than anything in this PR: in a compile flow two consecutive genuinely identical replies are not far-fetched, so that sink can drop a real turn.

The retry-lifetime section and the new "Images and rollout order" section cover the other two points, including `imagePullPolicy: Always` only applying at Pod creation.

### Test run

`npx tsc --noEmit` clean. `npm test`: 373 files passed, 2 failed — `src/core/model-usage.test.ts` and `src/core/pi-sdk.contract.test.ts`. **Neither file is in this PR's diff**; both come from the rebase onto the new main, which added `postinstall: node scripts/install/pi-usage-patch.mjs` to patch `pi-ai`, and my worktree reused an unpatched `node_modules`. Not attributable to this PR. Python tests still not run locally (no aiohttp); CI covers them.

### Remaining minor items from the previous review

Left as-is, which is fine for this PR: four readers/constants for `.citation-manifest.json`, the unbounded `unlabeledByFile` map, `knowledgeLabelSample` walking every page at each materialization, and the full-JSON log line per container observation.

jacoblee-io (approved):
Third pass, on `69f745d7`. The unmatched-provenance guard is exactly what was needed, and the placement is better than what I suggested. **Approving.**

### The new guard

```python
missing = sorted(produced_set.difference(pages))
if missing:
    sample = json.dumps([page[:160] for page in missing[:5]])
    raise CommandRejected(f"recovery provenance references {len(missing)} page(s) absent from candidate/: {sample}; ...", 409)
```

I checked the ordering claim in the doc ("before the command id is accepted or a model turn is dispatched") against `handle_command` rather than taking it on trust:

- 7297 `_prepare_command`
- 7298 `_render_command` ← the 409 originates here
- 7299 `run._command_context = context`
- 7300 `run._accepted_commands[command_id] = digest`
- 7302 `_dispatch_authoring_turn`

So a rejection leaves no accepted id and no pinned operation generation, which is what makes "corrected provenance can retry the rejected id" true rather than aspirational — and the test asserts all three (`_accepted_commands == {}`, `_command_context is None`, `dispatch.assert_not_awaited()`).

Putting the check next to the classification instead of in `_prepare_command` is the right call: `produced_set.difference(pages)` needs the same `candidate_pages()` snapshot the classification uses, so splitting them would allow the set to be validated against one snapshot and applied to another.

Three details worth noting as deliberate rather than accidental:

- `pages` is the `candidate_pages()` dict and `set.difference(dict)` compares against its keys — the same posix-relative keys the classification consumes, so validation and use cannot diverge.
- The pending batch plan still wins: `state == "plan"` returns before this branch, and the new assertion in `test_existing_batch_ownership_precedes_size_and_mode` pins it with a `complete` payload naming a page the batch train has not materialized.
- `test_complete_rejects_missing_pages_before_dispatch` parametrizes over both sources and covers the case I would have missed: *"A complete inline preview must not hide an invalid authoritative file."* A valid inline list next to an invalid provenance file is precisely the shape that would have slipped through a naive check.

The bounded diagnostic (5 entries, 160 chars each, asserted under 1200) keeps a pathological provenance file from turning an error string into the payload.

### Status of everything raised across the three passes

| Round | Finding | Status |
|---|---|---|
| 1 | Single-library mount misdetected as multi-library | Fixed, re-verified against a repro |
| 1 | New archive fixture fails on macOS | Fixed, suite green without `COPYFILE_DISABLE` |
| 1 | Three definitions of "a KB box", one naming a nonexistent profile | Fixed, derived from `BUILTIN_PROFILES` |
| 1 | Redundant `score === 0` branch | Fixed |
| 1 | `omitted` reporting trims that did not happen | Fixed |
| 1 | Tie threshold inconsistent with `LIBRARY_ROUTE_MARGIN` | Fixed |
| 1 | `turn_done` dedup window narrower than claimed | Answered; the doc now states at-least-once delivery and names the sink's content-comparison limitation |
| 1 | Two retry policies on one event stream | Answered; artifact policy and the window's start point documented |
| 1 | Three images, rollout notes covered two | Fixed; "Images and rollout order" added |
| 2 | `produced_pages` unvalidated against `candidate_pages()` | Fixed in `69f745d7` |

The minor items I flagged and did not press — four readers of `.citation-manifest.json`, the unbounded `unlabeledByFile` map, `knowledgeLabelSample` re-walking every page at each materialization, and the per-observation full-JSON log line — are all reasonable to leave for follow-up.

One thing worth a separate issue rather than this PR: the authoring sink's content-based dedup can suppress two genuinely distinct but identical assistant replies. The doc now says so plainly, and the fix is a durable turn identity on the consumer contract, which is out of scope here.

### Verification on this head

- TypeScript is unchanged between `e6c75b7f` and `69f745d7` (the diff is the protocol doc plus two Python files), so the run I did on `e6c75b7f` stands: `tsc --noEmit` clean, `npm test` 373 files passed with two failures in `src/core/model-usage.test.ts` and `src/core/pi-sdk.contract.test.ts` — neither file is in this PR's diff; both need the `pi-ai` postinstall patch the rebased main introduced, and my worktree reused an unpatched `node_modules`.
- **I did not run the Python suite locally** — `claude-agent-sdk` is not installable in my environment, so `compile_box` cannot be imported here. `test_resume_routing.py` and `test_event_relay.py` are both wired into `kbc-ci.yml`, so CI is the evidence for the box half, not me. Please confirm that job is green before merging.

### Deployment reminder

Three image roles, as the new rollout section says: Runtime, compiler box (`SICLAW_COMPILE_BOX_IMAGE`) and AgentBox (`SICLAW_AGENTBOX_IMAGE`). Drain active compiles across the protocol boundary, and recycle AgentBox Pods after bumping their image reference — `imagePullPolicy: Always` only pulls at Pod creation.

PR #587, opened 2026-09-11
runtime: reconnect a lost box event stream before failing the run

LikiosSedo:
## Why

The Runtime↔box relay is a single SSE connection with no reconnect. A transport error (SSE drop, apiserver-side proxy hiccup) ended the run as `relay_failed`, and the relay's `finally` stopped the box — throwing away the in-flight batch of a healthy box. The 2026-09-10 harness review ranked this the second stability risk after plan-less recovery.

## What

- `driveCapabilitySession` wraps the stream in a bounded reconnect loop: on a stream error, if the run is not already terminal and the box still answers `/health`, wait (2s→60s backoff, 6 attempts) and re-open `/events/:runId?replay=1` so an in-flight sync batch is delivered again (persistArtifacts is idempotent; duplicate ACKs are accepted). A dead box or a spent budget rethrows, so the caller's `relay_failed` path is unchanged. A stream error after the run already settled is ignored (no spurious relay_failed on a finished run).
- Policy is injectable (`reconnect` option) for tests; production uses the defaults.
- New counter `siclaw_gateway_capability_relay_reconnects_total`.

## Tests

`session-driver.test.ts`: reconnect with replay relays events from both connections; dead box gives up immediately; attempt budget exhausted rethrows; error after terminal is ignored. Capability suite: 100 passed.

Companion: sicore MR !1180 (recovery modes / breaker classes / sweeps) and siclaw #586 (plan-less `compile.resume`).

### DIFF
--- a/src/gateway/capability/capability-metrics.ts
+++ b/src/gateway/capability/capability-metrics.ts
@@ -37,3 +37,9 @@ export const capabilityRelayFailuresTotal = new Counter({
   help: "Capability box event relays that ended with an error",
   registers: [federationSelfRegistry],
 });
+
+export const capabilityRelayReconnectsTotal = new Counter({
+  name: "siclaw_gateway_capability_relay_reconnects_total",
+  help: "Capability box event stream reconnects (stream lost while the box was still alive)",
+  registers: [federationSelfRegistry],
+});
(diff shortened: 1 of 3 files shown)


LikiosSedo:
Review follow-up pushed (`20860738`): an absent run record now counts as settled (endRun drops it), the check repeats after the health probe and after the backoff, and only transport errors are retried — relay-originated errors still fail the run. The reviewer's four real-manager cases pass unchanged.

PR #589, opened 2026-09-11
fix(core): ground task plans and preserve thinking defaults

jacoblee-io:
## Summary

Investigations were forced to create task lists before understanding the procedure, and late model binding could leave the intended `high` thinking level at `off`. Form and revise verifiable milestones from available context, and preserve the thinking default through an instance-only settings override that respects explicit preferences.

### Solution

- Allow focused discovery before detailed planning, describe concrete completion criteria, refine unfinished milestones, and reduce repeated bookkeeping. Existing ledger and UI contracts remain compatible.
- Preserve the session thinking default across deferred model binding without writing a global user preference; cover the transition through the installed SDK and actual HTTP serialization.
- Add a synthetic MCP/Skill replay harness with resident-instruction checks and terminal-response validation. The [source comparison and deployment report](docs/design/2026-09-11-planning-alignment.md) records the approach, evidence, and limitations.

## Test Plan

- [x] `npm test`: 7,111 passed, 2 skipped.
- [x] `npx tsc --noEmit` and `npx tsc -p tsconfig.agentbox.json --noEmit`.
- [x] `npm run build`, smoke-script syntax checks, and `git diff --check`.
- [x] Replay CLI regression: late SSE errors fail and stop the batch; an earlier error followed by a completed response can recover. Both event orders pass regression tests, including exit status and saved artifacts; the new test also passes a strict TypeScript check.
- [x] Audited 133 historical replay traces, including the final 32-case matrix: no previously accepted replay is invalidated by the stream-error fix.
- [x] Installed-SDK regression: late model binding retains `high`, the request serializes the intended effort, explicit overrides win, and the default is not persisted globally.
- [x] Isolated K8s deployment and real-model replay: two models, eight scenarios each, with actual image and resident-instruction verification. All 16 candidate cases returned final text; the main comparison retains one empty-final-response failure and a separately retried provider HTTP 400.
- [ ] Full behavioral acceptance: Sonnet still marked failed incomplete-trace verification completed in 3/3 runs. GPT showed improved discovery order, scope revisions, and failed-verification status. Broader parity with Codex and Claude Code remains unverified.

## General Checklist

- [x] Changes are focused on the planning behavior and its session configuration.
- [x] Commit message follows Conventional Commits.
- [x] No unrelated changes included.


scripts/smoke/planning-replay.mjs
```diff
@@ -0,0 +1,147 @@
...
+          buffer = buffer.slice(split + 2);
+        }
+      }
+      buffer += decoder.decode();
+      if (buffer.trim()) accept(buffer);
+      const assistantMessages = events.filter((e) => e.data?.type === "message_end"
+        && e.data.message?.role === "assistant").map((e) => e.data.message);
+      const lastAssistant = assistantMessages.at(-1);
+      const streamError = lastAssistant?.stopReason === "error"
+        || (events.some((e) => e.event === "error") && lastAssistant?.stopReason !== "stop");
```
LikiosSedo: [P2] Treat stream errors after the final answer as failures

The `lastAssistant?.stopReason !== "stop"` condition suppresses every SSE `error` whenever the last assistant response completed, including errors that arrive **after** that response. This is reachable: Runtime forwards `message_end` before finishing SSE consumption; a subsequent stream or deferred persistence failure emits `stream_error` and `prompt_done` (`src/gateway/server.ts:1255-1271`), which Portal translates to SSE `error` and `done`.

I reproduced the sequence `message_end(stop, text) -> error(STREAM_INTERRUPTED) -> prompt_done -> done` by executing the unchanged runner with in-memory fetch/filesystem bindings. It exits with code **0**, records `errors: 1`, and leaves `failure` unset, so a failed replay is counted as successful and the batch can continue.

Please classify errors by event ordering: an error after the final successful response must fail the replay; an earlier error followed by a successful recovery may pass. Cover both sequences in the regression check.

LikiosSedo: Verified fixed in 9bc62879221225c237b352f04824b27a82819d9a. I reran the original in-memory reproduction against the updated, unchanged runner: an error after the final answer now produces exit code 1 and records failure in both the artifact and summary; an earlier error followed by a completed answer still passes with exit code 0.

The two new CLI regression tests also passed in CI, covering batch stop/continue behavior and persisted results. The new test file passes a strict TypeScript check. No new findings in the follow-up diff; this P2 is addressed.

LikiosSedo (approved):
Reviewed at 9bc62879221225c237b352f04824b27a82819d9a. The reported P2 is verified fixed: terminal stream errors after a completed response fail the replay, while successful recovery after an earlier error still passes. The new regression tests and all four CI checks pass. No remaining findings from this review.

PR #595, opened 2026-09-12
Record model usage across chat, analysis and compiler execution

LikiosSedo:
Record independent model-call observations for chat, analysis, routing attempts, compaction and compiler roles. Capture provider-native numeric evidence before normalization and preserve missing fields across supported API and subscription models. SDK-only usage remains unknown, and interruption outcomes remain distinct.

AgentBox delivers through its durable outbox. Compiler observations bypass the diagnostic queue, enter a durable Runtime outbox and retry independently of run completion. The consuming control plane owns workload, Agent, user, repository/profile and frozen model attribution.

Validation: TypeScript build, focused parser/recording tests and compiler measurement tests passed. Broader regressions retain previously reproduced self-check failures. Integration checks used synthetic inputs for real chat, analysis and knowledge compilation, reconciling recorded evidence and attribution. Deploy the paired ingestion backend before the execution components.

Coverage begins with instrumented dispatches. Compiler delivery before Runtime receipt, surviving outbox storage and SDK-internal HTTP retries remain boundaries. This measures observed token consumption; invoice pricing and remaining subscription allowance are separate.

### DIFF
--- a/.github/workflows/kbc-ci.yml
+++ b/.github/workflows/kbc-ci.yml
@@ -4,10 +4,10 @@ name: KBC Box CI
 on:
   push:
     branches: [main]
-    paths: ["kbc/**", "src/core/**", "src/kbc/**", "package*.json", "tsconfig.json", ".dockerignore", ".github/workflows/kbc-ci.yml"]
+    paths: ["kbc/**", "src/core/**", "src/kbc/**", "package*.json", "patches/**", "scripts/install/**", "tsconfig.json", ".dockerignore", ".github/workflows/kbc-ci.yml"]
   pull_request:
     branches: [main]
-    paths: ["kbc/**", "src/core/**", "src/kbc/**", "package*.json", "tsconfig.json", ".dockerignore", ".github/workflows/kbc-ci.yml"]
+    paths: ["kbc/**", "src/core/**", "src/kbc/**", "package*.json", "patches/**", "scripts/install/**", "tsconfig.json", ".dockerignore", ".github/workflows/kbc-ci.yml"]
 
 # Least-privilege: this job only reads the repo (no PR writes, no packages).
 permissions:
--- a/Dockerfile.agentbox
+++ b/Dockerfile.agentbox
@@ -20,6 +20,8 @@ WORKDIR /app
 
 # Dependencies layer (cached unless package*.json changes)
 COPY package*.json ./
+COPY patches/ ./patches/
+COPY scripts/install/ ./scripts/install/
 RUN npm ci
 
 # Copy backend source only (AgentBox runtime graph: agentbox-main + core + tools + knowledge).
@@ -97,6 +99,8 @@ WORKDIR /app
 
 # Production deps (cached unless package*.json changes)
 COPY package*.json ./
+COPY patches/ ./patches/
+COPY scripts/install/ ./scripts/install/
 RUN npm ci --omit=dev
 
 # Copy compiled JS + skills + entrypoint (knowledge synced at runtime via bundle API)
--- a/Dockerfile.portal
+++ b/Dockerfile.portal
@@ -18,6 +18,8 @@ FROM node:22.19-slim AS backend-builder
 WORKDIR /app
 
 COPY package*.json ./
+COPY patches/ ./patches/
+COPY scripts/install/ ./scripts/install/
 RUN npm ci
 
 COPY tsconfig.json ./
@@ -34,6 +36,8 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 WORKDIR /app
 
 COPY package*.json ./
+COPY patches/ ./patches/
+COPY scripts/install/ ./scripts/install/
 RUN npm ci --omit=dev
 
 # Backend JS
--- a/Dockerfile.runtime
+++ b/Dockerfile.runtime
@@ -8,6 +8,8 @@ WORKDIR /app
 
 # Dependencies layer (cached unless package*.json changes)
 COPY package*.json ./
+COPY patches/ ./patches/
+COPY scripts/install/ ./scripts/install/
 RUN npm ci
 
 # Copy source and compile
@@ -26,6 +28,8 @@ WORKDIR /app
 
 # Production deps
 COPY package*.json ./
+COPY patches/ ./patches/
+COPY scripts/install/ ./scripts/install/
 RUN npm ci --omit=dev
 
 # Copy compiled JS
--- a/docs/design/model-usage-observations.md
+++ b/docs/design/model-usage-observations.md
@@ -5,12 +5,20 @@ and subscription models use the same observation contract. Routing retries
 create separate calls; HTTP retries hidden inside Pi remain one logical call.
 This is runtime consumption telemetry, not an invoice or remaining allowance.
 
+Chat, system-analysis sessions and compiler workers share this observation
+contract. Workload ownership is resolved by the consuming control plane from
+the authenticated session or capability run. Compiler sessions retain their
+native identity and executor role; they do not impersonate a chat Agent.
+
 ## Evidence and identity
 
 Each call has a UUID, immutable dispatch timestamp, session and trace identity,
 model configuration snapshot, routing attempt and agent/auxiliary kind.
 Started and finished observations share that identity. Failed, cancelled and
 auxiliary calls are observed independently of visible assistant messages.
+Request IDs group calls within a prompt. Chat uses the root trace when present;
+compiler roles use their turn ID. Parent-call identity is optional and is not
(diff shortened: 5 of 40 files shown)


jacoblee-io:
Reviewing this alongside #597 (and its sicore counterpart !1209), which attacks the same problem from the other end. Capturing inside pi, before `parseChunkUsage` normalises, is the better cut — it removes the raw-vs-normalised question entirely, and that question is not academic: #597 got it wrong three times in review, because pi's `openai-responses-shared.js:445` does `Math.max(0, input_tokens - cached - cacheWrite)` and an envelope's `input` is therefore cache-exclusive while the provider's raw `input_tokens` is not.

Three observations from the aggregation side, all grounded in the diff rather than a run.

**1. `partial` rows are excluded from the sums, so real spend leaves the total silently.**

`SUM(CASE WHEN u.usage_status = 'reported' THEN u.<field>_tokens END)` is a safe rule — it never states a wrong number. But a call that reported input and nothing else DID consume input, and under this rule those tokens are absent from the headline while `partial` is the only hint that anything is missing. A window where most calls are partial shows a small, confident-looking total.

Worth considering: keep the strict sum, but put coverage next to it wherever the total is read, not only in the status row — the reader has to be able to tell "we spent little" from "we measured little".

**2. Is the headline `total` the provider's own `total_tokens`?**

If so it is the right choice where it exists, but Anthropic does not send it — its usage carries `input_tokens` / `output_tokens` / `cache_creation_input_tokens` / `cache_read_input_tokens` and no total. Anthropic traffic would then be missing from the headline card with nothing on screen saying so.

The obvious substitute is not safe either: `input + output` is correct for raw openai-shaped readings (cache already inside `input`) and **wrong for Anthropic**, where the cache figures are charged on top. A concrete case: `input=100 output=20 cacheRead=1000 cacheWrite=200` costs 1320, not 120. Raw capture makes this distinction matter more, not less, because the raw openai figure really does include the cache.

**3. Offer: the scenario axis, and a set of counter-examples.**

#597 carries an axis this PR does not — which part of the product spent it: conversation vs knowledge compilation vs system analysis — and `sourceKind: api|subscription` here is orthogonal to it, not a replacement. Knowledge-compile and analysis spend is platform overhead; summing it into user traffic makes a busy build night read as growth.

I also have a set of counter-examples that took several review rounds to arrive at, mostly about absence surviving aggregation: a call with no `usage` key at all (distinct from `usage:{}` — they take different code paths), two partial calls combining into a complete-looking total, an unreported field rendered as 0, and a workload with no producer yet rendered as free rather than unknown. Happy to port them against this implementation if useful.

**Coordination, and the reason for this comment:** #595/!1207 and #597/!1209 are two implementations of the same feature on the same base (`304290c8d`), touching the same files — `src/core/llm-call-recorder.ts` here, and on the sicore side `internal/db/migrate.go`, `metrics/handler.go`, `web/messages/*.json`, plus a package and an endpoint and a panel that differ only in name (`modelusage`/`llmusage`, `model_usage.go`/`token_usage.go`). Merging both would leave two answers to "how many tokens did this cost" with nothing to say which is authoritative.

My read is that this PR should be the base — the capture point is better, and the subscription dimension is missing from mine entirely — with the scenario axis and the absence-handling contracts folded in rather than kept as a parallel stack. Flagging it here so the decision is made deliberately rather than by merge order.

jacoblee-io:
Re-reviewed at `f36fd193` (with sicore `07851957b`). All three points from the earlier comment are addressed, and two of them are solved better than the version I was proposing — making `input` uniformly cache-inclusive at write time, rather than teaching every consumer the per-protocol rule, is the right call: one convention settled once beats a rule each reader has to remember.

I also ran the counter-examples that took several review rounds to arrive at against this implementation. **All pass**, including the one I most expected to break: Anthropic `input=100 output=20 cacheRead=1000 cacheWrite=200` prices at 1320. Also checked `modelusage` / `metrics` / `capability` — green.

One concern I had turned out to be unfounded and is worth recording so nobody re-raises it: I thought a cache figure dropped by the `invalid_*_subset` check would leave `input` inflated by a value that now reads as null. It cannot — once `input` is made cache-inclusive, `cache <= input` holds by construction, so the check never fires on that path.

---

## One thing I do think is worth fixing: a missing patch is indistinguishable from silent providers

The installer itself is careful — before/after SHA-256 on every file, idempotent re-runs, all files verified before any is written, hard failure on a version or hash mismatch, and it follows the dependency graph into nested copies. No complaints there.

But it runs as `postinstall`, and `npm ci --ignore-scripts` is ordinary practice in CI and hardened builds. When that happens:

1. the patch never applies, so pi emits no `output.providerUsageEvidence`
2. `llm-call-recorder.ts:544` — `message?.providerUsageEvidence ?? call.evidence` — yields `undefined`
3. every observation normalises to `status: "missing"`
4. the panel reads `0 reported / N finished`

So it is **visible but not diagnosable**. "This build shipped without the patch" and "every provider stopped reporting usage at once" render identically, and the two call for completely different responses — rebuild the image, versus go ask the upstream.

That is this project's own thesis playing out one level up: absence in the *data* is handled carefully, but absence of the *collector* has no representation at all. `grep` finds `providerUsageEvidence` only in `llm-call-recorder.ts` and the tests — there is no startup check.

Cheap fix: have the patch leave a sentinel in the modules it edits (a version constant, or an exported marker), read it once at startup, and carry it in `UsageHealth` — `collectorPatched: false` is worth far more than a screen of `missing`. Pinning `--no-ignore-scripts` in the Dockerfiles only protects your own images, not somebody installing the published package.

## Two nits

**`thinking_tokens` on the Anthropic branch.** `normalize.go` reads reasoning from `output_tokens_details.thinking_tokens` for `anthropic`, and Anthropic's usage carries no such path — thinking tokens are already inside `output_tokens`. Harmless (it resolves to nil), but the line's presence suggests Anthropic reasoning is obtainable. Either drop it or say in a comment that it is reserved and unreported today.

**Outbox drops the newest, not the oldest.** `record()` returns early past capacity, keeping what is already on disk. That is defensible — durable records survive, ordering is preserved — and the opposite choice (keep the freshest picture) is equally defensible. Worth one line of comment on which was intended, or the next reader will "fix" it. The accounting around it is fine: counted, warned, and carried in health.

---

For the record: #597 and sicore !1209 are being closed in favour of this pair. The capture point here is better, and the subscription dimension was missing from mine entirely. The prompt/token *efficiency* analysis that rode along in those branches — where the ~17k per-turn fixed overhead goes, and how Claude Code and codex keep cache bytes stable — is unrelated to metering and is being taken up separately rather than dropped.

jacoblee-io:
Correcting my own review above: **the postinstall concern was wrong.** I raised it from reading the installer alone, without checking how the images are actually built. Having now checked, there is nothing to fix — please don't spend time on it.

What refutes it:

- `--ignore-scripts` appears nowhere — not in any Dockerfile, not in the CI workflows, not in the Makefile. `npm ci` runs `postinstall` by default.
- `diff` is a **production** dependency (`dependencies`, `^8.0.3`), not a dev one, so `npm ci --omit=dev` still resolves the installer's own import.
- `Dockerfile.agentbox` copies `patches/` and `scripts/install/` **before** `RUN npm ci --omit=dev` — deliberately, so the patch is present when the hook fires.

And the installer has no silent-failure mode of its own: it pins the pi-ai version, hashes every file against the manifest, verifies the result of `applyPatch` before writing anything, and throws on any mismatch. A run either patches or fails loudly. The only way to get an unpatched install is for the hook never to run, and no path here does that.

So the "missing patch is indistinguishable from a silent provider" scenario needs someone to deliberately install with `--ignore-scripts`. That is worth a line in a contributing note at most, not a startup check.

I'm also withdrawing the `thinking_tokens` remark. Anthropic counts extended-thinking tokens inside `output_tokens` and emits no `output_tokens_details`, so that read resolves to nothing and leaves `Reasoning` unknown — which is the correct outcome. Since `anthropic-messages` is a protocol other vendors implement, a speculative read there is defensive rather than mistaken.

The outbox capacity remark stands as a nit only: dropping the newest observation when full is a defensible choice, and it is counted and surfaced rather than swallowed.

One unrelated thing that did turn out to be real, on the sicore side rather than here: compile usage was attributed to `attempt.OwnerID`, which is the knowledge base's owner rather than whoever opened the attempt — both attempt-creation sites set `OwnerID: ownerOf(lockedRepo)` alongside `CreatedBy: userID`. A per-user token view would have billed a repo's owner for someone else's compile. Fixed in `650cc69a1` on the matching MR, with the existing test's fixture given a `created_by` distinct from `owner_id` so it exercises the primary path rather than the fallback.

nightmeng (approved):

PR #597, opened 2026-09-13
Record every LLM call with its provenance intact

jacoblee-io:
## Why

Billing showed an unexplained `cache_write` share, and nothing in the product could answer where the tokens went. Message metadata never covered sub-agent calls, and the one usage figure Portal stored was never aggregated.

## The hard part is not summing tokens

It is keeping **absence** honest. Four different code paths produce an all-zero usage object and they are indistinguishable once written:

1. The SDK zero-fills `usage` **before the request leaves**, so at the streamFn boundary "reported 0" and "never reported" are the same value.
2. A rehydrated transcript fills `ZERO_USAGE` — and not only on handoff: eviction and a pod restart rebuild from the control plane too, so a per-session verdict is always wrong.
3. A derived figure gets `Math.max(0, …)` clamped.
4. A provider genuinely reports zero. `cache_read = 0` is an ordinary answer.

So provenance is **recorded at the point of production and never inferred**. Usage is read from the raw response at the HTTP boundary via `options.fetch` (a public SDK input; the response body is tee'd and the original bytes pass through untouched), and `null` and `0` stay different values end to end. A derived figure that comes out impossible — reasoning > output, cache > input — is reported as `null` plus an inconsistency record, never clamped: the clamp is exactly how the upstream adapter loses the same evidence.

## Correlation follows the same rule

`root_request_id` and `parent_call_id` are carried by the **producer**, because only the box knows which turn a call served and which of its calls spawned a child. Deriving either from session lineage collapses every request in a conversation into one id.

- `turnId` is reused rather than minting a second identifier — it already names one accepted prompt execution and survives model-routing retries.
- The **channel and scheduled entries now mint one at the entry**. They called `AgentBoxClient` directly and carried no id at all, so every call they made was uncorrelated. It cannot be minted inside the client: a turn id names a turn, while the client runs once per *attempt*, and a busy-retry would split one message into two correlations.
- `parent_call_id` is the agent call that just sealed — the round whose tools are executing. Aux calls are skipped: a summarisation issues no tools, so naming it would attribute a spawn to a call that could not have made it.

## Reading it back

Portal aggregates per provider × model and per user. Two rules carry the collection work through to the screen:

- **Every sum ships with the count of calls that reported it.** `SUM` skips NULLs, so a total over half the calls looks exactly like a complete one.
- **Ranking is server-side**, because the result is capped: the top 50 by total tokens is not the set holding the top 50 by cache write.

Untrustworthy rows (`sdk_default`, `rehydrated`, `unknown`) are excluded from the sums and counted separately, so the page states its coverage instead of implying completeness.

## Deploy order

Reverse of the data flow — receiver first: control plane (DDL + persist handler) → Runtime (route, validation, authorization) → AgentBox. Each stage degrades rather than breaking: a new box against an old Runtime retries, drops, and increments `siclaw_metering_unconfirmed_total`, and no conversation is affected. `imagePullPolicy: Always` only pulls on pod CREATE, so verifying against an existing session proves nothing.

Full analysis and contracts: `docs/design/2026-09-12-token-cost-analysis.zh-CN.md`.

## Testing

373 files / 7596 tests passing, `tsc --noEmit` clean. portal-web: 32 files / 280 tests, `tsc` clean.

Not yet verified end to end against a real box → Runtime; `cost_micros` stays `null` until unit prices are confirmed, because a wrong cost is worse than an absent one.

### DIFF
--- a/docs/design/2026-09-12-token-cost-analysis.zh-CN.md
+++ b/docs/design/2026-09-12-token-cost-analysis.zh-CN.md
@@ -1,0 +1,963 @@
+# siclaw Token 成本分析与优化方案
+
+> **修订 r6（2026-09-13）** —— 基线 `origin/main` @ `946e0675`。
+> **P0 已实现**，含接收端与事实表；实现记录与剩余缺口见 **§7.8.9**。
+> **P0 方案定稿见 §7.8**，五项决策与三处补正均已并入。
+> r1 有多处实质错误已被推翻，逐条列于「修订记录」。**r1 的诊断路径不可执行，勿按 r1 实施。**
+>
+> **r3 相对 r2 的改动**：①「子代理历史无可回灌」过满，改为「可回灌范围待查」（pi 的 JSONL 另有一份，且无正文的 assistant 也能带 usage）；② `PI_CACHE_RETENTION=long` 不等于直接拿到 24h（分支可能发 `ttl=30m`）；③「50 item = 50 次前缀 write」撤回为「前缀复用率未知」。
+
+## 修订记录：r1 被推翻的结论
+
+| r1 的说法 | 实际 | 证据 |
+|---|---|---|
+| 「数据完整，只缺聚合」「无需新增采集」「历史不丢」 | **错。** subagent 走独立持久化路径，assistant 行的 metadata 只有 `phase`/`stop_reason`/`assistant_item`，**没有 `llm_call`**；且 `if (text)` 守卫使无正文的调用不落行。**头号怀疑对象恰恰在 Portal 侧无数据可查**（r3 补正：pi 的 JSONL 另有一份完整 assistant 消息且可携带 usage，可回灌范围见 §7.7 待查清单）| `session.ts:2921-2945` |
+| §3.1 的 SQL 能区分三个假说 | **错。** `round` 是 **prompt 内**轮次、每个 prompt 重置，故「rnd=1 多」只说明 prompt 多，不能证明 subagent 多；首轮 `since_prev_ms` 是准备耗时而非距上次提问的间隔 | `llm-call-recorder.ts:49` |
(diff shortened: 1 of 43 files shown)


jacoblee-io:
Closing in favour of #595, which solves the same problem better.

**Why this one loses on the axis that matters.** Both capture provider usage; #595 captures it *inside* pi, before `parseChunkUsage` normalises. This one tees the SSE body through `options.fetch` and re-parses it from outside. The difference is not stylistic: pi's `openai-responses-shared.js:445` does `Math.max(0, input_tokens - cached - cacheWrite)`, so an envelope's `input` is cache-exclusive while the raw figure is not — and this PR got that distinction wrong three times across review before settling it. #595 never has the question, because it reads the provider's own words.

**And it covers ground this one does not.** Subscription calls (`sourceKind: api|subscription`, `codex_responses`) are absent here entirely — under a subscription the question is quota, not billed tokens. #595 also records started/finished as separate phases, so a call that never returns still leaves a trace; this PR only records at seal time, so such a call simply never existed.

**What was checked before closing, rather than assumed.** The three aggregation concerns raised on #595 have been addressed in `f36fd193` / sicore `07851957b`, and the counter-examples this PR accumulated over several review rounds were run against that implementation — all pass:

- a call with no `usage` key at all (a different code path from `usage:{}`, and the one that is easy to miss)
- a partial reading not producing a complete-looking total
- Anthropic `input=100 output=20 cacheRead=1000 cacheWrite=200` priced at 1320, not 120
- an intermediate reading not counted as final
- an unreported field staying unknown rather than rendering as 0

Their fix for the Anthropic case is better than the one here: instead of making every consumer protocol-aware, they normalise `input` to a cache-inclusive figure at write time, so `input + output` holds everywhere. One convention settled once beats a rule each reader has to remember.

**Not carried over, and deliberately so:** `docs/design/2026-09-12-token-cost-analysis.zh-CN.md` §1–6 and `2026-09-12-token-efficiency-comparison.zh-CN.md` are about prompt/token *efficiency* — where the ~17k per-turn fixed overhead goes, and how Claude Code and codex keep their cache bytes stable. That work is unrelated to metering and is being taken up separately rather than dropped.

Two implementations of one feature would leave "how many tokens did this cost" with two answers and nothing to say which is authoritative. Closing so the answer is #595.

PR #599, opened 2026-09-14
feat(qa): enable scoped research tools and Skill discovery

LikiosSedo:
Knowledge QA can discover Skills, execute local Skill helpers, save working results, plan research and dispatch scoped research sub-agents. The preset adds local script execution independently of node, Pod and host script tools; memory remains disabled.

## Scope and execution boundaries

- Agent-scoped sessions filter ambient user-global Skills before the first resource sync, even when the resolved directory does not exist. This shared rule covers QA, SRE and Custom. Standalone unscoped SRE/Custom CLI sessions retain their compatibility fallback.
- Bundled and platform discovery remains independent of tool permissions. An older Runtime's read-only QA whitelist cannot disable the scope filter. Explicit inheritance switches and per-Skill masks remain authoritative.
- Operational Safety follows execution capability, including QA local scripts. Prompt inspection reports missing safety guidance. QA working-file guidance names the actual user-data directory.
- Factory Write/Edit reject paths outside their allowed workspace and pre-existing symbolic links within the write path. Real factory tests exercise the resource loader and SDK file operations, with no provider request.
- Children retain the parent capability/resource scope; existing Session and registry gates control recursion and planning. Channel work remains foreground-only. The registry fixture's test name describes its actual coverage.

## Validation

- Before the fix, real factory regressions failed for ambient Skill exposure in QA/SRE/Custom and for a workspace symlink redirecting a write. They pass after the fix; unscoped CLI discovery and normal workspace edits still pass.
- Final focused checks: 47 tests across four files pass; TypeScript no-emit checking passes.
- Full backend suite: 377 files and 7,605 tests pass. TypeScript build passes. The first sandboxed run could not exercise local HTTP servers/process-group checks; the complete rerun with those permissions passed.
- No new capability keys in this follow-up; the paired generated capability list still matches all 14 keys.

## Rollout

Deploy matching Runtime and AgentBox builds, and rebuild the standalone Portal frontend for the preset and picker. New and existing QA sessions receive the preset after rollout and session rebuild. Local Skill scripts remain trusted code using existing resource bindings; prompt safety is not an OS sandbox. No schema migration is added. Deployed conversation and real-provider acceptance remain pending.


docs/design/invariants.md
```diff
@@ -116,10 +116,13 @@ For any scoped Agent (Portal or Gateway materialization), pi-coding-agent's
...
 harness: Portal/Gateway materialized bindings, repo-bundled operational skills
-when execution is permitted, and platform authoring skills when write/preview
-capabilities are permitted. QA, Coordinator, delegated read-only, and unresolved
-harnesses therefore cannot inherit ambient SRE skill context. Standalone,
-unscoped SRE headless CLI sessions retain the legacy repo/global skill fallback.
+for operational and Knowledge QA sessions, and platform skills for Knowledge QA
+or sessions with write/preview capabilities. Explicit Skill inheritance switches
+and per-name masks still apply. Unresolved harnesses cannot inherit ambient
+Skill context. Filtering applies to every Agent-scoped session even before its
+first Skill sync creates `resolved/`; bundled discovery does not disable it.
```
jacoblee-io, line 123: This sentence, and the matching "repo-bundled operational skills for operational and Knowledge QA sessions" above it, describe a contract the code does not implement in steady state.

`resolveSkillDirectories` makes bundled roots an **`else` branch** of `resolved/`, not a sibling:

```ts
} else if (fs.existsSync(resolvedSkillsDir)) { push(resolved) }
else if (input.includeBundledSkills)     { push(core, extension) }
```

So for a Knowledge QA box the observable effect of `includeBundledSkills: true` is: `skills/core` (27 SRE diagnostics) is in the prompt **only before the first Skill sync**, and never after it. I measured this on a materialized tree:

```
modelVisibleRoots:      [".siclaw/skills/resolved"]
bundledRootVisible:     false
localScriptCanRunBundled: true
```

Two consequences worth deciding on deliberately:

1. The window where the docs are true is the transient one, and it is also the least desirable configuration (a QA agent advertising volcano/pod/node diagnostics it has no cluster binding for).
2. `local_script` resolves image built-ins through its **own** fallback (`script-resolver.ts` `getSkillScriptDirs` step 4, gated only by `.inherit-builtins.json`), so post-sync a QA agent can *execute* a bundled Skill script it cannot *see*. That asymmetry predates this PR, but QA reaching it is new.

Either intent is defensible — bundled Skills alongside `resolved/`, or `includeBundledSkills: false` for QA so the pre-sync window stops leaking 27 descriptions. What should not ship is the third state, where the design doc asserts one and the resolver implements the other. Per CLAUDE.md, design docs record contracts; this one currently records a contract that holds for a few seconds at box startup.

src/core/agent-factory.ts
```diff
@@ -344,6 +344,24 @@ function assertToolPathAllowed(
   blockedDirs: Array<{ dir: string; reason: string }>,
 ): void {
   assertPathAllowed(absolutePath, allowedDirs, operation);
+  if (operation === "write" || operation === "edit") {
+    // New files may not exist yet. Check each component beneath the allowed
+    // workspace before mkdir/read/write so a pre-existing link cannot redirect
+    // a file tool into another resource directory.
+    const root = allowedDirs.find(dir => isPathInsideDir(absolutePath, dir))!;
```
jacoblee-io, line 351: The non-null assertion rests on two predicates agreeing, and they are not the same function: `assertPathAllowed` (line 325) compares against the raw `dir`, while `isPathInsideDir` (line 336) compares against `path.resolve(dir)`.

Today they agree because `writeAllowedDirs` is `[userDataDir]` and `userDataDir` is already `path.resolve`d, so `find` always hits. But if an unresolved or trailing-slash directory ever enters `allowedDirs`, `find` returns `undefined` and the next line throws `TypeError: The "path" argument must be of type string` instead of the clean denial `assertPathAllowed` would have produced — and the `!` makes TS silent about it.

Cheap fix: drop the `!` and throw the ordinary `${operation} blocked: ...` denial when no root matches.

src/core/agent-factory.ts
```diff
@@ -344,6 +344,24 @@ function assertToolPathAllowed(
...
+    let current = path.resolve(root);
+    for (const part of ["", ...path.relative(current, path.resolve(absolutePath)).split(path.sep)]) {
```
jacoblee-io, line 353: The leading `""` makes the first iteration `path.join(root, "")` — i.e. this lstats the **allowed root itself**, not only components beneath it, which is what the comment above says.

That is a deployment constraint rather than a bug: if `.siclaw/user-data` is ever materialized as a symlink (aliasing a mount), every write and edit fails with `symbolic links are not writable workspace paths`, and the message will not suggest the workspace root as the cause. Worth making explicit — either state in the comment that the root must be a real directory, or start the walk at the first component below it.

(Ancestors above the root are fine: `lstat` only declines to dereference the final component, so a symlinked `cwd` still resolves normally.)

src/core/agent-context.ts
```diff
@@ -153,12 +153,14 @@ export function resolveAgentHarness(
       resolution === "resolved" &&
       input.memoryConfigured &&
       hasAnyTool(allowedTools, ["memory_search", "memory_get"]),
+    // QA consumes Skill instructions through Read or configured MCP tools.
+    // Skill discovery is independent of command execution and authoring rights.
```
jacoblee-io, line 157: This rationale was written in `9496c87` and invalidated by `707895d`: QA now holds `local_script`, so it does not consume Skill instructions "through Read or configured MCP tools" — it executes them, and Skill discovery is no longer independent of command execution for this type.

The `agentType === "knowledge_qa" ||` clauses on the next lines should stay — I initially read them as dead code, but that is only true under a matched deployment. With an old Runtime pushing the read-only QA whitelist, `canOperate` is false and the clause is what keeps bundled/platform discovery on; your new boundaries test pins exactly that case. It is the comment that needs to catch up with the second commit.

src/core/agent-types.ts
```diff
@@ -204,9 +204,9 @@ export const AGENT_TYPES: Record<AgentType, AgentTypeDef> = {
   knowledge_qa: {
     label: "Knowledge Q&A Agent",
     description: "Researches bound knowledge bases and answers with synthesized, source-backed information.",
-    capabilities: ["read_files"],
+    capabilities: ["read_files", "run_local_scripts", "write_sandbox", "plan_tasks", "spawn_subagents", "session_output"],
```
jacoblee-io, line 207: `write_sandbox` was presumably chosen for `write`/`edit`, but it also carries `skill_preview`, and `skill_preview` in turn switches on `includeSkillAuthoring` (`agent-context.ts`), which injects `# Skill Authoring` into the prompt.

Confirmed in the manifest emitted by the new boundaries test — a Knowledge QA session carries `platform.skill_authoring`, 446 chars, telling it: *"Whenever you create, modify, optimize, or rewrite a skill, you MUST output the result via `skill_preview` ... write ALL files to `.siclaw/user-data/skill-drafts/<name>/`"*. Sub-agents inherit it too.

No permission consequence — `skill_preview` only renders — but a question-answering role being instructed on Skill authoring is off-role prompt weight in every QA turn. Either split the write tools from the authoring tool into separate groups, or accept it deliberately and say so in `agent-prompt-lifecycle.md` (the table row currently summarizes this preset as "files", which does not hint at Skill authoring).

portal-web/src/lib/toolCapabilities.ts
```diff
@@ -23,6 +23,7 @@ export const CAPABILITY_GROUPS: CapabilityGroup[] = [
   { key: "write_sandbox", name: "Write & author skills", description: "Write/edit scratch files and author skills (sandboxed to user-data)", tools: ["write", "edit", "skill_preview"] },
   { key: "inspect_infra", name: "Inspect infrastructure", description: "Read-only discovery of bound clusters and hosts", tools: ["cluster_list", "host_list"] },
   { key: "run_commands", name: "Run commands", description: "Execute whitelisted shell commands (kubectl read-only)", tools: ["bash", "node_exec", "pod_exec", "host_exec", "k8s_inspect"] },
+  { key: "run_local_scripts", name: "Run local Skill scripts", description: "Execute available Skill scripts locally", tools: ["local_script"] },
```
jacoblee-io, line 26: Resolution and the UI count both dedupe correctly, and the parity test plus the reworked overlap assertion cover drift — no correctness issue here.

One UI nit: these are now two adjacent checkboxes where selecting "Run scripts" already grants everything "Run local Skill scripts" grants, with nothing on screen saying so. An operator narrowing an agent's capabilities could plausibly uncheck the subset expecting it to remove `local_script`. Suggest the description carry the relationship, e.g. "Execute available Skill scripts locally (also included in Run scripts)".

jacoblee-io (approved):
Approving. The two blocking issues from the earlier pass are fixed, and the fixes are better scoped than what I suggested.

**Verified locally on `27a2b855`** (fresh worktree): `npx tsc --noEmit` clean; 9 touched test files / 173 tests green; plus two probes of my own comparing `main` and this branch.

### What the fix closes

**Ambient Skill discovery.** On `707895d`, `includeBundledSkills: true` for QA flipped `filterSkillsToHarness` to `false` whenever there was no `portalSkillsDir` and no `resolved/` — which is exactly the state of a K8s AgentBox before its first Skill sync (`session.ts` leaves `skillsDir` unset in K8s, and the startup budget explicitly allows coming up without a sync). Measured then: filter off, `~/.pi/agent/skills` admitted, 27 bundled SRE skills loaded into a QA prompt. The `unscopedCli` gate is the right shape — it names the real condition (standalone CLI, no agent identity, sre/custom) instead of inferring it from a discovery flag. `agent-factory-boundaries.test.ts` earns its keep: it drives the real factory against a real `agent-config/skills` tree rather than asserting flag values, and it covers the old-Runtime read-only-QA whitelist case explicitly.

**Operational Safety.** `sre || canOperate` is correct and equivalent for `custom`. Confirmed in the emitted manifest that a QA session now reports `"operationalSafety": true`.

**Write boundary.** The symlink walk plus the new end-to-end test (real Write/Edit tools, real symlinks, asserting the outside file is byte-unchanged) is a solid pin for the `write_sandbox` claim in the description, which previously had no test behind it.

### Non-blocking

One substantive item — the bundled-Skill discovery semantics now contradict the docs this PR edits — plus four smaller ones, all inline. None is a regression against `main` or a permission hole, so they read as follow-ups rather than merge blockers.

Rollout ordering degrades safely in both directions: `allowedTools` is computed by the Runtime (`internal-api.ts`) and pushed, so an old box receives tool names it already implements, and an old Runtime leaves QA on the read-only list. The Skill-scope filter is now on for every scoped session regardless, which is what makes the mixed-deploy window harmless.

(Unrelated to this PR: `src/core/pi-sdk.contract.test.ts` has a usage-evidence assertion that already fails on `main` in my environment, and `propose_execution` is listed in the `session_output` group with no implementation anywhere in the tree.)

PR #601, opened 2026-09-14
fix(kbc): bound worker shutdown and preserve recovery ownership

LikiosSedo:
Planner and batch sessions previously replaced the persistent authoring client and shared reply/completion state. Give each internal step its own client, task and reply. SDK connection and teardown stay in the same task; confirmed teardown permits the existing batch retry, while unconfirmed teardown prevents workspace reuse, synchronization, planner fallback and finalization.

Bound the interrupt call and worker cleanup so a stalled SDK cannot hold execution indefinitely. Cleanup failure emits `worker_teardown_failed`; Runtime persists and acknowledges the fatal error, then closes the relay so its existing cleanup disposes the box. It does not consume later artifacts from that worker. A late interrupt failure cannot reap a successor client. User cancellation stays cancelled. Close Claude subprocess pipes after SDK disconnect, including unread output from an aborted tool turn.

The command protocol and artifact/checkpoint commit contract remain compatible. Recovery schedules and ticket-repair admission remain control-plane responsibilities. This change does not add a persistent ticket-repair queue. Two existing self-check assertions are also aligned with the current fixture line numbering and bounded-repair contract.

Validation:
- Regression tests first reproduced unbounded cleanup, a hanging interrupt call, artifact consumption after a fatal error, Claude's unclosed stdout after cancellation, and a late watchdog failure cancelling a successor.
- `npm run build` and final `tsc --noEmit` passed; `npm test`: 7,609 passed.
- Combined KBC domain, ownership, Pi/Claude SDK, file-tool, event-relay and recovery suites: 293 passed. SDK tests use local synthetic model servers.
- Standalone `test_compile_box.py` HTTP/SSE protocol smoke passed.
- Real Pi and Claude SDK tests cover host-tool writes and cancellation through the internal-step driver. Claude cancellation also asserts process exit and closed stdout.

Rollout: ship matching Runtime and KBC box images; Runtime may be upgraded first. Existing boxes retain their original image. No image publication, cluster deployment or hosted-model acceptance was performed for this revision.


LikiosSedo:

